PR #522, opened 2014-08-11
Fix alien cache catalog updates, move cvmfschecksum to local cache

DrDaveD:
This has now been tested on largish lustre alien cache installation and it works fine.  This is tracked in https://sft.its.cern.ch/jira/browse/CVM-653

### DIFF
--- a/cvmfs/cache.cc
+++ b/cvmfs/cache.cc
@@ -856,8 +856,7 @@ catalog::LoadError CatalogManager::LoadCatalog(const PathString  &mountpoint,
   }
 
   // Happens only on init/remount, i.e. quota won't delete a cached catalog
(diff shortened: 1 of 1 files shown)


jblomer:
Turns out that this change breaks the use case when clients have no network connectivity but only access to the alien cache.  In this case, they have to take .cvmfschecksum from the alien cache where it has been put by `cvmfs_swissknife pull`.  I'm going to add an additional check: if the .cvmfschecksum file is not found in the client cache, try to open it from the alien cache.  That (hopefully) shouldn't affect your use case as the .cvmfschecksum file is never written to the alien cache.

DrDaveD:
Ah, you're talking about cvmfs_swissknife pull -c for fully pre-populated alien caches.  Your planned change sounds good.

PR #534, opened 2014-08-27
New feature: files already in S3 are not reuploaded

ssheikki:
Default behaviour is now that HEAD request is first done to all data files before possible PUT request. PUT request is done only if the file did not exist. This should speedup the upload process in case there is lots of duplicate files.


ssheikki:
Same fix is already in commit 2ce1b734.

ssheikki:
Same fix is already in commit 2ce1b734.

PR #553, opened 2014-09-23
Fix s3 thread safety

ssheikki:
We figured out that `S3Uploader` is not thread safe and thus fails when used as a stratum 1 backend. Fixed.


ssheikki:
The 078504ad commit have been fixed through other pull request. The other one is not (probably) necessary anymore.

PR #562, opened 2014-10-14
TEST: Proxy Failover Scenarios

reneme:
This test cases uses [WireMock](http://wiremock.org/index.html) to simulate various proxy failure scenarios and verify CVMFS's failover behaviour. WireMock needs a bit of setting up, though. I'll adapt the cloud platform contextualisation scripts soon, to allow the test case to run properly.

Also this adds `CURLE_GOT_NOTHING` and `CURLE_RECV_ERROR` to the error handling in the download module.


reneme:
I've put the wiremock specific shell functions in a wrapper file `wiremock_wrapper.sh` that can potentially be used by other test cases as well. First thing for each test case is to call `init_wiremock()` which will first check if `java` is available and if `wiremock` is already installed. If the latter is not the case, it will download the wiremock standalone jar from here: [https://ecsft.cern.ch/dist/cvmfs/wiremock/](https://ecsft.cern.ch/dist/cvmfs/wiremock/) and install it.

I tested if wiremock works on the essential platforms, namely SLC5/6 x64 and Ubuntu. Furthermore I updated the cloud testing contextualisation scripts to install java on these platforms.

PR #644, opened 2014-12-18
Refactor: Extract and Generalize the ObjectFetcher<> from CatalogTraversal<>

reneme:
This decouples the `ObjectFetcher<>` strategy classes from the `CatalogTraversal<>` and generalises it. This way it can be reused in various places throughout the swissknife code. Now the `CatalogTraversal<>` does take an `ObjectFetcher<>` type as template parameter and acts accordingly.

The `ObjectFetcher<>` is a CRTP implemented in `AbstractObjectFetcher<>` which has various implementations inheriting from it (currently `LocalObjectFetcher<>`, `HttpObjectFetcher<>` and `MockObjectFetcher`). Both `LocalObjectFetcher<>` and `HttpObjectFetcher<>` can be configured to emit different `Catalog` and `History` objects. Furthermore all `ObjectFetchers<>` encapsulate the loading of the `Manifest` object of a given repository.

In conclusion, `CatalogTraversals<>` can be configured to retrieve repository objects from different locations and traverse different catalog types (i.e. `Catalog`, `SimpleCatalog`, `WritableCatalog`) with the (hopefully) maximal amount of code reuse. While the `ObjectFetcher<>` can be used standalone (ie. to facilitate loading manifest or history objects) in other places of the code.

TODO: Unit tests for the `ObjectFetcher<>`. I'll append this pull request.


reneme:
Finalised the unit test for `MockObjectFetcher`, `LocalObjectFetcher<>` and `HttpObjectFetcher<>`. The latter is a the full blown production class and performs signature checks of a (mocked) manifest and whitelist. The test case `T_ObjectFetcher` contains code to mock properly signed `.cvmfspublished` and `.cvmfswhitelist` files along with a temporary backend storage that might come in handy for unit tests of other components. Can go in now...

cvmfs/catalog_traversal.h
```diff
@@ -147,26 +104,70 @@ class ObjectFetcher;
...
-template<class CatalogT, class ObjectFetcherT = ObjectFetcher>
-class CatalogTraversal : public Observable<CatalogTraversalData<CatalogT> > {
+template<class ObjectFetcherT>
+class CatalogTraversal
+  : public Observable<CatalogTraversalData<typename ObjectFetcherT::catalog_t> >
+{
  public:
-  typedef CatalogTraversalData<CatalogT> CallbackData;
-  typedef CatalogT                       Catalog;
+  typedef ObjectFetcherT                      object_fetcher_t;
```
jblomer: Why the switch from camel case to snake case?

cvmfs/object_fetcher.h
```diff
@@ -0,0 +1,387 @@
...
+#ifndef OBJECT_FETCHER_H
+#define OBJECT_FETCHER_H
+
+#include <string>
+
+#include "manifest.h"
+#include "manifest_fetch.h"
+#include "history_sqlite.h"
+#include "catalog.h"
+
```
jblomer: The blank line can be removed.  In general it would be better to sort the includes alphabetically within each group (system includes, language includes, project includes) so that with many includes we can quickly find if a particular include exists or not.

jblomer:
There is one specific comment but in general I do not quite understand in which cases camel case and in which cases snake case is used.  Shouldn't all type names be in camel case?

cvmfs/object_fetcher.h
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  ~HttpObjectFetcher() {
+    signature_manager_.Fini();
+    download_manager_.Fini();
+  }
+
+ protected:
+  bool Initialize() {
+    download_manager_.Init(1, true);
```
jblomer: The download manager should be taken in a way it is externally constructed.  There should be one central place where a download manager is configured.  This place can then deal with tricky configuration issues, e.g. if proxies are used or not.

cvmfs/object_fetcher.h
```diff
@@ -0,0 +1,387 @@
...
+    signature_manager_.Init();
```
jblomer: Same here.  We currently only use public rsa key pair but this does not necessarily remains the only way to validate a manifest.  There is already code to do it by PKI verification.  The exact mode in which the signature manager operates should not be hard coded into the object fetcher.

cvmfs/util.cc
```diff
@@ -163,6 +163,18 @@ bool IsAbsolutePath(const std::string &path) {
 }
 
 
+bool IsRemotePath(const std::string &path) {
```
jblomer: Better name might be IsHttpUrl

cvmfs/util.h
```diff
@@ -274,7 +275,7 @@ class UniquePtr : SingleCopy {
 
   inline operator bool() const        { return IsValid(); }
   inline T& operator*() const         { return *ref_; }
-  inline UniquePtr& operator=(T* ref) { ref_ = ref; return *this; }
+  inline UniquePtr& operator=(T* ref) { delete ref_; ref_ = ref; return *this; }
```
jblomer: Shouldn't this capture ref_ == ref?

cvmfs/catalog_traversal.h
reneme: This is somewhat inspired by the STL convention. I use it like this:

* CamelCase = Classes and structures
* CamelCaseT = Template parameters (note the suffix T)
* snake_case_t = Derived or dependent types

In STL you could do something like:

```C++
typedef std::vector<Foo> FooVector;

FooVector vector;
FooVector::value_type foo();
vector.push_back(foo);
FooVector::pointer p = &vector[0];
```

cvmfs/object_fetcher.h
reneme: Adding this to the list of code cleanup tasks... I agree, but consistency for that is probably broken anyway right now...

jblomer: At the moment the order is random so we can just start to introduce an order for all new things.

cvmfs/catalog_traversal.h
jblomer: I find this confusing because `some_name_t` looks like something from a C library.  I like the idea of being able to see a derived type but please think about a way of naming them in camel case, like it is done for template parameters.

cvmfs/object_fetcher.h
reneme: That is a good point, in fact. It adds some [boiler plate](https://github.com/reneme/cvmfs/commit/7732ec1c5c4069db2659490f19a9a9441f2a165f) to the user side, since one needs to initialise these objects every time an `HttpObjectFetcher<>` is needed. However, the gained flexibility is a real advantage. [Fixed](https://github.com/reneme/cvmfs/commit/f7dea8145991de0f4093f8a9911c6fce3002ade0).

cvmfs/util.cc
reneme: [Fair enough](https://github.com/reneme/cvmfs/commit/ce5ae5e901ae4371f1b7472c8fcbadee4854dd60).

cvmfs/util.h
reneme: Ooops! Very good catch. Thanks. [Fixed](https://github.com/reneme/cvmfs/commit/1f6d5003876d5b3560db8a66d1801a4ba0a2740c) and [tested](https://github.com/reneme/cvmfs/commit/875ca799b275dabb3c4d972a1fdfd677aaa282a3).

cvmfs/object_fetcher.h
reneme: [Fixed](https://github.com/reneme/cvmfs/commit/3b4b6ac4950f297d6ec280d6e7eb7ca58a4e0c36) for newly added files.... Doing this consistently everywhere is [left for later](https://sft.its.cern.ch/jira/browse/CVM-740)...

cvmfs/catalog_traversal.h
reneme: What about `CatalogTN` for `typename`. I feel that any other suffix would be misinterpreted as an abbreviation that causes confusion. With `T + <something>` there is at least a reference to template type stuff...

jblomer: Yep, that sounds good!

reneme: I'm preparing a patch... but I will send you another Pull Request for it... it's going to be quite many changes everywhere... I feel that this should not be mixed with the stuff in here.

jblomer: Yep, merging then

jblomer:
One more thing: I get a compile errors
    t_catalog_traversal.cc:16: error: using 'typename' outside of template
    t_garbage_collector.cc:20: error: using 'typename' outside of template

PR #651, opened 2015-01-02
Add support for fallback proxies

DrDaveD:
This pull request adds support for fallback proxies as called for in https://sft.its.cern.ch/jira/browse/CVM-708.  I chose the word "fallback" instead of "backup" because the backup proxy term was already used for another purpose.  There's a new config variable CVMFS_FALLBACK_PROXY to set a semicolon-separated list of fallback proxies, a new cvmfs_talk command 'proxy fallback' to set them on the command line, and a new libcvmfs option fallback_proxies to set them from the library.   Geo-api probing sorts them in addition to the server hosts.  SetProxyChain() sets both regular proxies and fallback proxies and GetProxyInfo() retrieves an additional piece of info that indicates the index into the proxies at which fallback proxies start.  This includes an update to the geoapi test function to also test setting and geo-sorting fallback proxies.


cvmfs/download.cc
```diff
@@ -245,10 +245,10 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
     // Write to memory
     if (info->destination_mem.pos + num_bytes > info->destination_mem.size) {
       if (info->destination_mem.size == 0)
-	LogCvmfs(kLogDownload, kLogDebug, "Content-Length was missing or zero, but %d bytes were received", info->destination_mem.pos + num_bytes);
+        LogCvmfs(kLogDownload, kLogDebug, "Content-Length was missing or zero, but %d bytes were received", info->destination_mem.pos + num_bytes);
```
jblomer: Please use a maximum line length of 80 characters

cvmfs/download.cc
```diff
@@ -1844,11 +1848,21 @@ bool DownloadManager::ProbeHostsGeo() {
...
     host_names.push_back(dns::ExtractHost(host_chain[i]));
   SortTeam(&host_names, &host_chain);
+
+  // add fallback proxy names to the end of the host list
+  unsigned first_geo_fallback = host_names.size();
+  for (unsigned i = fallback_group; i < proxy_chain.size(); ++i) {
+    // note that there's only ever one fallback proxy name per group
+    host_names.push_back(proxy_chain[i][0].host.name());
+  }
+  // There's no reason to sort fallbacks, they're set by a widely shared config
```
jblomer: Still it can't hurt to sort them upfront.

cvmfs/download.cc
```diff
@@ -1913,63 +1940,96 @@ bool DownloadManager::SortWrtGeoReply(
...
     vector<string> this_group = SplitString(proxy_groups[i], '|');
     for (unsigned j = 0; j < this_group.size(); ++j) {
-      // Note: DIRECT strings will be "extracted" to an empty string and not
-      // resolved.
-      urls.push_back(dns::ExtractHost(this_group[j]));
+      // Note: DIRECT strings will be "extracted" to an empty string.
+      string hostname = dns::ExtractHost(this_group[j]);
+      if ((hostname == "") && (fallback_proxy_list != "")) {
+        // skip DIRECTs if there are fallback proxies
+        opt_proxy_groups_fallback_--;
```
jblomer: That would mean if there is a DIRECT within a load-balance group, the group index of the fallback group changes?
Isn't this wrong, e.g. consider `proxy_list == "DIRECT|DIRECT|DIRECT" && fallback_proxy_list="foo"`?

cvmfs/download.cc
```diff
@@ -1913,63 +1940,96 @@ bool DownloadManager::SortWrtGeoReply(
...
+    }
+  }
+  vector<string> fallback_groups;
+  if (fallback_proxy_list != "")
+    fallback_groups = SplitString(fallback_proxy_list, ';');
+  for (unsigned i = 0; i < fallback_groups.size(); ++i) {
+    string fallback_proxy = fallback_groups[i];
+    size_t barpos = fallback_proxy.find('|');
+    if (barpos != std::string::npos) {
+      // ignore anything after a vertical bar, it's not supported in fallbacks
```
jblomer: Why wouldn't it be supported?  Wouldn't it be easier because the exact same code can be used to deal with the fallback proxies?

cvmfs/talk.cc
```diff
@@ -308,10 +311,56 @@ static void *MainTalk(void *data __attribute__((unused))) {
...
+            vector< vector<download::DownloadManager::ProxyInfo> > proxy_chain;
+            unsigned fallback_group;
+            cvmfs::download_manager_->GetProxyInfo(&proxy_chain, NULL, &fallback_group);
+            std::string fallback_proxies;
+            for (unsigned i = fallback_group; i < proxy_chain.size(); ++i) {
+              if (fallback_proxies.size() != 0)
+                fallback_proxies.append(";");
+              if (proxy_chain[i].size() > 0)
+                // fallback proxies may have only one per group
+                fallback_proxies.append(proxy_chain[i][0].Print());
```
jblomer: This will result in a different fallback proxy configuration if one of the fallback proxies got converted into a load-balance group because it uses round-robin DNS.

cvmfs/talk.cc
```diff
@@ -308,10 +311,56 @@ static void *MainTalk(void *data __attribute__((unused))) {
...
+        } else {
+          string fallback_proxies = line.substr(15);
+          vector< vector<download::DownloadManager::ProxyInfo> > proxy_chain;
+          unsigned fallback_group;
+
+          // get existing non-fallback proxies
+          cvmfs::download_manager_->GetProxyInfo(&proxy_chain, NULL, &fallback_group);
+          std::string proxies;
+          for (unsigned i = 0; i < fallback_group; ++i) {
+            if (proxy_chain[i].size() == 0) {
```
jblomer: If DIRECT proxies are skipped when the chain is set, is this condition still necessary?

jblomer:
Thanks!  I still have the feeling that it might be more complicated than necessary but I need to think about it a bit more (and perhaps I'm wrong).  Perhaps things could be simpler in the setters and getters if the original strings are kept in addition.

Regarding skip of DIRECT, it would be good to log a skipped DIRECT to syslog and also to warn about it in `cvmfs_config chksetup`.

cvmfs/download.cc
DrDaveD: When I wrote that I thought it would be significant extra work to sort, because it would require also sorting the corresponding subset of the proxy chain while not sorting the pre-fallback part of the proxy chain.   Now looking back I guess it would just require one more step of extracting the fallback part of the proxy chain.  Do you think I should do it?

Thanks for making me look at this again because now I realize that there's a problem below.  I'll add my own note in the appropriate place.

cvmfs/download.cc
```diff
@@ -1879,15 +1900,26 @@ bool DownloadManager::ProbeHostsGeo() {
...
+  vector< vector<ProxyInfo> > tmp_fallback_groups((*opt_proxy_groups_).begin() + opt_proxy_groups_fallback_,
+                                                  (*opt_proxy_groups_).end());
+  unsigned hosti = 0;
+  unsigned proxyi = opt_proxy_groups_fallback_;
+  for (unsigned i = 0; i < geo_order.size(); ++i) {
+    uint64_t orderval = geo_order[i];
+    if (orderval < (uint64_t) first_geo_fallback)
+      (*opt_host_chain_)[hosti++] = host_chain[orderval];
+    else
+      (*opt_proxy_groups_)[proxyi++] = tmp_fallback_groups[orderval - first_geo_fallback];
```
DrDaveD: I just realized that it isn't safe to simply overwrite the indexes in opt_host_chain_ and opt_proxy_groups_ because their lengths may have been changed by another thread in the meanwhile.  I need to recreate them the way that opt_host_chain_ was re-created before I changed it.

cvmfs/download.cc
DrDaveD: I think you're right.  I hadn't considered that possibility because it's wrong to have DIRECT in a load balance group in the first place.  It doesn't make sense to randomly alternate between a proxy and DIRECT.  I think the thing to do is to remove DIRECTs from a group.  As a concession if there's only DIRECTs in a group as in your example, it could be translated it into one DIRECT.

cvmfs/download.cc
DrDaveD: There's no need for it anymore now that round-robins are fully supported as groups.  The code has to be separate at this point anyway because of some other differences, and leaving out support here makes things simpler here and elsewhere.   In fact I would like to deprecate use of vertical bars if I could now that round-robin is supported, but because of backward compatibility it will probably never happen.  In any case I'd rather not introduce a new use for it where it isn't needed.

cvmfs/talk.cc
DrDaveD: I see.  I didn't realize that round-robin got expanded into multiple entries in the load-balance group.  I think I can use .url instead of .Print(),  I will try it. There's also a problem with setting a fallback proxy the way I do when there's a round-robin DNS based proxy group; that ends up with duplicates.  I suppose the way to detect that is to look for the same .host in the first two entries of a group.

cvmfs/talk.cc
DrDaveD: In this case there may not have been a fallback set earlier, so the original proxy list may still have a DIRECT.

DrDaveD:
> Perhaps things could be simpler in the setters and getters if the original strings are kept in addition.

Yes that's an interesting idea.  I guess for now the original strings per ProxyInfo structure is in "url" but I take it you're thinking about the whole original proxy & fallback string.   I'll look into that.

> Regarding skip of DIRECT, it would be good to log a skipped DIRECT to syslog and also to warn about it in cvmfs_config chksetup.

I don't consider it any kind of a bad condition deserving of warning for a DIRECT to be skipped when there's a fallback set; it is just slightly redefining DIRECT from the user's point of view to mean to connect "directly" to a fallback proxy instead of directly to a stratum 1.  Both conditions are bypassing the local proxies.  I'll put in a debug log message, but I do not think it should be a warning in chksetup or go to /var/log/messages by default.

cvmfs/talk.cc
DrDaveD: I decided to instead use your idea of keeping track of the original proxy and fallback proxy strings.

cvmfs/download.cc
DrDaveD: I decided to instead keep track of whether there were any hostnames in a group and only remove the group if there were none.

DrDaveD:
Jakob,

I believe I have addressed all of the issues.  Please check again.

Dave

cvmfs/download.cc
jblomer: From point of view of cvmfs, the vertical bars are by no means deprecated but a fully supported feature.  It provides high-availability even if you don't want or cannot edit DNS entries.

jblomer:
Hm, 'cvmfs_talk proxy fallback new' crashes with a SEGFAULT for me.

jblomer:
Since I started process of merging, I'll look into it and let you know if I get stuck.  I wanted to make a few more changes anyway.

cvmfs/download.cc
```diff
@@ -1879,14 +1902,38 @@ bool DownloadManager::ProbeHostsGeo() {
...
     return false;
   }
 
   pthread_mutex_lock(lock_options_);
   delete opt_host_chain_;
+  opt_host_chain_ = new vector<string>(host_chain.size());
+
+  vector< vector< ProxyInfo> > *proxy_groups = 
+        new vector< vector<ProxyInfo> > (
+            opt_proxy_groups_fallback_ + proxy_chain.size() - fallback_group);
```
jblomer: This I don't understand.  Isn't `opt_proxy_groups_fallback_ + proxy_chain.size() - fallback_group == proxy_chain.size()`?

cvmfs/download.cc
DrDaveD: Sure, but I'm thinking especially about the proxy auto-config case where it can't be supported.   I'd like to deprecate vertical bars even though I realize it's never going to happen.

cvmfs/download.cc
DrDaveD: It's possible that opt_proxy_groups_fallback_ might have changed while the lock wasn't held.  A comment would be in order there.  Since you're merging, I'll let you add it.

DrDaveD:
I don't see a SEGFAULT with 'proxy fallback new' when I re-built the branch of this pull request.   I did these commands:
```
$ ls /cvmfs/grid.cern.ch
$ sudo cvmfs_talk -i grid.cern.ch proxy info
Load-balance groups:
[0] http://131.225.153.35:3128 (squid1.fnal.gov, +5h), http://131.225.153.36:3128 (squid2.fnal.gov, +5h)
Active proxy: [0] http://131.225.153.35:3128
$ sudo cvmfs_talk -i grid.cern.ch proxy fallback new
OK
```

Of course SEGFAULT can be caused by uninitialized variables so it may depend on a lot of factors.  Or it may have something to do with the merge which at this point I assume required some conflict resolution.

jblomer:
Hm, possibly due to my merge then.  Sorry for the noise.  I'm going to list my adjustment in the JIRA ticket soon.

cvmfs/download.cc
jblomer: Ah, I see.  That means that `opt_proxy_fallback_list_`, `opt_num_proxies_`, `opt_proxy_groups_current_burned_`, and `opt_proxy_groups_current_` are possibly out of sync, too.  But should be possible to reset them here.

jblomer:
Ah, there's one more problem.  When only fallback proxies are used, these proxies must not be used in the Geo-API.  Instead the uuid must be used.  Fixing it.

jblomer:
Hm, actually no.  It's correct to use fallback proxies in the geo api.

DrDaveD:
> Ah, there's one more problem. When only fallback proxies are used, these proxies must not be used in the Geo-API. Instead the uuid must be used. Fixing it.  ... Hm, actually no. It's correct to use fallback proxies in the geo api. In principle. But if both are sorted at the same time and only fallback proxies are in use, the result depends on the initial order of fallback proxies. And what should happen is that fallback proxies are ordered first using the uuid in the Geo-API and only then hosts are ordered using the first fallback proxy in the Geo-API.

You've brought up a good point here.  My intention is that these fallback proxies will become configured by default for cern.ch, opensciencegrid.org, and egi.eu.  That means that if we don't do a dual Geo-API lookup, all the independent users who aren't at sites with proxies, who set CVMFS_HTTP_PROXY=DIRECT and get a uuid in their geo-api call, will connect to the closest fallback proxy and from there they'll connect to their closest stratum 1 which may not be the closest to the fallback proxy.   Ideally I suppose the thing to do would be to always do a separate Geo-API call for the servers whenever a new fallback proxy is selected for the first time, basing it on the fallback proxy.  Oh well, it's an optimization we might want to think about later.  I'll make a ticket so we don't forget about it.

PR #657, opened 2015-01-05
FIX: download of local files into memory

jblomer:
Fix the destination_mem.size field of the JobInfo struct.


reneme:
Fixed my issue as well! Thanks.

PR #667, opened 2015-01-07
FIX: ObjectFetcher<> Leaks Temporary Files

reneme:
This add an auto-manage functionality to the `Database<>` class that unlinks the underlying SQLite file when the database is destructed. By default, this  feature is disabled (legacy compatibility) but it exposes an API to control it:
```C++
void Database<>::TakeFileOwnership();
void Database<>::DropFileOwnership();
void Database<>::OwnsFile();
```

The `ObjectFetcher<>` is hiding the file dependency of `Catalog` and `History` objects. Hence if returned by `ObjectFetcher<>` they have the auto-management enabled by default for convenience. If the using code of `ObjectFetcher<>` wishes to manage the database file manually, both `Catalog` and `History` expose a facade API for controlling the underlying database:
```C++
void History::TakeDatabaseFileOwnership();
void History::DropDatabaseFileOwnership();
void History::OwnsDatabaseFile();
```

Furthermore this contains a number of unit test cases to verify the proper functionality of the auto-management feature.

### DIFF
--- a/cvmfs/history.h
+++ b/cvmfs/history.h
@@ -165,12 +165,11 @@ class History {
    */
   virtual bool GetHashes(std::vector<shash::Any> *hashes) const = 0;
 
-  /**
-   * If the concrete history implementation deals with local files internally
-   * the cleanup responsability is retained by the concrete implementation of
-   * this class.
-   */
-  virtual void TakeFileOwnership() {};
+  // database file management controls
+  virtual void TakeDatabaseFileOwnership() {};
+  virtual void DropDatabaseFileOwnership() {};
+  virtual bool OwnsDatabaseFile() const { return false; }
+
 
   const std::string& fqrn() const { return fqrn_; }
 
--- a/cvmfs/history.h
+++ b/cvmfs/history.h
@@ -166,10 +166,9 @@ class History {
   virtual bool GetHashes(std::vector<shash::Any> *hashes) const = 0;
 
   // database file management controls
-  virtual void TakeDatabaseFileOwnership() {};
-  virtual void DropDatabaseFileOwnership() {};
-  virtual bool OwnsDatabaseFile() const { return false; }
-
+  virtual void TakeDatabaseFileOwnership() = 0;
+  virtual void DropDatabaseFileOwnership() = 0;
+  virtual bool OwnsDatabaseFile() const    = 0;
 
   const std::string& fqrn() const { return fqrn_; }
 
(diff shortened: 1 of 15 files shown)


cvmfs/history.h
```diff
@@ -165,6 +165,12 @@ class History {
    */
   virtual bool GetHashes(std::vector<shash::Any> *hashes) const = 0;
 
+  // database file management controls
+  virtual void TakeDatabaseFileOwnership() {};
```
jblomer: Wouldn't it be better to make them abstract?  Otherwise one might forget to implement them in a derived class.

cvmfs/object_fetcher.h
```diff
@@ -195,8 +210,10 @@ class LocalObjectFetcher :
     file_path->clear();
 
     const std::string source = BuildPath(object_hash, hash_suffix);
-    const std::string dest   = temporary_directory_ + "/" +
-                               object_hash.ToString();
+    const std::string dest   = CreateTempPath(temporary_directory_ + "/" +
```
jblomer: Would be more secure to use CreateTempFile and to use the Decompress function working on files.  That also saves one pair of open/close functions.

cvmfs/object_fetcher.h
```diff
@@ -319,7 +336,9 @@ class HttpObjectFetcher :
     object_file->clear();
 
     const std::string url  = BuildUrl(object_hash, hash_suffix);
-    const std::string dest = temporary_directory_ + "/" + object_hash.ToString();
+    const std::string dest = CreateTempPath(temporary_directory_ + "/" +
```
jblomer: Would be more secure to use CreateTempFile and to use the JobInfo constructor working with `FILE *`. That also saves one pair of open/close functions.

cvmfs/history.h
reneme: Not sure if there will be another implementation of `History` soon, apart from `SqliteHistory` and `MockHistory`. But I agree, it cannot hurt to [make it pure virtual](https://github.com/reneme/cvmfs/commit/5d26bca7d3ba6f757e36a9f56cfb3c25c375c495).

cvmfs/object_fetcher.h
reneme: Good point. Didn't know that this decompress flavour exists. [Fixed](https://github.com/reneme/cvmfs/commit/50fc4501aabe4830a5c5246c1886fa9edac89508).

cvmfs/object_fetcher.h
reneme: Right, same here. [Fixed](https://github.com/reneme/cvmfs/commit/6d79d47136f688bf4f4d8444a6d84f8cb76a2902).

PR #668, opened 2015-01-07
FIX: Compiler Warnings in t_uuid.cc and t_dns.cc

reneme:
This fixes a couple of compiler warnings in `t_uuid.cc` and `t_dns.cc`. Googletest is pedantic when it comes to type safe comparisons. Furthermore I replaced some `assert()` calls by googletest's `ASSERT_TRUE()` with the advantage that this fails only the current test and doesn't kill the `cvmfs_unittests` binary entirely.

Some general remarks to the usage of googletest that I've seen in the code base:
* `ASSERT_EQ(<expected>, <actual>)` --> The first parameter is the expected value!
* `ASSERT_TRUE(a == b)` is bad practise, because there is an `ASSERT_EQ()`

### DIFF
--- a/test/unittests/t_uuid.cc
+++ b/test/unittests/t_uuid.cc
@@ -52,7 +52,7 @@ TEST(T_Uuid, FromCache) {
 
 TEST(T_Uuid, FailWrite) {
   UniquePtr<Uuid> uuid(Uuid::Create("/no/such/path"));
-  EXPECT_TRUE(uuid.IsValid());
+  EXPECT_FALSE(uuid.IsValid());
 }
 
 TEST(T_Uuid, FailRead) {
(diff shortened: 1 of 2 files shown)


test/unittests/t_uuid.cc
```diff
@@ -62,7 +62,7 @@ TEST(T_Uuid, FailRead) {
   fclose(f);
   UnlinkGuard unlink_guard(path);
   UniquePtr<Uuid> uuid(Uuid::Create(path));
-  EXPECT_TRUE(uuid == NULL);
+  EXPECT_TRUE(uuid.IsValid());
```
jblomer: That should be EXPECT_FALSE(uuid.IsValid()), no?

reneme: True... Thanks for reviewing carefully. [Fixed](https://github.com/cvmfs/cvmfs/pull/668), also the one above was quirked. However, I wonder, because yesterday the test didn't fail, I believe...

jblomer: Ah, that's important.  We should be confident that the test fails with the reversed condition.  Can follow it up?

reneme: Mhh, reverting the latest fix and re-running the test cases produces errors as expected. Perhaps I mixed up a `--gtest_filter=...`. Sorry for the fuzz.

PR #671, opened 2015-01-08
Fix s3 coverity bug fixes

ssheikki:
Fixes to coverity found defects in S3 code.

### DIFF
--- a/cvmfs/s3fanout.cc
+++ b/cvmfs/s3fanout.cc
@@ -456,7 +456,7 @@ int S3FanoutManager::InitializeDnsSettings(CURL *handle,
 
   // We need to resolve the hostname
   // TODO(ssheikki): support ipv6 also...  if (opt_ipv4_only_)
-  dns::Host host = resolver->Resolve(remote_host);
+  dns::Host host = resolver_->Resolve(remote_host);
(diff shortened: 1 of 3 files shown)


cvmfs/s3fanout.cc
```diff
@@ -356,10 +356,12 @@ void S3FanoutManager::ReleaseCurlHandle(JobInfo *info, CURL *handle) const {
   assert(elem != pool_handles_inuse_->end());
 
   if (pool_handles_idle_->size() > pool_max_handles_) {
-    CURL *handle = *elem;
-    curl_easy_setopt(handle, CURLOPT_SHARE, NULL);
-    curl_easy_cleanup(*elem);
-    curl_sharehandles_->erase(handle);
+    CURLcode retval = curl_easy_setopt(handle, CURLOPT_SHARE, NULL);
+    assert(retval == CURLE_OK);
+    curl_easy_cleanup(handle);
```
jblomer: That doesn't look right.  For handle, an option is set and then it is cleaned up.  Did you mean `curl_easy_cleanup(*elem)`?

jblomer: Ah, sorry, it does make sense. Never mind.

PR #674, opened 2015-01-08
Change default Apache configuration to set KeepAlive on

jblomer:


cvmfs/cvmfs_server
```diff
@@ -1681,6 +1681,7 @@ EOF
     local repository_dir=$(get_upstream_config $upstream)
     create_apache_config_file "cvmfs.${name}.conf" << EOF
 # Translation URL to real pathname
+KeepAlive On
```
DrDaveD: It's probably not so good to insert this between a comment ("# Translation URL to real pathname") and the code that the comment is about ("Alias ...").  Likewise below.

jblomer: Indeed. Fixed.

PR #677, opened 2015-01-08
Feature: Configure Maximal Concurrent Upload Count

reneme:
This is related to [CVM-703](https://sft.its.cern.ch/jira/browse/CVM-703) and allows to configure the number of maximal concurrent writes in the `server.conf` using `CVMFS_MAXIMAL_CONCURRENT_WRITES=...`. However, I would not sell this as a real feature, yet. It configures the maximal number of upload jobs in the queue rather than actual concurrency. At least for the local uploader (into POSIX based backends) there is no real write concurrency right now. S3 does a better job with the cURL multi API, I believe.

Anyway, it doesn't hurt to have this configurable. We'll anyway need to have a deeper look into optimising the write backend so real concurrency, async I/O or similar is left to be implemented...

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -3568,6 +3568,10 @@ publish() {
     if [ "x$CVMFS_GARBAGE_COLLECTION" = "xtrue" ]; then
       sync_command="$sync_command -g"
     fi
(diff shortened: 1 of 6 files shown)


jblomer:
Please also add the parameter to the table in the appendix of the technical report, mentioning also the default setting.

PR #686, opened 2015-01-12
Feature: Automatic Garbage Collection in `cvmfs_server snapshot`

reneme:
This (optionally) runs the automatic garbage collection after `cvmfs_server snapshot` fetched new revisions from the stratum 0. Additionally there is an integration test to verify the behaviour.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -3924,11 +3924,16 @@ __run_gc() {
   load_repo_config $name
 
   # sanity checks
-  is_garbage_collectable $name || return 1
+  is_garbage_collectable $name  || return 1
+  [ x"$repository_url" != x"" ] || return 2
   if [ $dry_run -eq 0 ]; then
-    is_in_transaction $name || is_stratum1 $name || return 2
+    is_in_transaction $name || is_stratum1 $name || return 3
   fi
 
+  if is_stratum0 $name; then
+    [ x"$manifest"  != x"" ] || return 4
+    [ x"$base_hash" != x"" ] || return 5
+  fi
 
   # do it!
   local user_shell="$(get_user_shell $name)"
@@ -3939,7 +3944,7 @@ __run_gc() {
                                         -k $CVMFS_PUBLIC_KEY       \
                                         -t ${CVMFS_SPOOL_DIR}/tmp/ \
                                         $additional_switches"
-  $user_shell "$gc_command" || return 3
+  $user_shell "$gc_command" || return 5
 
   if is_stratum0 $name && [ $dry_run -eq 0 ]; then
     tag_command="cvmfs_swissknife tag_empty_bin \
@@ -3951,7 +3956,7 @@ __run_gc() {
       -f $name                                  \
       -b $base_hash                             \
       -e $CVMFS_HASH_ALGORITHM"
-    $user_shell "$tag_command" || return 4
+    $user_shell "$tag_command" || return 6
   fi
 
   return 0
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -3617,10 +3617,10 @@ publish() {
       [ $? -eq 0 ] || { publish_failed $name; die "Cannot parse time stamp '$timestamp_threshold'"; }
       local dry_run=0
       __run_gc $name       \
-               $manifest   \
-               $trunk_hash \
                $stratum0   \
                $dry_run    \
+               $manifest   \
+               $trunk_hash \
                -z $tst || { local err=$?; publish_failed $name; die "Garbage collection failed ($err)"; }
     fi
 
@@ -3893,10 +3893,10 @@ gc() {
     # run the garbage collection
     echo "Running Garbage Collection"
     __run_gc $name           \
-             $manifest       \
-             $base_hash      \
              $repository_url \
              $dry_run        \
+             $manifest       \
+             $base_hash      \
              $additional_switches || die "Fail ($?)!"
 
     # sign the result
@@ -3914,10 +3914,10 @@ gc() {
 
 __run_gc() {
   local name="$1"
-  local manifest="$2"
-  local base_hash="$3"
-  local repository_url="$4"
-  local dry_run="$5"
+  local repository_url="$2"
+  local dry_run="$3"
+  local manifest="$4"
+  local base_hash="$5"
   shift 5
   local additional_switches="$*"
 
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -3995,6 +3995,7 @@ snapshot() {
     user=$CVMFS_USER
     spool_dir=$CVMFS_SPOOL_DIR
     stratum0=$CVMFS_STRATUM0
+    stratum1=$CVMFS_STRATUM1
     upstream=$CVMFS_UPSTREAM_STORAGE
     num_workers=$CVMFS_NUM_WORKERS
     public_key=$CVMFS_PUBLIC_KEY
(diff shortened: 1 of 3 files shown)


cvmfs/cvmfs_server
```diff
@@ -4046,8 +4052,20 @@ snapshot() {
       -i ${spool_dir}/tmp/last_snapshot \
       -o .cvmfs_last_snapshot"
 
-    # remove magic file after successful replication
-    $user_shell "$delete_command"
+    # run the automatic garbage collection (if configured)
+    if is_garbage_collectable $name && [ ! -z "$CVMFS_GC_TIMESTAMP_THRESHOLD" ]
```
DrDaveD: Doesn't a default need to be set somewhere for CVMFS_GC_TIMESTAMP_THRESHOLD?

reneme: For newly created repositories with enabled garbage collection (`cvmfs_server mkfs -z`) [this config variable defaults to](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/cvmfs_server#L1668) `3 days ago`. Additionally, I will cover how to configure legacy repositories for garbage collection in a documentation section. This is to be written still...

I don't think that defaulting `CVMFS_GC_TIMESTAMP_THRESHOLD` here is a good idea as there could be repositories that are garbage collectable (`CVMFS_GARBAGE_COLLECTION=true`) but don't want to make use of the automatic garbage collection feature (`CVMFS_GC_TIMESTAMP_THRESHOLD` is set in the configuration).

DrDaveD: > For newly created repositories with enabled garbage collection (cvmfs_server mkfs -z) this config variable defaults to 3 days ago. 

That default is only used on repository servers, in mkfs() and import() functions, not on stratum 1s' add-replica().  There's no default for new replicas on stratum 1s. 

> I don't think that defaulting CVMFS_GC_TIMESTAMP_THRESHOLD here is a good idea as there could be repositories that are garbage collectable (CVMFS_GARBAGE_COLLECTION=true) but don't want to make use of the automatic garbage collection feature (CVMFS_GC_TIMESTAMP_THRESHOLD is set in the configuration).

Oh, so you're overloading the meaning of "threshold" to also mean don't do any collection if it isn't set, and requiring stratum 1 administrators to set a value in order to get collection done.  The CVMFS_GARBAGE_COLLECTION flag is only indicating that it is collectable, not that collection is happening.  That's a way to do it, but it seems misleading to me.  Why bother to have the CVMFS_GARBAGE_COLLECTION flag at all if it doesn't do anything but make it possible to use the threshold?  I'd rather have add-replica() not set CVMFS_GARBAGE_COLLECTION=true on collectible repos and have a default for the threshold.

DrDaveD: If you want some indication that it the repo is collectible you could add a default CVMFS_GARBAGE_COLLECTION=false on those that are collectible, and leave it out completely on those that aren't.

reneme: > Why bother to have the CVMFS_GARBAGE_COLLECTION flag at all if it doesn't do anything but make it possible to use the threshold?

In the long-term we might have the `cvmfs_server gc` as a public command, to explicitly run the garbage collection instead of running it automatically after each publish. With `CVMFS_GARBAGE_COLLECTION=true` the basic functionality of garbage collection is switched on. `CVMFS_GC_TIMESTAMP_THRESHOLD` is meant to additionally enable the *automatic* garbage collection on `publish`/`snapshot`. Though, I agree that this is not obvious and propose to rename the `CVMFS_GC_TIMESTAMP_THRESHOLD` to `CVMFS_AUTO_GC_TIMESPAN`. Does that seem more clear to you?

> That default is only used on repository servers, in mkfs() and import() functions, not on stratum 1s' add-replica(). There's no default for new replicas on stratum 1s.

You are right, that is actually a flaw and I am going to fix it.

DrDaveD: > In the long-term we might have the cvmfs_server gc as a public command, to explicitly run the garbage collection instead of running it automatically after each publish.  With CVMFS_GARBAGE_COLLECTION=true the basic functionality of garbage collection is switched on. 

Couldn't/shouldn't the functionality of garbage collection be turned on by looking for 'Gyes' in the .cvmfspublished file?  Or are you saying that if 'cvmfs_server gc' became a public function you wouldn't require that but instead go by the environment variable?  It seems to me that if somebody ran the gc function by hand, it ought to be done whether or not an environment variable setting was turned on.

> CVMFS_GC_TIMESTAMP_THRESHOLD is meant to additionally enable the automatic garbage collection on publish/snapshot. Though, I agree that this is not obvious and propose to rename the CVMFS_GC_TIMESTAMP_THRESHOLD to CVMFS_AUTO_GC_TIMESPAN. Does that seem more clear to you?

That's a bit of an improvement but it still kind of implies to me that it's the timespan for auto gc and not that it enables auto gc.  I would rather replace CVMFS_GARBAGE_COLLECTION with CVMFS_AUTO_GARBAGE_COLLECTION to enable automatic garbage collection, and have a default for the threshold.  I don't see why most people should care about the threshold, a default of 3 days will be fine.

reneme: The `Gyes` in *.cvmfspublished* indicates that a repository is garbage collectible to the outside world (for example stratum1s). On the stratum 0 the garbage collectibility is determined by `CVMFS_GARBAGE_COLLECTION=true` and the flag in *.cvmfspublished* is set accordingly.

I agree that stratum 1s should be able to determine the garbage collectibility solely based on the `Gyes` in *.cvmfspublished* omitting the `CVMFS_GARBAGE_COLLECTION=true` all together and make this a stratum0-only configuration variable. I'll look into that.

reneme: Update: Added a [Pull Request](https://github.com/cvmfs/cvmfs/pull/694) that both renames `CVMFS_GC_TIMESTAMP_THRESHOLD` and removes the necessity of setting `CVMFS_GARBAGE_COLLECTION=true` on stratum 1s.

reneme: Coming back to the configuration variables... There is already a sensitive default of '3 days ago' when you create a stratum1 from a garbage collectible repository. Hence, `cvmfs_server add-replica <...> <...>` will place `CVMFS_AUTO_GC_TIMESPAN='3 days ago'` in the config automatically.

We see the garbage collection as an experimental feature addressing mainly the nightly integration build use-case for now. The implementation and configuration of this features will almost certainly be subject to change in future versions. I suggest that we should first gain some experience with it and later see which changes are really necessary.

DrDaveD: > There is already a sensitive default of '3 days ago' when you create a stratum1 from a garbage collectible repository

It wasn't there when I checked 11 days ago, but I do see that you added it now because that's the way that the GC gets enabled on the replica.  In any case, if somebody converts an existing repo to be garbage collectible and then somebody converts the corresponding existing replica, it has to be added by hand.  That part's not so bad, it can be in the instructions, but I do still object to the double purpose for the one variable.   It's confusing, and there's no reason for most people to set a timespan.

> We see the garbage collection as an experimental feature addressing mainly the nightly integration build use-case for now. The implementation and configuration of this features will almost certainly be subject to change in future versions. I suggest that we should first gain some experience with it and later see which changes are really necessary.

The problem is that once you've got some usage in production you've always got to be backward compatible.  That constrains what you can change, especially things in configuration files.

reneme:
You are right, I [changed the configuration](https://github.com/cvmfs/cvmfs/pull/739) accordingly. Having a somewhat 'hidden' default value for `CVMFS_AUTO_GC_TIMESPAN` and switching the feature on with `CVMFS_AUTO_GC=true` is clearer than the previously implemented double meaning.

PR #691, opened 2015-01-13
Fix: Memory Error in Sql::Bind<>

reneme:
This fixes the 'sporadic SQL insert failure' that showed up in some unit test cases. There is a convenience method `Sql::Bind<>()` that binds a specific value to an internal SQLite query placeholder and uses template parameter deduction to figure out which concrete `Sql::Bind...()` method to use (f.e. `BindInteger()`, `BindDouble()`, ...). For `std::string` it was using `BindText()` which just binds a pointer to the internal string buffer. This is fine for strings that stay around until `Sql::Execute()` is called on the same query object. Hence, it was causing undefined behaviour when using `Sql::Bind<>()` with temporary strings like:

```C++
const int bar = 1337;
Sql q(db, "INSERT INTO foo (:bar);")
q->Bind(1, "foo" + StringifyInt(bar)); // binding a temporary (resulted in undefined behaviour)
q->Execute();
```

The `Sql::Bind<std::string>()` method now uses `Sql::BindTextTransient()` to avoid this problem. However, this comes at the cost of copying strings in any case, even if it is not necessary. In general this shouldn't be a big deal, but if binding large strings to a query it is probably beneficial to use the concrete `Sql::BindText()` instead of the convenience method. I've added a comment to `Sql::Bind<>()` for clarification.

### DIFF
--- a/cvmfs/sql.h
+++ b/cvmfs/sql.h
@@ -300,6 +300,14 @@ class Sql {
     last_error_code_ = sqlite3_bind_text(statement_, index, value, size, dtor);
     return Successful();
   }
+
+  /**
+   * Figures out the type to be bound by template parameter deduction
+   * NOTE: For strings or char* buffers this is suboptimal, since it needs to
+   *       assume that the provided buffer is transient and copy it to be sure.
+   *       Furthermore, for char* buffers we need to assume a null-terminated
+   *       C-like string to obtain its length using strlen().
+   */
   template <typename T>
   inline bool Bind(const int index, const T value);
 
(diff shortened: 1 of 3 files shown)


jblomer:
Please create a ticket in the code cleanup section in order to remind us to review where we bind strings in the code.

reneme:
We don't use this convenience wrapper in a lot of places yet anyway. I mainly needed it to simplify the `sqlite::Database::SetProperty<>()` and `sqlite::Database::GetProperty<>()` interface. Setting properties should be a rather sparse task anyway, the performance penalty shouldn't hurt too much for this case.

PR #694, opened 2015-01-14
Fix: Rename Timestamp Threshold and Remove CVMFS_GARBAGE_COLLECTION from Stratum 1

reneme:
As @DrDaveD correctly pointed out, `CVMFS_GARBAGE_COLLECTION=true` is not necessary on stratum 1 as this information is already embedded in the manifest file put together and signed by stratum 0. Therefore, this makes `CVMFS_GARBAGE_COLLECTION` a stratum-0-only configuration variable and takes only the *.cvmfspublished* file into account on stratum 1. Furthermore this renames `CVMFS_GC_TIMESTAMP_THRESHOLD` to `CVMFS_AUTO_GC_TIMESPAN` for better usability.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -715,6 +715,16 @@ is_garbage_collectable() {
   [ x"$CVMFS_GARBAGE_COLLECTION" = x"true" ]
 }
 
+# checks if a repository has automatic garbage collection enabled
+#
+# @param name  the name of the repository to be checked
+# @return      0 if automatic garbage collection is enabled
+has_auto_garbage_collection_enabled() {
+  local name=$1
+  load_repo_config $name
+  is_garbage_collectable $name && [ ! -z "$CVMFS_AUTO_GC_TIMESPAN" ]
+}
+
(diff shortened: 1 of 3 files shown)


DrDaveD:
I'm still uncomfortable with "CVMFS_AUTO_GC_TIMESPAN" having the double meaning of enabling auto GC and setting a timespan.  One issue is that when you decide to enable "cvmfs_server gc" as a separate public function, it would make sense for it to use the same timespan variable if set, but not if it has "AUTO" in the name.

I think it would better to go with CVMFS_AUTO_GARBAGE_COLLECTION or CVMFS_AUTO_GC as true/false for both stratum 0 and stratum 1, stick with a timespan/timestamp variable that does not have "AUTO" in the name, and let it have a default value of '3 days ago' if it is not set.    Let "cvmfs_server gc" not require any variables, but honor the timespan/timestamp variable if it is set and otherwise use the default.

DrDaveD:
Another issue is that I think it would be better for most people to not have to bother with setting a timespan.  They probably don't know what value makes sense.

PR #696, opened 2015-01-14
FIX: mktime Produced Non-Deterministic Results on SL6

reneme:
Turns out, that the `mktime()` function seems to react non-deterministic for semi-correct inputs. First, the [documentation](http://www.cplusplus.com/reference/ctime/tm/) says, that `tm_mon` needs to be from 0 to 11 (January == 0) and furthermore I specified the `tm_isdst` 'daylight saving time' flag. This seems to fix some unit tests that failed occasionally (not always) on SLC6 x64.... Enjoy the little things!


jblomer:
Very good!  Do we use mktime somewhere else in the code?

PR #698, opened 2015-01-15
Fix: Properly Close Database Files

reneme:
Since `sqlite::Database<>` used `sqlite3_close()` to close a database handle that was opened with `sqlite3_open_v2()` it failed. Unfortunately the failing happened silently, essentially leaking file descriptors. This uses `sqlite3_close_v2()` and does some debug error checking on the function.

### DIFF
--- a/cvmfs/sql.h
+++ b/cvmfs/sql.h
@@ -68,6 +68,27 @@ class Sql;
  */
 template <class DerivedT>
 class Database : SingleCopy {
+ private:
+  struct DatabaseRaiiWrapper {
+    DatabaseRaiiWrapper(const std::string   &filename,
+                        Database<DerivedT>  *delegate)
+      : sqlite_db(NULL)
+      , db_file_guard(filename, UnlinkGuard::kDisabled)
+      , delegate_(delegate) {}
+    ~DatabaseRaiiWrapper();
+
+    sqlite3*           database() const { return sqlite_db;            }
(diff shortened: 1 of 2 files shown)


jblomer:
Whether a database connection is opened by `sqlite3_open` or `sqlite3_open_v2` shouldn't matter.  But `sqlite_close_v3` (behaves differently with respect to outstanding SQL statements)[https://sqlite.org/c3ref/close.html].  If you see `sqlite_close_v3` and `sqlite_close` behaving differently, it is pretty likely that the underlying problem is somewhere else.

reneme:
Admitted. That is actually a good point. The root cause is, that `Database<>` contains some prepared statements that are managed as `UniquePtr<>`. Hence, they are wiped *after* the d'tor of `Database<>` closed the database connection and (potentially) unlinked the database file.

Now, one could argue, that this is a proper usage of `sqlite_close_v2()` as it defers the destruction of the SQLite context until the last prepared statement goes belly up. Thus `UniquePtr<>` would take care of those statements right after the d'tor ran cleaning up the SQLite context implicitly. However, I consider that quite implicit and also black magic.

My solution is, to build a thin `struct DatabaseRaiiWrapper` around the SQLite context and the `UnlinkGuard` for the database file. As this object will be destructed *after* the `UniquePtr<>` destructed the prepared statements, everything works fine.

PR #708, opened 2015-01-20
Fix IP resolving: prefer IPv4 over IPv6

jblomer:
At least VMware Fusion, possibly other clients too, have problems connecting to IPv6 servers.  Hence IPv4 addresses are preferred if available.


jblomer:
Not complete yet.

PR #732, opened 2015-01-26
Fix: Catalog Defragmentation

reneme:
The [recently implemented optimisation](https://github.com/cvmfs/cvmfs/pull/727) of the catalog defragmentation query was flawed. The *chunks* table has a foreign key into the *catalog* table. Therefore, the - intermediate - sweeping of the *catalog* table caused a foreign key violation. I fixed it by temporarily disabling the foreign key checks in SQLite (`PRAGMA foreign_keys = OFF`). Furthermore I added a large file to the catalog defragmentation integration test (553) to actively check for this problem.


jblomer:
Good catch!

PR #734, opened 2015-01-26
Fix: Proxy Failover Test

reneme:
Seems that `wiremock` needs a considerable amount of time to start up as a background service. The proxy-failover integration test case (572) fails occasionally because it tries to mount the CernVM-FS client through `wiremock` before it came up properly and fails. Java... ;-)

Furthermore the same test case is disabled for the S3 tests, because it relies on a local backend.


reneme:
Addition: Netcat (`nc`) seems to be available on SLC5/6 and Ubuntu by default...

PR #735, opened 2015-01-26
Fix: Migration Test on SLC5 and SLC6

reneme:
This adds a partition for `/var/lib/cvmfs`. It seems that the migration test fails on SLC 5 and 6 because the cache partition runs full. Furthermore this disables the migration test on CentOS 7, because there is no package of CernVM-FS 2.1.19 for it. Keeping my fingers crossed...

### DIFF
--- a/test/cloud_testing/platforms/centos7_x86_64_test.sh
+++ b/test/cloud_testing/platforms/centos7_x86_64_test.sh
@@ -32,7 +32,8 @@ export CVMFS_TEST_SERVER_CACHE='/srv/cache' &&         \
                           src/024-reload-during-asetup \
                           src/5* || it_retval=$?
 
-echo "running CernVM-FS migration test cases..."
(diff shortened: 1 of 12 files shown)


reneme:
Please wait... still need to check the SLC 5 yum stuff... meh!

reneme:
Implemented a fallback version that removes previous version packages and conflicting packages on old yum versions... I believe, most of this code can be dropped after the update campaign for 2.1.20 is done.

reneme:
Can be merged, built and stuffed into the cloud tests for a new iteration now...

DrDaveD:
It might be a good idea to be cautious about using the term "fallback" for things other than the fallback proxy, in order to avoid confusion especially when doing greps to see where it's referenced.

reneme:
Well... I am very much in favour of clear naming, but this is a fairly general term for something being replaced by some alternative method... Furthermore it doesn't even reside in the production code tree...

PR #747, opened 2015-01-29
Fix s3 unit tests

ssheikki:
Local upload unit tests work now also for S3. Tests are basically copy paste, but a simple S3 mockup server was added to make the tests possible. Few relatively small fixes also to the actual code due to that. Unfortunately also one config file spelling mistake was found, PARALLELL should be PARALLEL.

### DIFF
--- a/cvmfs/s3fanout.cc
+++ b/cvmfs/s3fanout.cc
@@ -772,14 +772,17 @@ bool S3FanoutManager::VerifyAndFinalize(const int curl_error, JobInfo *info) {
     try_again = CanRetry(info);
   }
   if (try_again) {
-    LogCvmfs(kLogDownload, kLogDebug, "Trying again to upload %s",
-             info->object_key.c_str());
-    // Reset origin
-    if (info->origin == kOriginMem)
-      info->origin_mem.pos = 0;
-    if (info->origin == kOriginPath)
-      rewind(info->origin_file);
-
+    if (info->request == JobInfo::kReqPut) {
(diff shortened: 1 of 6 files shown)


reneme:
Thanks! That provides some nice extra test coverage.

One major comment though:
Please have a look at googletest's [TypedTest](https://code.google.com/p/googletest/wiki/AdvancedGuide#Typed_Tests) that allows you to share unit test code for different implementations of the same module. I have done the same for the `ObjectFetcher<>` and `History` tests and it works pretty well. Have a look in `t_object_fetcher.cc` and/or `t_history.cc` for examples.

One idiomatic thing that you will probably need is a type-based overloaded template method inspired from [this StackOverflow post](http://stackoverflow.com/questions/5512910/explicit-specialization-of-template-class-member-function). It allows you to rather simply abstract away the differences between the setup procedures of `LocalUploader` and `S3Uploader`.

Using this googletest feature, we should be able to merge both `t_local_uploader.cc` and `t_s3_uploader.cc` in one `t_uploaders.cc` unit test set. With the main advantage that changes or additions to the unit test set are only needed once.

Please feel free to come back to me if things are unclear, I'm happy to have a closer look on that as well.

reneme:
One more general thing: You made also some patches to the production code. Since we are targeting for a code freeze still this week: Is there anything (except the config variable typo fix) that is worth to go into the 2.1.20 release? 

Since the code runs pretty much flawlessly on the LHCb nightlies, I am a bit reluctant to potentially break it in the last second. ;-)

PR #755, opened 2015-01-30
Fix: Read Permission Problems when Copying from /usr/bin/*

reneme:
Same problem [as before](https://github.com/cvmfs/cvmfs/pull/753), but with `/usr/bin/*`. This is not critical but just to be sure.


reneme:
wait please! :)

PR #778, opened 2015-02-12
Fix: T_ManagedExec.ExecuteBinaryDoubleFork Crashes Hard

reneme:
While testing some Jenkins specific things @Moliholi found that `T_ManagedExec.ExecuteBinaryDoubleFork` fails occasionally  with *SIGPIPE*. This test double forks and executes `gdb` and tries to communicate with the new process through a pipe. Turns out, that `gdb` was not installed on the machine.

I suppose, that there is a race between the parent process and the forked one when the parent is writing to the pipe. If the child process didn't invoke `execvp(gdb)` yet, it will still be around and the write to the pipe works fine. However, if the child already invoked `execvp(gdb)` (and thus is dead already since `gdb` is not found) the parent receives *SIGPIPE* because of the write to a broken pipe and crashes hard as well.

I added `GetExecutablePath()` to `util.h` that resembles the `which` utility in bash. Using that, the test case can check if `gdb` is installed before trying to execute it. Furthermore there is a test `T_Util.GetExecutablePath` to verify the function.

### DIFF
--- a/cvmfs/util.cc
+++ b/cvmfs/util.cc
@@ -1373,6 +1373,64 @@ bool ManagedExec(const vector<string>  &command_line,
   return true;
 }
 
+
+/**
+ * Traverses the $PATH environment variable to find the absolute path of a given
+ * program name. Pretty much what `which` in bash would do.
+ *
+ * Inspired by a solution to this bulletin board question:
+ *  http://www.linuxquestions.org/questions/programming-9/g
+ *         et-full-path-of-a-command-in-c-117965/#post611028
+ *
+ * @param exe_name  the name of the program to search the $PATH for
+ * @return          absolute path to the program or empty string if not found
+ */
+std::string GetExecutablePath(const std::string &exe_name) {
+  std::string result;
+
+  if (exe_name.empty() || exe_name.find('/') != std::string::npos) {
+    return result;
+  }
+
+  const char *searchpath = getenv("PATH");
+  if (NULL == searchpath) {
+    return result;
+  }
+
+  const std::vector<std::string> paths = SplitString(searchpath, ':');
+        std::vector<std::string>::const_iterator i    = paths.begin();
+  const std::vector<std::string>::const_iterator iend = paths.end();
+  for (; i != iend; ++i) {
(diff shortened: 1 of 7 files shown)


jblomer:
Since `GetExecutablePath()` is currently only needed by the tests, it should rather be placed in test_util for the time being.

cvmfs/util.cc
```diff
@@ -1373,6 +1373,64 @@ bool ManagedExec(const vector<string>  &command_line,
...
+    return result;
+  }
+
+  const std::vector<std::string> paths = SplitString(searchpath, ':');
+        std::vector<std::string>::const_iterator i    = paths.begin();
+  const std::vector<std::string>::const_iterator iend = paths.end();
+  for (; i != iend; ++i) {
+    const std::string candidate_path = *i + "/" + exe_name;
+
+    char *real_path = realpath(candidate_path.c_str(), NULL);
```
jblomer: Real path can be put on the stack as char[MAX_PATH], which should avoid a bunch of `malloc()` and `free()` calls.

cvmfs/util.cc
```diff
@@ -1373,6 +1373,64 @@ bool ManagedExec(const vector<string>  &command_line,
...
+    if (NULL == real_path) {
+      continue;
+    }
+
+    struct stat statinfo;
+    const int res = stat(real_path, &statinfo);
```
jblomer: `stat()` in cvmfs is handled by the portable (64bit libc interface on Linux and OS X) `platform_stat`

reneme:
Agreed and [moved to testutil.h](https://github.com/reneme/cvmfs/commit/0bda5fad30348aa7d0b13e8ecaa3528c5c5200d8). Avoiding `malloc()` is a good idea [as well](https://github.com/reneme/cvmfs/commit/e179166eca7b94ff2b5435571f4ae4cc65922231). Meh, forgot about the platform `stat()`, [fixed](https://github.com/reneme/cvmfs/commit/8ab48c4cdb3af5313aea3f9f9cbc694eaf7cfcb8) as well.

PR #793, opened 2015-02-18
Added fast parse options feature. Changed options interface.

Moliholy:
Also added a markdown file in /doc explaining the problems we found out while running with IgProf


cvmfs/loader.cc
```diff
@@ -55,6 +55,7 @@ struct CvmfsOptions {
   int grab_mountpoint;
   int cvmfs_suid;
   int disable_watchdog;
+  int fast_parse;
```
reneme: Maybe clarify in the name of that variable that you deal with the option parsing? `fast_option_parsing` or `minimal_option_parsing`?

jblomer:
Many thanks, that looks already quite good!  

Since this is your first larger commit, we'll probably take it through a few rounds before merging.  I'd like to ask you to create a feature branch or a bugfix branch for every pull request you're working on.  Also, please create separate pull request for
  - The changes to .gitignore
  - The igprof document
  - The interface change to options and the unit test
  - The additional logic for fast_parse (which we should probably rename to simple_parse)

We'll then work on the details of the individual pull requests.

cvmfs/talk.cc
```diff
@@ -510,11 +511,12 @@ static void *MainTalk(void *data __attribute__((unused))) {
 /**
  * Init the socket.
  */
-bool Init(const string &cachedir) {
+bool Init(const string &cachedir, OptionsManager *optionsManager) {
```
Moliholy: I forgot to change the variable name to something like options_manager, according to the Google C++ style guide.

jblomer:
As discussed, closing this one in lieu for the four separate ones.

PR #794, opened 2015-02-18
Changes in gitignore

Moliholy:
Eclipse generates in the root folder the following files and directories, which have been added to the .gitignore:
- .metadata/
- .settings/
- .RemoteSystemsTempFiles/


jblomer:
It would be good to have comments explaining where the excluded files and directories are produced.

PR #810, opened 2015-03-03
Fix: /* in .cvmfsdirtab Causes Crash

reneme:
This fixes a problem reported by @DrDaveD causing `cvmfs_swissknife dirtab` to crash if the `.cvmfsdirtab` file contains `/*`. The root cause was a filtering check to rule out the `.` and `..` directories from the dirtab application rules. The check is done using `std::substr` but unfortunately in the wrong order resulting in `"/.".substr(3)` and a crash.

### DIFF
--- a/cvmfs/swissknife_sync.cc
+++ b/cvmfs/swissknife_sync.cc
@@ -326,7 +326,7 @@ void swissknife::CommandApplyDirtab::FilterCandidatesFromGlobResult(
     }
 
     // check if the path is a meta-directory (. or ..)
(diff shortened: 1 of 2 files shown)


jblomer:
Please make a cpplint run

jblomer:
Please add a commit to this pull request after a cpplint run.

reneme:
Fixed the whitespace in the `assert` call. Interesting: I actually did that white space deliberately and exclusively for `assert` to emphasize that this is actually _not_ a real function call.

PR #811, opened 2015-03-04
Fix compile error on mac

jblomer:


reneme:
Ah! Thanks. I actually stumbled over that yesterday evening and forgot to tell you. :-)

PR #812, opened 2015-03-07
Use custom sqlite vfs for reading file catalogs in the client

jblomer:
This sqlite vfs is tuned for read-only, immutable files.  There is no locking and once opened, only the file descriptor is used.


cvmfs/sqlitevfs.cc
```diff
@@ -0,0 +1,494 @@
...
+ * TODO(jblomer): the reads seem to be rather small.  Investigate buffered read.
+ */
+static int VfsRdOnlyRead(
+  sqlite3_file *pFile,
+  void *zBuf,
+  int iAmt,
+  sqlite_int64 iOfst
+) {
+  VfsRdOnlyFile *p = reinterpret_cast<VfsRdOnlyFile *>(pFile);
+  ssize_t got = pread(p->fd, zBuf, iAmt, iOfst);
```
reneme: style guide: shouldn't these variable names be `snake_case`?

jblomer: Perhaps.  Borderline because they connect to sqlite and reflect the variable names from the sqlite documentation.  We have a similar situation with the calling conventions in tbb.  Please feel free to change if you think it makes the code more readable.

cvmfs/statistics.h
```diff
@@ -45,6 +45,13 @@ class Counter {
   atomic_int64 counter_;
 };
 
+// perf::Func(Counter) is more clear to read in the code
+inline void Dec(class Counter *counter) { counter->Dec(); }
```
reneme: why not simply `counter++`?

jblomer: With counter++ you'd need to think if this is atomic or not when you see it in the code.  `Inc()` or `Dec()` can be immediately recognized.

cvmfs/util.cc
```diff
@@ -886,6 +886,35 @@ bool HasPrefix(const string &str, const string &prefix,
 }
 
 
+bool HasSuffix(
```
reneme: What about:
``` C++
#include <algorithm>

struct IgnoreCaseComperator {
  bool operator()(const std::string::value_type a,
                  const std::string::value_type b) const {
    return std::tolower(a) == std::tolower(b);
  }
};

bool HasSuffix(const std::string &value,
               const std::string &ending,
               const bool         ignore_case = false) {
  if (ending.size() > value.size()) return false;
  const IgnoreCaseComperator icmp;
  return (ignore_case)
           ? std::equal(ending.rbegin(), ending.rend(), value.rbegin(), icmp)
           : std::equal(ending.rbegin(), ending.rend(), value.rbegin());
}
```

Less lines, easier to understand, use of STL and probably faster, because it only checks `ignore_case` once and potentially inlines the rest.

jblomer: According to my mini benchmark: no significant difference in size or performance (gcc 4.4 and OS X 10.9 clang).  Changed.

PR #813, opened 2015-03-09
Added benchmarks

Moliholy:
Added a benchmark_tests folder including the tests we've been dealing with. Modified run.sh and test_functions in consequence


test/benchmark_tests/001-atlas/main
```diff
@@ -0,0 +1,26 @@
+
+cvmfs_test_name="ATLAS benchmark"
+cvmfs_test_autofs_on_startup=false
+cvmfs_benchmark="yes"
+
+FQRN=atlas.cern.ch
+
+
+cvmfs_run_benchmark() {
+  export ATLAS_LOCAL_ROOT_BASE=/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase
```
jblomer: Can you check if the test runs without `ATLAS_LOCAL_ROOT_BASE` or perhaps with `local ATLAS_LOCAL_ROOT_BASE=...`.  That would avoid a possible source of environment pollution.

test/run.sh
```diff
@@ -104,6 +104,11 @@ setup_environment() {
     fi
   fi
 
+  # if the test is a benchmark we have to configure the environment
+  if [ "$cvmfs_benchmark" = "yes" ]; then
```
jblomer: That breaks if `cvmfs_benchmark` is not defined.  The trick to get around it is 

    if [ "x$cvmfs_benchmark" = "xyes" ]; then

(notice the additional x)

test/run.sh
```diff
@@ -162,6 +167,11 @@ do
          exit \$retval" >> $logfile 2>&1
   RETVAL=$?
 
+  # if the test is a benchmark we have to collect the results before removing the folder
+  if [ "$cvmfs_benchmark" = "yes" ]; then
```
jblomer: [Same as above]

test/test_functions
```diff
@@ -28,6 +28,13 @@ CVMFS_TEST_S3_CONFIG=${CVMFS_TEST_S3_CONFIG:=}
 CVMFS_TEST_STRATUM0=${CVMFS_TEST_STRATUM0:=}
 CVMFS_TEST_SERVER_CACHE=${CVMFS_TEST_SERVER_CACHE:=}
 
+# This OPT variables are for the benchmark options under the benchmark_tests folder
+CVMFS_OPT_WARM_CACHE=${CVMFS_OPT_WARM_CACHE:=yes}
+CVMFS_OPT_ITERATIONS=${CVMFS_OPT_ITERATIONS:=1}
+CVMFS_OPT_TEST_TYPE=${CVMFS_OPT_TEST_TYPE:=callgrind} # callgrind / memcheck
+CVMFS_OPT_CONFIG_FILE=${CVMFS_OPT_CONFIG_FILE:=}
+CVMFS_OPT_OUTPUT_DIR=${CVMFS_OPT_OUTPUT_DIR:=~/cvmfs_benchmarks}
```
jblomer: Perhaps it would be better to use $workspace/cvmfs_benchmarks as a default.

test/test_functions
```diff
@@ -1181,3 +1188,125 @@ cp_usrbin() {
     sudo chmod 0444 $f || return 3
   done
 }
+
+
+# waits until cvmfs is mounted
+wait_fqrn_mount() {
+  local waits=0
+  while ! mount | grep "/cvmfs/${FQRN}" > /dev/null 2>&1; do
```
jblomer: easier: `grep -q "/cvmfs/${FQRN}"`, so you don't need the redirect (q for quiet).

test/test_functions
```diff
@@ -1181,3 +1188,125 @@ cp_usrbin() {
...
+}
+
+
+# runs the command once in order to get a warm cache
+create_warm_cache() {
+  local workspace=$1
+  cvmfs2 -f -o config=cvmfs.conf,disable_watchdog,simple_options_parsing $FQRN /cvmfs/$FQRN &> /dev/null &
+  wait_fqrn_mount
+
+  bash -c "cvmfs_run_benchmark" > /dev/null 2>&1
```
jblomer: We probably do want to keep the stdout and stderr messages.  Have a look in the integration tests, at some point the output of the testcases is redirected into the logfile specified as a parameter to `run.sh`.  We'd probably like to have the same behavior for the benchmarks.

test/test_functions
```diff
@@ -1181,3 +1188,125 @@ cp_usrbin() {
...
+  return 0
+}
+
+
+# creates the default cvmfs.conf file for normal testing
+#
+# @param destination_file  absolute path of the directory
+create_default_cvmfs_config() {
+  local destination_directory="$1"
+  echo 'CVMFS_CACHE_BASE=/root/cvmfs_testing/cache
```
jblomer: that wouldn't quite work because now only root can run the benchmarks.  Hmm.. how about putting it into /var/lib/cvmfs/benchmark by default and adding an additional parameter `CVMFS_OPT_RAMCACHE` to put the cache on a ramdisk (i.e. a tmpfs mountpoint).  In order to comply with SElinux, the cache directory has to be labeled in addition with `chcon -Rv --type=cvmfs_cache_t <cachedir>`.

test/test_functions
```diff
@@ -1181,3 +1188,125 @@ cp_usrbin() {
...
+CVMFS_RELOAD_SOCKETS=/root/cvmfs_testing/cache
```
jblomer: Can be put in the cache directory as well to not bind it to root.

test/benchmark_tests/001-atlas/main
Moliholy: It's necessary to define that variable inside the test, otherwise it doesn't run properly. Concerning the pollution, for the warm cache creation it runs on purporse like `bash -c "cvmfs_run_benchmark"`, and during the normal run it's also the same by default. This way it doesn't propagate the value of the variable.


Just in case I've just tried it out with the following script:

bash -c 'export OH_GOD="ohmygod"'
echo $OH_GOD

... and it prints nothing.

test/test_functions
```diff
@@ -1181,3 +1188,125 @@ cp_usrbin() {
...
+CVMFS_SERVER_URL=http://cvmfs-stratum-one.cern.ch/cvmfs/@fqrn@
+CVMFS_SHARED_CACHE=no
+CVMFS_HTTP_PROXY=http://ca-proxy.cern.ch:3128
```
jblomer: That can be set to the `CVMFS_TEST_PROXY` defined above

test/test_functions
```diff
@@ -1181,3 +1188,125 @@ cp_usrbin() {
...
+      OPTIONS=$CVMFS_OPT_MEMCHECK
+    ;;
+  esac
+}
+
+
+# creates the environment to run the benchmark
+#
+# @param workspace  absolute path to the directory where the benchmark will put the files in
+setup_benchmark_environment() {
```
jblomer: The /etc/mtab trick could be part of the setup, too.

test/test_functions
```diff
@@ -1181,3 +1188,125 @@ cp_usrbin() {
...
+  wait $PID
+}
+
+
+# collects the files and pastes them into $CVMFS_OPT_OUTPUT_DIR/$FQRN
+collect_benchmark_results() {
+  local benchmark_output_dir="$CVMFS_OPT_OUTPUT_DIR/$FQRN"
+  local workspace=$workdir
+  rm -rf $benchmark_output_dir/*
+  cp  -rf $workdir/$FQRN.*.log $workdir/callgrind.out.* $workdir/$FQRN.memcheck.xml $workdir/cvmfs.conf $benchmark_output_dir > /dev/null 2>&1
```
jblomer: Here, a few more things can be collected.  To my mind comes `uname -a`, `cvmfs_talk -i <fqrn> internal affairs` (the statistics counters), `cvmfs_config stat <fqrn>` (some nicely prepared statistics for the user), `cvmfs2 --version`, the md5 hash of the cvmfs2 binary.

test/test_functions
Moliholy: In run.sh, line 178 (sudo rm -rf "$workdir" >> $logfile) it removes everything inside the workspace. That's the reason I wanted a place that I know it won't be deleted.

jblomer: I see.  In this case, I'd let the user decide via a `CVMFS_OPT_...` parameter which defaults to somewhere in /tmp.  I'm not sure if the tilde (~) is interpreted correctly in all cases.

Moliholy: In that case now:

CVMFS_OPT_OUTPUT_DIR=${CVMFS_OPT_OUTPUT_DIR:=/tmp/cvmfs_benchmarks}

and in any case in the benchmark log the final output is specified

test/run.sh
Moliholy: Oh, I didn't know it crashed. I take very good note, thanks!

test/test_functions
reneme: I would rather create a general parameter for the cache placement like `CVMFS_OPT_CACHE_DIR` and default it to */var/lib/cvmfs/benchmark*. That way you *can* place it on a ramdisk if you want to but its not necessary to put the actual ramdisk creation somewhere in the test system's code.

jblomer: Isn't the test system the best spot to handle the ramdisk creation?  Where else should be done?

reneme: If you frequently run it with a ramdisk, yes... only for occasional (manual) tests I would just do it manually and use the proposed `CVMFS_OPT_CACHE_DIR`. This would also be useful to put it into other places like alien caches (?), pre-populated cached, corrupted caches and what not...

jblomer: Having manual test would mean we almost never run it.  Anything that starts as a manual test and is not completely useless should become an automatic procedure.  The other cache flavors you mention should rather become part of the test suite, too (the alien cache is somewhat with the possibility to provide a custom cvmfs config file).

test/test_functions
Moliholy: Do you mean creating an empty /etc/mtab file?

jblomer: I was referring to our initial finding when dealing with igprof: if /etc/mtab is not a symlink, rename to /etc/mtab.save and create a symlink /etc/mtab --> /proc/mounts.

test/test_functions
reneme: > Having manual test would mean we almost never run it. Anything that starts as a manual test and is not completely useless should become an automatic procedure.

I couldn't agree more. However, I would add it as soon as it becomes actually necessary. Maybe in a wrapping script that takes care of the environment setup and is called by the CI? I would rather make the benchmark driver flexible but slim and allow for external configuration than extending it with environment setup functionality all the time. Just me 2 cents. ;-)

jblomer: That would be fine with me, as long as the benchmarks can be executed with a ramdisk cache from Jenkinks in one way or another.  I think it would be an important option for the benchmarks because it cancels out the uncertainties from accessing the hard disk.

Moliholy:
New commit pushed. Added full test support and fixed umounts

jblomer:
Uh, something is not right here.  This pull request comprises 474 files.  Can you have a look what happened?

Moliholy:
Totally right. By mistake there was a folder inside the alice and (specially) cms tests, which are now removed. I've re-run the tests to make sure they work properly, and they do.

reneme:
Hi José, could you please rewrite the history of your pull request with `git rebase -i` to remove these 400 unnecessary files altogether? Otherwise they will stay in the repository history forever and need to be downloaded with every `git clone`, which is bad. Thanks!

Moliholy:
Closing this one and opened a new one in https://github.com/cvmfs/cvmfs/pull/819

PR #814, opened 2015-03-09
cvmfs/cvmfs_config : Added a regularcheck() argument and function for cvmfs_fsck

tssavita:
Added a regularcheck() argument and function for cvmfs_fsck to check consistency of all the mounted files


jblomer:
Thanks!  This pull requests targets the master branch, which is reserved for stable releases.  Can you please rebase your patch on top of the devel branch and send a pull request for the devel branch.

PR #817, opened 2015-03-09
Use std:: functions to implement HasSuffix

jblomer:


cvmfs/util.cc
```diff
@@ -83,6 +83,22 @@ const signed char db64_table[] =
 
 static pthread_mutex_t getumask_mutex = PTHREAD_MUTEX_INITIALIZER;
 
+namespace {
+
+/**
+ * Used for cas  insensitive HasSuffix
+ */
+struct IgnoreCaseComperator {
+  IgnoreCaseComperator() { }
```
reneme: Is there a style guide requiring an empty c'tor?

jblomer: Not really.  But clang complain without.

PR #819, opened 2015-03-17
Test benchmark

Moliholy:
New pull request containing the same information than https://github.com/cvmfs/cvmfs/pull/813


jblomer:
Thanks, this looks already quite nice!  Let's go through a few iterations.  For the naming, just `benchmarks` would be more concise than `benchmark_tests`.  There are still a few files that can be removed (some of them come from me, sorry about that):
003-alice/ppbench/CMakeLists.txt
003-alice/ppbench/QAImageRec0.ps
003-alice/ppbench/monrec.C
003-alice/ppbench/monsim.C
Please also remove them from the history.  Probably this will require a new pull request so I'll wait with the inline comments.

Moliholy:
Concerning the "benchmark" folder, I did it this way at the beginning, but it is included in the .gitignore, so I renamed it. I don't really know if nowadays "benchmark" should be in the .gitignore, but "benchmarks" (note the final S) is not ignored, so there wouldn't be any problem.

Moliholy:
It's done. I have just amended the last commit, so now mentioned files wouldn't be there.

test/benchmarks/001-atlas/main
```diff
@@ -0,0 +1,29 @@
...
+  export ATLAS_LOCAL_ROOT_BASE=/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase
+  source ${ATLAS_LOCAL_ROOT_BASE}/user/atlasLocalSetup.sh
+  . ${ATLAS_LOCAL_ROOT_BASE}/x86_64/AtlasSetup/V00-04-09/AtlasSetup/scripts/asetup.sh 19.0.0
+  /cvmfs/atlas.cern.ch/repo/sw/software/x86_64-slc6-gcc47-opt/19.0.0/AtlasCore/19.0.0/InstallArea/share/bin/athena.py AthExHelloWorld/HelloWorldOptions.py
+}
+
+cvmfs_run_test() {
+  logfile=$1
+
+  mkdir -p /cvmfs/sft.cern.ch
```
jblomer: This requires sudo

test/benchmarks/001-atlas/main
```diff
@@ -0,0 +1,29 @@
...
+  mount -t cvmfs sft.cern.ch /cvmfs/sft.cern.ch
```
jblomer: Requires sudo

test/benchmarks/003-alice/main
```diff
@@ -0,0 +1,34 @@
...
+
+cvmfs_run_benchmark() {
+  set -e
+  source /cvmfs/alice.cern.ch/etc/login.sh
+  eval `alienv printenv AliRoot/v5-05-Rev-30`
+  cd $test_location/ppbench
+  aliroot -b -q sim.C
+  aliroot -q -b rec.C
+  aliroot -b -q ${ALICE_ROOT}/STEER/CheckESD.C
+  rm -rf *.root *.dat *.log fort* hlt hough raw* recraw/*.root recraw/*.log GRP QAImageSim0.ps
```
jblomer: In addition, QAImageRec0.ps needs to be removed.  I forgot about that.

test/benchmarks/003-alice/main
```diff
@@ -0,0 +1,34 @@
...
+  cd $test_location
+}
+
+cvmfs_run_test() {
+  logfile=$1
+  test_location=$2
+
+  mkdir -p /cvmfs/alice-ocdb.cern.ch
```
jblomer: sudo ...

test/benchmarks/003-alice/main
```diff
@@ -0,0 +1,34 @@
...
+  mount -t cvmfs alice-ocdb.cern.ch /cvmfs/alice-ocdb.cern.ch
```
jblomer: sudo ...

test/run.sh
```diff
@@ -162,6 +167,13 @@ do
          exit \$retval" >> $logfile 2>&1
   RETVAL=$?
 
+  # if the test is a benchmark we have to collect the results before removing the folder
+  if [ x"$cvmfs_benchmark" = x"yes" ]; then
+    collect_benchmark_results
+    umount /cvmfs/$FQRN > /dev/null 2>&1
```
jblomer: Can you use the umount function from the test utilities here?  The one that waits until the umount is completed?

test/test_functions
```diff
@@ -28,6 +28,14 @@ CVMFS_TEST_S3_CONFIG=${CVMFS_TEST_S3_CONFIG:=}
 CVMFS_TEST_STRATUM0=${CVMFS_TEST_STRATUM0:=}
 CVMFS_TEST_SERVER_CACHE=${CVMFS_TEST_SERVER_CACHE:=}
 
+# This OPT variables are for the benchmark options under the benchmark_tests folder
```
jblomer: benchmark_tests --> benchmarks

Moliholy:
`[root@cvm-molina ppbench]# ls -lah / | grep cvmfs`
`drwxr-xr-x   8 root root 4.0K Mar 13 17:07 cvmfs`

Directly mounting under /cvmfs requires sudo, so being superuser seems to be necessary anyway, isn't it?

test/test_functions
```diff
@@ -28,6 +28,14 @@ CVMFS_TEST_S3_CONFIG=${CVMFS_TEST_S3_CONFIG:=}
...
+CVMFS_OPT_WARM_CACHE=${CVMFS_OPT_WARM_CACHE:=yes}
+CVMFS_OPT_ITERATIONS=${CVMFS_OPT_ITERATIONS:=1}
+CVMFS_OPT_TEST_TYPE=${CVMFS_OPT_TEST_TYPE:=callgrind} # callgrind / memcheck
+CVMFS_OPT_CONFIG_FILE=${CVMFS_OPT_CONFIG_FILE:=}
+CVMFS_OPT_OUTPUT_DIR=${CVMFS_OPT_OUTPUT_DIR:=/tmp/cvmfs_benchmarks}
+CVMFS_OPT_RAMCACHE=${CVMFS_OPT_RAMCACHE:=/var/lib/cvmfs/benchmark}
```
jblomer: Perhaps better named `CVMFS_OPT_CACHEDIR`.  If it is on a ramdisk or not is another question.

test/test_functions
```diff
@@ -1181,3 +1193,195 @@ cp_usrbin() {
...
+#
+# @param workspace  absolute path to the directory where the benchmark will put the files in
+setup_benchmark_environment() {
+  local workspace="$1"
+  mkdir -p /cvmfs/$FQRN
+  mkdir -p $CVMFS_OPT_OUTPUT_DIR/$FQRN
+  mkdir -p $workspace/cache
+
+  # cleaning the cache
+  rm -rf ${CVMFS_OPT_RAMCACHE}/${FQRN}/*
```
jblomer: That is very dangerous.  CVMFS_OPT_RAMCACHE and FQRN might not be defined, in which case the whole node is removed.  Please check before if both parameters are defined.

test/test_functions
```diff
@@ -1181,3 +1193,195 @@ cp_usrbin() {
...
+
+  # mtab trick
+  if [ ! -h /etc/mtab ]; then
```
jblomer: This should probably be reverted at the end of the test in order to leave the node in a clean state.

test/test_functions
```diff
@@ -1181,3 +1193,195 @@ cp_usrbin() {
...
+  # check that the cache directory exists, is empty and is correctly labeled for SElinux
+  if [ ! -d $CVMFS_OPT_RAMCACHE ]; then
+    mkdir -p $CVMFS_OPT_RAMCACHE
+    has_selinux
+    selinux=$?
+    if [ "$selinux" -eq 0 ]; then
+      chcon -Rv --type=cvmfs_cache_t $CVMFS_OPT_RAMCACHE
+    fi
+  else
+    rm -rf $CVMFS_OPT_RAMCACHE/*
```
jblomer: Same as above, please check that CVMFS_OPT_RAMCACHE is defined.  Also, please use quotes in case there is a blank in the path.

jblomer:
Yep, the mounting requires super user privileges, just running the benchmark does not.

test/test_functions
Moliholy: In the line 1290 of the same file that checking is done. I have added quotes to "$CVMFS_OPT_RAMCACHE" (actually now renamed to "$CVMFS_OPT_CACHEDIR") just in case

jblomer: The point of checking here is to prevent a disaster even when the code gets restructured later.  This checking should be done right before every `rm -rf ${VARIABLE}/*`.  The problem of the undefined $VARIABLE is so common, it probably happened to everybody at least once.

Moliholy: Updated then:
```
    if [ -d "$CVMFS_OPT_CACHEDIR" ]; then
      rm -rf "$CVMFS_OPT_CACHEDIR"/*
    fi
```

jblomer: The problem still remains: if ${CVMFS_OPT_CACHEDIR} is empty, you will remove every file on that computer with this code (don't try).  Please test if the variable is empty, not if the directory exists.

Moliholy: Ah, now I see it. I thought it was enough with `[ -d "$CVMFS_OPT_CACHEDIR" ]`, which is supposed to return 1. Yes, it's always better to be sure

test/run.sh
Moliholy: Sure, fixed!

Moliholy:
Pull request updated

reneme:
I have a bit of a meta-comment: Please make sure to keep your commit messages short (like 50 to 80 characters) so that they easily fit into user interfaces like `tig` or here in the commit-list on GitHub. Thanks! :-)

Moliholy:
Done ;-)

test/test_functions
```diff
@@ -1181,3 +1194,210 @@ cp_usrbin() {
...
+# @param workspace  absolute path to the directory where the benchmark will put the files in
+setup_benchmark_environment() {
+  local workspace="$1"
+  mkdir -p /cvmfs/$FQRN
+  mkdir -p "$CVMFS_OPT_OUTPUT_DIR/$FQRN"
+  mkdir -p $workspace/cache
+
+  # cleaning the cache
+  local cachedir="${CVMFS_OPT_CACHEDIR}/${FQRN}"
+  if [ -d $cachedir ] && [ x$cachedir != x ]; then
```
jblomer: Please use quotes: `-d "$cachdir"` and `"x$cachedir" != x`

test/test_functions
```diff
@@ -1181,3 +1194,210 @@ cp_usrbin() {
...
+  local benchmark_output_dir="$CVMFS_OPT_OUTPUT_DIR/$FQRN"
+
+  # undo the mtab trick
+  if [ x"$CVMFS_OPT_MTAB_MODIFIED" = x"true" ]; then
+    mv /etc/mtab.save /etc/mtab
+  fi
+
+  # collect the results
+  local workspace=$workdir
+  if [ -d $benchmark_output_dir ] && [ x$benchmark_output_dir != x ]; then
```
jblomer: Same as above: `-d "$benchmark_output_dir"` and `"x$benchmark_output_dir" != x`

test/test_functions
```diff
@@ -1181,3 +1194,210 @@ cp_usrbin() {
...
+    rm -rf $benchmark_output_dir/*
```
jblomer: `"$benchmark_output_dir"/*`

Moliholy:
Added quotes

jblomer:
Quotes from Mr. Root :) Merging.

PR #820, opened 2015-03-18
Unit tests for "smalloc.h"

dimart:
Unit tests for the memory allocation wrappers in "cvmfs/smalloc.h".


jblomer:
Thanks!  That looks alright.  Looking forward to the proposal.  I'll close this particular pull request but we'll commit a combined patch of several similar pull requests.

PR #824, opened 2015-03-22
Test cases for smalloc.h

prango:
Here is my solution for the smalloc.h test case.


jblomer:
Thanks!  This currently wouldn't compile with the cvmfs sources, unfortunately.  Please take another look at the build system, and please also check the contributing guidelines once more for style and naming.

prango:
Hi,
Surely I will make the changes ! I had modified the build system locally but forgot to change it in the pull request. It might me not compiling as the last test had declaration error with int64.

jblomer:
Closing, merged with other similar PRs.

PR #826, opened 2015-03-24
Added statistics 1

Moliholy:
Here I have modified some files with the new statistics system. There are more to be modified, but I think I have enough changes to take a look at them so that the next ones are done according to what is discussed here.

### DIFF
--- a/cvmfs/cache.h
+++ b/cvmfs/cache.h
@@ -82,8 +82,8 @@ class CatalogManager : public catalog::AbstractCatalogManager {
     return result;
   }
   std::string GetCertificateStats() {
-    return "hits: " + certificate_hits_->Print() + "    " +
-    "misses: " + certificate_misses_->Print() + "\n";
+    return "hits: " + n_certificate_hits_->Print() + "    " +
+    "misses: " + n_certificate_misses_->Print() + "\n";
   }
   bool offline_mode() const { return offline_mode_; }
   uint64_t all_inodes() const { return all_inodes_; }
@@ -115,8 +115,8 @@ class CatalogManager : public catalog::AbstractCatalogManager {
   signature::SignatureManager *signature_manager_;
   download::DownloadManager *download_manager_;
   bool offline_mode_;  /**< cached copy used because there is no network */
-  perf::Counter *certificate_hits_;
-  perf::Counter *certificate_misses_;
+  perf::Counter *n_certificate_hits_;
+  perf::Counter *n_certificate_misses_;
   uint64_t all_inodes_;
   uint64_t loaded_inodes_;
   BackoffThrottle backoff_throttle_;
--- a/cvmfs/cache.h
+++ b/cvmfs/cache.h
@@ -81,10 +81,6 @@ class CatalogManager : public catalog::AbstractCatalogManager {
     Unlock();
     return result;
   }
-  std::string GetCertificateStats() {
-    return "hits: " + n_certificate_hits_->Print() + "    " +
-    "misses: " + n_certificate_misses_->Print() + "\n";
-  }
   bool offline_mode() const { return offline_mode_; }
   uint64_t all_inodes() const { return all_inodes_; }
   uint64_t loaded_inodes() const { return loaded_inodes_; }
--- a/cvmfs/cache.h
+++ b/cvmfs/cache.h
@@ -19,6 +19,7 @@
 #include "manifest_fetch.h"
 #include "shortstring.h"
 #include "signature.h"
+#include "statistics.h"
 
 namespace catalog {
 class DirectoryEntry;
@@ -68,7 +69,8 @@ class CatalogManager : public catalog::AbstractCatalogManager {
  public:
   CatalogManager(const std::string &repo_name,
                  signature::SignatureManager *signature_manager,
-                 download::DownloadManager *download_manager);
+                 download::DownloadManager *download_manager,
+                 perf::Statistics *statistics);
   virtual ~CatalogManager();
 
   bool InitFixed(const shash::Any &root_hash);
@@ -80,8 +82,8 @@ class CatalogManager : public catalog::AbstractCatalogManager {
     return result;
   }
   std::string GetCertificateStats() {
-    return "hits: " + StringifyInt(atomic_read32(&certificate_hits_)) + "    " +
-    "misses: " + StringifyInt(atomic_read32(&certificate_misses_)) + "\n";
+    return "hits: " + certificate_hits_->Print() + "    " +
+    "misses: " + certificate_misses_->Print() + "\n";
   }
   bool offline_mode() const { return offline_mode_; }
   uint64_t all_inodes() const { return all_inodes_; }
@@ -113,8 +115,8 @@ class CatalogManager : public catalog::AbstractCatalogManager {
   signature::SignatureManager *signature_manager_;
   download::DownloadManager *download_manager_;
   bool offline_mode_;  /**< cached copy used because there is no network */
-  atomic_int32 certificate_hits_;
-  atomic_int32 certificate_misses_;
+  perf::Counter *certificate_hits_;
+  perf::Counter *certificate_misses_;
   uint64_t all_inodes_;
   uint64_t loaded_inodes_;
   BackoffThrottle backoff_throttle_;
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -23,6 +23,7 @@
 #include "file_chunk.h"
 #include "hash.h"
 #include "logging.h"
+#include "statistics.h"
 #include "util.h"
 
 class XattrList;
@@ -67,38 +68,30 @@ inline const char *Code2Ascii(const LoadError error) {
 
 
 struct Statistics {
-  atomic_int64 num_lookup_inode;
-  atomic_int64 num_lookup_path;
-  atomic_int64 num_lookup_path_negative;
-  atomic_int64 num_lookup_xattrs;
-  atomic_int64 num_listing;
-  atomic_int64 num_nested_listing;
+  perf::Counter *num_lookup_inode;
+  perf::Counter *num_lookup_path;
+  perf::Counter *num_lookup_path_negative;
+  perf::Counter *num_lookup_xattrs;
+  perf::Counter *num_listing;
+  perf::Counter *num_nested_listing;
 
-  Statistics() {
-    atomic_init64(&num_lookup_inode);
-    atomic_init64(&num_lookup_path);
-    atomic_init64(&num_lookup_path_negative);
-    atomic_init64(&num_lookup_xattrs);
-    atomic_init64(&num_listing);
-    atomic_init64(&num_nested_listing);
+  explicit Statistics(perf::Statistics *statistics) {
+    num_lookup_inode = statistics->Register("catalog_mgr.n_", "");
+    num_lookup_path = statistics->Register("catalog_mgr.n_", "");;
+    num_lookup_path_negative = statistics->Register("catalog_mgr.n_", "");;
+    num_lookup_xattrs = statistics->Register("catalog_mgr.n_", "");;
+    num_listing = statistics->Register("catalog_mgr.n_", "");;
+    num_nested_listing = statistics->Register("catalog_mgr.n_", "");;
   }
 
   std::string Print() {
     return
-      "lookup(inode): " + StringifyInt(atomic_read64(&num_lookup_inode)) +
-      "    " +
-      "lookup(path-all): " + StringifyInt(atomic_read64(&num_lookup_path)) +
-      "    " +
-      "lookup(path-negative): " +
-        StringifyInt(atomic_read64(&num_lookup_path_negative)) +
-      "    " +
-      "lookup(xattrs): " +
-        StringifyInt(atomic_read64(&num_lookup_xattrs)) +
-      "    " +
-      "listing: " + StringifyInt(atomic_read64(&num_listing)) +
-      "    " +
-      "listing nested catalogs: " +
-        StringifyInt(atomic_read64(&num_nested_listing)) + "\n";
+      "lookup(inode): " + num_lookup_inode->Print() + "    " +
+      "lookup(path-all): " + num_lookup_path->Print() + "    " +
+      "lookup(path-negative): " + num_lookup_path_negative->Print() + "    " +
+      "lookup(xattrs): " + num_lookup_xattrs->Print() + "    " +
+      "listing: " + num_listing->Print() + "    " +
+      "listing nested catalogs: " + num_nested_listing->Print() + "\n";
   }
 };
 
@@ -157,7 +150,7 @@ class RemountListener {
 class AbstractCatalogManager : public SingleCopy {
  public:
   static const inode_t kInodeOffset = 255;
-  AbstractCatalogManager();
+  explicit AbstractCatalogManager(perf::Statistics *statistics);
   virtual ~AbstractCatalogManager();
 
   void SetInodeAnnotation(InodeAnnotation *new_annotation);
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -68,30 +68,37 @@ inline const char *Code2Ascii(const LoadError error) {
 
 
 struct Statistics {
-  perf::Counter *num_lookup_inode;
-  perf::Counter *num_lookup_path;
-  perf::Counter *num_lookup_path_negative;
-  perf::Counter *num_lookup_xattrs;
-  perf::Counter *num_listing;
-  perf::Counter *num_nested_listing;
+  perf::Counter *n_lookup_inode;
+  perf::Counter *n_lookup_path;
+  perf::Counter *n_lookup_path_negative;
+  perf::Counter *n_lookup_xattrs;
+  perf::Counter *n_listing;
+  perf::Counter *n_nested_listing;
 
   explicit Statistics(perf::Statistics *statistics) {
-    num_lookup_inode = statistics->Register("catalog_mgr.n_", "");
-    num_lookup_path = statistics->Register("catalog_mgr.n_", "");;
-    num_lookup_path_negative = statistics->Register("catalog_mgr.n_", "");;
-    num_lookup_xattrs = statistics->Register("catalog_mgr.n_", "");;
-    num_listing = statistics->Register("catalog_mgr.n_", "");;
-    num_nested_listing = statistics->Register("catalog_mgr.n_", "");;
+    n_lookup_inode = statistics->Register("catalog_mgr.n_lookup_inode",
+        "Number of inode lookups");
+    n_lookup_path = statistics->Register("catalog_mgr.n_lookup_path",
+        "Number of path lookups");
+    n_lookup_path_negative = statistics->Register(
+        "catalog_mgr.n_lookup_path_negative",
+        "Number of negative path lookups");
+    n_lookup_xattrs = statistics->Register("catalog_mgr.n_lookup_xattrs",
+        "Number of xattrs lookups");
+    n_listing = statistics->Register("catalog_mgr.n_listing",
+        "Number of listing");
+    n_nested_listing = statistics->Register("catalog_mgr.n_nested_listing",
+        "Number of nested listing");
   }
 
   std::string Print() {
     return
-      "lookup(inode): " + num_lookup_inode->Print() + "    " +
-      "lookup(path-all): " + num_lookup_path->Print() + "    " +
-      "lookup(path-negative): " + num_lookup_path_negative->Print() + "    " +
-      "lookup(xattrs): " + num_lookup_xattrs->Print() + "    " +
-      "listing: " + num_listing->Print() + "    " +
-      "listing nested catalogs: " + num_nested_listing->Print() + "\n";
+      "lookup(inode): " + n_lookup_inode->Print() + "    " +
+      "lookup(path-all): " + n_lookup_path->Print() + "    " +
+      "lookup(path-negative): " + n_lookup_path_negative->Print() + "    " +
+      "lookup(xattrs): " + n_lookup_xattrs->Print() + "    " +
+      "listing: " + n_listing->Print() + "    " +
+      "listing nested catalogs: " + n_nested_listing->Print() + "\n";
   }
 };
 
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -86,19 +86,9 @@ struct Statistics {
     n_lookup_xattrs = statistics->Register("catalog_mgr.n_lookup_xattrs",
         "Number of xattrs lookups");
     n_listing = statistics->Register("catalog_mgr.n_listing",
-        "Number of listing");
+        "Number of listings");
     n_nested_listing = statistics->Register("catalog_mgr.n_nested_listing",
-        "Number of nested listing");
-  }
-
-  std::string Print() {
-    return
-      "lookup(inode): " + n_lookup_inode->Print() + "    " +
-      "lookup(path-all): " + n_lookup_path->Print() + "    " +
-      "lookup(path-negative): " + n_lookup_path_negative->Print() + "    " +
-      "lookup(xattrs): " + n_lookup_xattrs->Print() + "    " +
-      "listing: " + n_listing->Print() + "    " +
-      "listing nested catalogs: " + n_nested_listing->Print() + "\n";
+        "Number of listings of nested catalogs");
   }
 };
 
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1501,11 +1501,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   } else if (attr == "user.maxfd") {
     attribute_value = StringifyInt(max_open_files_ - kNumReservedFd);
   } else if (attr == "user.usedfd") {
-    attribute_value = no_open_files_->Print();
+    attribute_value = no_open_files_->ToString();
   } else if (attr == "user.useddirp") {
-    attribute_value = no_open_dirs_->Print();
+    attribute_value = no_open_dirs_->ToString();
   } else if (attr == "user.nioerr") {
-    attribute_value = n_io_error_->Print();
+    attribute_value = n_io_error_->ToString();
   } else if (attr == "user.proxy") {
     vector< vector<download::DownloadManager::ProxyInfo> > proxy_chain;
     unsigned current_group;
@@ -1547,9 +1547,9 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     const int num_catalogs = catalog_manager_->GetNumCatalogs();
     attribute_value = StringifyInt(num_catalogs);
   } else if (attr == "user.nopen") {
-    attribute_value = n_fs_open_->Print();
+    attribute_value = n_fs_open_->ToString();
   } else if (attr == "user.ndiropen") {
-    attribute_value = n_fs_dir_open_->Print();
+    attribute_value = n_fs_dir_open_->ToString();
   } else if (attr == "user.ndownload") {
     attribute_value = StringifyInt(cache::GetNumDownloads());
   } else if (attr == "user.timeout") {
@@ -1561,11 +1561,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     download_manager_->GetTimeout(&seconds, &seconds_direct);
     attribute_value = StringifyInt(seconds_direct);
   } else if (attr == "user.rx") {
-    int64_t rx = download_manager_->GetCounters().no_transferred_bytes->Get();
+    int64_t rx = cvmfs::statistics_->Lookup("n_transfered_bytes")->Get();
     attribute_value = StringifyInt(rx/1024);
   } else if (attr == "user.speed") {
-    int64_t rx = download_manager_->GetCounters().no_transferred_bytes->Get();
-    int64_t time = download_manager_->GetCounters().sz_transfer_time->Get();
+    int64_t rx = cvmfs::statistics_->Lookup("n_transfered_bytes")->Get();
+    int64_t time = cvmfs::statistics_->Lookup("sz_transfer_time")->Get();
     if (time == 0)
       attribute_value = "n/a";
     else
@@ -2095,9 +2095,9 @@ static int Init(const loader::LoaderExports *loader_exports) {
 
   // Runtime counters
   cvmfs::n_fs_open_ = cvmfs::statistics_->Register("cvmfs.n_fs_open",
-      "Number of opened files");
+      "Overall number of file open operations");
   cvmfs::n_fs_dir_open_ = cvmfs::statistics_->Register("cvmfs.n_fs_dir_open",
-      "Number of opened directories");
+      "Overall number of directory open operations");
   cvmfs::n_fs_lookup_ = cvmfs::statistics_->Register("cvmfs.n_fs_lookup",
       "Number of lookups");
   cvmfs::n_fs_lookup_negative_ = cvmfs::statistics_->Register(
@@ -2109,7 +2109,7 @@ static int Init(const loader::LoaderExports *loader_exports) {
   cvmfs::n_fs_readlink_ = cvmfs::statistics_->Register("cvmfs.n_fs_readlink",
       "Number of links read");
   cvmfs::n_fs_forget_ = cvmfs::statistics_->Register("cvmfs.n_fs_forget",
-      "Number of forgets");
+      "Number of inode forgets");
   cvmfs::n_io_error_ = cvmfs::statistics_->Register("cvmfs.n_io_error",
       "Number of I/O errors");
 
@@ -2280,9 +2280,9 @@ static int Init(const loader::LoaderExports *loader_exports) {
   }
   cvmfs::max_open_files_ = monitor::GetMaxOpenFiles();
   cvmfs::no_open_files_ = cvmfs::statistics_->Register("cvmfs.no_open_files",
-      "Total number of opened files");
+      "Number of currently opened files");
   cvmfs::no_open_dirs_ = cvmfs::statistics_->Register("cvmfs.no_open_dirs",
-      "Total number of opened directories");
+      "Number of currently opened directories");
 
   // Control & command interface
   if (!talk::Init(".", cvmfs::options_manager_)) {
@@ -2628,8 +2628,7 @@ static int AltProcessFlavor(int argc, char **argv) {
     return quota::MainCacheManager(argc, argv);
   }
   if (strcmp(argv[1], "__wpad__") == 0) {
-    return download::MainResolveProxyDescription(argc, argv,
-        cvmfs::statistics_);
+    return download::MainResolveProxyDescription(argc, argv);
   }
   return 1;
 }
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -223,19 +223,19 @@ uint64_t next_directory_handle_ = 0;
 ChunkTables *chunk_tables_;
 
 perf::Statistics *statistics_ = NULL;
-perf::Counter *num_fs_open_ = NULL;
-perf::Counter *num_fs_dir_open_ = NULL;
-perf::Counter *num_fs_lookup_ = NULL;
-perf::Counter *num_fs_lookup_negative_ = NULL;
-perf::Counter *num_fs_stat_ = NULL;
-perf::Counter *num_fs_read_ = NULL;
-perf::Counter *num_fs_readlink_ = NULL;
-perf::Counter *num_fs_forget_ = NULL;
-perf::Counter *num_io_error_ = NULL;
+perf::Counter *no_fs_open_ = NULL;
+perf::Counter *no_fs_dir_open_ = NULL;
+perf::Counter *no_fs_lookup_ = NULL;
+perf::Counter *no_fs_lookup_negative_ = NULL;
+perf::Counter *no_fs_stat_ = NULL;
+perf::Counter *no_fs_read_ = NULL;
+perf::Counter *no_fs_readlink_ = NULL;
+perf::Counter *no_fs_forget_ = NULL;
+perf::Counter *no_io_error_ = NULL;
 /**< number of currently open files by Fuse calls */
-perf::Counter *open_files_ = NULL;
+perf::Counter *no_open_files_ = NULL;
 /**< number of currently open directories */
-perf::Counter *open_dirs_ = NULL;
+perf::Counter *no_open_dirs_ = NULL;
 unsigned max_open_files_; /**< maximum allowed number of open files */
 /**
  * Number of reserved file descriptors for internal use
@@ -326,7 +326,7 @@ std::string GetOpenCatalogs() {
 
 
 void ResetErrorCounters() {
-  num_io_error_->Set(0);
+  no_io_error_->Set(0);
 }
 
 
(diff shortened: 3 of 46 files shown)


jblomer:
Alright, looking through it.  Please have a look at cpplint errors first.

jblomer:
I think it would make sense to make the variable names consistent with the counter name.  So for instance the `certificate_hits_` counter variable would become `n_certificate_hits_`.

cvmfs/cvmfs.cc
```diff
@@ -2105,15 +2104,24 @@ static int Init(const loader::LoaderExports *loader_exports) {
...
-  atomic_init64(&cvmfs::num_fs_stat_);
-  atomic_init64(&cvmfs::num_fs_read_);
-  atomic_init64(&cvmfs::num_fs_readlink_);
-  atomic_init64(&cvmfs::num_fs_forget_);
-  atomic_init32(&cvmfs::num_io_error_);
+  cvmfs::num_fs_open_ = cvmfs::statistics_->Register("cvmfs.no_fs_open",
+      "Number of opened files");
+  cvmfs::num_fs_dir_open_ = cvmfs::statistics_->Register("cvmfs.no_fs_dir_open",
+      "Number of opened directories");
+  cvmfs::num_fs_lookup_ = cvmfs::statistics_->Register("cvmfs.no_fs_loopup",
```
jblomer: That would be `n_fs_lookup`, same for the other counters below (ever growing counters)

cvmfs/cvmfs.cc
```diff
@@ -2626,7 +2637,7 @@ static int AltProcessFlavor(int argc, char **argv) {
     return quota::MainCacheManager(argc, argv);
   }
   if (strcmp(argv[1], "__wpad__") == 0) {
-    return download::MainResolveProxyDescription(argc, argv);
+    return download::MainResolveProxyDescription(argc, argv, cvmfs::statistics_);
```
jblomer: Here we should create the statistics class in MainResolveProxyDescription.  This is one of the "special modes" in which the client can be started.  In this particular special mode, it doesn't act as a fuse module but only resolves the proxy definition and prints the result.

cvmfs/cvmfs.cc
```diff
@@ -2811,8 +2822,8 @@ static bool RestoreState(const int fd_progress,
 
     if (saved_states[i]->state_id == loader::kStateOpenFilesCounter) {
       SendMsg2Socket(fd_progress, "Restoring open files counter... ");
-      cvmfs::open_files_ = *(reinterpret_cast<uint32_t *>(
-        saved_states[i]->state));
+      cvmfs::open_files_->Set(*(reinterpret_cast<uint64_t *>(
```
jblomer: Ah, here we have a slightly bigger problem... let's discuss.

Moliholy:
Added the rest of the files to the new statistics system

cvmfs/download.cc
```diff
@@ -422,8 +422,8 @@ void *DownloadManager::MainDownload(void *data) {
     } else {
       timeout = -1;
       gettimeofday(&timeval_stop, NULL);
-      download_mgr->statistics_->transfer_time +=
-        DiffTimeSeconds(timeval_start, timeval_stop);
+      perf::Xadd(download_mgr->counters_->sz_transfer_time,
+        1000 * DiffTimeSeconds(timeval_start, timeval_stop));
```
Moliholy: Now we are counting miliseconds, not seconds. This is because now the counter is not a double, but a long, so in order be more accurate I had to change it

cvmfs/download.cc
```diff
@@ -1516,10 +1517,10 @@ Failures DownloadManager::Fetch(JobInfo *info) {
     int retval;
     do {
       retval = curl_easy_perform(handle);
-      statistics_->num_requests++;
+      perf::Inc(counters_->n_num_requests);
```
Moliholy: In principle this operation was not atomic, but now it is

cvmfs/download.cc
```diff
@@ -1516,10 +1517,10 @@ Failures DownloadManager::Fetch(JobInfo *info) {
...
       double elapsed;
       if (curl_easy_getinfo(handle, CURLINFO_TOTAL_TIME, &elapsed) == CURLE_OK)
-        statistics_->transfer_time += elapsed;
+        perf::Xadd(counters_->sz_transfer_time, (int64_t)(elapsed * 1000));
```
Moliholy: Mandatory cast of a double into a int64_t to get the miliseconds

Moliholy:
Sorry, I forgot to add the catalog_mgr_*** statistics. Now it is fixed.

jblomer:
Before every pull request, please make a cpplint run

jblomer:
The t_lru.cc unit test needs to be adapted to the new LruCache interface.

Moliholy:
Fixed talk.cc to not to print old statistics.
Fixed all unittests, making them compatible with the new statistics. I have just run the 475 tests and they all successfully passed.
I have also tried the program out (only basic operations, nothing big) and I found no crashes.

cvmfs/cache.h
```diff
@@ -80,8 +82,8 @@ class CatalogManager : public catalog::AbstractCatalogManager {
     return result;
   }
   std::string GetCertificateStats() {
```
jblomer: This function can be removed I guess.

cvmfs/catalog_mgr.h
```diff
@@ -67,38 +68,37 @@ inline const char *Code2Ascii(const LoadError error) {
...
+        "Number of inode lookups");
+    n_lookup_path = statistics->Register("catalog_mgr.n_lookup_path",
+        "Number of path lookups");
+    n_lookup_path_negative = statistics->Register(
+        "catalog_mgr.n_lookup_path_negative",
+        "Number of negative path lookups");
+    n_lookup_xattrs = statistics->Register("catalog_mgr.n_lookup_xattrs",
+        "Number of xattrs lookups");
+    n_listing = statistics->Register("catalog_mgr.n_listing",
+        "Number of listing");
```
jblomer: "Number of listing" --> "Number of listings"

cvmfs/catalog_mgr.h
```diff
@@ -67,38 +68,37 @@ inline const char *Code2Ascii(const LoadError error) {
...
+    n_nested_listing = statistics->Register("catalog_mgr.n_nested_listing",
+        "Number of nested listing");
```
jblomer: "Number of nested listing" --> Number of listings of nested catalogs

cvmfs/catalog_mgr.h
```diff
@@ -67,38 +68,37 @@ inline const char *Code2Ascii(const LoadError error) {
...
   }
 
   std::string Print() {
```
jblomer: This function can be removed

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
-atomic_int64 num_fs_read_;
-atomic_int64 num_fs_readlink_;
-atomic_int64 num_fs_forget_;
-atomic_int32 num_io_error_;
-atomic_int32 open_files_; /**< number of currently open files by Fuse calls */
-atomic_int32 open_dirs_; /**< number of currently open directories */
+perf::Statistics *statistics_ = NULL;
+perf::Counter *no_fs_open_ = NULL;
+perf::Counter *no_fs_dir_open_ = NULL;
+perf::Counter *no_fs_lookup_ = NULL;
```
jblomer: `no_fs_lookup` --> `n_fs_lookup`

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *no_fs_lookup_negative_ = NULL;
```
jblomer: `no_fs_lookup_negative_` --> `n_fs_lookup_negative_`

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *no_fs_stat_ = NULL;
```
jblomer: `no_fs_stat_` --> `n_fs_stat_`

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *no_fs_read_ = NULL;
```
jblomer: `no_fs_read_` --> `n_fs_read`

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *no_fs_readlink_ = NULL;
```
jblomer: `no_fs_readlink_` --> `n_fs_readlink`

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *no_fs_forget_ = NULL;
```
jblomer: `no_fs_forget_` --> `n_fs_forget_`

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *no_io_error_ = NULL;
```
jblomer: `no_io_error_` --> n_io_error_

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
-atomic_int64 num_fs_stat_;
-atomic_int64 num_fs_read_;
-atomic_int64 num_fs_readlink_;
-atomic_int64 num_fs_forget_;
-atomic_int32 num_io_error_;
-atomic_int32 open_files_; /**< number of currently open files by Fuse calls */
-atomic_int32 open_dirs_; /**< number of currently open directories */
+perf::Statistics *statistics_ = NULL;
+perf::Counter *no_fs_open_ = NULL;
+perf::Counter *no_fs_dir_open_ = NULL;
```
jblomer: `no_fs_dir_open_` --> `n_fs_dir_open_`

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
-atomic_int64 num_fs_lookup_negative_;
-atomic_int64 num_fs_stat_;
-atomic_int64 num_fs_read_;
-atomic_int64 num_fs_readlink_;
-atomic_int64 num_fs_forget_;
-atomic_int32 num_io_error_;
-atomic_int32 open_files_; /**< number of currently open files by Fuse calls */
-atomic_int32 open_dirs_; /**< number of currently open directories */
+perf::Statistics *statistics_ = NULL;
+perf::Counter *no_fs_open_ = NULL;
```
jblomer: `no_fs_open_` --> `n_fs_open_`

Moliholy:
All current fixes applied

cvmfs/catalog_mgr.h
```diff
@@ -67,38 +68,27 @@ inline const char *Code2Ascii(const LoadError error) {
...
+        "Number of path lookups");
+    n_lookup_path_negative = statistics->Register(
+        "catalog_mgr.n_lookup_path_negative",
+        "Number of negative path lookups");
+    n_lookup_xattrs = statistics->Register("catalog_mgr.n_lookup_xattrs",
+        "Number of xattrs lookups");
+    n_listing = statistics->Register("catalog_mgr.n_listing",
+        "Number of listings");
+    n_nested_listing = statistics->Register("catalog_mgr.n_nested_listing",
+        "Number of nested listings");
```
jblomer: Description: number of listings of nested catalogs

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *n_fs_open_ = NULL;
+perf::Counter *n_fs_dir_open_ = NULL;
+perf::Counter *n_fs_lookup_ = NULL;
+perf::Counter *n_fs_lookup_negative_ = NULL;
+perf::Counter *n_fs_stat_ = NULL;
+perf::Counter *n_fs_read_ = NULL;
+perf::Counter *n_fs_readlink_ = NULL;
+perf::Counter *n_fs_forget_ = NULL;
+perf::Counter *n_io_error_ = NULL;
+/**< number of currently open files by Fuse calls */
```
jblomer: I think the Doxygen `/**<` syntax only works for comments that are in the same line.  For comments in front of a code line, the

    /**
     * comment
    */

syntax should be used.

cvmfs/cvmfs.cc
```diff
@@ -222,18 +222,20 @@ uint64_t next_directory_handle_ = 0;
...
+perf::Counter *no_open_files_ = NULL;
+/**< number of currently open directories */
```
jblomer: Same as above

cvmfs/cvmfs.cc
```diff
@@ -339,12 +341,12 @@ static bool UseWatchdog() {
 }
 
 
-void GetLruStatistics(lru::Statistics *inode_stats, lru::Statistics *path_stats,
-                      lru::Statistics *md5path_stats)
+void GetLruStatistics(lru::Counters *inode_stats, lru::Counters *path_stats,
```
jblomer: This function can be removed

cvmfs/cvmfs.cc
```diff
@@ -368,20 +370,15 @@ catalog::Statistics GetCatalogStatistics() {
   return catalog_manager_->statistics();
 }
 
-string GetCertificateStats() {
-  return catalog_manager_->GetCertificateStats();
-}
-
 string GetFsStats() {
```
jblomer: This function can be removed

cvmfs/cvmfs.cc
```diff
@@ -1070,7 +1067,7 @@ static void cvmfs_open(fuse_req_t req, fuse_ino_t ino,
     return;
   }
 
-  atomic_inc64(&num_fs_open_);  // Count actual open / fetch operations
+  perf::Inc(n_fs_open_);;  // Count actual open / fetch operations
```
jblomer: double semicolon

cvmfs/cache.h
Moliholy: Removed, even though it's not shown here

cvmfs/cvmfs.cc
```diff
@@ -1520,11 +1517,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   } else if (attr == "user.maxfd") {
     attribute_value = StringifyInt(max_open_files_ - kNumReservedFd);
   } else if (attr == "user.usedfd") {
-    attribute_value = StringifyInt(atomic_read32(&open_files_));
+    attribute_value = no_open_files_->Print();
```
jblomer: At the moment, `Print()` does the right thing.  However, `Print()` in cvmfs is used for human-readable output, so the exact format of the string might change in the future.  Please add a function `ToString()` to the `perf::Counter` class, which transforms the number into a string.

cvmfs/cvmfs.cc
```diff
@@ -1520,11 +1517,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
   } else if (attr == "user.useddirp") {
-    attribute_value = StringifyInt(atomic_read32(&open_dirs_));
+    attribute_value = no_open_dirs_->Print();
```
jblomer: See above.

cvmfs/cvmfs.cc
```diff
@@ -1520,11 +1517,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
   } else if (attr == "user.nioerr") {
-    attribute_value = StringifyInt(atomic_read32(&num_io_error_));
+    attribute_value = n_io_error_->Print();
```
jblomer: See above.

cvmfs/cvmfs.cc
```diff
@@ -1566,9 +1563,9 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     const int num_catalogs = catalog_manager_->GetNumCatalogs();
     attribute_value = StringifyInt(num_catalogs);
   } else if (attr == "user.nopen") {
-    attribute_value = StringifyInt(atomic_read64(&num_fs_open_));
+    attribute_value = n_fs_open_->Print();
```
jblomer: See above

cvmfs/cvmfs.cc
```diff
@@ -1566,9 +1563,9 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
   } else if (attr == "user.ndiropen") {
-    attribute_value = StringifyInt(atomic_read64(&num_fs_dir_open_));
+    attribute_value = n_fs_dir_open_->Print();
```
jblomer: See above

cvmfs/cvmfs.cc
```diff
@@ -1580,11 +1577,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     download_manager_->GetTimeout(&seconds, &seconds_direct);
     attribute_value = StringifyInt(seconds_direct);
   } else if (attr == "user.rx") {
-    int64_t rx = uint64_t(download_manager_->GetStatistics().transferred_bytes);
+    int64_t rx = download_manager_->GetCounters().no_transferred_bytes->Get();
```
jblomer: This can be replaced by `Statistics->Lookup("n_transferred_bytes")`, in order to get rid of the `GetCounters()` method.

cvmfs/cvmfs.cc
```diff
@@ -1580,11 +1577,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
     attribute_value = StringifyInt(rx/1024);
   } else if (attr == "user.speed") {
-    int64_t rx = uint64_t(download_manager_->GetStatistics().transferred_bytes);
-    int64_t time = uint64_t(download_manager_->GetStatistics().transfer_time);
+    int64_t rx = download_manager_->GetCounters().no_transferred_bytes->Get();
```
jblomer: See above.

cvmfs/cvmfs.cc
```diff
@@ -1580,11 +1577,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+    int64_t time = download_manager_->GetCounters().sz_transfer_time->Get();
```
jblomer: See above.

cvmfs/cvmfs.cc
```diff
@@ -2105,15 +2110,24 @@ static int Init(const loader::LoaderExports *loader_exports) {
...
-  atomic_init64(&cvmfs::num_fs_dir_open_);
-  atomic_init64(&cvmfs::num_fs_lookup_);
-  atomic_init64(&cvmfs::num_fs_lookup_negative_);
-  atomic_init64(&cvmfs::num_fs_stat_);
-  atomic_init64(&cvmfs::num_fs_read_);
-  atomic_init64(&cvmfs::num_fs_readlink_);
-  atomic_init64(&cvmfs::num_fs_forget_);
-  atomic_init32(&cvmfs::num_io_error_);
+  cvmfs::n_fs_open_ = cvmfs::statistics_->Register("cvmfs.n_fs_open",
+      "Number of opened files");
```
jblomer: Description: Overall number of file open operations

cvmfs/cvmfs.cc
```diff
@@ -2105,15 +2110,24 @@ static int Init(const loader::LoaderExports *loader_exports) {
...
+  cvmfs::n_fs_dir_open_ = cvmfs::statistics_->Register("cvmfs.n_fs_dir_open",
+      "Number of opened directories");
```
jblomer: Description: Overall number of directory open operations

cvmfs/cvmfs.cc
```diff
@@ -2105,15 +2110,24 @@ static int Init(const loader::LoaderExports *loader_exports) {
...
+  cvmfs::n_fs_lookup_negative_ = cvmfs::statistics_->Register(
+      "cvmfs.n_fs_lookup_negative", "Number of negative lookups");
+  cvmfs::n_fs_stat_ = cvmfs::statistics_->Register("cvmfs.n_fs_stat",
+      "Number of stats");
+  cvmfs::n_fs_read_ = cvmfs::statistics_->Register("cvmfs.n_fs_read",
+      "Number of files read");
+  cvmfs::n_fs_readlink_ = cvmfs::statistics_->Register("cvmfs.n_fs_readlink",
+      "Number of links read");
+  cvmfs::n_fs_forget_ = cvmfs::statistics_->Register("cvmfs.n_fs_forget",
+      "Number of forgets");
```
jblomer: Description: number of inode forgets

cvmfs/cvmfs.cc
```diff
@@ -2281,8 +2295,10 @@ static int Init(const loader::LoaderExports *loader_exports) {
     g_monitor_ready = true;
   }
   cvmfs::max_open_files_ = monitor::GetMaxOpenFiles();
-  atomic_init32(&cvmfs::open_files_);
-  atomic_init32(&cvmfs::open_dirs_);
+  cvmfs::no_open_files_ = cvmfs::statistics_->Register("cvmfs.no_open_files",
+      "Total number of opened files");
```
jblomer: Description: Number of currently open files

cvmfs/cvmfs.cc
```diff
@@ -2281,8 +2295,10 @@ static int Init(const loader::LoaderExports *loader_exports) {
...
+  cvmfs::no_open_dirs_ = cvmfs::statistics_->Register("cvmfs.no_open_dirs",
+      "Total number of opened directories");
```
jblomer: Description: number of currently open directories

cvmfs/cvmfs.cc
```diff
@@ -2626,7 +2644,8 @@ static int AltProcessFlavor(int argc, char **argv) {
     return quota::MainCacheManager(argc, argv);
   }
   if (strcmp(argv[1], "__wpad__") == 0) {
-    return download::MainResolveProxyDescription(argc, argv);
+    return download::MainResolveProxyDescription(argc, argv,
+        cvmfs::statistics_);
```
jblomer: This is a special mode of operation.  The Statistics object should be created on the fly in `download::MainResolveProxyDescription()`

cvmfs/cvmfs.h
```diff
@@ -46,8 +46,8 @@ std::string GetOpenCatalogs();
 unsigned GetMaxTTL();  // in minutes
 void SetMaxTTL(const unsigned value);  // in minutes
 void ResetErrorCounters();
-void GetLruStatistics(lru::Statistics *inode_stats, lru::Statistics *path_stats,
-                      lru::Statistics *md5path_stats);
+void GetLruStatistics(lru::Counters *inode_stats, lru::Counters *path_stats,
```
jblomer: Not needed anymore

cvmfs/download.h
```diff
@@ -83,25 +84,31 @@ inline const char *Code2Ascii(const Failures error) {
...
-    transferred_bytes = 0.0;
-    transfer_time = 0.0;
-    num_requests = 0;
-    num_retries = 0;
-    num_proxy_failover = 0;
-    num_host_failover = 0;
+struct Counters {
+  perf::Counter *no_transferred_bytes;
+  perf::Counter *sz_transfer_time;  // measured in miliseconds
+  perf::Counter *n_num_requests;
```
jblomer: `n_num_requests` --> `n_requests`

cvmfs/download.h
```diff
@@ -83,25 +84,31 @@ inline const char *Code2Ascii(const Failures error) {
...
+  perf::Counter *n_num_retries;
```
jblomer: `n_num_retries` --> `n_retries`

cvmfs/download.h
```diff
@@ -83,25 +84,31 @@ inline const char *Code2Ascii(const Failures error) {
...
-
-  Statistics() {
-    transferred_bytes = 0.0;
-    transfer_time = 0.0;
-    num_requests = 0;
-    num_retries = 0;
-    num_proxy_failover = 0;
-    num_host_failover = 0;
+struct Counters {
+  perf::Counter *no_transferred_bytes;
```
jblomer: `no_transferred_bytes` --> `sz_transferred`

cvmfs/download.h
```diff
@@ -83,25 +84,31 @@ inline const char *Code2Ascii(const Failures error) {
...
+        "Number of requests");
+    n_num_retries = statistics->Register("download.n_num_retries",
+        "Number of retries");
+    n_proxy_failover = statistics->Register("download.n_num_proxy_failover",
+        "Number of proxy failovers");
+    n_host_failover = statistics->Register("download.n_num_host_failover",
+        "Number of host failovers");
   }
 
   std::string Print() const;
```
jblomer: Shouldn't be required anymore

cvmfs/download.h
```diff
@@ -301,7 +309,7 @@ class DownloadManager {
   void SetTimeout(const unsigned seconds_proxy, const unsigned seconds_direct);
   void GetTimeout(unsigned *seconds_proxy, unsigned *seconds_direct);
   void SetLowSpeedLimit(const unsigned low_speed_limit);
-  const Statistics &GetStatistics();
+  const Counters &GetCounters();
```
jblomer: Not needed anymore

cvmfs/lru.h
```diff
@@ -71,49 +72,56 @@ namespace lru {
...
  * Counting of cache operations.
  */
-struct Statistics {
-  int64_t size;
-  atomic_int64 num_hit;
-  atomic_int64 num_miss;
-  atomic_int64 num_insert;
-  atomic_int64 num_insert_negative;
+struct Counters {
+  perf::Counter *size;
```
jblomer: `size` --> `sz_size`

cvmfs/lru.h
```diff
@@ -71,49 +72,56 @@ namespace lru {
...
-  Statistics() {
-    size = 0;
+  perf::Counter *n_update;
+  perf::Counter *n_replace;
+  perf::Counter *n_forget;
+  perf::Counter *n_drop;
+  perf::Counter *sz_allocated;
+
+  Counters(perf::Statistics *statistics, const std::string &name) {
+    size = statistics->Register(name + ".no_size", "Size for " + name);
```
jblomer: Name: `no_size` --> `sz_size`

cvmfs/lru.h
```diff
@@ -71,49 +72,56 @@ namespace lru {
...
+        "Number of replaces for " + name);
+    n_forget = statistics->Register(name + ".n_forget",
+        "Number of forgets for " + name);
+    n_drop = statistics->Register(name + ".n_drop",
+        "Number of drops for " + name);
+    sz_allocated = statistics->Register(name + ".sz_allocated",
+        "Number of allocated bytes for " + name);
   }
 
   std::string Print() {
```
jblomer: Not needed anymore

cvmfs/nfs_shared_maps.cc
```diff
@@ -206,20 +208,17 @@ bool GetPath(const uint64_t inode, PathString *path) {
   path->Assign(raw_path, strlen(raw_path));
   sqlite3_reset(stmt_get_path_);
   pthread_mutex_unlock(&lock_);
-  atomic_inc64(&dbstat_inode_found_);
+  perf::Inc(n_dbstat_inode_found_);
   return true;
 }
 
 
 string GetStatistics() {
```
jblomer: Not needed anymore

cvmfs/shortstring.h
```diff
@@ -175,15 +176,34 @@ class ShortString {
     return ShortString(this->GetChars() + start_at, length-start_at);
   }
 
-  static uint64_t num_instances() { return atomic_read64(&num_instances_); }
-  static uint64_t num_overflows() { return atomic_read64(&num_overflows_); }
+  static void RegisterCounters(perf::Statistics *statistics) {
+    if (StackSize == kDefaultMaxPath && Type == 0) {
```
jblomer: Checking on `Type` should be sufficient

cvmfs/shortstring.h
```diff
@@ -175,15 +176,34 @@ class ShortString {
     return ShortString(this->GetChars() + start_at, length-start_at);
   }
 
-  static uint64_t num_instances() { return atomic_read64(&num_instances_); }
-  static uint64_t num_overflows() { return atomic_read64(&num_overflows_); }
+  static void RegisterCounters(perf::Statistics *statistics) {
```
jblomer: Hm, that's not too nice yet.  If one forgets to use `RegisterCounters`, it will result in a crash.  I'll have to think about it.

jblomer: Let's try like this: add a function `perf::TryInc` that will only increase the counter if it is not `NULL`.  This new function should be used instead of `perf::Inc` in the `ShortString` class (but only there).

cvmfs/statistics.cc
```diff
@@ -7,12 +7,40 @@
...
 
 using namespace std;  // NOLINT
 
+#ifdef CVMFS_NAMESPACE_GUARD
+namespace CVMFS_NAMESPACE_GUARD {
+#endif
+
 namespace perf {
 
+Counter::Counter() { atomic_init64(&counter_); }
```
jblomer: This needs to go back to the header, otherwise it can't be inlined.

cvmfs/statistics.cc
```diff
@@ -7,12 +7,40 @@
...
+void Counter::Inc() { atomic_inc64(&counter_); }
```
jblomer: See above.

cvmfs/statistics.cc
```diff
@@ -7,12 +7,40 @@
...
+void Counter::Dec() { atomic_dec64(&counter_); }
```
jblomer: See above

cvmfs/statistics.cc
```diff
@@ -7,12 +7,40 @@
...
+int64_t Counter::Get() { return atomic_read64(&counter_); }
```
jblomer: See above

cvmfs/statistics.cc
```diff
@@ -7,12 +7,40 @@
...
+void Counter::Set(const int64_t val) { atomic_write64(&counter_, val); }
```
jblomer: See above

cvmfs/statistics.cc
```diff
@@ -7,12 +7,40 @@
...
+int64_t Counter::Xadd(const int64_t delta) {
```
jblomer: See above

cvmfs/statistics.cc
```diff
@@ -40,7 +68,7 @@ string Statistics::PrintList(const PrintOptions print_options) {
   for (map<string, CounterInfo *>::const_iterator i = counters_.begin(),
        iEnd = counters_.end(); i != iEnd; ++i)
   {
-    result += i->first + "|" + StringifyInt(i->second->counter.Get()) +
+    result += i->first + "|" + i->second->counter.Print() +
```
jblomer: Better use the `ToString()` method (see above)

cvmfs/cvmfs.cc
jblomer: Hm, on a second thought: it is probably better to leave it in.  This way, the compiler will complain whenever we change the name of the counter in the download manager.  Never mind.

cvmfs/wpad.cc
```diff
@@ -226,7 +226,8 @@ static void AltCvmfsLogger(const LogSource source, const int mask,
 }
 
 
-int MainResolveProxyDescription(int argc, char **argv) {
+int MainResolveProxyDescription(int argc, char **argv,
+    perf::Statistics *statistics) {
```
jblomer: Here you can create the `Statistics` object on the fly (see above)

mount/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ set (CVMFS_MOUNT_SOURCES
   ../cvmfs/logging.h ../cvmfs/logging_internal.h ../cvmfs/logging.cc
   ../cvmfs/options.h ../cvmfs/options.cc
   ../cvmfs/sanitizer.h ../cvmfs/sanitizer.cc
+  ../cvmfs/statistics.h ../cvmfs/statistics.cc
```
jblomer: Is this really needed?

cvmfs/nfs_shared_maps.cc
Moliholy: It is used in the line 259 of the nfs_maps.cc file:

```
string GetStatistics() {
  if (use_shared_db_)
    return nfs_shared_maps::GetStatistics();

  string result = "Total number of issued inodes: " +
                  StringifyInt(seq_-root_inode_) + "\n";

  string stats;
  db_inode2path_->GetProperty(leveldb::Slice("leveldb.stats"), &stats);
  result += "inode --> path database:\n" + stats + "\n";

  db_path2inode_->GetProperty(leveldb::Slice("leveldb.stats"), &stats);
  result += "path --> inode database:\n" + stats + "\n";

  return result;
}
```

Shall we remove that if (used_shared_db_) ?

jblomer: Not only the if but the whole `GetStatistics()` in nfs_maps.cc can go away, no?

Moliholy: It is used in swissknife_pull.cc, line 558. However, it's just a log that can be perfectly removed

jblomer: This comment seems to be misplaced (is for `DownloadManager::GetStatistics()`, right?).  Please see my follow-up comment.  It probably makes sense to keep the method in `DownloadManager`.

cvmfs/shortstring.h
Moliholy: Fixed in the last update

mount/CMakeLists.txt
Moliholy: Actually no, it isn't. My mistake. Now it is fixed

jblomer:
Hm, something is still not quite right.  If I try to mount, I get the error "/usr/lib64/libcvmfs_fuse.so: undefined symbol: _ZN8nfs_maps13GetStatisticsEv".  Can you confirm?

Also, please always use cpplint.py before submitting the pull request.

Moliholy:
Yes, there was a mistake. I called a function defined in the .h that was recently deleted by mistake in the corresponding .cc
After the rebase it should be fixed (for me it works now), as well as the cpplint.

cvmfs/cache.h
jblomer: When I try to merge, this function is still present.  Can you have a look?

cvmfs/cvmfs.h
```diff
@@ -46,8 +46,6 @@ std::string GetOpenCatalogs();
 unsigned GetMaxTTL();  // in minutes
 void SetMaxTTL(const unsigned value);  // in minutes
 void ResetErrorCounters();
-void GetLruStatistics(lru::Statistics *inode_stats, lru::Statistics *path_stats,
-                      lru::Statistics *md5path_stats);
 std::string PrintInodeTrackerStatistics();
 std::string PrintInodeGeneration();
 catalog::Statistics GetCatalogStatistics();
```
jblomer: This declaration should be removed (I think there is at least one more that got removed in the .cc file but not here)

Moliholy:
Fixed the unnecessary method in cache.h

cvmfs/cvmfs.cc
```diff
@@ -339,15 +345,6 @@ static bool UseWatchdog() {
...
-void GetLruStatistics(lru::Statistics *inode_stats, lru::Statistics *path_stats,
-                      lru::Statistics *md5path_stats)
-{
-  *inode_stats = inode_cache_->statistics();
-  *path_stats = path_cache_->statistics();
-  *md5path_stats = md5path_cache_->statistics();
-}
-
-
 string PrintInodeTrackerStatistics() {
```
jblomer: This function and similar ones can be removed, too.

Moliholy: Removed as well

jblomer:
There is one bigger problem: running the unit tests results in a memory corruption.  Can you have a look?

Moliholy:
I have just run the unittests three times without apparently any problem

Moliholy:
So, I have fixed the bug! It was a problem with the initialization of the counters, apparently. This is because the "statistics objects" are created depending on the binary in different places, so it's quite difficult to keep track of all of them. I would suggest to create a _super_ global singleton object in statistics.h so that every class who wants to register its counters just have to import this file and call the object method.

reneme:
I've been working with @Moliholy on this today and I really wonder why the central `Statistics` class isn't designed as a singleton for three reasons:

1. The statistics class is a cross-cutting concern and therefore should be accessible from essentially everywhere.
2. A singleton inherently does its memory management automagically via lazy initialisation even before `main()` runs, if necessary.
3. It would simplify the API of using classes be entirely decoupling the `Statistics` class from them.
4. Naïvely, I don't see a reason for more than one `Statistics` object per process.

Especially point (2) is interesting, because it would allow us to get rid of the `ShortString<>::RegisterCounter()` methods that initialise global counter state in a rather icky way. Furthermore, the global `Statistics` object doesn't need to be handled by some high-level program code anymore.

I just wanted to cross-check if there is a reason to not make it a singleton.

jblomer:
You need multiple statistics objects for the different contexts in the library.  Otherwise there is a name clash.  Other than that I have no objection against singleton.

reneme:
Statistics are per repository?

jblomer:
Every counter has a name.  That would clash.

reneme:
Yes, I got that... What about prefixing this name with the repo's FQRN?

jblomer:
Fine with me, if you find a way to figure out the fqrn at early initialization time of the statistics object.

Moliholy:
I have a expected problem with some counters. When executing the following in libcvmfs_test:

```
/home/molina/cvmfs/build/cvmfs/test_libcvmfs: initializing with options: cache_directory=/tmp/test-libcvmfs-cache
cvmfs> attach sft
attaching repo with options: repo_name=sft.cern.ch,url=http://cvmfs-stratum-one.cern.ch/opt/sft;http://cernvmfs.gridpp.rl.ac.uk/opt/sft;http://cvmfs.racf.bnl.gov/opt/sft,pubkey=/etc/cvmfs/keys/cern.ch.pub
CernVM-FS: using public key(s) /etc/cvmfs/keys/cern.ch.pub
cvmfs> attach atlas
attaching repo with options: repo_name=atlas.cern.ch,url=http://cvmfs-stratum-one.cern.ch/opt/atlas;http://cernvmfs.gridpp.rl.ac.uk/opt/atlas;http://cvmfs.racf.bnl.gov/opt/atlas,pubkey=/etc/cvmfs/keys/cern.ch.pub
test_libcvmfs: /home/molina/cvmfs/cvmfs/statistics.cc:72: perf::Counter* perf::Statistics::Register(const std::string&, const std::string&): Assertion `counters_.find(name) == counters_.end()' failed.
```

There is (obviously) a name clash when registering a counter. To fix it there are two quick options:

1) Identify the classes that are invoking these methods and force them to add a "surname" (scope) to the variable. In a nutshell, these classes would be the followings:
  * cache::CatalogManager
  *  download::DownloadManager
  *  lru::Md5PathCache
However, this would make the functionality for these classes more complicated and it would add more characters to the counter name, which is maybe not desirable.

2) A even quicker and simpler workaround would be to remove the restriction in the Counter::Register method so that it now returns the instance that has already been created. This way these "numbers" would be modified by two different context, but in any case at the moment this is only possible in libcvmfs, whose statistics are not shown. For a future pull request this could (should) probably be modified adding a different solution.


What's your opinion?

jblomer:
The second option looks appealing but it is also dangerous if at any point we use the same name for different counters by mistake.  I think there is also another issue with the ShortString class and the hotpatching.  In fact it would be better not to touch the memory layout of this class, otherwise we need to code the memory layout upgrade.  I'd suggest we have a little white board session next week, now that we know where all the dead bodies are hidden.

jblomer:
As discussed, to be replaced by two smaller pull requests.  Closing this one.

PR #832, opened 2015-04-02
FIX: Upstream OverlayFS Integration 

reneme:
This fixes some minor things to resurrect the OverlayFS support in the CernVM-FS server. Some things seemed to have changed in the OverlayFS implementation since it went into upstream:

* Kernel module is called `overlay` instead of `overlayfs`
* Same with file system type name
* OverlayFS now requires a `workdir` along with the `upperdir` and `lowerdir`


jblomer:
Can you figure out what the workdir is being used for?  In particular if it needs to be wiped out after publishing.

reneme:
The `workdir` is comparable to the `data/txn` directory we use in the _CAS_. It is [used to prepare files](http://lwn.net/Articles/600106/) before moving them atomically in place. Therefore it needs to be on the same file system as the `upperdir`. To me it seems that overlayfs takes care of the management automagically.

PR #835, opened 2015-04-06
Fix CVMFS libraries cannot find

HarryWei:
For Ubuntu 14.04 i386 and x86_64, `cvmfs_config chksetup` cannot find cvmfs libraries. Commit d3abc28 fixes this bug.


jblomer:
Thanks!  As discussed, a slightly modified fix was committed.

PR #838, opened 2015-04-07
Fix: S3-backend caching disabled for mutable objects (CVM-808)

ssheikki:
Mutable objects are uploaded now with "Cache-control: no-cache" header. This is needed at least with Google cloud storage to execute consecutive changes to the repository (e.g. twice mkfs). The problem is described in more detailed at Jira ticket CVM-808.

### DIFF
--- a/cvmfs/upload_s3.cc
+++ b/cvmfs/upload_s3.cc
@@ -361,11 +361,14 @@ void S3Uploader::FileUpload(
                             const_cast<void*>(
                                 static_cast<void const*>(callback)),
                             local_path);
+
+  if (remote_path.substr(0, 1) == ".") {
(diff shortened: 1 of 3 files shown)


cvmfs/upload_s3.cc
```diff
@@ -361,11 +361,14 @@ void S3Uploader::FileUpload(
                             const_cast<void*>(
                                 static_cast<void const*>(callback)),
                             local_path);
+
+  if (remote_path.substr(0, 1) == ".") {
```
jblomer: Probably works but would also match "./data/<hash>".  Perhaps it's better to check for ".cvmfs" here

ssheikki: True, if all special files start with ".cvmfs". I will fix it.

PR #852, opened 2015-04-17
FIX: Kernel Deadlock in CentOS 7 with OverlayFS + Hardlink Stresstest in Ubuntu 12.04

reneme:
Seems that OverlayFS suffers from the same problem as old AUFS versions. :-( This adds the [kernel deadlock workaround](http://cernvm.cern.ch/portal/cvmfs/workaround-krnl-deadlock) to the CentOS 7 integration test context. 

Furthermore this disables the hardlink stresstest for Ubuntu 12.04 since the AUFS version is too old to handle it correctly in this distribution.


jblomer:
We should report this upstream.

PR #854, opened 2015-04-20
FIX: Dirtab Application Leaks Temporary Catalog Database Files

reneme:
As described in [CVM-818](https://sft.its.cern.ch/jira/browse/CVM-818) the application of `.cvmfsdirtab` doesn't properly remove temporarily downloaded catalog database files. This fixes it by adding an auto-management feature to `SimpleCatalogManager`. It uses the `TakeDatabaseFileOwnership()` method of `Catalog` to automatically clean up the underlying SQLite database files as soon as a catalog is closed.

Furthermore this fixes a malformed shell quotation in `cvmfs_server` that would have acted as a safe guard to clear out the temp directory after each publishing. During normal operation the procedures in `cvmfs_swissknife` should not leak any tmp files. However, crashes or user aborts could still cause it (cf. [CVM-816](https://sft.its.cern.ch/jira/browse/CVM-816)), hence I'd like to keep this safe guard in place.

Also this introduces checks for an empty tmp directory in various integration test cases.

### DIFF
--- a/cvmfs/catalog.h
+++ b/cvmfs/catalog.h
@@ -163,7 +163,7 @@ class Catalog : public SingleCopy {
   void TakeDatabaseFileOwnership();
   void DropDatabaseFileOwnership();
   bool OwnsDatabaseFile() const {
-    return NULL != database_ && database_->OwnsFile();
+    return ((database_ != NULL) && database_->OwnsFile()) || managed_database_;
   }
 
   uint64_t GetTTL() const;
@@ -256,6 +256,7 @@ class Catalog : public SingleCopy {
   PathString path_;
   bool volatile_flag_;
   const bool is_root_;
+  bool managed_database_;
 
   Catalog *parent_;
   NestedCatalogMap children_;
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -899,7 +899,7 @@ close_transaction() {
     cvmfs_suid_helper rdonly_umount $name
   fi
   cvmfs_suid_helper clear_scratch $name
-  rm -rf ${tmp_dir}/*
+  [ -z "$tmp_dir" ] && rm -fR "${tmp_dir}"/*
   cvmfs_suid_helper rdonly_mount $name > /dev/null
   cvmfs_suid_helper rw_mount $name
   rmdir "$tx_lock"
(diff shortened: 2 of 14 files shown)


cvmfs/catalog.h
```diff
@@ -163,7 +163,7 @@ class Catalog : public SingleCopy {
   void TakeDatabaseFileOwnership();
   void DropDatabaseFileOwnership();
   bool OwnsDatabaseFile() const {
-    return NULL != database_ && database_->OwnsFile();
+    return ((database_ != NULL) && database_->OwnsFile()) || managed_database_;
```
jblomer: Can we just use `database_->OwnsFile()` and avoid double-bookkeeping of the state in `managed_database_`?

reneme: Unfortunately not, I don't like it either. But a `Catalog` object can exist without a database object attached to it (yet). Therefore a `Catalog` needs to hold this state in order to forward it to its `CatalogDatabase` as soon as it gets attached to it.

cvmfs/cvmfs_server
```diff
@@ -899,7 +899,7 @@ close_transaction() {
     cvmfs_suid_helper rdonly_umount $name
   fi
   cvmfs_suid_helper clear_scratch $name
-  rm -rf "${tmp_dir}/*"
+  rm -rf ${tmp_dir}/*
```
jblomer: That should be prepended with a test if `${tmp_dir}` is defined.  Hm, I wonder why it does not work with the quotes?  If for some reason we ever have a space in `$tmp_dir`, this command will surprise us, too.

cvmfs/catalog.h
reneme: mmh, double dispatch might work... I would vote for revisiting the `Catalog` resp. `CatalogManager` facilities on a more basic level though. Not sure what that means yet, but they are not unit tested and hence need some care anyways.

cvmfs/cvmfs_server
reneme: I believe, any spaces in paths or repo names or anything might surprise us anyways. ;-) The quoted thing doesn't work, because it intercepts the shell globbing. Probably something like `rm -rf "${tmp_dir}"/*` would work as expected...? Shell is evil...

reneme:
added a check for $tmp_dir emptiness in `cvmfs_server`. Also it should allow for spaces in the path now.

reneme:
is there still something that prevents it from being merged?

PR #858, opened 2015-04-23
Second statistics set of changes

Moliholy:
It includes:

1) All changes in the main source code except the ShortString class.
2) Fixes for the unit tests
3) Fix for libcvmfs to allow mounting more than one repository. However this should be considered just a workaround

NOTE: unit tests have all passed

### DIFF
--- a/cvmfs/libcvmfs.cc
+++ b/cvmfs/libcvmfs.cc
@@ -22,10 +22,13 @@
 #include "libcvmfs_int.h"
 #include "logging.h"
 #include "smalloc.h"
+#include "statistics.h"
 #include "util.h"
 
 using namespace std;  // NOLINT
 
+
+perf::Statistics statistics_;
 
 int set_option(char const *name, char const *value, bool *var) {
   if (*value != '\0') {
--- a/cvmfs/libcvmfs_int.cc
+++ b/cvmfs/libcvmfs_int.cc
@@ -284,7 +284,8 @@ cvmfs_context* cvmfs_context::Create(const options &opts) {
   cvmfs_context *ctx = new cvmfs_context(opts);
   assert(ctx != NULL);
 
-  if (ctx->Setup(opts, &statistics_) != 0) {
+  perf::Statistics *statistics = new perf::Statistics(); // provisional
+  if (ctx->Setup(opts, statistics) != 0) {
     delete ctx;
     ctx = NULL;
   }
--- a/cvmfs/libcvmfs_int.cc
+++ b/cvmfs/libcvmfs_int.cc
@@ -73,6 +73,7 @@
 #include "hash.h"
 #include "logging.h"
 #include "lru.h"
+#include "libcvmfs.h"
 #include "monitor.h"
 #include "murmur.h"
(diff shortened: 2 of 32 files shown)


jblomer:
Ah, was just about to work on the merging of this but it conflicts with René's latest fix on the leakage of temporary files.  Could you please rebase to the latest devel?

Moliholy:
Done! Also I have just run the unit tests again and they all passed

cvmfs/libcvmfs.cc
```diff
@@ -22,11 +22,14 @@
 #include "libcvmfs_int.h"
 #include "logging.h"
 #include "smalloc.h"
+#include "statistics.h"
 #include "util.h"
 
 using namespace std;  // NOLINT
 
 
+perf::Statistics statistics_;
```
jblomer: Is there actually still a need for the global?

cvmfs/libcvmfs_int.cc
```diff
@@ -282,7 +284,8 @@ cvmfs_context* cvmfs_context::Create(const options &opts) {
   cvmfs_context *ctx = new cvmfs_context(opts);
   assert(ctx != NULL);
 
-  if (ctx->Setup(opts) != 0) {
+  perf::Statistics *statistics = new perf::Statistics(); // provisional
```
jblomer: Is this objected deleted anywhere?

cvmfs/libcvmfs.cc
Moliholy: That was the source of the previous problem, where by mistake I was using different statistics objects. After creating that global everything inside libcvmfs uses that statistics object and there are no clashes.
However now there is no problem with the ShortString class, so maybe we could think about playing with different statistics objects, but I have the sensation it would be playing with fire.

cvmfs/libcvmfs_int.cc
Moliholy: No, it isn't. I mean, it is in the end of the program. It is just a workaround to avoid clashes between different repositories (contexts) mounted at the same time.

cvmfs/swissknife_sync.cc
```diff
@@ -328,9 +329,8 @@ void swissknife::CommandApplyDirtab::FilterCandidatesFromGlobResult(
     }
 
     // check if the path is a meta-directory (. or ..)
-    assert(candidate_rel.size() >= 2);
-    if (candidate_rel.substr(candidate_rel.size() - 2) == "/." ||
-        candidate_rel.substr(candidate_rel.size() - 3) == "/..") {
+    if (candidate_rel.substr(candidate_rel.size() - 3) == "/.." ||
```
jblomer: Is this change supposed to be part of this PR?

cvmfs/libcvmfs_int.cc
```diff
@@ -73,6 +73,7 @@
 #include "hash.h"
 #include "logging.h"
 #include "lru.h"
+#include "libcvmfs.h"
```
jblomer: style, order of includes

PR #859, opened 2015-04-23
Fix S3 and download HTTP redirects

ssheikki:
This fix enables CVMFS to follow HTTP redirect requests. This is useful when using repositories with several buckets and Squid. The only change required in the Squid configuration is that mapped url is returned with "302:"-prefix, which in effect redirects the CVMFS to download the requested file directly from the cloud storage.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -1054,7 +1054,7 @@ get_item() {
   local noproxy="$2"
 
   if [ x"$noproxy" != x"" ]; then
-    unset http_proxy && curl "$url" 2>/dev/null
+    unset http_proxy && curl -L "$url" 2>/dev/null
   else
     curl "$url" 2>/dev/null
   fi
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -793,12 +793,17 @@ is_master_replica() {
   local name=$1
   local is_master_replica
 
+  load_repo_config $name
+  local swissknife_info_params="-m"
+  if [ x"$CVMFS_FOLLOW_REDIRECTS" == x"yes" ]; then
+    swissknife_info_params="${swissknife_info_params} -L"
+  fi
+
   if [ $(echo $name | cut --bytes=1-7) = "http://" ]; then
-    is_master_replica=$(cvmfs_swissknife info -r $name -m 2>/dev/null)
+    is_master_replica=$(cvmfs_swissknife info -r $name ${swissknife_info_params} 2>/dev/null)
   else
     is_stratum0 $name || return 1
-    load_repo_config $name
-    is_master_replica=$(cvmfs_swissknife info -r $CVMFS_STRATUM0 -m 2>/dev/null)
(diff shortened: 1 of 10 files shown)


jblomer:
This looks fine except that I am still a bit reluctant to enable HTTP redirects for the download by default.  It might just work but I'm not convinced yet that we won't see any unwanted consequences.  For the moment, you can add CVMFS_FOLLOW_REDIRECTS=yes to the client configuration.

ssheikki:
If I understand correctly, the client.conf that is generated by mkfs should have in default "CVMFS_FOLLOW_REDIRECTS=yes"? I cannot see any other way to set that value because the client.conf is created by mkfs. In any case, this parameter is not however read by "cvmfs_swissknife info" or "cvmfs_swissknife sync". We could however parse it in cvmfs_server and pass it as a parameter (e.g. -L to enable redirects) to the "cvmfs_swissknife info" and "cvmfs_swissknife sync" commands. Would this perhaps make sense? Other option would be to give the client.conf file to the "cvmfs_swissknife info" and "cvmfs_swissknife sync".

ssheikki:
The commit "6bb1f53" does what I described. There is only one place (when the repository name is not known, cvmfs_server line 2569) that the default behaviour has to be to follow redirects (because we cannot know if they should be followed or not). Please don't hesitate to say if this was not perhaps the best approach!

jblomer:
Looks ok to me.  @reneme can you take a look?

reneme:
The changes in `cvmfs_server` for the calls to `cvmfs_swissknife info` look rather repetitive to me, though. Can we encapsulate the calls to `cvmfs_swissknife info` into a helper shell function and do the parameter tricks only once? Something along those lines:

```bash
get_repo_info() {
  local repo_name=$1
  # optionally add -L here
  cvmfs_swissknife info $optional_minus_L $repo_name $@
}

# foblaghr...

hash=$(get_repo_info $repo_name -h)
```

ssheikki:
There is also "tag" and "sync" which need very similar parsing so what if there would be only function to get the redirect flag?
```bash
get_follow_http_redirects_flag() {
  if [ x"$CVMFS_FOLLOW_REDIRECTS" == x"yes" ]; then
    return "-L"
  fi
  return " "
}
cvmfs_swissknife info -r $CVMFS_STRATUM0 $(get_follow_http_redirects_flag)
```

All the cvmfs_swissknife flags are btw with small letters (e.g. -r). However, "cvmfs_swissknife sort" uses already all the small letters from a-z so I chose initially "-L", which is the flag used by curl. Is this ok or could there be some other solution that would not break the small letter convention?

jblomer:
Eventually we should parse the options files directly instead of reformatting them as command line options. But for the moment I think -L is as good as any other choice.  Go for it.

By the way, the current pull request can't be merged.  Could you rebase to the latest devel branch?

ssheikki:
Hmm, download.cc seems to have EnableRedirects(); set, I am going to fix that still!

ssheikki:
Ready! Tested with S3 unit tests and with 30,000 files that were successfully stored to a real S3 backend (speed around 600 files/s).

PR #863, opened 2015-04-30
STATISTICS: fixed bugs in the benchmarks. Added statistics_collector.py script

Moliholy:


test/test_functions
```diff
@@ -267,8 +269,12 @@ cvmfs_umount() {
     sudo umount /cvmfs/$r
     local result=$?
     if [ $result -ne 0 ]; then
-      lsof /cvmfs/$r
-      return 200
+      sudo umount -l /cvmfs/$r
```
jblomer: That's dangerous because the fuse process might stay alive, which results in unexpected behavior for the next mount.  Can we revert this?

test/test_functions
```diff
@@ -204,7 +206,7 @@ contains() {
 
   for elem in $haystack
   do
-    if [ $(basename $elem) = $(basename $needle) ]; then
+    if [ $(readlink --canonicalize $elem) = $(readlink --canonicalize $needle) ]; then
```
Moliholy: This is wrong, I've just realised. I'll revert it very soon

Moliholy:
Changes applied!

PR #864, opened 2015-05-01
Fix gitignore

PerilousApricot:
Add more matches to keep the repository "clean" with a currently built
working area

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,33 @@ python/MANIFEST
 .metadata
 .settings
 RemoteSystemsTempFiles/
+
+# Externals
+externals/build_*/
+externals/.decompressionDone
+
+# CMake stuff
+cvmfs_config.h
+CMakeCache.txt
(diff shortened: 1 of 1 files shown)


jblomer:
Thank you for the patch!  Now, I wonder if this is the best default for all the developers.  We normally use a different directory for building so that the source tree doesn't get polluted in the first place.  Would that work for you?

PerilousApricot:
I don't think it hurts to have these in the .gitignore since they are all specific files and not wildcards (i.e. adding *.cmake would be overly broad).

jblomer:
That's fine for merging if it solves a particular problem and doesn't introduce a new one.  @reneme do you have any objections?

@PerilousApricot Can you explain why building out of the source tree doesn't work for you?

reneme:
No objections from my side. Sound reasonable to me to 'support' in-place build to give developers a choice.

PR #865, opened 2015-05-02
Make alien_cache usable for libcvmfs

bbockelm:
In libcvmfs, there's no way to separate the instance / quota lock directory from the alien cache directory.

This causes a flock to be taken within the alien cache directory, meaning that only one instance of libcvmfs can be active per alien cache (somewhat defeating the purpose of alien cache!).

This adversely affects parrot integration with CVMFS.  Discovered while trying to use Parrot at NERSC for CMS.

This PR adds an option to have a separate lock directory.

@btovar @dthain - if alien_cache is enabled for libcvmfs2, you probably will be interested in this option (or wherever the conversation goes...).  It probably should default to the cvmfs_locks directory created by parrot.

### DIFF
--- a/cvmfs/libcvmfs.cc
+++ b/cvmfs/libcvmfs.cc
@@ -113,6 +113,7 @@ struct cvmfs_repo_options : public cvmfs_context::options {
 struct cvmfs_global_options : public cvmfs_globals::options {
   int set_option(char const *name, char const *value) {
     CVMFS_OPT(cache_directory);
+    CVMFS_OPT(lock_directory);
     CVMFS_OPT(change_to_cache_directory);
     CVMFS_OPT(alien_cache);
     CVMFS_OPT(log_syslog_level);
@@ -208,6 +209,7 @@ typedef cvmfs_options<cvmfs_global_options> global_options;
 
             "global options are:\n"
             " cache_directory=DIR        Where to store disk cache\n"
+            " lock_directory=DIR         Where to store the per-instance locks (should be unique for alien cache)\n"
             " change_to_cache_directory  Performs a cd to the cache directory (performance tweak)\n"
             " alien_cache                Treat cache directory as alien cache\n"
             " log_syslog_level=LEVEL     Sets the level used for syslog to DEBUG (1), INFO (2), or NOTICE (3).\n"
--- a/cvmfs/libcvmfs_int.cc
+++ b/cvmfs/libcvmfs_int.cc
@@ -148,6 +148,10 @@ cvmfs_globals::~cvmfs_globals() {
 int cvmfs_globals::Setup(const options &opts) {
   // Fill cvmfs option variables from arguments
   cache_directory_ = opts.cache_directory;
+  lock_directory_ = opts.lock_directory;
+  if (!lock_directory_.size()) {
+    lock_directory_ = cache_directory_;
+  }
   uid_ = getuid();
   gid_ = getgid();
   options_ready_ = true;
@@ -208,6 +212,12 @@ int cvmfs_globals::Setup(const options &opts) {
     return -2;
   }
 
+  // Create cache directory, if necessary
+  if (! MkdirDeep(lock_directory_, 0700)) {
+    PrintError("cannot create lock directory " + lock_directory_);
+    return -2;
+  }
+
   // Try to jump to cache directory.  This tests, if it is accassible.
   // Also, it brings speed later on.
   if (opts.change_to_cache_directory &&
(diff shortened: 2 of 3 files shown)


jblomer:
Interesting, that used to work with libcvmfs 2.1.19 where alien_cachedir=DIR would allow you to specify a separate cache directory (where no locks are taken).  Can you try if libcvmfs 2.1.19 works for you?

@reneme What was the reason to remove the alien_cachedir option?

reneme:
I skimmed over the associated Pull Requests to both [libcvmfs](https://github.com/cvmfs/cvmfs/pull/502) and [parrot](https://github.com/cooperative-computing-lab/cctools/pull/488). I found the redundant cache_dirs (`cachedir`) and (`alien_cachedir`) confusing and [changed the latter to a flag](https://github.com/cvmfs/cvmfs/pull/502/files#diff-5ed2af3388a077bbf8d5576ce9238fddR229) (`alien_cache`). That should treat the cache directory given through `cache_directory` as an alien cache (and hence not `flock()`).

bbockelm:
Hi,

Yes - I can confirm this is a regression from 2.1.19 to 2.1.20.  The flock() is now done in a file in the alien cache.

(Additionally, the renaming of various parameters such as cache_dir -> cache_directory seemed a bit arbitrary and annoying in a bugfix release.)

Brian

jblomer:
Agreed.  Let me see how to bring the interface back to the 2.1.19 state.  I think what we will do is to make a libcvmfs branch where we can fix and release more quickly than the cvmfs fuse module, which requires updating all of grid.  We'll speak to cctools guys about how to come up with a common channel for parrot/cvmfs.  Perhaps we just put a pre-compiled `parrot_run` online.

For this particular problem, would you like us to build something immediately or can you survive for another week with your patch?

bbockelm:
Hi Jakob,

I have staged a private build for us to use.  So, that decreases the urgency a bit.

As far as the interface changes done - I was just commenting that they were annoying, not that they need to be reverted.  I've already adopted all our scripts to them.

Brian

dthain:
How can we help on the parrot side of things?

bbockelm:
Well, there's no need for me to post my corresponding patch for the parrot codebase if Jakob will be fixing up the interface.

So, from the parrot side, I would ask:
1.  It seems that the 4.4.0 release isn't tagged in github?  Something break in the release process?
   * It would be beneficial if "parrot_run -v" recorded the configure flags used for CVMFS.
2.  Can we get some unit test coverage for the various configuration options?  Particularly:
   * Make sure alien cache can be specified; detect configuration key name changes.
   * Validate that multiple instance of parrot can use alien cache at the same time.

jblomer:
Hm, now it would be helpful to know who else had scripts depending on the parameters.  

Sorry about that, it was not our intention to change the interface, this got sort of messed up.  I need to check if other parts of the code rely on the fact that alien cache directory and cache directory are separated.  Perhaps we can support both options now.

jblomer:
@bbockelm Your fixes, together with a few more ones, are in the branch fix-libcvmfs and I'll rebase them into libcvmfs-stable branch so that they can be picked up by cctools.  For the testing, could you tell us how you start parrot_run? Is it like

    PARROT_CVMFS_CONFIG="cache_directory=...,alien_cache,lock_directory=..." parrot_run ...

bbockelm:
@jblomer - it's a bit of mess right now as we're on a private build for this bug, but here's the environment variables we use:

https://github.com/tiradani/parrot_glidein_wrapper/blob/master/parrot_cfg_cern

In particular,

```
PARROT_CVMFS_REPO="cms.cern.ch:timeout=3600,timeout_direct=3600,pubkey=${pubkey},url=${CVMFS_STRATUM1},proxies=${CVMFS_PROXIES} grid.cern.ch:timeout=3600,timeout_direct=3600,pubkey=${pubkey},url=http://cvmfs-stratum-one.cern.ch/opt/grid,proxies=${CVMFS_PROXIES}"

PARROT_CVMFS_CONFIG="cache_directory=${alien_cache},alien_cache"
```

Then, the parrot_run invocation looks like this (https://github.com/tiradani/parrot_glidein_wrapper/blob/master/parrot_setup#L259):

```
$GLIDEIN_PARROT/parrot_run" -d cvmfs -t "$GLIDEIN_PARROT_TMP"
```
(the actual invocation before the job is a bit more circuitous; the above is roughly equivalent)

jblomer:
Thanks!  So where does the value for `lock_directory` come from?

bbockelm:
oh - that's a separate patch for parrot itself; it will create (and cleanup) a subdirectory of the parrot temporary directory and append as appropriate to the PARROT_CVMFS_CONFIG.

jblomer:
Alright, so as far as I can see, you can use the libcvmfs-stable branch and a parrot version with this other patch included.  If there are no objections, I'll close this pull request.

bbockelm:
No objections.  Depending on workload, I'll try to switch and drop our private patch next week.

PR #867, opened 2015-05-05
bump version to 2.2.0, adjust bump_version script

jblomer:
Updates to MAJOR.MINOR.PATCH versioning.  Heads-up: there might be hickups in the ci-scripts

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2.1.21:
+2.2.0:
   * Disable caching for mutable objects in S3 backend (CVM-808)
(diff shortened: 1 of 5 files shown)


ChangeLog
```diff
@@ -1,4 +1,4 @@
-2.1.21:
+2.2.0:
```
reneme: you might want to specifically mention a change in the version number scheme in the changelog, no?

jblomer: Yep, done.  As always, I do the changelog changes after the merging.

PR #869, opened 2015-05-05
FIX: CMake 2.6.3 required for unset()

reneme:
This is a follow up for the [last pull request](https://github.com/cvmfs/cvmfs/pull/868). Turns out, that `unset()` [needs](http://www.cmake.org/pipermail/cmake/2010-January/034307.html) at least CMake 2.6.3 thus I'm bumping the required version. We have it installed on all our build machines - after I updated the package on SLES11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@
 # See externals/CMake-Register_External_Lib.txt for details on external inclusion
 #
 
-cmake_minimum_required (VERSION 2.6.2)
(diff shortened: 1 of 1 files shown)


jblomer:
is cmake 2.6.3 part of standard SLES11?

reneme:
I got it from an external RPM, so most likely not...

jblomer:
Then it would be better to work around unset().  We have users who build on SLES11.

PR #870, opened 2015-05-05
Revert: CMake version 2.6.2 due to SLES 11

reneme:
Going back to CMake 2.6.2 to support vanilla SLES 11. I replaced `unset()` by `set(XXX "")` hoping to produce the same effect as before. (Testing now...)


reneme:
... seems to work.

PR #871, opened 2015-05-05
Update: LevelDB to 1.18

reneme:
Currently we use LevelDB 1.15 which doesn't compile vanilla on AArch64. With 1.18 it works fine.


jblomer:
Before we add any new tarballs, it is now a good moment to look into submodules.  I think we should fork leveldb from github into our project, create a submodule, and change `bootstrap.sh` to initialize the submodule.

reneme:
Jira: [CVM-754](https://sft.its.cern.ch/jira/browse/CVM-754). I wonder if we could clean the repository history from these tar balls, otherwise we don't gain that much. `git filter-branch` might do the trick... taking the discussion over to Jira.

jblomer:
Looking at the tarball, it's small enough to make the situation significantly worse.  Merging.

PR #873, opened 2015-05-06
add cvmfs_rsync script to cvmfs-server package

DrDaveD:
As discussed in https://sft.its.cern.ch/jira/browse/CVM-814

### DIFF
--- a/cvmfs/cvmfs_rsync
+++ b/cvmfs/cvmfs_rsync
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Rsync directories into cvmfs.
 
-ME=cvmfsrsync
+ME=cvmfs_rsync
 
 usage()
 {
--- a/cvmfs/cvmfs_rsync
+++ b/cvmfs/cvmfs_rsync
@@ -1,0 +1,87 @@
+#!/bin/bash
+# Rsync directories into cvmfs.
+
+ME=cvmfsrsync
+
+usage()
+{
+    echo "$ME [rsync_options] srcdir /cvmfs/reponame[/destsubdir]" >&2
+    echo "  Rsync to cvmfs repositories.  Avoids .cvmfscatalog files while allowing"
+    echo "  directories containing them to be deleted when their source goes away." >&2
+    exit 1
+}
+
+if [ $# -lt 2 ]; then
+    usage
+fi
+
+# readlink -m canonicalizes paths, removing extra slashes, etc.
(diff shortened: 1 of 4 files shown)


DrDaveD:
Please don't forget this pull request

cvmfs/cvmfs_rsync
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+# Rsync directories into cvmfs.
+
+ME=cvmfsrsync
```
jblomer: This should be `cvmfs_rsync`, no?

jblomer:
Apologies for the delay.  Looks fine to me.  @reneme any comments? 

Can you add a few integration tests for this new tool?

reneme:
Why doesn't the script work with remote hosts? Is there a technical reason? I'm not sure if it is an operational problem, just curious.

cvmfs/cvmfs_rsync
DrDaveD: You're right, I corrected it to cvmfs_rsync

DrDaveD:
> Can you add a few integration tests for this new tool?
Is it alright if I do that in a later pull request?  I have other priorities at the moment and we're in conversation about a possible alternative that would make this obsolete anyway.

> Why doesn't the script work with remote hosts? Is there a technical reason? 
Yes, it needs to look for missing directories in the source and .catalog files in the destination and it's not practical to do either remotely.  This limitation is one of the reasons (along with the script's general complexity) that I asked in the ticket https://sft.its.cern.ch/jira/browse/CVM-814 about the possibility of an alternative that bypasses the intermediate step of creating .cvmfscatalog files from .cvmfsdirtab, going instead directly from .cvmfsdirtab to catalog.  (Please respond to the discussion there because I think it will be easier to locate in the future than this pull request).

jblomer:
If you need this urgently merged, I can do it of course.  Otherwise, I wouldn't mind leaving the pull request open for amendment.

DrDaveD:
It's not that it's urgently needed, but it would be helpful for it to be in a place that I could easily refer people to the latest version.  Also in general it is a pain to let new code branches languish as common files (in this case the cmake file and packaging descriptors) get modified by other updates and need later to be merged.  Frequently merging smaller working pieces into the common branch is the easiest way to do joint development.

DrDaveD:
I want to move this out of my devel branch so I'm closing this request and I'll open a new one

PR #876, opened 2015-05-07
FIX: Legacy Catalog Migration

reneme:
This fixes integration tests 519, 550, 563 and 568 that are all related to the legacy catalog migration from 2.0 catalogs. With the [introduction of `xattr`](https://github.com/cvmfs/cvmfs/pull/768) in the catalog database schema, the migration code was not updated accordingly and therefore crashed.
Additionally this adds a check for such situations to catch them easier next time.

Furthermore this fixes an issue introduced by @ssheikki's [S3 Redirects Pull Request](https://github.com/cvmfs/cvmfs/pull/859) that was using a repository URL as the repository name, leading to a failure in `cvmfs_server` - `load_repo_config()`.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -827,10 +827,10 @@ is_master_replica() {
   local name=$1
   local is_master_replica
 
-  load_repo_config $name
   if [ $(echo $name | cut --bytes=1-7) = "http://" ]; then
     is_master_replica=$(__swissknife info -r $name -m $(get_follow_http_redirects_flag) 2>/dev/null)
   else
+    load_repo_config $name
     is_stratum0 $name || return 1
     is_master_replica=$(__swissknife info -r $CVMFS_STRATUM0 -m $(get_follow_http_redirects_flag) 2>/dev/null)
   fi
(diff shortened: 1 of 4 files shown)


cvmfs/cvmfs_server
```diff
@@ -787,16 +787,16 @@ is_stratum1() {
...
-# @param name   the repository name to be checked
+# @param name   the repository name or URL to be checked
 # @return       0 if it is a stratum0 repository and replicable
 is_master_replica() {
   local name=$1
   local is_master_replica
 
-  load_repo_config $name
   if [ $(echo $name | cut --bytes=1-7) = "http://" ]; then
     is_master_replica=$(cvmfs_swissknife info -r $name -m $(get_follow_http_redirects_flag) 2>/dev/null)
```
ssheikki: It looks like we do not know the repository name necessarily and thus the load_repo_config cannot be called. This means that
$(get_follow_http_redirects_flag)
will never return anything. Should we remove it or have "-L" as a default (to follow the redirects)?

reneme: Very good point, actually. Yes. I am amending this pull request. Thanks for the review!

ssheikki:
It is line 2560, the only other "-L" in the file.

jblomer:
Please use cpplint

reneme:
cpplint cleaned and rebased to latest devel...

PR #879, opened 2015-05-08
TEST: added t_smalloc.cc

Moliholy:
There are basically two type of tests:
1) Normal functionality, using normal numbers and checking that it doesn't crash.
2) The opposite: making reservations of a huge amount of memory that should force CVMFS to crash.

The tests passed, by the way.

### DIFF
--- a/test/unittests/t_smalloc.cc
+++ b/test/unittests/t_smalloc.cc
@@ -1,0 +1,71 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+
+#include <gtest/gtest.h>
+
+#include "../../cvmfs/smalloc.h"
+
+
+const size_t kSmallAllocation = 1024UL;
+const size_t kBigAllocation = 1125899906842624UL;  // 1024⁵ Bytes -> 1024 TB
+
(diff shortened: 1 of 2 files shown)


test/unittests/t_smalloc.cc
```diff
@@ -0,0 +1,71 @@
...
+ * This file is part of the CernVM File System.
+ */
+
+
+#include <gtest/gtest.h>
+
+#include "../../cvmfs/smalloc.h"
+
+#define SMALL_ALLOCATION 1024UL
+#define BIG_ALLOCATION 1125899906842624UL // 1024⁵ Bytes -> 1024 TB
```
reneme: Why not:
```C++
const unsigned long kSmallAllocation = ...;
const unsigned long kBigAllocation = ...;
```
Please avoid the pre-processor wherever possible.

Moliholy: Fixed!

jblomer:
Please use cpplint before submitting PRs.

Moliholy:
Fixed!

PR #882, opened 2015-05-08
FIX: Compile on 32bit Platforms

reneme:
Turns out that 32bit platforms typedef `size_t` to a 32bit integer which raises a compile error since the numerical literal introduced [here](https://github.com/cvmfs/cvmfs/pull/879) is not representable by it. The testee `smalloc()` expects a `size_t` so I changed it to `std::numeric_limits<size_t>::max()` and hope that 4GiB are enough to cause a runtime error on `smalloc()`.

### DIFF
--- a/test/unittests/t_smalloc.cc
+++ b/test/unittests/t_smalloc.cc
@@ -4,12 +4,13 @@
 
 
 #include <gtest/gtest.h>
+#include <limits>
 
 #include "../../cvmfs/smalloc.h"
 
 
 const size_t kSmallAllocation = 1024UL;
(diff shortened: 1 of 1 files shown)


jblomer:
I don't think we can count on only 4GiB for the big allocation.  How about using an uint64_t expression (like 4G * 4G * 4G...)

reneme:
I'd like to try it first. Seems that `size_t` is a 64bit integer on 64bit systems (otherwise they wouldn't have compiled either). Even though a 32bit system will only be asked for 4GiB of memory, it should still fail - since it's every single bit of virtual memory it can possibly address with 32bit.

jblomer:
That's true, of course the unit tests run on the same platform they were built for.  So should be fine as is.

PR #883, opened 2015-05-13
FIX: improved readability in StringifyTime function

Moliholy:
I did it while I was writing the unittests for util.cc and the result is the same than before

### DIFF
--- a/cvmfs/util.cc
+++ b/cvmfs/util.cc
@@ -772,21 +772,16 @@ string StringifyDouble(const double value) {
  * Converts seconds since UTC 0 into something readable
  */
 string StringifyTime(const time_t seconds, const bool utc) {
-  struct tm timestamp;
+  time_t now = time(NULL);
+  char buf[80];
+  struct tm ts;
   if (utc) {
-    localtime_r(&seconds, &timestamp);
+    ts = *localtime(&now);
   } else {
-    gmtime_r(&seconds, &timestamp);
+    ts = *gmtime(&now);
   }
-
-  const char *months[] = {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul",
-    "Aug", "Sep", "Oct", "Nov", "Dec"};
-  char buffer[21];
-  snprintf(buffer, sizeof(buffer), "%d %s %d %02d:%02d:%02d", timestamp.tm_mday,
-           months[timestamp.tm_mon], timestamp.tm_year + 1900,
-           timestamp.tm_hour, timestamp.tm_min, timestamp.tm_sec);
-
-  return string(buffer);
+  strftime(buf, sizeof(buf), "%d %h %Y %H:%M:%S", &ts);
+  return string(buf);
 }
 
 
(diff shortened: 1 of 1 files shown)


cvmfs/util.cc
```diff
@@ -772,21 +772,15 @@ string StringifyDouble(const double value) {
...
-
-  const char *months[] = {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul",
-    "Aug", "Sep", "Oct", "Nov", "Dec"};
-  char buffer[21];
-  snprintf(buffer, sizeof(buffer), "%d %s %d %02d:%02d:%02d", timestamp.tm_mday,
-           months[timestamp.tm_mon], timestamp.tm_year + 1900,
-           timestamp.tm_hour, timestamp.tm_min, timestamp.tm_sec);
-
-  return string(buffer);
+  strftime(buf, sizeof(buf), "%d %b %Y %H:%M:%S", &ts);
```
reneme: Only minor concern is, that `%b` produces the abbreviated name of the month depending on the configured locale. But since this function should only produce 'something readable' as stated in the comment above, I believe its fine. :o)

jblomer:
Good!  Can you figure out where this function is used?

Moliholy:
Yes!
```
cache.cc -> 957
cvmfs.cc -> 2413 & 2416
swissknife_history.cc -> 881, 902 & 994
talk.cc -> 457 & 464
whitelist.cc -> 220 & 227
````

jblomer:
Right, I mean: in which contexts is it used?  Is it everywhere for debug logging?  Or is there an occurrence where the date string is exposed in such a way that an external script might rely on the date format for parsing?

Moliholy:
screen shot 2015-05-13 at 11 50 22

Most of the times it's for logging. However, a few times it's used for other purposes. Concretely:

*cvmfs.cc* -> for an error message
*swissknife.cc* -> inside CommandListTags::PrintHumanReadableList
*talk.cc* -> inside the "hotpatch history" option, concretely to show the date of the last boot and the followings

Moliholy:
In any case the format is exactly the same than before, so even if some script is using it there would be no difference

reneme:
Having a look at the usages it is used for error messages related to repository tags. I highly doubt that somebody wrote a script against those. In the server, it is used to produce a human-readable list of the tags, with a machine readable option (both included in 2.1.20). Whoever wrote a script against that human-readable list anyway deserves a little pain. ;-) The rest is debug output and `tracer.cc` which is dead in the water anyway.

The only one that might be problematic is `cvmfs_talk hotpatch history` printing the timestamps of the client hot patching runs.

jblomer:
Thanks, looks all fine to me.  Just wanted to make sure...

jblomer:
(I'll merge it after the corresponding test cases get merged)

Moliholy:
Right now I am dealing with the RfcTimestamp function, which for consistency should be written in the same way. I will add it soon to this pull request to see if it looks better

jblomer:
All right, but don't get lost in details.  The aim is foremost to increase the test coverage, brushing up is only second priority.

Moliholy:
Closed because this method cannot ensure consistency in the output.

PR #884, opened 2015-05-13
FIX: add support for socket in fs_traversal.h

Moliholy:
When running the unittests I realized that a socket I created was never deleted using the FileSystemTraversal class. I have just tested it now and it works.

### DIFF
--- a/cvmfs/fs_traversal.h
+++ b/cvmfs/fs_traversal.h
@@ -45,6 +45,7 @@ class FileSystemTraversal {
   VoidCallback fn_leave_dir;
   VoidCallback fn_new_file;
   VoidCallback fn_new_symlink;
+  VoidCallback fn_new_socket;
 
   /**
    * Optional callback for all files during recursion to decide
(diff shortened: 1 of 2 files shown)


jblomer:
For the sake of this class, we should ignore special files (FIFO, socket, device files, ...) because we cannot handle them in cvmfs anyway.  Cvmfs internally uses FIFOs and sockets, but this is another story.  This class is ok in only recognizing regular files, symbolic links, and directories.

reneme:
This doesn't contradict with the change @Moliholy proposes here. If the callback for sockets in `FileSystemTraversal` is not set by the using code, it will just ignore sockets. For the `RemoveTree()` it is actually handy for the sake of generality.

jblomer:
But then shouldn't we have rather a callback for "other" file types?  What I find troubling is to introduce a special case (sockets) even though there is no place we can use this information yet.

reneme:
There is, in the `RemoveTree()` (as proposed here). Furthermore, I will need to introduce a *character device* callback for the support of OverlayFS soon. This is already implemented in one of my branches. 

However, you have a point in saying, that we should support all possible cases maybe. `S_ISBLK` and `S_ISFIFO` are the missing pieces, I believe. Or at least treat them as *other*.

jblomer:
Superseded by [pull request 892](https://github.com/cvmfs/cvmfs/pull/892)

PR #885, opened 2015-05-15
Feature: Bulk Change of File Owner on Catalog Level

reneme:
This allows to map the 'user ID' and 'group ID' meta-data fields for an entire repository on catalog level using a simple mapping description (Jira: [CVM-836](https://sft.its.cern.ch/jira/browse/CVM-836)). It is based on map files similar to the ones for `CVMFS_UID_MAP` and `CVMFS_GID_MAP` on the client. For large repositories, a simple `chown -R foo:bar /cvmfs/foo.bar.ch` is not practical, as it would re-compress and re-hash the entire repository content.

Doing this on catalog level is still a heavy-weight operation and is not supposed to be done regularly. The feature is meant to simplify migration of existing repositories to new infrastructure with different user ID requirements.

To run it on an existing repository, one first needs to create two map files, both for UIDs and GIDs and afterwards run:
```bash
cvmfs_server catalog-chown -u $uid_map -g $gid_map $repo_name
``` 
This will recreate the entire catalog tree with the updated UID and GID configuration.

The format of the map files looks like this:
```bash
# map UIDs 137 and 138 to 1001
137 1001
138 1001

# swap UID 1002 and 1001
1001 1002
1002 1001

# wildcard: map all others to 1000
*   1000
```

*Note:* I added the class `IntegerMap<>` to encapsulate the handling of such map files. The file format is backward compatible to the files used in the client, but a bit more agnostic and versatile. In particular it can deal with multiple spaces as field separator and knows about a 'map all to this' wild-card. Furthermore it comes with unit tests.

**TODO:** Potentially this should be re-used for the aforementioned client-side UID/GID maps which currently use a pretty simplistic read function.

### DIFF
--- a/cvmfs/uid_map.h
+++ b/cvmfs/uid_map.h
@@ -5,10 +5,12 @@
 #ifndef CVMFS_UID_MAP_H_
 #define CVMFS_UID_MAP_H_
 
+#include <sys/types.h>
+
 #include <cerrno>
 #include <map>
+#include <string>
 #include <vector>
-#include <sys/types.h>
 
 #include "util.h"
 
@@ -37,12 +39,12 @@ class IntegerMap {
   }
 
   bool Contains(const T k) const {
-    assert (IsValid());
+    assert(IsValid());
     return map_.find(k) != map_.end();
   }
 
   T Map(const T k) const {
-    assert (IsValid());
+    assert(IsValid());
     typename map_type::const_iterator i = map_.find(k);
     if (i != map_.end()) {
       return i->second;
@@ -79,7 +81,7 @@ class IntegerMap {
       }
 
       std::vector<std::string> components = SplitString(line, ' ');
-      FilterEmptyStrings(components);
+      FilterEmptyStrings(&components);
       if (components.size() != 2    ||
           !IsNumeric(components[1]) ||
           (components[0] != "*" && !IsNumeric(components[0]))) {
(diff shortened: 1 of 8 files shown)


cvmfs/uid_map.h
```diff
@@ -0,0 +1,183 @@
...
+  size_t RuleCount()  const { return map_.size();        }
+
+  T GetDefault() const { assert(has_default_value_); return default_value_; }
+  const map_type& GetRuleMap() const { return map_; }
+
+ protected:
+  bool ReadFromFile(const std::string &path) {
+    FILE *fmap = fopen(path.c_str(), "r");
+    if (!fmap) {
+      LogCvmfs(kLogUtility, kLogDebug, "failed to open %s (errno: %d)",
```
jblomer: This requires `#include <logging.h>`

cvmfs/uid_map.h
```diff
@@ -0,0 +1,183 @@
...
+      ++line_number;
+      line = Trim(line);
+      if (line.empty() || line[0] == '#') {
+        continue;
+      }
+
+      std::vector<std::string> components = SplitString(line, ' ');
+      FilterEmptyStrings(&components);
+      if (components.size() != 2    ||
+          !IsNumeric(components[1]) ||
```
jblomer: I think this has just been removed from util.cc

reneme: Fixed by using `sanitizer::IntegerSanitizer`. That will allow negative UIDs which are apparently common on some UNIX systems like HP-UX. Questionable if we will ever support that, but who knows.

cvmfs/uid_map.h
reneme: Fixed, thanks.

PR #887, opened 2015-05-19
Bugfix/osx 1010 defines

PerilousApricot:
Like I said in the commit message, this is almost certainly the wrong place. If you want, I can figure out the cmake-fu to test for the existence of these #defines.


jblomer:
It might be better to disable building the server tools on OS X altogether.  The binaries won't make much sense anyway, for various reasons OS X is not a supported platform for the cvmfs server tools.  For now, you can use `-DBUILD_SERVER=no` in cmake.  @reneme Can you disable the server on OS X?

jblomer:
Closing as [Pull Request 889](https://github.com/cvmfs/cvmfs/pull/889) was merged.

PR #892, opened 2015-05-20
util.cc tests

Moliholy:
It includes the fix in fs_traversal.h so that I can remove sockets from the directory where I am working.


cvmfs/util.cc
```diff
@@ -1125,6 +1095,9 @@ string ReplaceAll(const string &haystack, const string &needle,
   string result(haystack);
   size_t pos = 0;
   const unsigned needle_size = needle.size();
+  if(needle == "")
+    return result;
```
Moliholy: Otherwise it enters in an infinite loop

jblomer:
Looks good in a first skim!

Now, before we go into details, there are the usual things before submitting a pull request:
1. This one you'll most likely figure out yourself
2. At least on my machine, the T_Util unit tests do not run but produce the error

    [ RUN      ] T_Util.BlockSignal
    User defined signal 1

Regarding the `fs_traversal.h`, I'll close the other (pull request)[https://github.com/cvmfs/cvmfs/pull/884] but please take the comments into account and make the list of supported file types complete including `S_ISBLK` and `S_ISFIFO`

cvmfs/util.cc
```diff
@@ -927,15 +906,6 @@ bool HasSuffix(
 }
 
 
-bool IsNumeric(const std::string &str) {
```
reneme: In a currently pending [Pull Request](https://github.com/cvmfs/cvmfs/pull/885) I am actually [using this function](https://github.com/cvmfs/cvmfs/pull/885/files#diff-e97df2d1b1380e814faaf512cc866e46R142). Please don't remove it.

test/unittests/t_util.cc
```diff
@@ -209,3 +307,853 @@ TEST(T_Util, IsHttpUrl) {
...
+  ReadPipe(fd_stdout, buffer, path.length());
+  string result(buffer);
+
+  EXPECT_EQ(path, result);
+  close(fd_stdin);
+  close(fd_stdout);
+  close(fd_stderr);
+}
+
+TEST_F(T_Util, ManagedExec) {
```
reneme: To my own little surprise, I found [this](https://github.com/cvmfs/cvmfs/blob/devel/test/unittests/t_managed_exec.cc). O.o I actually jotted together a couple of unit tests for `ManagedExec` back in 2013.

jblomer:
I think we have this already in the sanitizer classes.  Wouldn't it make more sense to take it from there?

cvmfs/util.cc
jblomer: I think we have this already in the sanitizer classes. Wouldn't it make more sense to take it from there?

reneme: Ah, sweet... I didn't know about that thing. Jep, I'm going to switch to that then.

Moliholy:
Updated to the latest version of devel, including René's PR (https://github.com/cvmfs/cvmfs/pull/906).
Now all tests pass.

Moliholy:
Passed cpplint through fs_traversal.h

cvmfs/util.cc
```diff
@@ -822,13 +823,6 @@ string StringifyTimeval(const timeval value) {
 }
 
 
-string StringifyIpv4(const uint32_t ip4_address) {
```
jblomer: Should be removed from the header, too

cvmfs/util.cc
```diff
@@ -875,21 +869,6 @@ uint64_t String2Uint64(const string &value) {
 }
 
 
-uint64_t HexString2Uint64(const string &value) {
```
jblomer: Should be removed from the header, too

cvmfs/util.cc
```diff
@@ -875,21 +869,6 @@ uint64_t String2Uint64(const string &value) {
...
-  uint64_t result;
-  sscanf(value.c_str(), "%"PRIx64, &result);
-  return result;
-}
-
-
-int HexDigit2Int(const char digit) {
```
jblomer: Should be removed from the header, too

cvmfs/util.cc
jblomer: Should be removed from the header, too

test/unittests/t_util.cc
```diff
@@ -32,7 +38,94 @@ class ThreadDummy {
 };
 
 
-TEST(T_Util, ThreadProxy) {
+class T_Util : public ::testing::Test {
+ protected:
+  virtual void SetUp() {
+    const bool retval = MkdirDeep(string(sandbox), 0700);
```
jblomer: It would be better to use CreateTempDir here for the sandbox, which ensures a fresh directory is used

test/unittests/t_util.cc
```diff
@@ -32,7 +38,94 @@ class ThreadDummy {
...
+  }
+
+  virtual void TearDown() {
+    const bool retval = RemoveTree(string(sandbox));
+    ASSERT_TRUE(retval) << "failed to remove string(sandbox)";
+  }
+
+ protected:
+  static void WriteBuffer(int fd) {
+    sleep(1);
```
jblomer: Is this required?  Can it be replaced by `SafeSleepMs`?

test/unittests/t_util.cc
```diff
@@ -32,7 +38,94 @@ class ThreadDummy {
...
+    return t;
+  }
+
+  static void SendSignal(int signal) {
+    BlockSignal(signal);
+    sleep(1);
+    kill(getpid(), signal);
+  }
+
+  static const char sandbox[];
```
jblomer: This one and the following don't need to be static, do they?  Can't this just be strings that are initialized in the test fixture setup?

test/unittests/t_util.cc
```diff
@@ -215,3 +314,855 @@ TEST(T_Util, IsHttpUrl) {
...
+
+  EXPECT_STREQ(to_write, static_cast<const char*>(buffer_output));
+  ASSERT_DEATH(MakePipe(static_cast<int*>(NULL)), ".*");
+  free(buffer_output);
+  ClosePipe(fd);
+}
+
+TEST_F(T_Util, WritePipe) {
+  int fd[2];
+  void *buffer_output = calloc(20, sizeof(char));
```
jblomer: Here and for the other uses of `calloc`, better use scalloc from smalloc.h

test/unittests/t_util.cc
Moliholy: Yes, of course they can. However they will always be initialized the same way and remain constant... so just for performance I thought to put them as constant strings, but this triggered warnings from cpplint, so I changed them to const char[]. But yes, they could perfectly be const string attributes, which maybe would make things easier in the rest of the code, avoiding many conversions to string instances

test/unittests/t_util.cc
```diff
@@ -215,3 +314,855 @@ TEST(T_Util, IsHttpUrl) {
...
+TEST_F(T_Util, LockFile) {
+  string filename = string(sandbox) + "/lockfile.txt";
+  int retval;
+  int fd;
+
+  EXPECT_EQ(-1, LockFile("/fakepath/fakefile.txt"));
+  EXPECT_LE(0, fd = LockFile(filename));
+
+  tbb::tbb_thread thread(LockFileTest, filename, &retval);
+  sleep(2);
```
jblomer: That's a bit long.  Use `SafeSleepMs` with 100ms.  I don't see a way to avoid the waiting altogether, unfortunately.  We should consider adding a `#define CVMFS_TESTING` to cvmfs.

test/unittests/t_util.cc
Moliholy: Yes, it could. I was trying not to use functions that are going to be tested in order to avoid bug creations, but with this one there shouldn't be any problem.

test/unittests/t_util.cc
```diff
@@ -215,3 +314,855 @@ TEST(T_Util, IsHttpUrl) {
...
+  free(buffer_output);
+  ClosePipe(fd);
+}
+
+TEST_F(T_Util, ReadHalfPipe) {
+  int fd[2];
+  void *buffer_output = calloc(20, sizeof(char));
+  MakePipe(fd);
+
+  tbb::tbb_thread writer(WriteBuffer, fd[1]);
```
jblomer: What happens if you write without sleeping?

test/unittests/t_util.cc
```diff
@@ -215,3 +314,855 @@ TEST(T_Util, IsHttpUrl) {
...
+
+  EXPECT_FALSE(ParseKeyvalPath("/path/that/does/not/exists.txt", &map));
+  EXPECT_FALSE(ParseKeyvalPath(string(sandbox) + "/" + big_file, &map));
+  EXPECT_TRUE(ParseKeyvalPath(string(sandbox) + "/" + content_file, &map));
+}
+
+TEST_F(T_Util, DiffTimeSeconds) {
+  struct timeval start = CreateTimeval(2, 900000);
+  struct timeval end = CreateTimeval(3, 200000);
+  EXPECT_EQ(0.3, DiffTimeSeconds(start, end));
```
jblomer: There are special macros for comparing floating point numbers.  Due to rounding errors, you can never by sure about the exact result of floating point arithmetic.

test/unittests/t_util.cc
```diff
@@ -215,3 +314,855 @@ TEST(T_Util, IsHttpUrl) {
...
+
+  start = CreateTimeval(8, 400000);
+  end = CreateTimeval(1, 800000);
+  EXPECT_NE(-6.6, DiffTimeSeconds(start, end));
```
jblomer: What is this supposed to test?

test/unittests/t_util.cc
```diff
@@ -32,7 +38,94 @@ class ThreadDummy {
...
+  static timeval CreateTimeval(int64_t tv_sec, int64_t tv_usec) {
+    timeval t;
+    t.tv_sec = tv_sec;
+    t.tv_usec = tv_usec;
+    return t;
+  }
+
+  static void SendSignal(int signal) {
+    BlockSignal(signal);
+    sleep(1);
```
jblomer: It's risky to rely on sleep.  Try to figure out if you find an order of events and threads that doesn't require sleep.

test/unittests/t_util.cc
jblomer: Very well, @moliholy can you make these unit tests part of t_util.cc?

test/unittests/t_util.cc
```diff
@@ -215,3 +314,855 @@ TEST(T_Util, IsHttpUrl) {
...
+TEST_F(T_Util, StopWatch) {
+  StopWatch sw;
+  ASSERT_DEATH(sw.Stop(), ".*");
+  sw.Start();
+  ASSERT_DEATH(sw.Start(), ".*");
+  sw.Stop();
+  ASSERT_LT(0.0, sw.GetTime());
+  ASSERT_DEATH(sw.Stop(), ".*");
+  sw.Reset();
+  ASSERT_EQ(0.0, sw.GetTime());
```
jblomer: See above on floating point comparison

jblomer:
Looks quite good!  Please check the minor comments.

test/unittests/t_util.cc
Moliholy: That the difference is not negative. In principle one would expect to find a negative difference, but the functions is not supposed to return a negative value. However it could be changed for EXPECT_GT(0, DiffTimeSeconds(start, end)

test/unittests/t_util.cc
reneme: Couldn't you just put a `Future<bool>` into LockFileTest that gets filled once the thread has started executing and put a `Future<bool>::Get()` right under the thread creation here. Or am I missing something?

test/unittests/t_util.cc
Moliholy: Inside WriteBuffer it waits, but in the other thread.
If it wouldn't wait then the writer could write everything before the reader reads it. The idea of the ReadHalfPipe function is to read something from the pipe even if the writer hasn't started to write yet. For that there _must_ be something in the pipe, and the idea of this test was to check what happens if there is nothing written. Of course it indefinitely waits until the writer (atomically) flushes what it has written, and then the reader consumes it.

test/unittests/t_util.cc
Moliholy: Updated using EXPECT_DOUBLE_EQ macro. Thanks for the warning!

test/unittests/t_util.cc
Moliholy: You don't want to release the main thread when the other is active, but when the other is blocked in the function we are testing here. But if the other is blocked... how is he going to tell the main thread that he actually is? They both have to wait until each other is ready, but at the same time there is no way to warn them.

test/unittests/t_util.cc
Moliholy: Funny thing I've realized right now when using CreateTempDir: it doesn't use the second parameter. Temporary directories are always created using by default 0700 permissions.

test/unittests/t_util.cc
Moliholy: Fixed using scalloc.

test/unittests/t_util.cc
Moliholy: Fixed the way we mentioned before

Moliholy:
Updated.
I have removed most of the "sleeps" because I found other ways to implement it (for signal blocking concretely) or because they were not really necessary (ReadHalfPipe).
Unittests and cpplint passed

test/unittests/t_util.cc
Moliholy: Ops, I skipped this comment, I'll do it right now!

Moliholy:
Removed t_managed_exec.cc. Its tests have been added to t_util.cc.
Cpplint and unittests successfully passed

cvmfs/fs_traversal.h
```diff
@@ -182,6 +185,18 @@ class FileSystemTraversal {
...
                  path.c_str(), dit->d_name);
         Notify(fn_new_symlink, path, dit->d_name);
+      } else if (S_ISSOCK(info.st_mode)) {
+        LogCvmfs(kLogFsTraversal, kLogVerboseMsg, "passing socket %s/%s",
+                 path.c_str(), dit->d_name);
+        Notify(fn_new_socket, path, dit->d_name);
+      } else if (S_ISBLK(info.st_mode)) {
+        LogCvmfs(kLogFsTraversal, kLogVerboseMsg,
+            "passing blocked device %s/%s", path.c_str(), dit->d_name);
+        Notify(fn_new_socket, path, dit->d_name);
```
jblomer: this should not call `fn_new_socket` but `fn_new_block`.

cvmfs/fs_traversal.h
```diff
@@ -182,6 +185,18 @@ class FileSystemTraversal {
...
+      } else if (S_ISFIFO(info.st_mode)) {
+        LogCvmfs(kLogFsTraversal, kLogVerboseMsg, "passing FIFO %s/%s",
+                 path.c_str(), dit->d_name);
+        Notify(fn_new_socket, path, dit->d_name);
```
jblomer: Same as above, but for fifo.  Please also add test cases that catch this problem.

Moliholy:
fs_traversal modified and the corresponding tests have been added.
Cpplint and unitests passed.

test/unittests/t_fs_traversal.cc
```diff
@@ -26,7 +27,10 @@ class T_FsTraversal : public ::testing::Test {
       File,
       Symlink,
       Untouched,
-      Unspecified
+      Unspecified,
+      Socket,
+      BlockedDevice,
```
jblomer: Not a show stopper for the merging, but it is actually a "block device" (because data is transferred in blocks, it's not blocked).

test/unittests/t_util.cc
```diff
@@ -215,3 +328,960 @@ TEST(T_Util, IsHttpUrl) {
...
+  sw.Stop();
+  ASSERT_LT(0.0, sw.GetTime());
+  ASSERT_DEATH(sw.Stop(), ".*");
+  sw.Reset();
+  ASSERT_DOUBLE_EQ(0.0, sw.GetTime());
+}
+
+TEST_F(T_Util, SafeSleepMs) {
+  StopWatch sw;
+  unsigned time = 1000;
```
jblomer: This can be decreased to 50 so that the unit tests run faster.  I'll do it during merging.

Moliholy:
Renamed to block_dev instead of blocked_dev

PR #908, opened 2015-05-26
help me clean up no-op commits

DrDaveD:
I don't know how else to remove these commits + reverts from continuing to show up in my devel branch than to get them into the upstream devel branch.  If you know another way, please let me know, otherwise please merge this no-op pull request.


reneme:
I'm frankly not really sure what the problem is. But can't you just discard your devel branch and recreate it starting based on the upstream devel branch?

```bash
git checkout origin/devel
git branch -d devel
git checkout -b devel
```

DrDaveD:
Thanks, Rene, that basic idea worked. I had already pushed it to origin, so instead I checked it out from upstream/devel.  I had to temporarily change the settings on my github repo to have a different default branch than devel in order to push the delete of the origin branch.

The problem I was having was that every time I create a new branch it was off my devel branch so these would keep popping up.

PR #918, opened 2015-05-28
Cleanup: Content Hash and Derived Storage Path Handling

reneme:
This is the follow-up to a [Pull Request](https://github.com/cvmfs/cvmfs/pull/602) that went into 2.1.20 which made the `Digest<>` aware of the hash suffix (i.e. **C**, **H**, **X**, ...).

Currently, in many places you would see code like:
```C++
// const shash::Any hash("b09b1bcf01ee6414525f3307d1f0bbf7a6d8ade3");
const string url = "/data" + hash.MakePathExplicit(1, 2) + "C";
```
This patch aims for the elimination of shared knowledge of how CAS paths are derived from hashes. Namely, the two magic numbers `1` and `2` in `MakePathExplicit()` and the need of appending a suffix to it, depending on what is going to be fetched. Furthermore `MakePathExplicit()` did emit an absolute path like */b0/9b1...* for unclear reasons.

This patch changes the above call to:
```C++
// const shash::Any hash("b09b1bcf01ee6414525f3307d1f0bbf7a6d8ade3", shash::kSuffixCatalog);
const string url = "/data/" + hash.MakePath();
```
The hash suffix is specified only once when creating the hash object, because this should be the place where one knows most about the object that will be referenced by this specific hash. Furthermore hashes are never created 'out of the blue', but fall out of other data structures like `Catalog`, `Manifest` or `History`. These classes can take care of creating hash objects with the right suffixes internally, benefitting information hiding.

Furthermore the magic numbers (`1`, `2`) are removed and hidden inside `MakePath()` because they are the same throughout the code base anyway. Also `MakePath()` now generates a relative path like *b0/9b1...C* that can be easily prepended by any other paths or URLs in the using code.

There is a second method `Digest<>::MakePathWithoutSuffix()` that omits the hash suffix (if there is one specified) which is necessary for the local file cache accesses.

Stringifying hashes with `Digest<>::ToString()` does omit the suffix while `Digest<>::ToStringWithSuffix()` appends it. I argue that the suffix is primarily interesting for the CAS path creation and should not appear in the default string representation of a hash.

I am aware that this is quite a surgery in the code, but it greatly improves readability and flexibility. For instance, passing an `shash::Any` to an `ObjectFetcher::Fetch()` method would now be universal, since the caller doesn't need to be aware of the type of object to be fetched anymore. Also it simplifies many method signatures where `shash::Any` and `shash::Suffix` were handed in separately.

Throughout the code `Digest<>::MakePathExplicit()` was used before and I changed the signature of this method, making sure to properly find all occurrences.

### DIFF
--- a/cvmfs/swissknife_pull.cc
+++ b/cvmfs/swissknife_pull.cc
@@ -76,25 +76,21 @@ string              *preload_cachedir = NULL;
 }  // anonymous namespace
 
 
-static bool Peek(const string &remote_path, const char suffix)
-{
+static bool Peek(const string &remote_path) {
   if (preload_cache) {
     // Strip "data"
+    // TODO: Fix that! It's ugly and error prone!
     return FileExists(*preload_cachedir + remote_path.substr(4));
   } else {
-    string real_remote_path = remote_path;
-    if (suffix != 0)
-      real_remote_path.push_back(suffix);
-    return spooler->Peek(real_remote_path);
+    return spooler->Peek(remote_path);
   }
 }
 
 
-static void Store(const string &local_path, const string &remote_path,
-                  const char suffix)
-{
+static void Store(const string &local_path, const string &remote_path) {
   if (preload_cache) {
     // Strip "data"
+    // TODO: Fix that! taking assumptions on the path? Wut?!
     const string dest_path = *preload_cachedir + remote_path.substr(4);
     string tmp_dest;
     FILE *fdest = CreateTempFile(dest_path, 0660, "w", &tmp_dest);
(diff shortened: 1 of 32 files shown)


reneme:
Rebased to latest devel (merge conflicts after merging [919](https://github.com/cvmfs/cvmfs/pull/919)), fixed an issue with the generation of chunked file names and removed the last superfluous hash suffix usages.

cvmfs/swissknife_pull.cc
```diff
@@ -76,25 +76,21 @@ string              *preload_cachedir = NULL;
 }  // anonymous namespace
 
 
-static bool Peek(const string &remote_path, const char suffix)
-{
+static bool Peek(const string &remote_path) {
   if (preload_cache) {
     // Strip "data"
+    // TODO(rene): Fix that! It's ugly and error prone!
     return FileExists(*preload_cachedir + remote_path.substr(4));
```
jblomer: I haven't tried it but I think things will break here.  When the cache is preloaded, path names are always without suffix.  This is different from Stratum 1 replication.

reneme: Seems to be true, the 'preload cache' integration test 541 fails... I'm having a look.

PR #920, opened 2015-05-29
Solved some compiling and test problems in some platforms

Moliholy:
* Overflow in T_Util.StringifyTime
* \0 character in T_Util.Shell
* Check that a directory is readable in T_FsTraversal.BlockDevice to avoid insufficient permissions errors.

By the way, when running the non-slow tests now I see this in the end:

[----------] Global test environment tear-down
[==========] 492 tests from 50 test cases ran. (51419 ms total)
[  PASSED  ] 492 tests.

  YOU HAVE 1 DISABLED TEST

And it looks like that disabled test is not one of the new ones. Am I missing something?

### DIFF
--- a/cvmfs/fs_traversal.h
+++ b/cvmfs/fs_traversal.h
@@ -46,7 +46,7 @@ class FileSystemTraversal {
   VoidCallback fn_new_file;
   VoidCallback fn_new_symlink;
   VoidCallback fn_new_socket;
-  VoidCallback fn_new_blocked_dev;
+  VoidCallback fn_new_block_dev;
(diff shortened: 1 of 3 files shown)


reneme:
The disabled test is `T_Dns::CaresResolverBadResolver` which did show unclear behaviour that's why we decided to disable it for the moment.

reneme:
By the way: The Mac [still doesn't work](https://phsft-jenkins.cern.ch/view/CVMFS/job/CvmfsIncrementalBuild/) and there is a new CppLint warning. ;-)

PR #926, opened 2015-06-01
FIX: Snapshotting/Preloading Code after Hash Cleanup

reneme:
This retro-fixes the problem @jblomer pointed out pointed out [here](https://github.com/cvmfs/cvmfs/pull/918#discussion_r31390901). It ended up in a new pull request since the fix requires some refactoring which is worth reviewing individually.

Changes done:

1. `Catalog::AllChunksNext()` now returns hashes with proper suffix annotation instead of `hash` and `type` individually
2. `catalog::ChunkTypes` was removed and replaced by `shash::Suffix` directly (this depends on `shash::Suffix` being a `char` and should be fixed in the future, see TODO)
3. Consolidate `Peek()` and `Store()` functions in `swissknife pull` that depend on `preload_cache`
4. Extend `struct ChunkJob` to encapsulate the hash (de)serialisation

### DIFF
--- a/cvmfs/swissknife_pull.cc
+++ b/cvmfs/swissknife_pull.cc
@@ -76,41 +76,49 @@ string              *preload_cachedir = NULL;
 }  // anonymous namespace
 
 
+static std::string MakePath(const shash::Any &hash) {
+  return (preload_cache)
+    ? *preload_cachedir + "/" + hash.MakePathWithoutSuffix()
+    : "data/"           + hash.MakePath();
+}
+
+
 static bool Peek(const string &remote_path) {
-  if (preload_cache) {
-    // Strip "data"
-    // TODO(rene): Fix that! It's ugly and error prone!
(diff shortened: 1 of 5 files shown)


cvmfs/swissknife_pull.cc
```diff
@@ -76,36 +107,40 @@ string              *preload_cachedir = NULL;
...
+
 static bool Peek(const string &remote_path) {
-  if (preload_cache) {
-    // Strip "data"
-    // TODO(rene): Fix that! It's ugly and error prone!
-    return FileExists(*preload_cachedir + remote_path.substr(4));
-  } else {
-    return spooler->Peek(remote_path);
-  }
+  return (preload_cache) ? FileExists(remote_path)
```
jblomer: Here, `FileExists(remote_path)` would still mistakenly add the suffix, no?

jblomer: Ah, never mind.  I see it.

PR #927, opened 2015-06-01
TEST: IntegerSanitizer

reneme:
Added a test for `IntegerSanitizer` and fixed its implementation to allow for negative numbers.

### DIFF
--- a/cvmfs/sanitizer.h
+++ b/cvmfs/sanitizer.h
@@ -52,21 +52,18 @@ class InputSanitizer {
 class AlphaNumSanitizer : public InputSanitizer {
  public:
   AlphaNumSanitizer() : InputSanitizer("az AZ 09") { }
-  virtual ~AlphaNumSanitizer() { }
 };
 
 
 class RepositorySanitizer : public InputSanitizer {
  public:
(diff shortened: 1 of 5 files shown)


cvmfs/sanitizer.h
```diff
@@ -55,10 +63,25 @@ class RepositorySanitizer : public InputSanitizer {
...
-  IntegerSanitizer() : InputSanitizer("09") { }
+  PositiveIntegerSanitizer() : InputSanitizer("09") { }
+
+ protected:
+  virtual bool Sanitize(std::string::const_iterator   begin,
+                        std::string::const_iterator   end,
+                        std::string                  *filtered_output) const;
+};
+
+class IntegerSanitizer : public PositiveIntegerSanitizer {
```
jblomer: That's a strange class hierarchy.  An `IntegerSanitizer` is not a `PositiveIntegerSanitizer`.

reneme: My rationale was that positive integers are somewhat a subset of all integers. But I agree, its somewhat counter-intuitive... going to change it.

reneme: Fixed. Also I removed the superfluous d'tors from the class definitions.

PR #934, opened 2015-06-04
Cleanup: Use Database<>::GetProperty<>() and Database<>::SetProperty()

reneme:
This replaces the explicit SQL queries in `catalog.cc` and `catalog_rw.cc` to access information in the `properties` table with calls to `Database<>::GetProperty<>()` and `Database<>::SetProperty()` as requested in [CVM-720](https://sft.its.cern.ch/jira/browse/CVM-720).

### DIFF
--- a/cvmfs/catalog.h
+++ b/cvmfs/catalog.h
@@ -231,6 +231,7 @@ class Catalog : public SingleCopy {
   Counters& GetCounters() { return counters_; }
 
   inline const CatalogDatabase &database() const { return *database_; }
+  inline       CatalogDatabase &database()       { return *database_; }
   inline void set_parent(Catalog *catalog) { parent_ = catalog; }
 
   void ResetNestedCatalogCache();
--- a/cvmfs/catalog.cc
+++ b/cvmfs/catalog.cc
@@ -177,14 +177,10 @@ bool Catalog::OpenDatabase(const string &db_path) {
 
   // Get root prefix
   if (IsRoot()) {
-    Sql sql_root_prefix(database(), "SELECT value FROM properties "
-                                    "WHERE key='root_prefix';");
-    if (sql_root_prefix.FetchRow()) {
-      root_prefix_.Assign(
-        reinterpret_cast<const char *>(
-          sql_root_prefix.RetrieveText(0)),
-        strlen(reinterpret_cast<const char *>(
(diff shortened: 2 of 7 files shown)


cvmfs/catalog.h
```diff
@@ -231,6 +231,7 @@ class Catalog : public SingleCopy {
   Counters& GetCounters() { return counters_; }
 
   inline const CatalogDatabase &database() const { return *database_; }
+  inline       CatalogDatabase &database()       { return *database_; }
```
jblomer: Would be better to return a pointer.  What comes back from this function has value syntax but pointer semantics.

reneme: Those are *protected* accessors to be used by subclasses of `Catalog`. For example all `Sql` classes take a `Database&` in their constructors. I believe rather should the `database_` member in `Catalog` not be a pointer as it is now.

I need to look into the feasibility of this change, since currently one can have a `Catalog` object without an attached `Database`, which I find odd anyway. But I believe it was the reason for defining the `database_` member as a pointer.

jblomer: Yes, I am aware they are protected.  What's the problem in returning a pointer `CatalogDatabase *`?

reneme: I find it odd to dereference these pointer to pass them into the `Sql` query classes c'tors... just a matter of taste.

reneme: I just find the 'value semantics' for a *protected* accessor less odd than dereferencing pointers to gain `value semantics` selectively.

jblomer: Please let's avoid non-const references in general.  Otherwise, when you read the code in a derived class, you don't know if what comes out of `database()` is modified or not.

reneme: Fine. However, I'm not sure what you mean. In a derived class `database()` is used to get an inherited field, either `mutable` or `const` depending on the calling method's const-ness. Or do you mean something else?

jblomer: Let me think a little bit about this and let's discuss this after the DPHEP workshop.

jblomer: I think I would prefer using `Database *` in general (e.g. also in the `Sql` interface).  But this is a slightly larger change and not for this PR.

PR #936, opened 2015-06-05
Statistics comparison

Moliholy:
Updated the python scripts to:

* Enable statistics comparisons between two branches
* Improve readability
* Fix bugs
* Allow users to generate csv files with their benchmark results


python/statistics_comparison.py
```diff
@@ -0,0 +1,169 @@
...
+                           help="Original branch to compare to")
+    argparser.add_argument("--external_branch",
+                           default="devel",
+                           required=False, type=str,
+                           help="External branch to compare to")
+    argparser.add_argument("external_repo", type=str)
+    return argparser.parse_args()
+
+
+def get_benchmark_list(benchmark_string):
```
jblomer: Would it be possible to make this in a way that doesn't require knowledge about the list of benchmarks?  Essentially I would provide `benchmarks/XYZ-mybenchmark` which you wouldn't need to further parse or modify,

Moliholy: Of course, it would actually be simpler. I'll modify it

Moliholy:
Updated.
After some more clean-up steps I have successfully tested the functionality and fixed some errors in the docker container (this time not creating new ones in the process). Now it works for all four benchmarks we have. A sample generated file is pasted here:

image

PR #938, opened 2015-06-08
FIX: T_FsTraversal unit test

reneme:
On some platforms both `T_FsTraversal::CharacterDevice` and `T_FsTraversal::BlockDevice` were failing due to `stat()`-ing dangling symlinks in `/dev/**`.

### DIFF
--- a/test/unittests/t_fs_traversal.cc
+++ b/test/unittests/t_fs_traversal.cc
@@ -686,7 +686,7 @@ class CustomDelegate {
                const std::string  &obj_name,
                struct stat        *s) const {
     const std::string file = root_path + "/" + relative_path + "/" + obj_name;
-    const int retval = stat(file.c_str(), s);
+    const int retval = lstat(file.c_str(), s);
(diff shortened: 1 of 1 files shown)


test/unittests/t_fs_traversal.cc
```diff
@@ -686,8 +686,8 @@ class CustomDelegate {
                const std::string  &obj_name,
                struct stat        *s) const {
     const std::string file = root_path + "/" + relative_path + "/" + obj_name;
-    const int retval = stat(file.c_str(), s);
-    ASSERT_EQ(0, retval) << "cannot stat '" << file << "'";
+    const int retval = lstat(file.c_str(), s);
```
jblomer: Might be better to use `platform_lstat` anyway.

reneme: [Absolutely](https://github.com/cvmfs/cvmfs/pull/939)! Forgot about that...

PR #950, opened 2015-06-10
Archlinux dev

cburgard:
Refactored a couple of lines/paths to allow smooth compilation and installation on Archlinux.
Tested to work flawlessly under recent archlinux, not tested on any other system.


CMakeLists.txt
```diff
@@ -93,10 +97,17 @@ endif (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^arm.*$")
...
 else (MACOSX) # --> Linux
-  if (DEBIAN)
+  if (DEBIAN OR ARCHLINUX)
+    if (ARCHLINUX)
+      set (CMAKE_INSTALL_BINDIR "/usr/bin")
+    else (ARCHLINUX)
+      set (CMAKE_INSTALL_BINDIR "/sbin")
+    endif (ARCHLINUX)
     set (CMAKE_INSTALL_LIBDIR "lib")
   else (DEBIAN) # --> RedHat, Fedora, CentOS, SuSE
```
reneme: For some obscure reason CMake wants the same condition to be repeated in `else` and `endif`, please add your `OR ARCHLINUX` here.

CMakeLists.txt
```diff
@@ -93,10 +97,17 @@ endif (${CMAKE_SYSTEM_PROCESSOR} MATCHES "^arm.*$")
 #
 if (MACOSX)
   set (CMAKE_INSTALL_LIBDIR "lib")
+  set (CMAKE_INSTALL_BINDIR "/sbin")
```
reneme: Can you rename that from `CMAKE_INSTALL_BINDIR` to `CMAKE_MOUNT_INSTALL_BINDIR`, since it only changes the install location of the mount helper, right?

reneme:
Thanks for the contribution! I've added some minor comments. You can simply add a commit to your local branch and push again, that will automatically update your pull request here.

cburgard:
Done. If you have any other change requests or spot any other potential sources of problems, please let me know.

reneme:
@jblomer: from my side this is ready to merge.

cvmfs/cvmfs_config
```diff
@@ -254,13 +257,14 @@ cvmfs_chksetup() {
     fi
   fi
 
+  AUTOMASTER=$(find /etc -name "auto.master" 2>/dev/null | head -1)
```
jblomer: This can potentially take a long time.  I assume that auto.master is not directly under /etc on archlinux, right?  Can we just test for /etc/auto.master and then for the alternative location?

cvmfs/cvmfs_config
```diff
@@ -1231,9 +1235,13 @@ add_user_to_group_fuse() {
 check_auto_mounter() {
   case $sys_arch in
     Linux )
-      $service autofs status > /dev/null 2>&1
-      return $? ;;
-
+        if [[ $service == *systemctl ]]; then
```
jblomer: Is the compatibility symlink from `service` to `systemctl` missing on archlinux?  If it's there, I'd prefer for the time being to stay with it.  In the longer run, I think this script should have wrapper functions for start/stop/reload services that cover both the binary name of the system manager and the correct parameter syntax.

jblomer:
Many thanks, Carsten, I think we're almost ready to merge.  One thing that's missing is in the `chksetup` routine in cvmfs_config.  Here, the script is explicitly testing for /sbin/mount.cvmfs and will raise a false positive on arch linux.

cvmfs/cvmfs_config
cburgard: Archlinux does not know any binary named `service`. In addition to that, the ordering of arguments accepted by systemctl seems to be different than the one used by `service`:
`service status autofs` vs. `systemctl autofs status`.

cvmfs/cvmfs_config
cburgard: Sure. The path is `/etc/autofs/auto.master`. However, using `find` is more flexible, I think. Would it be an option to run the `find` during installation and save the found path in some text file? Otherwise, of course, simply hardcoding the path is simple enough as well.

cvmfs/cvmfs_config
jblomer: I see, alright then let's leave this spot as is.  We'll probably later make another round of modifications to hide the specifics of service management.

cvmfs/cvmfs_config
jblomer: I'd still prefer to use the hard coded paths.  I agree that `find` adds flexibility but I also find it a bit dangerous.  You could imagine funny setups where /etc or some sub path is on a large network volume.  Also, by failing we get to know if a distribution changes the location of auto.master in /etc, which is a good hint for us to take a second look whether something else changed, too.

cburgard: I've changed the code such that it visits the two presupplied paths and should be easy to extend for more.

jblomer:
Merged with minor modifications.  Apologies for the delay.

PR #951, opened 2015-06-10
FIX: cms benchmark termination fixed

Moliholy:
Due to a race condition the cms.cern.ch repository couldn't always unmount. Now there is a 5-second-delay to allow the mounted repository to finish its stuff. 
Also added descriptions when a benchmark fails.

### DIFF
--- a/test/benchmarks/004-cms/main
+++ b/test/benchmarks/004-cms/main
@@ -16,7 +16,7 @@ cvmfs_run_benchmark() {
   cmsRun TTbar_Tauola_8TeV_cfi_GEN_SIM.py # raises a segmentation fault error and says there is no disk space available, but returns 0
   cmsDriver.py TTbar_Tauola_8TeV_cfi -s DIGI,L1,DIGI2RAW,RAW2DIGI,L1Reco,RECO --conditions auto:startup --eventcontent RECOSIM --datatier RECO -n -1 --no_exec --filein file:./TTbar_Tauola_8TeV_cfi_GEN_SIM.root
   cmsRun TTbar_Tauola_8TeV_cfi_DIGI_L1_DIGI2RAW_RAW2DIGI_L1Reco_RECO.py
-  cd ..
+  cd -
 }
 
 cvmfs_run_test() {
--- a/test/test_functions
+++ b/test/test_functions
@@ -1484,7 +1484,6 @@ execute_statistics_loop() {
   local code=0
   local iteration=1
   local PID=0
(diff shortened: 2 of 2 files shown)


jblomer:
Hm, I don't quite understand the `sleep`.  Isn't `cvmfs_unmount` anyway trying for a while until unmounting succeeds?

Moliholy:
```
cvmfs_umount() {
  repositories=$1

  for r in $(echo $repositories | tr , " "); do
    sudo umount /cvmfs/$r
    local result=$?
    if [ $result -ne 0 ]; then
      lsof /cvmfs/$r
      return 200
    fi

    timeout=5
    while cat /proc/mounts | grep -q " /cvmfs/$r "; do
      if [ $timeout -eq 0 ]; then
        return 101
      fi
      timeout=$(($timeout-1))
      sleep 1
    done
  done

  return 0
}
```

It doesn't. It fails in the first line and returns before reaching the second part of the function and printing the result of lsof /cvmf$r

jblomer:
Ah, I see, it just gives the umount more time.  

Still, this 5 second sleep shouldn't be the right solution.  We need to figure out which process it is that is still sitting on /cvmfs.  Does `fuser` just before the umount return something useful?

Moliholy:
Updated.
I've been running in troubles while trying to fix this. There is clearly a race condition there that doesn't allow the umount, and reproducing the situation is not so easy. For instance, this very morning the Jenkins Bechmark project succeeded! (https://phsft-jenkins.cern.ch/job/CvmfsBenchmark/39/). The problem to find out what is the "problem" there is that after executing lsof, fuser, ps -a and other commands; the race condition disappears, not allowing me to detect it.
After talking to @reneme we thought an intermediate solution would be faster and simpler, not costing that much effort compared to debug this. Basically now the cvmfs_umount function has a few tries with a 1-second-delay between them.

 I have just tested it and works. Here is the output:
```
+ for r in '$(echo $repositories | tr , " ")'
+ ((  result != 0 && times > 0   ))
+ sudo umount /cvmfs/cms.cern.ch
umount: /cvmfs/cms.cern.ch: device is busy.
        (In some cases useful info about processes that use
         the device is found by lsof(8) or fuser(1))
+ result=1
+ times=4
+ '[' 1 -ne 0 ']'
+ '[' 4 -gt 0 ']'
+ sleep 1
+ ((  result != 0 && times > 0   ))
+ sudo umount /cvmfs/cms.cern.ch
+ result=0
```

PR #952, opened 2015-06-10
Python: Verify Signatures of Manifest and Whitelist

reneme:
This allows the verification of the signatures both in the `Manifest` and the `Whitelist`. I am using `M2Crypto` for this, because its both available for RedHat and provides a simple wrapper around libopenssl. What is missing, is the integration of these verifications into a simple method call.

In detail, this introduces `Certificate` as python class, so that one can simply do:
```python
import cvmfs
repo = cvmfs.open_repository("http://cvmfs-stratum-one.cern.ch/cvmfs/altas.cern.ch")
cert = repo.retrieve_certificate()
repo.manifest.verify_signature(cert)

# missing:
repo.verify_whitelist("/etc/cvmfs/keys/cern.ch/cern.ch.pub")
# and/or:
repo.open_repository("http://cvmfs-stratum-one.cern.ch/cvmfs/atlas.cern.ch", "/etc/cvmfs/keys/cern.ch/cern.ch.pub")
```

As always: Additional tests added to the python unit test suite.

### DIFF
--- a/python/cvmfs/__init__.py
+++ b/python/cvmfs/__init__.py
@@ -8,6 +8,7 @@ This file is part of the CernVM File System auxiliary tools.
 from root_file    import IncompleteRootFileSignature
 from manifest     import *
 from whitelist    import *
+from certificate  import *
 from repository   import *
 from availability import *
 from _common      import _split_md5
--- a/python/cvmfs/certificate.py
+++ b/python/cvmfs/certificate.py
@@ -18,7 +18,7 @@ class Certificate(CompressedObject):
         self.openssl_certificate = cert
 
     def __str__(self):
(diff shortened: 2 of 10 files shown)


jblomer:
Just keep in mind that we support RIPEMD-160, too, and soon also SHA-256.  Not necessary for the first iteration, just for not introducing road blocks.

reneme:
Yep, I have it in mind, just want to make it work in the first place. Thanks. I believe the client can also use other means of verification than X.509, no?

jblomer:
That is pretty much fixed, but the RSA signature might at some point be exchanged for PKCS7 signatures.  There is some code but that's not officially supported yet.

PR #966, opened 2015-06-15
FIX: manual cache cleanup for the benchmarks

Moliholy:
Removing the whole cache directory only works if cvmfs is fully stopped. Otherwise removing the cache produces errors that made the last benchmarks fail (https://phsft-jenkins.cern.ch/view/CVMFS/job/CvmfsBenchmark/).  With this fix it should work.

### DIFF
--- a/test/test_functions
+++ b/test/test_functions
@@ -1351,7 +1351,10 @@ EOF
 clean_benchmark_cache() {
   local cachedir="${CVMFS_OPT_CACHEDIR}/${FQRN}"
   if [ -d "$cachedir" ] && [ "x$cachedir" != x ]; then
-    rm -rf $cachedir/*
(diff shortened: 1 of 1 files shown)


test/test_functions
```diff
@@ -1351,7 +1351,10 @@ EOF
 clean_benchmark_cache() {
   local cachedir="${CVMFS_OPT_CACHEDIR}/${FQRN}"
   if [ -d "$cachedir" ] && [ "x$cachedir" != x ]; then
-    rm -rf $cachedir/*
+  list=$(ls $cachedir | grep -P "^[0-9a-f]{2}$")
+    for dir in $list; do
+      rm -rf "$cachedir/$dir"/*
```
reneme: please use `rm -f` only. there will be no directories to delete

PR #975, opened 2015-06-24
TEST: included most of the sources for the cvmfs_unittests_debug target

Moliholy:
Some of them could not be included because they contain a main function or rely on globals


test/unittests/CMakeLists.txt
```diff
@@ -167,19 +174,109 @@ set (CVMFS_UNITTEST_SOURCES
...
+  ${CVMFS_SOURCE_DIR}/catalog.cc
+  ${CVMFS_SOURCE_DIR}/catalog_counters.cc
+  ${CVMFS_SOURCE_DIR}/catalog_mgr.cc
+  ${CVMFS_SOURCE_DIR}/catalog_mgr_ro.cc
+  ${CVMFS_SOURCE_DIR}/catalog_mgr_rw.cc
+  ${CVMFS_SOURCE_DIR}/catalog_rw.cc
+  ${CVMFS_SOURCE_DIR}/catalog_sql.cc
+  ${CVMFS_SOURCE_DIR}/compat.cc
+  ${CVMFS_SOURCE_DIR}/compression.cc
+  #${CVMFS_SOURCE_DIR}/cvmfs.cc
```
Moliholy: I have maintained the commented lines in case in the future we can remove the comment, but is that a good idea?

test/unittests/CMakeLists.txt
```diff
@@ -195,6 +292,7 @@ endif (BUILD_UNITTESTS)
 if (BUILD_UNITTESTS_DEBUG)
   add_executable (${PROJECT_TEST_DEBUG_NAME} ${CVMFS_UNITTEST_DEBUG_SOURCES})
   add_dependencies (${PROJECT_TEST_DEBUG_NAME} googletest)
+  add_dependencies (${PROJECT_TEST_DEBUG_NAME} libvjson)
```
Moliholy: This is a necessary dependency to compile json_document.cc, which is anyway at the moment not compiled.

cvmfs/CMakeLists.txt
```diff
@@ -354,10 +354,7 @@ if (BUILD_SERVER)
     add_dependencies (geoip python-geoip)
   endif (GEOIP_BUILTIN)
 
-  include (${VJSON_BUILTIN_LOCATION}/CVMFS-CMakeLists.txt)
```
jblomer: Is it correct to remove the include path but not the dependency?

test/unittests/CMakeLists.txt
jblomer: That looks fine to me.

cvmfs/CMakeLists.txt
Moliholy: It is included in the top-level CMakelist now (https://github.com/cvmfs/cvmfs/pull/975/files#diff-af3b638bc2a3e6c650974192a53c7291R336), so there are no clashes

PR #978, opened 2015-06-25
Channel all accesses to the client cache through a virtual cache manager class

jblomer:
Change the cache and quota module to a cache manager, a quota manager, and a fetcher class.  All accesses to the cache are channeled through the cache manger including sqlite file catalog reads.  To do so, the sqlite vfs can deal with "file names" in the form of "@<file descriptor>".  Also, the download manager can write to an abstract "sink", which can represent an open transaction in a cache manager.

The fetcher encapsulates the logic of downloading and storing a file into the cache if it is not present already. The cache manager owns a quota manager, which can be the NoopQuotaManager. The cache::CatalogManager class moves to the catalog::ClientCatalogManager class.  The ClientCatalogManager uses file catalogs from the local cache, i.e. it uses the fetcher to retrieve file descriptors to the file catalogs.

Add many unit tests to code that was so far not covered by unit tests.


cvmfs/cache.cc
```diff
@@ -107,178 +111,86 @@ atomic_int32 CallGuard::num_inflight_calls_ = 0;
...
-  const string path = GetPathInCache(id);
-  int result = ::open(path.c_str(), O_RDONLY);
-
-  if (result >= 0) {
-    LogCvmfs(kLogCache, kLogDebug, "hit %s", path.c_str());
-    // platform_disable_kcache(result);
-  } else {
-    result = -errno;
-    LogCvmfs(kLogCache, kLogDebug, "miss %s (%d)", path.c_str(), result);
+  void *txn = alloca(this->SizeOfTxn());
```
reneme: Is there a reason that a `Transaction` is not represented as a struct or maybe a hierarchy of structs? Wouldn't that make all this transaction allocation transparent (`SizeOfTxn()`, `alloca()` and `void*`)?

cvmfs/quota.h
```diff
@@ -5,56 +5,437 @@
...
+
+  /**
+   * The last bit in the sequence number indicates if an entry is volatile.
+   * Such sequence numbers are negative and they are preferred during cleanup.
+   * Volatile entries are used for instance for ALICE conditions data.
+   */
+  static const uint64_t kVolatileFlag = 1ULL << 63;
+
+
+  bool InitDatabase(const bool rebuild_database);
```
reneme: Might be useful to encapsulate this into `sql::Database` since the quota database is the only one not managed through this abstract base class, to my knowledge.

cvmfs/catalog_mgr_client.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  if (!attached) {
+    LogCvmfs(kLogCatalog, kLogDebug, "failed to initialize root catalog");
+  }
+
+  return attached;
+}
+
+
+LoadError ClientCatalogManager::LoadCatalog(
```
reneme: This is tightly coupled with the `CacheManager` and even the `QuotaManager` class. I thought the `Fetcher` should actually break this coupling?

jblomer: Tell me how.

jblomer: Without rewriting all of cvmfs.

cvmfs/sink.h
```diff
@@ -0,0 +1,36 @@
...
+namespace cvmfs {
+
+/**
+ * A data sink that behaves like a writable file descriptor with a custom
+ * implementation.
+ *
+ * Currently used by the Fetcher class to redirect writing into a cache manager.
+ *
+ * TODO(jblomer): can all download destinations be implemented by inheriting
+ * from this class?
```
reneme: That'd be awesome!

cvmfs/catalog_mgr_client.cc
reneme: By properly wrapping this into the `Fetcher` object. `CatalogManager` would request a `Catalog` (note: it should obtain a `Catalog` object from `Fetcher` not a file descriptor or a file path!) which it could use right away. `Fetcher` needs to do the right things for retrieving and setting up a catalog like *pin* it in the cache.

The `Catalog` itself should be initialised with a file object (maybe the `Sink`?) pointing to the SQLite file to be used. This way also the interface between `Fetcher` and `Catalog` is minimal and could even be abstracted by templating the `Fetcher` with `CatalogT`. Hence, one could easily reuse the `Fetcher` for `WritableCatalogs` in the server, for example.

When the `CatalogManager` needs to detach a loaded `Catalog`, it would remove it from its internal structures and just *delete* it. Since `Catalog` has its SQLite file wrapped by some smart object (somewhat similar to the `Sink` introduced further down) it can delegate the cache cleanup (*unpin* and friends) to a central place. Note that this way also the `Catalog` doesn't need to have the information of how to deal with its underlying SQLite file.

This is a bit out of the blue of course, but it would potentially decouple the `CatalogManager` completely from the fetching logic (and also the internal `Catalog` logic by the way). I don't see why this would require a rewrite of CVMFS.

jblomer: Yeah, maybe, maybe not. Looks to me mostly like abstraction for the sake of it.  Retrieving a catalog on the server is different from retrieving a catalog on the client.  We don't need to squeeze this into the same generic class.

reneme: I would call it 'loose coupling' or 'separation of concerns'. In fact the fetching is not so much different on the server, only that we have no cache there. Download and cleanup could very well be shared, since the server uses the `DownloadManager` already.

Right now the `ClientCatalogManager` is connected to pretty much everything, since it knows about (at least some of) the internals of `CacheManager`, `PosixCacheManager`, `QuotaManager`, `Fetcher`, `Catalog`, `Manifest`, `manifest::Fetch` and `SignatureManager`. It even knows about the `.cvmfschecksum` file that resides in certain cache directories and that CVMFS usually switches to the cache directory at startup - admittedly, this is marked as a *TODO hack*. Pretty many concerns and lots of knowledge for a single class - or even a single method - if you ask me.

Please don't get me wrong: I just wanted to point that out; I agree that changing all that would be another major effort. The changes in this Pull Request are certainly a step in the right direction.

PR #994, opened 2015-06-29
TEST: added debug mode option for coverage

Moliholy:
It's possible to run gcovr with many different binaries, so this way we could compile and run the integration tests and see how much they actually cover. Of course it could also be added to our Jenkins project.

Just as an example, check out http://cvm-stat01/integration/ 
I only ran the first 5 tests there, so there is not so much coverage.


Moliholy:
I have just removed old test/unittests/CMakeList.txt code and added the new one. This way there are no problems in the source compilation

PR #998, opened 2015-07-02
TEST: benchmarks no longer require to be executed as sudo

Moliholy:
- Moved the cache directory to ~/.benchmark
- User added to the fuse group
- Granted ownership of /cvmfs/\<repo\> to the current user
- Added "sudo" to some commands

### DIFF
--- a/test/test_functions
+++ b/test/test_functions
@@ -1384,12 +1384,14 @@ set_valgrind_options() {
 setup_benchmark_environment() {
   local workspace="$1"
   local me=$(whoami)
-  sudo sh -c "mkdir -p /cvmfs/$FQRN && chown $me /cvmfs/$FQRN"
   mkdir -p "$CVMFS_OPT_OUTPUT_DIR/$FQRN"
   mkdir -p $workspace/cache
   
   # make sure the user can execute /bin/fusermount, write into /dev/fuse and write into /cvmfs/<repo>
-  sudo sh -c "usermod -a -G fuse $me"
+  sudo chmod 666 /dev/fuse
+  sudo chmod 4755 /bin/fusermount
+  sudo mkdir -p /cvmfs/$FQRN
+  sudo chown $me /cvmfs/$FQRN
   
   # mtab trick
   if [ ! -h /etc/mtab ]; then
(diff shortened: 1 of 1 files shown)


test/test_functions
```diff
@@ -1473,7 +1474,7 @@ statistics_daemon() {
   touch $dump_file
   while [ -f /proc/${PID}/smaps ]; do
     # memory in KB
-    echo 0 $(awk '/Rss/ {print "+", $2}' /proc/${PID}/smaps) | bc >> $dump_file
+    echo 0 $(sudo awk '/Rss/ {print "+", $2}' /proc/${PID}/smaps) | bc >> $dump_file
```
Moliholy: Access to /proc/<pid>/smaps is forbidden to non-root users

test/test_functions
```diff
@@ -1383,17 +1383,18 @@ set_valgrind_options() {
...
+  sudo sh -c "mkdir -p /cvmfs/$FQRN && chown $me /cvmfs/$FQRN"
   mkdir -p "$CVMFS_OPT_OUTPUT_DIR/$FQRN"
   mkdir -p $workspace/cache
-
-  # cleaning the cache
-  clean_benchmark_cache
-
+  
+  # make sure the user can execute /bin/fusermount, write into /dev/fuse and write into /cvmfs/<repo>
+  sudo sh -c "usermod -a -G fuse $me"
```
Moliholy: There is a problem with this: if the user was not in the "fuse" group then it's not effectively added in the current session, so the execution will fail the very first time because it won't be able to use /bin/fusermount. Any clues about how to solve this?

jblomer: I think it's better to change the mode of /bin/fusermount and /dev/fuse to be world accessible.  You then don't have to care about the group membership of sftnight.

Moliholy: In that case, I've been trying setting up file permissions of /dev/fuse and /bin/fusermount, but I had no success so far, even trying with 777 permissions for both. It means there is something more missing. Any idea?

Moliholy:
Now it works fine and there is no need to touch any group

PR #1002, opened 2015-07-03
FIX: remove unnecessary dependency in unittests/CMakeLists.txt

Moliholy:
This is no longer necessary since the target cvmfs_unittests_debug no longer has to compile that library.


jblomer:
I think we can leave it, we need it once we have a unit test for json_document

Moliholy:
Fine, then I close this PR

PR #1006, opened 2015-07-06
switch cern.ch stratum 1 URLs to port 8000

DrDaveD:
Port 8000 avoids transparent proxies.  The osg and egi domain configurations use port 8000 and all the stratum 1s support it.

### DIFF
--- a/mount/domain.d/cern.ch.conf
+++ b/mount/domain.d/cern.ch.conf
@@ -15,6 +15,6 @@
 # Use cvmfs_config showconfig to get the effective parameters.
 #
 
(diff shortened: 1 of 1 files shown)


jblomer:
Thanks for the PR!  For the default configuration, I'm not quite convinced by this change.  The grid is only one of the use cases of cvmfs.  In other uses cases, port 8000 can be blocked.  This should be handled by the "happy eyeballs" logic that tries to connect to both ports and uses whichever replies first.  We have to implement this logic anyway for the IPv4/IPv6 selection [CVM-876](https://sft.its.cern.ch/jira/browse/CVM-876).

I'll close the PR for the time being.

PR #1029, opened 2015-07-31
Extra functionalities to the catalog downloader add-on

Moliholy:
Now it not only downloads the catalogs, but stores them in a cache-like structure, i.e., inside a data/ folder and with subfolders from 00 to ff. Also, it downloads the .cvmfspublished from the repository and stores it in the selected folder.

These changes have as their main target to create a catalog structure that can be immediately used by the cvmfs.LocalRepository class of the python library. This last functionality has already been tested and it works without (apparently) any problem.

### DIFF
--- a/add-ons/tools/download_catalog_graph.py
+++ b/add-ons/tools/download_catalog_graph.py
@@ -2,6 +2,7 @@
 
 import cvmfs
 import sys
+import os
 
 class MerkleCatalogTreeIterator(cvmfs.CatalogTreeIterator):
     def __init__(self, repository, root_catalog, visited_hashes = set()):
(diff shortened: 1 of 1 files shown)


reneme:
Awesome! Actually I was always annoyed after downloading ATLAS or CMS and needing to write a shell script to put the files in the right directories. ;-)
Though, please consider to rebase your changes onto [devel](https://github.com/cvmfs/cvmfs/tree/devel), since this contains a slightly updated version of the script.

Moliholy:
Oh, my... I was working keeping in mind that this was for the python library, when it's actually in the add-ons folder... nevermind, I'll create a new pull request in the devel branch

PR #1043, opened 2015-08-12
Python new Repository class

Moliholy:
This pull request changes the Repository class in the python API. Before it was necessary to instantiate a sub-class of Repository, i.e., RemoteRepository or LocalRepository. The first one always downloaded everything and didn't use any permanent cache, while the second one only could work in local.

In this context, I thought it was better to have just a Repository class so that the user can specify the cache directory and the remote directory. This way he can abstract from the use of full-cached, full-remote and the combination of these two, which is the way the client actually works.

Catalog class has been modified in consequence since it doesn't require a compressed file to be initialised, but an already decompressed one.

Updated example usage:
```
import cvmfs

   repo = cvmfs.Repository('http://cvmfs-stratum-one.cern.ch/opt/boss', cache_dir='/home/user/cvmfs_cache')
   print 'Last Revision:' , repo.manifest.revision , repo.manifest.last_modified
   root_catalog = repo.retrieve_root_catalog()
   print 'Catalog Schema:' , root_catalog.schema
   for nested_catalog_ref in root_catalog.list_nested():
       print 'Nested Catalog at:' , nested_catalog_ref.root_path
   print 'Listing repository'
   for full_path, dirent in repo:
       print full_path
```

### DIFF
--- a/python/cvmfs/_common.py
+++ b/python/cvmfs/_common.py
@@ -72,9 +72,9 @@ def _binary_buffer_to_hex_string(binbuf):
 def _split_md5(md5digest):
     hi = lo = 0
     for i in range(0, 8):
-        lo = lo | (ord(md5digest[i]) << (i * 8))
+        lo |= (ord(md5digest[i]) << (i * 8))
     for i in range(8,16):
-        hi = hi | (ord(md5digest[i]) << ((i - 8) * 8))
+        hi |= (ord(md5digest[i]) << ((i - 8) * 8))
     return ctypes.c_int64(lo).value, ctypes.c_int64(hi).value  # signed int!
 
 def _combine_md5(lo, hi):
--- a/python/cvmfs/_common.py
+++ b/python/cvmfs/_common.py
@@ -75,17 +75,17 @@ class CompressedObject:
 
 
 
-class DatabaseObject(CompressedObject):
+class DatabaseObject:
     db_handle_ = None
 
-    def __init__(self, compressed_db_file):
-        CompressedObject.__init__(self, compressed_db_file)
+    def __init__(self, db_file):
+        self.file_ = db_file
         self._open_database()
 
     def __del__(self):
         if self.db_handle_:
             self.db_handle_.close()
-        self._close()
+        self.file_.close()
 
     def _open_database(self):
         """ Create and configure a database handle to the Catalog """
--- a/python/cvmfs/_common.py
+++ b/python/cvmfs/_common.py
@@ -36,6 +36,8 @@ class DatabaseObject:
         self._open_database()
 
     def __del__(self):
+        if self._db_handle:
+            self._db_handle.close()
         self._file.close()
 
     def _open_database(self):
@@ -58,7 +60,9 @@ class DatabaseObject:
         """ Run an arbitrary SQL query on the catalog database """
         cursor = self._db_handle.cursor()
         cursor.execute(sql)
-        return cursor.fetchall()
+        data = cursor.fetchall()
+        cursor.close()
+        return data
 
     def open_interactive(self):
         """ Spawns a sqlite shell for interactive catalog database inspection """
--- a/python/cvmfs/_common.py
+++ b/python/cvmfs/_common.py
@@ -29,24 +29,22 @@ class CvmfsNotInstalled(Exception):
 
 
 class DatabaseObject:
-    db_handle_ = None
+    _db_handle = None
 
     def __init__(self, db_file):
-        self.file_ = db_file
+        self._file = db_file
         self._open_database()
 
     def __del__(self):
-        if self.db_handle_:
-            self.db_handle_.close()
-        self.file_.close()
+        self._file.close()
 
     def _open_database(self):
         """ Create and configure a database handle to the Catalog """
-        self.db_handle_ = sqlite3.connect(self.file_.name)
-        self.db_handle_.text_factory = str
+        self._db_handle = sqlite3.connect(self._file.name)
+        self._db_handle.text_factory = str
 
     def db_size(self):
-        return os.path.getsize(self.file_.name)
+        return os.path.getsize(self._file.name)
 
     def read_properties_table(self, reader):
         """ Retrieve all properties stored in the 'properties' table """
@@ -58,21 +56,15 @@ class DatabaseObject:
 
     def run_sql(self, sql):
         """ Run an arbitrary SQL query on the catalog database """
-        cursor = self.db_handle_.cursor()
+        cursor = self._db_handle.cursor()
         cursor.execute(sql)
         return cursor.fetchall()
 
     def open_interactive(self):
         """ Spawns a sqlite shell for interactive catalog database inspection """
-        subprocess.call(['sqlite3', self.file_.name])
+        subprocess.call(['sqlite3', self._file.name])
 
 
-class FileObject(CompressedObject):
-    def __init__(self, compressed_file):
-        CompressedObject.__init__(self, compressed_file)
-
-    def file(self):
-        return self.get_uncompressed_file()
 
 def _binary_buffer_to_hex_string(binbuf):
     return "".join(map(lambda c: ("%0.2X" % c).lower(),map(ord,binbuf)))
@@ -90,8 +82,8 @@ def _combine_md5(lo, hi):
                   '\x00','\x00','\x00','\x00','\x00','\x00','\x00','\x00' ]
     for i in range(0, 8):
         md5digest[i] = chr(lo & 0xFF)
-        lo = lo >> 8
+        lo >>= 8
     for i in range(8,16):
         md5digest[i] = chr(hi & 0xFF)
-        hi = hi >> 8
+        hi >>= 8
     return ''.join(md5digest)
--- a/python/cvmfs/_common.py
+++ b/python/cvmfs/_common.py
@@ -6,11 +6,8 @@ This file is part of the CernVM File System auxiliary tools.
 """
 
 import ctypes
-import tempfile
-import zlib
 import sqlite3
 import subprocess
-import shutil
 import os
 
 
@@ -28,50 +25,6 @@ _REPLICATING_NAME      = ".cvmfs_is_snapshotting"
 class CvmfsNotInstalled(Exception):
     def __init__(self):
         Exception.__init__(self, "It seems that cvmfs is not installed on this machine!")
-
-
-class CompressedObject:
-    file_            = None
-    compressed_file_ = None
-
-    def __init__(self, compressed_file):
-        self.compressed_file_ = compressed_file
-        self._decompress()
-
-    def get_compressed_file(self):
-        return self.compressed_file_
-
-    def get_uncompressed_file(self):
-        return self.file_
-
-    def save_to(self, path):
-        shutil.copyfile(self.get_compressed_file().name, path)
-
-    def save_uncompressed_to(self, path):
-        shutil.copyfile(self.get_uncompressed_file().name, path)
-
-    def size_compressed(self):
-        """ Size of the compressed file in bytes """
-        return os.path.getsize(self.compressed_file_.name)
-
-    def size_uncompressed(self):
-        """Size of the uncompressed file in bytes """
-        return os.path.getsize(self.file_.name)
-
-    def _decompress(self):
-        """ Unzip a file to a temporary referenced by self.file_ """
-        self.file_ = tempfile.NamedTemporaryFile('w+b')
-        self.compressed_file_.seek(0)
-        self.file_.write(zlib.decompress(self.compressed_file_.read()))
-        self.file_.flush()
-        self.file_.seek(0)
-        self.compressed_file_.seek(0)
-
-    def _close(self):
-        if self.file_:
-            self.file_.close()
-        if self.compressed_file_:
-            self.compressed_file_.close()
 
 
 
@@ -92,6 +45,8 @@ class DatabaseObject:
         self.db_handle_ = sqlite3.connect(self.file_.name)
         self.db_handle_.text_factory = str
 
+    def db_size(self):
+        return os.path.getsize(self.file_.name)
 
     def read_properties_table(self, reader):
         """ Retrieve all properties stored in the 'properties' table """
--- a/python/cvmfs/dirent.py
+++ b/python/cvmfs/dirent.py
@@ -92,8 +92,7 @@ class DirectoryEntry:
             raise Exception("Cannot retrieve symlink")
         elif self.is_directory():
             raise Exception("Cannot retrieve directory")
-        f = repository.retrieve_object(self.content_hash_string())
-        return f.file()
+        return repository.retrieve_object(self.content_hash_string())
 
     def is_directory(self):
         return (self.flags & _Flags.Directory) > 0
--- a/python/cvmfs/dirent.py
+++ b/python/cvmfs/dirent.py
@@ -57,7 +57,7 @@ class Chunk:
         return _binary_buffer_to_hex_string(self.content_hash) + suffix
 
     @staticmethod
-    def _catalog_db_fields():
+    def catalog_db_fields():
         return "md5path_1, md5path_2, offset, size, hash"
 
 
@@ -82,7 +82,7 @@ class DirectoryEntry:
                str(self.md5path_1) + "|" + str(self.md5path_2) + ">"
 
     @staticmethod
-    def _catalog_db_fields():
+    def catalog_db_fields():
         # see DirectoryEntry.__init__()
         return "md5path_1, md5path_2, parent_1, parent_2, hash, \
                 flags, size, mode, mtime, name, symlink"
@@ -130,11 +130,11 @@ class DirectoryEntry:
         bit_mask     = _Flags.ContentHashType
         right_shifts = 0
         while bit_mask & 1 == 0:
-            bit_mask = bit_mask >> 1
+            bit_mask >>= 1
             right_shifts += 1
         hash_type = ((self.flags & _Flags.ContentHashType) >> right_shifts) + 1
         self.content_hash_type = \
-            hash_type if hash_type > 0 and hash_type < ContentHashTypes.UpperBound \
+            hash_type if 0 < hash_type < ContentHashTypes.UpperBound \
                       else ContentHashTypes.Unknown
 
     # def BacktracePath(self, containing_catalog, repo):
--- a/python/cvmfs/dirent.py
+++ b/python/cvmfs/dirent.py
@@ -5,7 +5,7 @@ Created by René Meusel
 This file is part of the CernVM File System auxiliary tools.
 """
 
-from _common import _binary_buffer_to_hex_string, FileObject
+from _common import _binary_buffer_to_hex_string
 
 
 class _Flags:
@@ -92,7 +92,7 @@ class DirectoryEntry:
             raise Exception("Cannot retrieve symlink")
         elif self.is_directory():
             raise Exception("Cannot retrieve directory")
-        f = FileObject(repository.retrieve_object(self.content_hash_string()))
+        f = repository.retrieve_object(self.content_hash_string())
         return f.file()
 
     def is_directory(self):
--- a/python/cvmfs/repository.py
+++ b/python/cvmfs/repository.py
@@ -271,11 +271,10 @@ class Fetcher:
     @staticmethod
     def _write_content_into_file(content, opened_file):
         """ Writes content into the opened file. The file must have
-         write permission
+         write permission. The file is closed after after the operation
         """
         opened_file.write(content)
-        opened_file.seek(0)
-        opened_file.flush()
+        opened_file.close()
 
 
     def retrieve_file(self, file_name):
@@ -316,7 +315,7 @@ class LocalFetcher(Fetcher):
             compressed_file.close()
             cached_file = self.cache.add(file_name)
             self._write_content_into_file(decompressed_content, cached_file)
-            return cached_file
+            return self.cache.get(file_name)
         else:
             raise FileNotFoundInRepository(file_name)
 
@@ -328,7 +327,6 @@ class LocalFetcher(Fetcher):
             cached_file = self.cache.add(file_name)
             compressed_file = open(full_path, 'r')
             Fetcher._write_content_into_file(compressed_file.read(), cached_file)
-            cached_file.close()
             compressed_file.seek(0)
             return compressed_file
         raise FileNotFoundInRepository(file_name)
@@ -351,9 +349,7 @@ class RemoteFetcher(Fetcher):
         for chunk in response.iter_content(chunk_size=4096):
             if chunk:
                 cached_file.write(chunk)
-        cached_file.seek(0)
-        cached_file.flush()
-        return cached_file
+        cached_file.close()
 
 
     @staticmethod
@@ -363,19 +359,20 @@ class RemoteFetcher(Fetcher):
             raise FileNotFoundInRepository(file_url)
         decompressed_content = zlib.decompress(response.content)
         Fetcher._write_content_into_file(decompressed_content, cached_file)
-        return cached_file
 
 
     def _retrieve_file(self, file_name):
         file_url = self._make_file_uri(file_name)
         cached_file = self.cache.add(file_name)
-        return RemoteFetcher._download_content_and_decompress(cached_file, file_url)
+        RemoteFetcher._download_content_and_decompress(cached_file, file_url)
+        return self.cache.get(file_name)
 
 
     def retrieve_raw_file(self, file_name):
         cached_file = self.cache.add(file_name)
         file_url = self._make_file_uri(file_name)
-        return RemoteFetcher._download_content_and_store(cached_file, file_url)
+        RemoteFetcher._download_content_and_store(cached_file, file_url)
+        return self.cache.get(file_name)
 
 
 
--- a/python/cvmfs/repository.py
+++ b/python/cvmfs/repository.py
@@ -8,7 +8,6 @@ This file is part of the CernVM File System auxiliary tools.
 import abc
 import os
 import glob
-import urlparse
 import tempfile
 import requests
 import collections
@@ -18,7 +17,6 @@ from dateutil.tz import tzutc
 import zlib
 
 import _common
-import cvmfs
 from manifest import Manifest
 from catalog import Catalog
 from history import History
@@ -27,6 +25,7 @@ from certificate import Certificate
 
 class RepositoryNotFound(Exception):
     def __init__(self, repo_path):
+        Exception.__init__(self)
         self.path = repo_path
 
     def __str__(self):
@@ -34,6 +33,7 @@ class RepositoryNotFound(Exception):
 
 class UnknownRepositoryType(Exception):
     def __init__(self, repo_fqrn, repo_type):
+        Exception.__init__(self)
         self.fqrn = repo_fqrn
         self.type = repo_type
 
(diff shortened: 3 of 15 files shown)


python/cvmfs/_common.py
```diff
@@ -75,17 +75,17 @@ def _close(self):
 
 
 
-class DatabaseObject(CompressedObject):
+class DatabaseObject:
```
reneme: Please get rid of the `CompressedObject` class altogether. Have a look into the `History` class as well, since it works very similar to the `Catalog`.

python/cvmfs/_common.py
```diff
@@ -75,17 +75,17 @@ def _close(self):
...
     db_handle_ = None
 
-    def __init__(self, compressed_db_file):
-        CompressedObject.__init__(self, compressed_db_file)
+    def __init__(self, db_file):
+        self.file_ = db_file
```
reneme: General remark for consistency: We should agree on how to mark private members. I think I used a trailing underscore everywhere until now. In your `Cache` class I've seen a prepended underscore... Let's not mix that.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        if metadata_cleanup:
+            self._cleanup_metadata()
+
+    def _cleanup_metadata(self):
+        for f in glob.glob(self._cache_dir + '/' + '.cvmfs*'):
+            os.remove(f)
+
+
+    def _create_dir(self, path):
+        cache_full_path = '/'.join([self._cache_dir, path])
```
reneme: use `os.path.join()` for that, please. (platform independence)

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        if not os.path.exists(cache_full_path):
+            os.mkdir(cache_full_path, 0755)
+
+
+    def _create_cache_structure(self):
+        self._create_dir('data')
+        for i in range(0x00, 0xff + 1):
+            new_folder = '{0:#0{1}x}'.format(i, 4)[2:]
+            self._create_dir('data' + '/' + new_folder)
```
reneme: `os.path.join()`

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+
+
+    def cache_path(self):
+        return str(self._cache_dir)
+
+
+    def add(self, file_name):
+        full_path = '/'.join([self._cache_dir, file_name])
```
reneme: `os.path.join()`

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+
+    def get(self, file_name):
+        full_path = '/'.join([self._cache_dir, file_name])
+        if os.path.exists(full_path):
+            return open(full_path, 'rb')
+        return None
+
+
+    def evict(self):
+        for i in range(0x00, 0xff + 1):
```
reneme: maybe just recursively clean out the whole directory and rebuild with `_create_cache_structure()`?

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+            folder = '{0:#0{1}x}'.format(i, 4)[2:]
+            wildcard = '/'.join([self._cache_dir, folder, '*'])
+            os.remove(glob.glob(wildcard))
+
+
+
+class Fetcher:
+    """ Abstract wrapper around a Fetcher """
+
+    __metadata__ = abc.ABCMeta
```
reneme: Remark: I hate to say that, but Python 2.4 doesn't have the ABC module. Could be that we need to change that once it gets released. Depends a bit, if we need to support SLC5 or not.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        decompressed_data = zlib.decompress(compressed_file.read())
+        temp_file_path = tempfile.mktemp(dir='/tmp')
+        temp_file = open(temp_file_path, 'w+')
+        Fetcher._write_content_into_file(decompressed_data, temp_file)
+        return temp_file
+
+
+    @staticmethod
+    def _write_content_into_file(content, opened_file):
+        """ Writes content into the opened file. The file must have write permission """
```
reneme: Style: Please stay below 80 characters per line.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+
+
+
+class LocalFetcher(Fetcher):
+    """ Retrieves files only from the local cache """
+
+    def __init__(self, cache_dir=''):
+        Fetcher.__init__(self, cache_dir, metadata_cleanup=False)
+
+    def retrieve_file(self, file_name):
```
reneme: We discussed that offline, just for reference: This should expect to access compressed files that are stored locally. For example in the backend storage of a Stratum0 or Stratum1 server.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        """ Abstract method to retrieve a file from the repository """
+        pass
+
+
+
+class LocalFetcher(Fetcher):
+    """ Retrieves files only from the local cache """
+
+    def __init__(self, cache_dir=''):
+        Fetcher.__init__(self, cache_dir, metadata_cleanup=False)
```
reneme: `metadata_cleanup=True` should be the default for the `LocalFetcher` as well.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        """ The LocalFetcher works in read-only mode with its cache """
+        cached_file = self.cache.get(file_name)
+        if not cached_file:
+            raise FileNotFoundInRepository(self, file_name)
+        return cached_file
+
+
+
+class RemoteFetcher(Fetcher):
+    """ Retrieves files from the local cache if found, and from remote otherwise """
```
reneme: Style: maximal 80 characters per line.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        if response.status_code != requests.codes.ok:
+            raise FileNotFoundInRepository(self, file_url)
+        decompressed_content = zlib.decompress(response.content)
+        Fetcher._write_content_into_file(decompressed_content, cached_file)
+        return cached_file
+
+
+    def _retrieve_file(self, file_name):
+        file_url = self.__make_file_url(file_name)
+        cached_file = self.cache.add(file_name)
```
reneme: The cache handling should be centralized into `retrieve_file()`. See there for details.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        return self._download_content_and_decompress(cached_file, file_url)
+
+
+    def retrieve_raw_file(self, file_name):
```
reneme: Please make that an abstract method in `Fetcher` as well and implement it for the `LocalFetcher` too.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        cached_file = self.cache.add(file_name)
```
reneme: 'raw files' are supposed to be compressed files, right? Hence they shouldn't be written to the cache.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        file_url = self.__make_file_url(file_name)
+        return self._download_content_and_store(cached_file, file_url)
+
+
+    def retrieve_file(self, file_name):
```
reneme: Please centralize the cache interaction here, like:

1. check cache (and possibly directly return)
2. download missing file
3. populate cache with newly downloaded file
4. return newly downloaded file

Currently the cache is queried in this method but the cache population happens as a side effect of `self._retrieve_file()` which is potentially error prone.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        return self._retrieve_file(file_name)
 
-    __metaclass__ = abc.ABCMeta
 
-    def __init__(self):
+
+class Repository:
+    """ Wrapper around a CVMFS Repository representation """
+
+    def __init__(self, cache_dir='', repo_url=''):
```
reneme: Discussed offline. The parameters should be swapped and `repo_url` is supposed to be always required. However, it might contain different things:

* an HTTP URL (**http://cvmfs-stratum-zero.cern.ch/cvmfs/atlas.cern.ch**)
* a local path to a backend storage (**/srv/cvmfs/atlas.cern.ch**)
* a repository name (**atlas.cern.ch**) that needs a lookup in the config files

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+        if repo_url == '' and cache_dir == '':
+            raise Exception('repo_url and cache_dir cannot be empty at the same time')
+        if repo_url == '':
```
reneme: Discussed offline. Instantiation of `Fetcher` strategy class should depend on the type of `repo_url`. See above.

python/cvmfs/repository.py
```diff
@@ -198,12 +200,170 @@ def _pop_catalog(self):
...
+            self._fetcher = LocalFetcher(cache_dir)
+        else:
+            self._fetcher = RemoteFetcher(repo_url, cache_dir)
+        self._storage_location = self._fetcher.cache.cache_path()
```
reneme: wut?

python/cvmfs/repository.py
```diff
@@ -241,14 +402,20 @@ def _try_to_get_last_replication_timestamp(self):
...
+            with self._retrieve_raw_file(_common._REPLICATING_NAME) as rf:
                 timestamp = rf.readline()
                 self.replicating = True
-                self.replicating_since = self.__read_timestamp(timestamp)
+                self.replicating_since = Repository.__read_timestamp(timestamp)
         except FileNotFoundInRepository, e:
             pass
 
 
+    def _retrieve_raw_file(self, file_name):
```
reneme: That should be handled by polymorphism in the `Fetcher`. See above.

python/cvmfs/repository.py
```diff
@@ -292,10 +459,9 @@ def retrieve_certificate(self):
         return Certificate(certificate)
 
 
-    @abc.abstractmethod
     def retrieve_file(self, file_name):
-        """ Abstract method to retrieve a file from the repository """
-        pass
+        """ Method to retrieve a file from the repository """
```
reneme: This method can go completely, I think. Shouldn't it be replaced by a call to the `Fetcher`?

python/cvmfs/repository.py
```diff
@@ -344,156 +509,19 @@ def _retrieve_and_open_catalog(self, catalog_hash):
...
     d = _common._REPO_CONFIG_PATH
     if not os.path.isdir(d):
         raise _common.CvmfsNotInstalled
-    return [ LocalRepository(repo) for repo in os.listdir(d) if os.path.isdir(os.path.join(d, repo)) ]
+    return [ Repository(cache_dir=repo) for repo in os.listdir(d) if os.path.isdir(os.path.join(d, repo)) ]
 
 def all_local_stratum0():
     return [ repo for repo in all_local() if repo.type == 'stratum0' ]
 
 def open_repository(repository_path, public_key = None):
```
reneme: This was just a convenience function to distinguish between the old `LocalRepository` and `RemoteRepository` classes, based on the provided URL/path. This should be handled by the generic `Repository` class now (see above). Please remove it entirely.

python/cvmfs/repository.py
Moliholy: I don't agree with that. The user wants to download a file, and has a repository object. He doesn't need to know about the Fetcher object, and its structure, but just to tell the repository to download/fetch a file. The Fetcher object haven't been shown at any point to the user, why should we do it now?

python/cvmfs/repository.py
Moliholy: I take a very good note, thanks!

python/cvmfs/root_file.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, file_object):
         self._check_validity()
 
     @abc.abstractmethod
-    def _verify_signature(public_entity):
```
Moliholy: There was a mistake here, which is now fixed

python/cvmfs/root_file.py
```diff
@@ -70,7 +70,8 @@ def verify_signature(self, public_entity):
         return self.has_signature and self._verify_signature(public_entity)
 
 
-    def _hash_over_content(self, file_object):
+    @staticmethod
+    def _hash_over_content(file_object):
```
Moliholy: This method can be static since it doesn't use 'self'

python/cvmfs/repository.py
Moliholy: Or wait... you mean the user should only use the retrieve_object method to get files? Because that could be... retrieve_file gives total flexibility to fetch whatever you want, including metadata if necessary, but retrieve_object can only do it from the 'data' folder.
On a different matter, if retrieve_object is actually going to retrieve files, it should be called retrieve_file, isn't it?

reneme: Well, `retrieve_file()` downloads any URL without doing anything with it and shouldn't be exposed per-se. For downloading the manifest and whitelist we have specialised methods. On the other hand `retrieve_object()` gets and decompresses CAS objects in **data/../........** which might be useful indeed. Hence, I would just remove `retrieve_file()` from the public interface and keep `retrieve_object()` as is.

Moliholy:
New changes made. There is a general code cleanup in the last commit that touches many files

Moliholy:
Updated unit tests. All of them passed, of course.

python/cvmfs/repository.py
```diff
@@ -283,7 +475,7 @@ def retrieve_history(self):
 
 
     def retrieve_whitelist(self):
-        whitelist = self.retrieve_file(_common._WHITELIST_NAME)
+        whitelist = self._fetcher.retrieve_raw_file(_common._WHITELIST_NAME)
```
Moliholy: It has to be retrieved with this method because it's not a compressed file

python/cvmfs/_common.py
```diff
@@ -30,68 +27,24 @@ def __init__(self):
...
-        CompressedObject.__init__(self, compressed_db_file)
+    def __init__(self, db_file):
+        self._file = db_file
         self._open_database()
 
     def __del__(self):
-        if self.db_handle_:
-            self.db_handle_.close()
-        self._close()
+        self._file.close()
```
reneme: Is there a reason, that you don't close `db_handle_` anymore here?

python/cvmfs/dirent.py
```diff
@@ -131,11 +130,11 @@ def _read_content_hash_type(self):
         bit_mask     = _Flags.ContentHashType
         right_shifts = 0
         while bit_mask & 1 == 0:
-            bit_mask = bit_mask >> 1
+            bit_mask >>= 1
             right_shifts += 1
         hash_type = ((self.flags & _Flags.ContentHashType) >> right_shifts) + 1
         self.content_hash_type = \
-            hash_type if hash_type > 0 and hash_type < ContentHashTypes.UpperBound \
+            hash_type if 0 < hash_type < ContentHashTypes.UpperBound \
```
reneme: O Rly?

You can specify ranges like that in Python?

python/cvmfs/repository.py
```diff
@@ -25,6 +26,7 @@
 
 class RepositoryNotFound(Exception):
     def __init__(self, repo_path):
+        Exception.__init__(self)
```
reneme: Are these up-call necessary? I believe they are implicitly done anyway, no?

python/cvmfs/repository.py
```diff
@@ -198,38 +205,223 @@ def _pop_catalog(self):
...
+class Cache:
+    def __init__(self, cache_dir):
+        if not os.path.exists(cache_dir):
+            cache_dir = tempfile.mkdtemp(dir='/tmp', prefix='cache.')
+        self._cache_dir = cache_dir
+        self._create_cache_structure()
+        self._cleanup_metadata()
+
+    def _cleanup_metadata(self):
+        for f in glob.glob(self._cache_dir + '/' + '.cvmfs*'):
```
reneme: Perhaps its better to explicitly delete these files here. You never know, who or what is going to put a file called `.cvmfs*` in there and how upset it would be to loose it. ;-)

python/cvmfs/_common.py
Moliholy: There is absolutely no reason. It was my mistake when cleaning up code. I'll re-add it now

python/cvmfs/repository.py
```diff
@@ -198,38 +205,223 @@ def _pop_catalog(self):
...
+            new_folder = '{0:#0{1}x}'.format(i, 4)[2:]
+            self._create_dir(os.path.join('data', new_folder))
+
+
+    def cache_path(self):
+        return str(self._cache_dir)
+
+
+    def add(self, file_name):
+        full_path = os.path.join(self._cache_dir, file_name)
```
reneme: In the long run that could become a concurrency problem if two 'things' try to add the same file. The C++ client does that transactionally to get around it, like so:

1. Open a temporary file in **data/txn** (for TransaXioN)
2. Let the user write the file
3. Call `commit()` to move the file in place **data/aa/....**

That way two process can `add()` the same file at the same time and since `rename()` is atomic in POSIX, their `commit()`s will just overwrite each other which is harmless.

python/cvmfs/repository.py
```diff
@@ -198,38 +205,223 @@ def _pop_catalog(self):
...
+
+
+    def get(self, file_name):
+        full_path = os.path.join(self._cache_dir, file_name)
+        if os.path.exists(full_path):
+            return open(full_path, 'rb')
+        return None
+
+
+    def evict(self):
```
reneme: That is better called `purge()` or `clear()` since you empty the cache completely. `evict()` is usually used for removing only a single object from a cache.

python/cvmfs/repository.py
```diff
@@ -198,38 +205,223 @@ def _pop_catalog(self):
...
+            raise FileNotFoundInRepository(file_name)
+
+
+    def retrieve_raw_file(self, file_name):
+        """ Retrieves the file directly from the source """
+        full_path = self._make_file_uri(file_name)
+        if os.path.exists(full_path):
+            cached_file = self.cache.add(file_name)
+            compressed_file = open(full_path, 'r')
+            Fetcher._write_content_into_file(compressed_file.read(), cached_file)
```
reneme: `self._write_content_into_file()` please.

python/cvmfs/repository.py
```diff
@@ -198,38 +205,223 @@ def _pop_catalog(self):
...
+
+
+    def evict(self):
+        data_path = os.path.join(self._cache_dir, 'data')
+        shutil.rmtree(data_path)
+        self._create_cache_structure()
+
+
+
+class Fetcher:
```
reneme: Potential refactoring: I think its cleaner if the cache handling happens in the `Fetcher` base class, so that the concrete `Fetcher`s only deal with fetching locally or remotely and the `Cache` interaction is factored out into the common base class. Right now it's kind of duplicated in both concrete classes.

python/cvmfs/repository.py
```diff
@@ -198,38 +205,223 @@ def _pop_catalog(self):
...
+            raise Exception('source cannot be empty')
+        self.__determine_source(source, cache_dir)
+        self._storage_location = self._fetcher.cache.cache_path()
         self._opened_catalogs = {}
         self._read_manifest()
         self._try_to_get_last_replication_timestamp()
         self._try_to_get_replication_state()
 
 
+    def __determine_source(self, source, cache_dir):
```
reneme: Rename: `__init_fetcher()` maybe? to make more clear what happens here

python/cvmfs/repository.py
```diff
@@ -198,38 +205,223 @@ def _pop_catalog(self):
...
 
-    def __init__(self):
+
+class Repository:
+    """ Wrapper around a CVMFS Repository representation """
+
+    def __init__(self, source, cache_dir=''):
+        if source == '':
+            raise Exception('source cannot be empty')
+        self.__determine_source(source, cache_dir)
```
reneme: Maybe like so:
```python
self._fetcher = self.__init_fetcher(source, cache_dir)
```
Probably making `__init_fetcher()` static.

python/cvmfs/repository.py
Moliholy: It doesn't clean up the metadata, only the actual data... does that maybe suggest another name?

python/cvmfs/repository.py
Moliholy: Hmmm I don't fully agree in principle. A Fetcher fetches files, not stores them. A Cache object stores objects, but is not in charge of anything else. What I think you are proposing is multipurpose Fetcher-Cache-class that does both functions.
However they have clearly different functions (fetch files and store files), so why merging them in a single class?
On a sightly different matter, if this library is for people (or ourselves) to debug, isn't it better to provide independent classes that deal only with a concrete functionality? That way you can abstractly analyse (for example) your cache with a cache object, without caring about any Fetcher. Because... who would imagine that the fetcher acts also as a Cache? Not so intuitive, I think

Moliholy: Maybe this article makes my point clearer: https://en.wikipedia.org/wiki/Is-a
Inheritance should happen when an sub-entity can be considered a member of a top-level entity. For that, one has to ask himself if that sub-entity meets the relationship IS-A with the parent-entity. For example: a human IS-An animal, so human could inherit from animal.

But is a fetcher a cache? I don't think so.

Just as an absurd example that came up to my mind: it's like if you plan to make a bus driver inherit from bus and human at the same time. No, a bus driver has an attribute, which is his vehicle :P

python/cvmfs/repository.py
Moliholy: Changed ;)

python/cvmfs/repository.py
Moliholy: Totally true, it feels weird when the fetcher attribute appear from nowhere. Fixed! ;)

python/cvmfs/repository.py
Moliholy: That looks harder to implement. I'm on it ;)

python/cvmfs/repository.py
```diff
@@ -198,38 +198,233 @@ def _pop_catalog(self):
...
+
+
+
+class RemoteFetcher(Fetcher):
+    """ Retrieves files from the local cache if found, and from
+    remote otherwise
+    """
+
+    def __init__(self, repo_url, cache_dir=''):
+        Fetcher.__init__(self, repo_url, cache_dir)
```
reneme: maybe like this, to not repeat the inherited object?
```python
super(RemoteFetcher, self).__init__()
```

python/cvmfs/repository.py
```diff
@@ -198,38 +198,233 @@ def _pop_catalog(self):
...
+        self.cache = Cache(cache_dir)
+        self.source = source
+
+
+    def _make_file_uri(self, file_name):
+        return os.path.join(self.source, file_name)
+
+
+    @staticmethod
+    def _write_content_into_file(content, opened_file):
```
reneme: The method shouldn't `close()` the file. That's a nasty hidden side effect...

python/cvmfs/repository.py
```diff
@@ -198,38 +198,233 @@ def _pop_catalog(self):
...
+        cached_file.close()
+
+
+    @staticmethod
+    def _download_content_and_decompress(cached_file, file_url):
+        response = requests.get(file_url, stream=False)
+        if response.status_code != requests.codes.ok:
+            raise FileNotFoundInRepository(file_url)
+        decompressed_content = zlib.decompress(response.content)
+        Fetcher._write_content_into_file(decompressed_content, cached_file)
```
reneme:
```python
self._write_content_into_file()
```

Moliholy: Nop, self in an static method is not possible

python/cvmfs/repository.py
Moliholy: True, I'll fix that

python/cvmfs/repository.py
Moliholy: So far it doesn't inherit from object, so I would have to change that. But yes, no problem with that

Moliholy:
Added new considerable changes to the Repository class.

python/cvmfs/repository.py
Moliholy: This function is used in the tests, by the way...

python/cvmfs/repository.py
```diff
@@ -197,39 +196,230 @@ def _pop_catalog(self):
...
+            cached_file.write(decompressed_content)
+            cached_file.close()
+        else:
+            raise FileNotFoundInRepository(file_name)
+
+    def _retrieve_raw_file(self, file_name, cached_file):
+        """ Retrieves the file directly from the source """
+        full_path = self._make_file_uri(file_name)
+        if os.path.exists(full_path):
+            compressed_file = open(full_path, 'r')
```
reneme: Retrieving raw files doesn't necessarily mean that they are compressed, no? But the variable name suggests it.

python/cvmfs/repository.py
```diff
@@ -197,39 +196,230 @@ def _pop_catalog(self):
         return wrapper.get_catalog()
 
 
+class Resource(file):
```
reneme: Even though I like this [RAII](http://www.wikiwand.com/en/Resource_Acquisition_Is_Initialization) idea, I think it's overkill here and kind of adds unclear magic. Looking at `retrieve_file()` and `retrieve_raw_file()` it is completely unclear how the control flow happens on first glance. See there please.

python/cvmfs/repository.py
```diff
@@ -197,39 +196,230 @@ def _pop_catalog(self):
...
+        self.__cache = Cache(cache_dir)
+        self.source = source
+
+    def _make_file_uri(self, file_name):
+        return os.path.join(self.source, file_name)
+
+    def get_cache_path(self):
+        return self.__cache.get_cache_path()
+
+    def retrieve_file(self, file_name):
```
reneme: How about that to avoid the code duplication in `retrieve_file()` and `retrieve_raw_file()` and to make the cache control flow more obvious:

```python
def retrieve_file(self, file_name):
    return self._retrieve(file_name, self._retrieve_file)

def retrieve_raw_file(self, file_name):
    return self._retrieve(file_name, self._retrieve_raw_file)

def retrieve_file(self, file_name, retrieve_fn):
    cached_file = self.__cache.get(file_name)
    if not cached_file:
        cached_file = self.__cache.transaction(file_name)
        retrieve_fn(file_name, cached_file)
        cached_file = self.__cache.commit(file_name)
    return cached_file
```

You probably still need the `Resource` class to keep the final destination path but perhaps you can tightly couple it to `Cache` by making it a member class of it.

reneme:
Thanks, that should have been the last comments. ;-)

Moliholy:
Updated with the Cache transactions and exceptions

python/cvmfs/repository.py
reneme: Okay, I think with the `Cache::transaction()` and `Cache::commit()` its fine to use this class in an RAII sense. I'd just rename it to `CacheTransactionFile` or something. Resource is too generic for this. It is tightly coupled to the cache and should also reflect that in its name.

PR #1046, opened 2015-08-13
add autofs and apache modules directories for gentoo

sfabbro:
Gentoo distribution keeps the auto.master in /etc/autofs/auto.master and apache modules configurations in /etc/apache2/modules.d. Apache modules configuration files are actually prefixed with number such as: 45_mod_dav.conf, but I did not change it since the file name is hard coded in a few functions of cvmfs_server.


sfabbro:
That last commit makes sure cvmfs links with a system c-ares library which is needed for dynamic linking.
Without it the building fails with Dynamic Shared Object missing.

[snip]
/usr/lib64/ccache/bin/x86_64-pc-linux-gnu-g++   -O2 -pipe -g -fno-exceptions -fno-strict-aliasing -fasynchronous-unwind-tables -fno-omit-frame-pointer -fvisibility=hidden -Wall -D_REENTRANT -D__EXTENSIONS__ -D_LARGEFILE64_SOURCE -D__LARGE64_FILES -fno-optimize-sibling-calls   -Wl,-O1,--as-needed,--hash-style=gnu CMakeFiles/test_libcvmfs.dir/test_libcvmfs.cc.o  -o test_libcvmfs -rdynamic libcvmfs.a -lsqlite3 -lcurl -lpacparser -lz -lssl -lcrypto -lrt -luuid -lpthread -ldl 
/usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/../../../../x86_64-pc-linux-gnu/bin/ld: libcvmfs.a(libcvmfs.a_pub.o): undefined reference to symbol 'ares_process_fd'
/usr/lib64/libcares.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status

cvmfs/cvmfs_config
```diff
@@ -1437,12 +1438,23 @@ reload_service() {
...
 configure_autofs() {
+  for path in /etc/auto.master /etc/autofs/auto.master; do
+      if [ -f $path ]; then
+          AUTOMASTER=$path
+          break;
+      fi
+  done
+  if [ -z "$AUTOMASTER" ]; then
+      echo "Error: unable to find auto.master in any of the supplied paths!"
+      num_errors=$(($num_errors+1))
```
jblomer: This is only meant used in the context of chksetup.  At this point, you can just `exit 1`

cvmfs/cvmfs_config
```diff
@@ -1437,12 +1438,23 @@ reload_service() {
 
 
 configure_autofs() {
+  for path in /etc/auto.master /etc/autofs/auto.master; do
+      if [ -f $path ]; then
```
jblomer: Please use 2 spaces for indentation

cvmfs/cvmfs_server
```diff
@@ -484,19 +484,22 @@ check_multiple_repository_existence() {
 
 
 # checks if the aufs kernel module is present
-#
+# of if aufs is compiled in
 # @return   0 if the aufs kernel module is loaded
 check_aufs() {
-  /sbin/modprobe -q aufs || test -d /sys/fs/aufs
+  /sbin/modprobe -q aufs || test -d /sys/fs/aufs || \
+  test x"$(zgrep AUFS_FS /proc/config.gz | awk -F = '{print $2}')" = xy
```
jblomer: Shouldn't this case be already covered by `test -d /sys/fs/aufs` in the line above?

cvmfs/cvmfs_server
```diff
@@ -484,19 +484,22 @@ check_multiple_repository_existence() {
...
-#
+# checks if the overlayfs kernel module is present 
+# of if overlayfs is compiled in
 # @return   0 if the overlayfs kernel module is loaded
 check_overlayfs() {
+  
   /sbin/modprobe -q overlayfs || test -d /sys/module/overlayfs || \
-  /sbin/modprobe -q overlay   || test -d /sys/module/overlay
+  /sbin/modprobe -q overlay   || test -d /sys/module/overlay || \
+  test x"$(zgrep OVERLAY_FS /proc/config.gz | awk -F = '{print $2}')" = xy
```
jblomer: Should be actually covered by `test -d ...`

jblomer: reneme: do we support `overlay` instead of `overlayfs` already?

cvmfs/cvmfs_server
```diff
@@ -519,7 +522,8 @@ get_overlayfs_name() {
   check_overlayfs || die "overlayfs is not installed!"
   if /sbin/modprobe -q overlayfs || [ -d /sys/module/overlayfs ]; then
     echo "overlayfs"
-  elif /sbin/modprobe -q overlay || [ -d /sys/module/overlay ]; then
+  elif /sbin/modprobe -q overlay || [ -d /sys/module/overlay ] || \
+       test x"$(zgrep OVERLAY_FS /proc/config.gz | awk -F = '{print $2}')" = xy ; then
```
jblomer: see above

jblomer:
Thanks!  Looks fine to me.  Please see minor comments, in particular about modprobe / statically compiled union file system.

cvmfs/cvmfs_server
reneme: Its in the making. CentOS 7.1 uses **overlay** instead of **overlayfs** but since I found it to be broken it is not yet functional. It will need to be fully support at a later stage though.

sfabbro:
I fixed a few indents and the check_aufs which does construct a /sys/fs/aufs even with a built-in. But overlayfs does not, I checked on a Gentoo 4.1.4 hardened kernel, basically similar to a vanilla upstream kernel regarding overlayfs.
It looks like now overlay is the name of the module and the mount option for overlayfs also on the latest versions.

PR #1057, opened 2015-08-19
Catalog Manager Test 1

Moliholy:
Makes the AbstractCatalogManager templated. It also defines the new MockCatalogManager class.

PR #1058, opened 2015-08-19
Catalog Manager Test 2

Moliholy:
Completes the tests for the catalog manager class. It includes https://github.com/cvmfs/cvmfs/pull/1057


test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,174 @@
...
+
+using namespace std;  // NOLINT
+
+
+namespace catalog {
+
+class T_CatalogManager : public ::testing::Test {
+ protected:
+  void SetUp() {
+    catalog_mgr_ = new MockCatalogManager(&statistics_);
```
reneme: Could actually just be allocated on the stack to avoid the d'tor, no?

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,174 @@
...
+  void AddTree() {
+    const shash::Any empty_hash = shash::Any();
+    const int size = 32;
+    const size_t file_size = 4096;
+    const char suffix = shash::kSha1;
+    shash::Any hash;
+    shash::Any empty_content;
+    // adding ""
+    MockCatalog *root_catalog = catalog_mgr_->RetrieveRootCatalog();
+    root_catalog->AddFile(empty_content, file_size, "", "");
```
reneme: **File** and **Directory** are the same here?

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,174 @@
...
+  EXPECT_TRUE(catalog_mgr_->Init());
+  EXPECT_EQ(1, catalog_mgr_->GetNumCatalogs());
+  EXPECT_EQ(1u, catalog_mgr_->GetRevision());
+  EXPECT_FALSE(catalog_mgr_->GetVolatileFlag());
+  EXPECT_EQ(0u, catalog_mgr_->GetTTL());
+}
+
+TEST_F(T_CatalogManager, Statistics) {
+  EXPECT_TRUE(catalog_mgr_->Init());
+  Statistics st = catalog_mgr_->statistics();
```
reneme: Could actually be a `const Statistics&` both here and in the `CatalogManager` to avoid copying.

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,174 @@
...
+  catalog::DirectoryEntry dirent;
+  ASSERT_TRUE(catalog_mgr_->Init());
+  AddTree();
+  EXPECT_TRUE(catalog_mgr_->LookupPath("/dir", kLookupSole, &dirent));
+  EXPECT_TRUE(dirent.IsDirectory());
+  EXPECT_TRUE(catalog_mgr_->LookupPath("/dir/dir", kLookupSole, &dirent));
+  EXPECT_TRUE(dirent.IsDirectory());
+  EXPECT_TRUE(catalog_mgr_->LookupPath("/file1", kLookupSole, &dirent));
+  EXPECT_TRUE(dirent.IsRegular());
+  // the father directory belongs to the catalog, so there is no problem
```
reneme: ??? `kLookupFull` should work in any case, shouldn't it?

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,174 @@
...
+  EXPECT_TRUE(catalog_mgr_->LookupPath("/dir/dir/file2", kLookupFull, &dirent));
+  EXPECT_TRUE(dirent.IsRegular());
+  // /dir/dir/dir/file4 belongs to a catalog that is not mounted yet
```
reneme: same here, I believe...

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,174 @@
...
+  // the father directory should also belong to the nested catalo
+  EXPECT_TRUE(catalog_mgr_->LookupPath("/dir/dir/dir/file4", kLookupFull,
+                                       &dirent));
+  // it is not a symplink, so it should crash
+  EXPECT_DEATH(catalog_mgr_->LookupPath("/dir/dir/dir/file4", kLookupRawSymlink,
+                                         &dirent), ".*");
+  EXPECT_EQ(2, catalog_mgr_->GetNumCatalogs());
+}
+
+TEST_F(T_CatalogManager, Listing) {
```
reneme: Perhaps add a second `Listing` test that tries to list a non-existent directory inside the nested catalog and checks if the catalog was properly loaded.

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,174 @@
...
+                          size, suffix);
+    new_catalog->AddFile(hash, file_size, "/dir/dir/dir", "file3");
+    // adding "/dir/dir/dir/file4" to the new nested catalog
+    hash = shash::Any(shash::kSha1,
+                          reinterpret_cast<const unsigned char*>(hashes[3]),
+                          size, suffix);
+    new_catalog->AddFile(hash, file_size, "/dir/dir/dir", "file4");
+    // we haven't mounted the second catalog yet!
+    ASSERT_EQ(1, catalog_mgr_->GetNumCatalogs());
+  }
```
reneme: Please add a couple of more catalogs to make the test a bit more realistic. For example, a single lookup should cause a whole branch of catalogs to be loaded. Like:
```bash
ls /dir/dir/dir/dir/dir/dir
```
Should load **/dir/dir** and in the same lookup **/dir/dir/dir/dir**.

test/unittests/t_catalog_mgr.cc
Moliholy: Yes, they are

test/unittests/t_catalog_mgr.cc
Moliholy: There is a test where it looks up /dir/dir/dir/file4, which loads the nested catalog in /dir/dir/dir

test/unittests/t_catalog_mgr.cc
Moliholy: It should, but just in case I have added the same test with kLooupFull ;)

test/unittests/t_catalog_mgr.cc
Moliholy: Added the same look up operation but using kLookupFull

test/unittests/t_catalog_mgr.cc
Moliholy: Test added and passed

test/unittests/t_catalog_mgr.cc
Moliholy: Added test with "long" lookups and listdirs

jblomer:
Can you please rebase this PR to the latest devel which has the first PR from this series merged.

Moliholy:
PR updated to the latest devel

cvmfs/catalog_mgr.h
```diff
@@ -217,6 +217,7 @@ class AbstractCatalogManager : public SingleCopy {
                                 shash::Any   *catalog_hash) = 0;
   virtual void UnloadCatalog(const CatalogT *catalog) { }
   virtual void ActivateCatalog(CatalogT *catalog) { }
+  const std::vector<CatalogT*>& GetCatalogs() const { return catalogs_; }
```
jblomer: Is this used?

cvmfs/catalog_mgr_impl.h
```diff
@@ -40,6 +40,7 @@ AbstractCatalogManager<CatalogT>::AbstractCatalogManager(
   assert(retval == 0);
   retval = pthread_key_create(&pkey_sqlitemem_, NULL);
   assert(retval == 0);
+  remount_listener_ = NULL;
```
jblomer: This was previously removed.

cvmfs/catalog_mgr_impl.h
```diff
@@ -115,6 +116,8 @@ LoadError AbstractCatalogManager<CatalogT>::Remount(const bool dry_run) {
     return LoadCatalog(PathString("", 0), shash::Any(), NULL, NULL);
 
   WriteLock();
+  if (remount_listener_)
```
jblomer: see above

cvmfs/catalog_mgr_impl.h
```diff
@@ -115,6 +116,8 @@ LoadError AbstractCatalogManager<CatalogT>::Remount(const bool dry_run) {
...
+    remount_listener_->BeforeRemount(this);
```
jblomer: see above

test/unittests/t_catalog_manager.cc
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
jblomer: This file should be entirely removed, no?  The actual unit tests are in `t_catalog_mgr.cc`

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,237 @@
...
+
+
+namespace catalog {
+
+class T_CatalogManager : public ::testing::Test {
+ public:
+  T_CatalogManager() : statistics_(), catalog_mgr_(&statistics_) { }
+
+ protected:
+  void AddTree() {
```
jblomer: Please add a comment explaining which directory structure the method creates

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,237 @@
...
+     "6d7fce9fee471194aa8b5b6e47267f03000000",
+     "48a24b70a0b376535542b996af517398000000",
+     "1dcca23355272056f04fe8bf20edfce0000000"
+};
+
+
+
+TEST_F(T_CatalogManager, InitialConfiguration) {
+  EXPECT_TRUE(catalog_mgr_.Init());
+  EXPECT_EQ(1, catalog_mgr_.GetNumCatalogs());
```
jblomer: Shouldn't this be zero?

Moliholy: No, after init() it loads the root catalog

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,237 @@
...
+  const Statistics &st = catalog_mgr_.statistics();
+  EXPECT_EQ(0u, st.n_listing->Get());
+  EXPECT_EQ(0u, st.n_lookup_inode->Get());
+  EXPECT_EQ(0u, st.n_lookup_path->Get());
+  EXPECT_EQ(0u, st.n_lookup_path_negative->Get());
+  EXPECT_EQ(0u, st.n_lookup_xattrs->Get());
+  EXPECT_EQ(0u, st.n_nested_listing->Get());
+}
+
+TEST_F(T_CatalogManager, InodeConfiguration) {
```
jblomer: Here it would be nice to add some code to actually test if the inode annotation works as expected

test/unittests/t_catalog_mgr.cc
```diff
@@ -0,0 +1,237 @@
...
+  EXPECT_FALSE(catalog_mgr_.Listing("/dir/dir/dir/dir/dir/fakedir", &del));
+  EXPECT_EQ(0u, del.size());
+  EXPECT_EQ(3, catalog_mgr_.GetNumCatalogs());
+}
+
+TEST_F(T_CatalogManager, Remount) {
+  EXPECT_TRUE(catalog_mgr_.Init());
+  LoadError le;
+  EXPECT_EQ(kLoadNew, le = catalog_mgr_.Remount(true));
+  EXPECT_EQ("loaded new catalog", Code2Ascii(le));
```
jblomer: that can be removed, we might change the error explanation text

test/unittests/testutil.cc
```diff
@@ -245,7 +247,55 @@ MockCatalog* MockCatalog::AttachFreely(const std::string  &root_path,
...
 }
 
-void MockCatalog::RegisterChild(MockCatalog *child) {
+void MockCatalog::RemoveChild(MockCatalog *child) {
+  std::vector<NestedCatalog>::iterator iter = active_children_.begin();
+  bool found = false;
+  while (!found && iter != active_children_.end()) {
+    if (iter->hash == child->hash()) {
+      active_children_.erase(iter);
+      found = true;
```
jblomer: This is correct but a bit Pascal like.  You can actually use `break` once the catalog is found.

test/unittests/testutil.cc
```diff
@@ -245,7 +247,55 @@ MockCatalog* MockCatalog::AttachFreely(const std::string  &root_path,
...
+      *dirent = files_[i].ToDirectoryEntry();
+      return true;
+    }
+  }
+  return false;
+}
+
+bool MockCatalog::ListingPath(const PathString &path,
+                 catalog::DirectoryEntryList *listing) const {
+  listing->clear();
```
jblomer: Does the real catalog clear the listing, too?

test/unittests/t_catalog_manager.cc
Moliholy: No, in the last PR I already created only the skeleton of the file, but nothing else. However that extra star at the beginning of the file is a small error that is fixed in this PR

test/unittests/testutil.cc
```diff
@@ -245,7 +247,55 @@ MockCatalog* MockCatalog::AttachFreely(const std::string  &root_path,
...
+  bool found = false;
+  for (unsigned i = 0; i < files_.size(); ++i) {
+    shash::Md5 path_hash(path.GetChars(), path.GetLength());
+    if (files_[i].parent_hash == path_hash)
+      listing->push_back(files_[i].ToDirectoryEntry());
+    else if (files_[i].path_hash == path_hash)
+      found = true;
+  }
+  if (!found)  // the parent does not exists, nor the children then
```
jblomer: this is not necessary, is it?  I think the real catalog would return children even if the parent is missing.

test/unittests/testutil.cc
```diff
@@ -245,7 +247,55 @@ MockCatalog* MockCatalog::AttachFreely(const std::string  &root_path,
   return new_catalog;
 }
 
-void MockCatalog::RegisterChild(MockCatalog *child) {
+void MockCatalog::RemoveChild(MockCatalog *child) {
```
jblomer: `RemoveNestedCatalog`?  To keep the naming consistent with `RegisterNestedCatalog`.

test/unittests/t_catalog_manager.cc
jblomer: But what is supposed to come into this file?  Isn't everything supposed to be in this file now in `t_catalog_mgr.cc`?

test/unittests/testutil.cc
jblomer: Ah sorry, looks like `AddChild` / `RemoveChild` is different from `RegisterNestedCatalog`.  Actually, what is the difference?

cvmfs/catalog_mgr.h
Moliholy: Yes, here:
```
void WritableCatalogManager::Balance() {
  CatalogList catalog_list = GetCatalogs();
  for (unsigned i = 0; i < catalog_list.size(); ++i) {
    FixWeight(catalog_list[i]);
  }
}
```
I need to get the catalog list in order to balance them. I though it was better to return a constant reference instead of directly using the private attribute.

test/unittests/t_catalog_manager.cc
```diff
@@ -1,33 +0,0 @@
...
-
-
-
-
-
-
-
-
-
-}  // namespace catalog
```
Moliholy: That was an error I introduced in the last PR. The actual file where the catalog manager tests are is t_catalog_mgr.cc, and not t_catalog_manager.cc

test/unittests/testutil.cc
Moliholy: No, it doesn't. It directly pushes back the new DirectoryEntries. However, I don't really see the point of not clearing the list before writing to it. Actually, when making a lookup the DirectoryEntry is at the beginning initialised to an empty value. The same way it could (and in my opinion should) be done in the CatalogManager, isn't it?

test/unittests/testutil.cc
Moliholy: But that sounds like a huge lack of consistency... isn't it?

test/unittests/testutil.cc
Moliholy: In normal catalogs the information about direct nested catalogs is stored in the database, so that you can easily know what is the total list of children-catalog. Also, catalogs have a list of already attached children, which is always a subset of the ones mentioned before.

In the mock catalog there is sqlite database, so there are two lists: one for the available children and another for the mounted children. The first one is constant, but the second one can be modified. In any case it's necessary to have different methods to operate with each list, and that includes inserting and, only in the mount list, deleting elements.

jblomer: Ok, thanks!  Just add a comment to each one of the functions that explains which of the list the function modifies.

Moliholy:
Pull request updated

jblomer:
Thanks!  Please remember to add comments to `MockCatalog::RemoveChild`, `MockCatalog::AddChild`, and `MockCatalog::RegisterNested`

Moliholy:
Documentation added

PR #1060, opened 2015-08-20
TEST/FIX: GeoIP Database Update

reneme:
This is a followup to yesterday's [pull request](https://github.com/cvmfs/cvmfs/pull/1056). Mainly it adds an integration test to verify the proper updating behaviour of the GeoIP database on Stratum1s. Furthermore it fixes a couple of minor sanity check inconsistencies.

@DrDaveD: Could you have a specific look at [this commit](https://github.com/cvmfs/cvmfs/commit/2964e13ca5d13f1a75412524954170b36e8ad3ba). I needed to remove the `check_apache()` and `check_wsgi_module()` calls from `cvmfs_server snapshot` as they assume root privileges. Do you think, we can survive without checking the proper installation of apache and WSGI on every `snapshot`?

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -4379,15 +4379,16 @@ snapshot() {
     # more sanity checks
     is_owner_or_root $alias_name || { echo "Permission denied: Repository $alias_name is owned by $user"; retcode=1; continue; }
     check_repository_compatibility
-    if is_local_upstream $CVMFS_UPSTREAM_STORAGE; then
-        # try to update the geodb, but continue if it doesn't work
-        _update_geodb -l || true
-    fi
     [ ! -z $stratum1 ] || die "Missing CVMFS_STRATUM1 URL in server.conf"
     gc_timespan="$(get_auto_garbage_collection_timespan $alias_name)" || { retcode=1; continue; }
 
     # do it!
     local user_shell="$(get_user_shell $alias_name)"
+
+    if is_local_upstream $CVMFS_UPSTREAM_STORAGE; then
+        # try to update the geodb, but continue if it doesn't work
+        _update_geodb -l || true
+    fi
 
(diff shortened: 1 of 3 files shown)


DrDaveD:
I think we could survive without the check, although as the comment says then it might never get run.  On the other hand, if you did the logic as
```
if is_local_upstream ...; then
    if check_apache; then
        check_wsgi_module
    fi
    _update_geodb -l || true
fi
```
then it would only skip the wsgi module check when non-root because check_apache will just fail when non-root.  I don't know how else to make sure the check is ever done when somebody uses add-replica -a.  They would just have to somehow know that it was needed.

cvmfs/cvmfs_server
```diff
@@ -2714,7 +2711,12 @@ add_replica() {
...
       check_apache || die "Apache must be installed and running"
       check_wsgi_module
-      [ x"$cvmfs_user" = x"root" ] || echo "NOTE: If snapshot is not run regularly as root, do update-geodb monthly from cron"
+      if [ x"$cvmfs_user" != x"root" ]; then
+        echo "NOTE: If snapshot is not run regularly as root, the GeoIP database will not be updated."
+        echo "      You have two options, either:"
+        echo "      1. chown $CVMFS_UPDATEGEO_DIR and $CVMFS_UPDATEGEO_DB accordingly OR"
+        echo "      2. run update-geodb monthly as root"
+      fi
     else
```
DrDaveD: Another option would be to chown the $CVMFS_UPDATEGEO_DIR and $CVMFS_UPDATEGEO_DB to a different user id and run update-geodb monthly as that user id.

reneme: True. I'll add that, thanks.

reneme:
Added both Dave's proposals. Merging, building, kicking-off tests and calling it a day.

PR #1063, opened 2015-08-21
FIX: Revert Inadvertently Added Debug Code

reneme:
Memo to myself: Don't ever do `git commit -am`!


jblomer:
Hm, what was the problem with this code?  If I remember correctly, the idea was to print out the number of inodes cvmfs needs to shadow.  If this test results in a memory warning, the most likely explanation is that the page cache was not fully cleaned.

reneme:
I copied that code into a random test (002-dummy in that case) to test it. It resides in `test_functions` but didn't work properly in my tests. However, it turned out to be due to the changed `Statistics` class. I just inadvertently checked in my debugging stuff...

PR #1065, opened 2015-08-21
Update bootstrap.sh

MicBrain:
Updating bootrap.sh file for the nghttp2 library


jblomer:
This should contain the new tarball, too, otherwise the PR would be inconsistent in itself.

PR #1066, opened 2015-08-21
Update bootstrap.sh

MicBrain:
Updating the bootstrap file after includeing nghttp2 library.


jblomer:
Looks like a duplicate.

PR #1067, opened 2015-08-21
Update CMakeLists.txt

MicBrain:
Updating CmakeList.txt for new nghttp2 library.


jblomer:
Looks like a duplicate.

PR #1073, opened 2015-08-25
FIX: Unittest Sandboxing

reneme:
This fixes [yesterday's pull request](https://github.com/cvmfs/cvmfs/pull/1069) that tried to exploit GTest's feature to switch to the working directory of the parent process when spawning child processes for death tests. Unfortunately this doesn't work, because GTest determines this working directory before even `main()` is running in the parent process. Hence, death tests would not properly switch to the sandbox directory and start littering in the wrong location.

This fixes `CvmfsEnvironment` which now drops a file like **/tmp/cvmfs_ut_sandbox.pid.$PID** which can be found by the child processes using `getppid()`. The file contains the full path of the sandbox directory created by the main GTest process. This way we can actively switch to the sandbox before the death test is running.

### DIFF
--- a/test/unittests/env.cc
+++ b/test/unittests/env.cc
@@ -5,8 +5,13 @@
 #include "env.h"
 #include "../../cvmfs/util.h"
 
+#include <cassert>
 #include <cerrno>
+#include <cstdlib>
+
+#include <fstream>
 #include <string>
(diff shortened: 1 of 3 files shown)


jblomer:
I think it would be better to use an environment variable instead.

PR #1082, opened 2015-08-31
CI: Docker Based Build Scripts

reneme:
This adapts the continuous integration scripts to be run in a dockerized environment. Mainly this restructures the `ci` sub directory and adds a couple of wrapper scripts to be used from Jenkins.


jblomer:
We shouldn't blindly take the images from DockerHub.  God knows what runs inside.  We should instead upload required images to our docker registry.

ci/docker/el4/build.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
jblomer: This thing should take our EL4 packages

reneme:
For the moment I'm just testing with them, lets see what comes out.
On the other hand: These are the official images from CentOS, not sure if God is so naughty to create them in a weird way. Switching to a self-maintained source (if it becomes necessary) shouldn't be too hard to do.

jblomer:
It's a bit like doing 'curl http://probably.trustworthy.com/script.sh | sudo bash'.  We do a lot of effort to sign cvmfs packages.  It would already be a step forward if these docker images were signed by a well-known CentOS key and we checked the signature.  But I'd actually prefer more control over the build environment.  For instance, if we try to debug a problem related to the build environment 3 weeks after the release but CentOS happened to update the image, how would we be able to recreate the original build conditions...

reneme:
> It's a bit like doing `curl http://probably.trustworthy.com/script.sh | sudo bash`

Agreed. I'll have a look how to configure it to use only our own images.

PR #1087, opened 2015-09-01
CI: Cross-Compiling i386 on EL 6 and 7

reneme:


reneme:
In fact only on 6 at the moment. Because in 5 it doesn't work yet and 7 doesn't need it.

PR #1091, opened 2015-09-07
Catalog balancer feature

Moliholy:
This new feature allows the WritableCatalogManager to (if configured to do so), rebalance those catalogs that have more/less entries than a maximum/minimum threshold in an optimal way.
It has been introduced right before the actual publishing operation occurs, and it creates the corresponding ".cvmfscatalog" files plus the new ".cvmfsautocatalog" files.

If a catalog is autogenerated (from now on they will be called "autocatalogs") then it can be removed if necessary. The same way every catalog can be automatically rebalanced according to the provided parameters. However, a catalog that has been manually created by the user will never be deleted, including those included in the .cvmfsdirtab.

These already mentioned parameters can be specified in the server configuration file, being CVMFS_AUTOCATALOG the only mandatory one, and false by default. The two others will be CVMFS_AUTOCATALOG_MAX_WEIGHT for the overflow threshold, and CVMFS_AUTOCATALOG_MIN_WEIGHT for the underflow threshold.

New unit test and integration test have been added for this new feature.

This pull request uses https://github.com/cvmfs/cvmfs/pull/1057 and https://github.com/cvmfs/cvmfs/pull/1058 as starting point.


Moliholy:
I will wait for the resolution of https://github.com/cvmfs/cvmfs/pull/1058 to add changes

Moliholy:
This pull request is ready for revision

Moliholy:
Nevermind, there are still some incompatibilities with the new code

Moliholy:
OK, now it is finally ready. Both the unit tests and integration test have passed.
Cpplint passed as well

cvmfs/catalog_mgr_rw.cc
```diff
@@ -837,6 +847,34 @@ shash::Any WritableCatalogManager::SnapshotCatalog(WritableCatalog *catalog)
...
+  if (catalog->GetNumEntries() < min_weight_ &&
+      !catalog->IsRoot() &&
+      catalog->IsAutogenerated()) {
+    LogCvmfs(kLogCatalog, kLogStdout,
+             "deleting an autogenerated catalog in '%s'",
+             catalog->path().c_str());
+    // Remove the .cvmfscatalog and .cvmfsautocatalog files first
+    string path = catalog->path().ToString();
+    catalog->RemoveEntry(path + "/.cvmfscatalog");
+    catalog->RemoveEntry(path + "/.cvmfsautocatalog");
```
Moliholy: I think it's better if the actual deletion of these files is outside the RemoveCatalog method. It could interfiere with the manual deletion of other catalogs.

cvmfs/catalog_mgr.h
```diff
@@ -139,6 +139,7 @@ template <class CatalogT>
 class AbstractCatalogManager : public SingleCopy {
  public:
   typedef std::vector<CatalogT*> CatalogList;
+  typedef CatalogT catalog_t;
```
jblomer: What's the purpose of this typedef?

cvmfs/catalog_mgr_impl.h
```diff
@@ -550,8 +550,9 @@ CatalogT* AbstractCatalogManager<CatalogT>::FindCatalog(
  * @return true if catalog is already present, false otherwise
  */
 template <class CatalogT>
-bool AbstractCatalogManager<CatalogT>::IsAttached(const PathString &root_path,
-                                        CatalogT **attached_catalog) const
+bool AbstractCatalogManager<CatalogT>::IsAttached(
```
jblomer: Please don't change the style of existing code unless it violates the linter or you are actually modifying things.  Such changes easily take the focus away from the actual purpose of the patch.

cvmfs/catalog_mgr_impl.h
```diff
@@ -572,9 +573,10 @@ bool AbstractCatalogManager<CatalogT>::IsAttached(const PathString &root_path,
  * The final leaf nested catalog is returned.
  */
 template <class CatalogT>
-bool AbstractCatalogManager<CatalogT>::MountSubtree(const PathString &path,
-                                          const CatalogT *entry_point,
-                                          CatalogT **leaf_catalog)
+bool AbstractCatalogManager<CatalogT>::MountSubtree(
```
jblomer: See above

cvmfs/catalog_mgr_impl.h
```diff
@@ -665,8 +667,9 @@ CatalogT *AbstractCatalogManager<CatalogT>::MountCatalog(
  * @return true on success, false otherwise
  */
 template <class CatalogT>
-bool AbstractCatalogManager<CatalogT>::AttachCatalog(const string &db_path,
-                                           CatalogT *new_catalog)
+bool AbstractCatalogManager<CatalogT>::AttachCatalog(
```
jblomer: See above

cvmfs/catalog_mgr_rw.cc
```diff
@@ -12,9 +12,12 @@
 #include <cassert>
 #include <cstdio>
 #include <cstdlib>
+#include <ctime>
```
jblomer: Is the `ctime` header needed?

jblomer: Ah, indeed.  But that's a flaw already in the existing code.  Please make a separate pull request for fixes in which this missing include can be added.

cvmfs/catalog_mgr_rw.cc
```diff
@@ -12,9 +12,12 @@
...
 #include <string>
 
 #include "catalog_rw.h"
+#include "directory_entry.h"
```
jblomer: See above

cvmfs/catalog_mgr_rw.cc
```diff
@@ -12,9 +12,12 @@
...
+#include "hash.h"
```
jblomer: See above

cvmfs/catalog_mgr_rw.cc
```diff
@@ -837,6 +847,34 @@ shash::Any WritableCatalogManager::SnapshotCatalog(WritableCatalog *catalog)
...
+  // firstly check underflow because they can provoke overflows
+  if (catalog->GetNumEntries() < min_weight_ &&
+      !catalog->IsRoot() &&
+      catalog->IsAutogenerated()) {
+    LogCvmfs(kLogCatalog, kLogStdout,
+             "deleting an autogenerated catalog in '%s'",
+             catalog->path().c_str());
+    // Remove the .cvmfscatalog and .cvmfsautocatalog files first
+    string path = catalog->path().ToString();
+    catalog->RemoveEntry(path + "/.cvmfscatalog");
```
jblomer: This will not trigger the actual merging of the catalog with its parent, will it?

cvmfs/catalog_mgr_rw.cc
Moliholy: I just pass the cpplint... and this is what I get. Should they now be (un)fixed to later be (re)fixed?

cvmfs/catalog_mgr_rw.cc
```diff
@@ -837,6 +847,34 @@ shash::Any WritableCatalogManager::SnapshotCatalog(WritableCatalog *catalog)
   return hash_catalog;
 }
 
+void WritableCatalogManager::Balance() {
+  CatalogList catalog_list = GetCatalogs();
```
jblomer: Locking is required, I think.

cvmfs/catalog_mgr_rw.cc
jblomer: As I said, a separate pull request for necessary fixes which will be merged first.  The new feature should then be a patch on top of the clean(er) code base.

cvmfs/catalog_mgr.h
Moliholy: It was at some point, but it ended up not being, and forgotten. I'll fix it

cvmfs/catalog_mgr_rw.cc
jblomer: To be more precise: if you need the include for the patch, it is of course part of the patch pull request.  But if the include was missing in the first place, this should be a separate pull request with a fix.

cvmfs/catalog_mgr_rw.h
```diff
@@ -34,9 +34,11 @@
 
 #include <set>
 #include <string>
+#include <vector>
```
jblomer: This is not necessary in this header, I think

cvmfs/catalog_mgr_rw.h
```diff
@@ -34,9 +34,11 @@
...
 
 #include "catalog_mgr_ro.h"
 #include "catalog_rw.h"
+#include "directory_entry.h"
```
jblomer: Again, missing in the first place

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
    * Directories don't have extended attributes at this point.
    */
   XattrList empty_xattrs;
+
+  /**
+   * Balancer thresholds
```
jblomer: Please add more helpful comments as to what these members steer in the lifetime of the catalog manager.

cvmfs/catalog_mgr_rw.cc
Moliholy: No, it only removes the entry, as in the next line. Three lines down it actually merges both catalogs.
```
    string path = catalog->path().ToString();
    catalog->RemoveEntry(path + "/.cvmfscatalog");
    catalog->RemoveEntry(path + "/.cvmfsautocatalog");
    // Remove the actual catalog
    string catalog_path = catalog->path().ToString().substr(1);
    RemoveNestedCatalog(catalog_path);
```

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+   */
+  const bool is_balanced_;
+  const unsigned max_weight_;
+  const unsigned min_weight_;
+  const unsigned balance_weight_;
 };  // class WritableCatalogManager
 
+
+template <class CatalogMgrT>
```
jblomer: I think it's better to separate into to compilation units.  The `CatalogBalancer` has a right to get its own .cc and .h files.

cvmfs/catalog_mgr_rw.cc
Moliholy: Off topic, while I was coding this I thought it would be a good idea to have a class/namespace in a sole file that can be exported and would contain all this constants. It's not precisely a good practise to hardcode file names, because one day it can be renamed, and that day we would have to look in all the files to check wether they contain the ".cvmfscatalog" string or not. What do you think?

cvmfs/catalog_mgr_rw.h
jblomer: The `CatalogBalancer` should come with a nice comment paragraph.  Why does it exist, what's its relationship to other classes in cvmfs, how does it roughly solve the problem.

cvmfs/catalog_mgr_rw.cc
Moliholy: Is it possible to have two or more publish processes at the same time in the same process? At this point I thought catalogs are totally static...

cvmfs/catalog_mgr_rw.cc
jblomer: It would be nicer without magic strings, yes.  A separate file might be the right think if the magic strings are used in multiple classes, otherwise a constant member would work, too.  This requires a record of the status quo first (what are our magic strings and where).

cvmfs/catalog_mgr_rw.h
```diff
@@ -111,6 +123,8 @@ class WritableCatalogManager : public SimpleCatalogManager {
 
   manifest::Manifest *Commit(const bool     stop_for_tweaks,
                              const uint64_t manual_revision);
+  void Balance();
+  void FixWeight(WritableCatalog *catalog);
```
jblomer: Can this be private?

cvmfs/catalog_mgr_rw.cc
jblomer: You might use it in a correct way but it is a public method and this class is generally thread-safe.  So if you don't lock, there should be a good reason (e.g. locking has been done by the caller), which should be documented.

cvmfs/catalog_mgr_rw.h
Moliholy: CatalogBalancer is templated, so the files would be catalog_balancer.h and catalog_balancer_impl.h. Since WritableCatalogManager is not templated we could get rid of the catalog_mgr_rw_impl.h file, which is good.

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+template <class CatalogMgrT>
+class CatalogBalancer {
+ public:
+  typedef typename CatalogMgrT::catalog_t catalog_t;
+  explicit CatalogBalancer(CatalogMgrT *catalog_mgr)
+    : catalog_mgr_(catalog_mgr) { }
+
+  void Balance(catalog_t *catalog = NULL);
+
+
```
jblomer: One blank line should suffice

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+ protected:
+  struct VirtualNode {
```
jblomer: Please document.  What does a virtual node represent, what's its purpose in life...

cvmfs/catalog_mgr_rw.h
Moliholy: It could, but I thought it could be useful to be able to individually balance a node (catalog). If this is not the case it can be perfectly private

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+      : children(), weight(1), dirent(dirent), path(path),
+        is_new_nested_catalog(false) {
+      if (!IsCatalog() && IsDirectory())
+        ExtractChildren(catalog_mgr);
+    }
+    bool IsDirectory() { return dirent.IsDirectory(); }
+    bool IsCatalog() { return is_new_nested_catalog ||
+        dirent.IsNestedCatalogMountpoint(); }
+  };
+  typedef typename CatalogBalancer<CatalogMgrT>::VirtualNode virtual_node_t;
```
jblomer: In the implementation of the Balancer, can't you just use `VirtualNode`?

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+
+ private:
+  void OptimalPartition(VirtualNode *virtual_node);
```
jblomer: Should be a verb.  Perhaps `PartitionOptimally`?

cvmfs/catalog_mgr_rw.h
jblomer: Then better private by default. It can still be easily promoted to a public method if we need it at a later point.

cvmfs/catalog_mgr_rw_impl.h
```diff
@@ -0,0 +1,181 @@
...
+#include <unistd.h>
+
+#include <cassert>
+#include <cstdio>
+#include <cstdlib>
+#include <ctime>
+#include <string>
+#include <vector>
+
+#include "catalog_rw.h"
```
jblomer: It looks to me as there are many includes in this file which are not really necessary. Pleas re-check.

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+  void AddCvmfsCatalogFile(string path);
+  DirectoryEntryBase CreateEmptyContentDirectoryEntryBase(string name,
```
jblomer: That's a very long name (but precise).  Please think 5-10 minutes if you can find a shorter name.  If nothing comes up, we'll leave it as is.

cvmfs/catalog_mgr_rw_impl.h
```diff
@@ -0,0 +1,181 @@
...
+
+template <class CatalogMgrT>
+DirectoryEntryBase
+CatalogBalancer<CatalogMgrT>::CreateEmptyContentDirectoryEntryBase(
+    string name, uid_t uid, gid_t gid) {
+  shash::Algorithms algorithm = catalog_mgr_->spooler_->GetHashAlgorithm();
+  shash::Any file_hash(algorithm);
+  shash::HashString("", &file_hash);
+
+  DirectoryEntryBase deb;
```
jblomer: Many of the following initializations are already done by the default constructor.  It should be enough to only set `name_`, `mode_`, `checksum_`, `mtime_`,  `uid_` and `gid_`.

cvmfs/catalog_mgr_rw_impl.h
```diff
@@ -0,0 +1,181 @@
...
+  deb.mtime_ = time(NULL);
+  deb.inode_ = catalog::DirectoryEntry::kInvalidInode;
+  deb.parent_inode_ = catalog::DirectoryEntry::kInvalidInode;
+  deb.uid_ = uid;
+  deb.gid_ = gid;
+  return deb;
+}
+
+template <class CatalogMgrT>
+void CatalogBalancer<CatalogMgrT>::AddCvmfsCatalogFile(string path) {
```
jblomer: We call these usually `CatalogMarker` to avoid confusion with the file that actually contains the SQlite catalog.

cvmfs/catalog_mgr_rw_impl.h
```diff
@@ -0,0 +1,181 @@
...
+  }
+  virtual_node->CaltulateWeight();
+  while (virtual_node->weight > catalog_mgr_->balance_weight_) {
+    typename CatalogBalancer<CatalogMgrT>::VirtualNode *heaviest_node =
+        MaxChild(virtual_node);
+    // we directly add a catalog in this node
+    // TODO(molina) apply heuristics here
+    if (heaviest_node != NULL &&
+        heaviest_node->weight >= catalog_mgr_->min_weight_) {
+      AddCvmfsCatalogFile(heaviest_node->path);
```
jblomer: Couldn't that leave us with a nested catalog in the heaviest node that again is too big?

cvmfs/catalog_mgr_rw.cc
Moliholy: Which implies that each and every small fix like this one has to be carefully analysed before being fixed, "loosing" time in the process, and even more if these analysis were wrong. I doubt the time lost and effort put is worth having into account that the final status of the repository will be the same...

jblomer: Depends on the setup of your tool chain.  It shouldn't take you longer than a few minutes to submit a few fixes as a fix PR.  Otherwise please have a look how you can do this faster.  The point is: please don't mix fixes and new features.

jblomer: Or perhaps I don't understand... what do you need to analyse exactly?

cvmfs/catalog_mgr_rw.cc
Moliholy: It is a protected method, that's why I left it that way. In any case it returns a reference to the attribute, which could be anyway modified after some time in other parts of the code by other threads

jblomer: Sorry, I don't mean `GetCatalogs` here. I mean that `Balance()` needs to get proper locking.

cvmfs/catalog_mgr_rw.cc
Moliholy: The situation would be like:
"I'm done with the new feature, so I pass the cpplint and I get 18 warnings." Now my standard procedure with be to automatically fix them all, and not to guess which ones belong to my new feature because that:
1) Requires time
2) I can be mistaken and I don't include a fix that I should, or the other way around

For small fixes like this one I think it's not so important if they do not strictly belong to the PR itself since _anyway_ they will be included in the next PR.
But yes, I admit that the counterpart of this is that the reviewer has more stuff to analyse, but anyway he will have to do it sooner or later

jblomer: Well, in my workflow, I go through the PR line by line and, among other things, I mark spots that are fixes to the existing code (i.e. I do the analysis, you don't have to do it).  If by the end of the review, we have only two or three lines of pure fixes, we can be sloppy and piggyback on the same PR.  

If it is more than that or if the PR anyway requires careful review (like this one), every line of code we can remove from the core of the patch makes my life easier. It helps me to focus on the important parts and not get distracted by lots of other changes.

cvmfs/catalog_mgr_rw.cc
Moliholy: Ah, in that case you are totally right. It should be locked.
However it's important to notice that the whole catalog manager would be locked for some seconds... or even minutes

cvmfs/catalog_mgr_rw_impl.h
```diff
@@ -0,0 +1,181 @@
...
+  retval = catalog_mgr_->LookupPath(PathString(path), kLookupSole, &parent);
+  assert(retval);
+  DirectoryEntryBase cvmfscatalog =
+      CreateEmptyContentDirectoryEntryBase(".cvmfscatalog", parent.uid(),
+                                           parent.gid());
+  DirectoryEntryBase cvmfsautocatalog =
+      CreateEmptyContentDirectoryEntryBase(".cvmfsautocatalog", parent.uid(),
+                                           parent.gid());
+  string relative_path = path.substr(1);
+  catalog_mgr_->AddFile(cvmfscatalog, xattr, relative_path);
```
jblomer: There is a possibility that the user created a `.cvmfsautocatalog` file.  How would the code handle this?

cvmfs/catalog_mgr_rw.cc
jblomer: Indeed that's important information worth a comment in the code.

cvmfs/catalog_mgr_rw_impl.h
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+
+template <class CatalogMgrT>
+void CatalogBalancer<CatalogMgrT>::VirtualNode::ExtractChildren(
+    CatalogMgrT *catalog_mgr) {
+  DirectoryEntryList direntlist;
+  catalog_mgr->Listing(path, &direntlist);
+  for (unsigned i = 0; i < direntlist.size(); ++i) {
+    string children_path = path + "/" + direntlist[i].name().ToString();
```
jblomer: `children_path` --> `child_path`

cvmfs/catalog_mgr_rw_impl.h
```diff
@@ -0,0 +1,181 @@
...
+    children.push_back(virtual_node_t(
+        children_path, direntlist[i], catalog_mgr));
+    weight += children[i].weight;
+  }
+}
+
+template <class CatalogMgrT>
+void CatalogBalancer<CatalogMgrT>::VirtualNode::CaltulateWeight() {
```
jblomer: `CalculateWeight`?

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+class CatalogBalancer {
+ public:
+  typedef typename CatalogMgrT::catalog_t catalog_t;
+  explicit CatalogBalancer(CatalogMgrT *catalog_mgr)
+    : catalog_mgr_(catalog_mgr) { }
+
+  void Balance(catalog_t *catalog = NULL);
+
+
+ protected:
```
jblomer: Can all of this be private?

cvmfs/catalog_mgr_rw_impl.h
jblomer: Perhaps even better is `PropagateWeights` because the actual calculation is done in the recursion of `PartitionOptimize`.  So at this point, we know already that the direct children have their correct weights.  We only sum them up.  This information would also be important as a comment to this method.

cvmfs/cvmfs_server
```diff
@@ -1939,6 +1939,7 @@ CVMFS_HASH_ALGORITHM=$hash_algo
 CVMFS_AUTO_TAG=$autotagging
 CVMFS_GARBAGE_COLLECTION=$garbage_collectable
 CVMFS_AUTO_REPAIR_MOUNTPOINT=true
+CVMFS_AUTOCATALOG=false
```
jblomer: Probably better `CVMFS_AUTOCATALOGS` (plural) so that it is clear it's not just one.

cvmfs/swissknife_sync.cc
```diff
@@ -479,6 +479,7 @@ int swissknife::CommandSync::Main(const swissknife::ArgumentList &args) {
   params.stratum0 = *args.find('w')->second;
   params.manifest_path = *args.find('o')->second;
   params.spooler_definition = *args.find('r')->second;
+  params.is_balanced = *args.find('A')->second == "true";
```
jblomer: This should be a switch (a parameter without an argument).  So if you find the `-A` in the command line, it means that auto balancing is on.

cvmfs/swissknife_sync.h
```diff
@@ -28,7 +28,10 @@ struct SyncParameters {
     avg_file_chunk_size(8*1024*1024),
     max_file_chunk_size(16*1024*1024),
     manual_revision(0),
-    max_concurrent_write_jobs(0) {}
+    max_concurrent_write_jobs(0),
+    is_balanced(false),
+    max_weight(100000),
```
jblomer: Magic number, should become a constant class member.

cvmfs/swissknife_sync.h
```diff
@@ -28,7 +28,10 @@ struct SyncParameters {
...
+    min_weight(1000) {}
```
jblomer: Magic number, should become a constant class member.

cvmfs/sync_mediator.cc
```diff
@@ -226,6 +226,8 @@ manifest::Manifest *SyncMediator::Commit() {
     return NULL;
   }
 
+  if (catalog_manager_->IsBalanced())
```
jblomer: Should this be `if (!catalog_manager_->IsBalanced())`?  If so, how could this slip through the tests?

jblomer: What I mean is: if this is wrong, please add a test first that triggers the issue before fixing it.

test/unittests/t_catalog.cc
```diff
@@ -23,7 +23,7 @@ class T_Catalog : public ::testing::Test {
   virtual void SetUp() {
     catalog = NULL;
     nested = NULL;
-    sandbox = CreateTempDir("./cvmfs_ut_catalog");
+    sandbox = CreateTempDir("/tmp/cvmfs_ut_catalog");
```
jblomer: This reverts @reneme's changes to the sandbox of the unit tests, no?

test/src/596-catalogbalancer/main
```diff
@@ -0,0 +1,108 @@
...
+}
+
+cvmfs_run_test() {
+  apache_switch on
+  local repo="test.local"
+  local repo_dir="/cvmfs/$repo"
+
+  create_repo "$repo" $(whoami)
+
+    sudo bash -c "echo 'CVMFS_AUTOCATALOG=true
```
jblomer: Looks like wrong indentation.

test/unittests/t_catalog_mgr.cc
```diff
@@ -20,7 +20,6 @@ using namespace std;  // NOLINT
 namespace catalog {
 
 class T_CatalogManager : public ::testing::Test {
-
```
jblomer: All the changes in this file, do they belong to the new feature or are they fixes?

test/unittests/testutil.cc
```diff
@@ -168,31 +168,53 @@ shash::Any h(const std::string &hash, const shash::Suffix suffix) {
...
-DirectoryEntry DirectoryEntryTestFactory::RegularFile() {
+DirectoryEntry DirectoryEntryTestFactory::RegularFile(const string &name,
+                                                      unsigned size,
+                                                      shash::Any hash) {
   DirectoryEntry dirent;
   dirent.mode_ = 33188;
+  dirent.name_ = NameString(name);
+  dirent.checksum_ = hash;
+  dirent.size_ = size;
+  dirent.is_nested_catalog_mountpoint_ = false;
```
jblomer: this comes already from the default constructor.

test/unittests/testutil.cc
```diff
@@ -168,31 +168,53 @@ shash::Any h(const std::string &hash, const shash::Suffix suffix) {
...
-DirectoryEntry DirectoryEntryTestFactory::Symlink() {
+DirectoryEntry DirectoryEntryTestFactory::Symlink(const string &name,
+                                                  unsigned size,
+                                                  const string &symlink_path) {
   DirectoryEntry dirent;
   dirent.mode_ = 41471;
+  dirent.name_ = NameString(name);
+  dirent.size_ = size;
+  dirent.symlink_ = LinkString(symlink_path);
+  dirent.is_nested_catalog_mountpoint_ = false;
```
jblomer: See above

test/unittests/testutil.cc
```diff
@@ -168,31 +168,53 @@ shash::Any h(const std::string &hash, const shash::Suffix suffix) {
...
 }
 
 
-DirectoryEntry DirectoryEntryTestFactory::ChunkedFile() {
+DirectoryEntry DirectoryEntryTestFactory::ChunkedFile(shash::Any content_hash) {
   DirectoryEntry dirent;
   dirent.mode_ = 33188;
   dirent.is_chunked_file_ = true;
+  dirent.checksum_ = content_hash;
+  dirent.is_nested_catalog_mountpoint_ = false;
```
jblomer: See above

test/unittests/testutil.cc
```diff
@@ -209,6 +231,7 @@ catalog::DirectoryEntry catalog::DirectoryEntryTestFactory::Make(
   dirent.linkcount_  = metadata.linkcount;
   dirent.has_xattrs_ = metadata.has_xattrs;
   dirent.checksum_   = metadata.checksum;
+  dirent.is_nested_catalog_mountpoint_ = false;
```
jblomer: See above

jblomer:
Thanks!

The fixes mentioned in the comments plus the fixes to the unit tests (include move of `MockCatalog` to the `catalog` namespace) should be a separate PR which we will go through first.

jblomer:
For the `CatalogBalancer` please also add unit tests.

cvmfs/catalog_mgr_rw.h
Moliholy: No, I have to specify "which" virtual node I want, depending on the template. I could call it 'typename CatalogBalancer<CatalogMgrT>::VirtualNode', but virtual_node_t is shorter

cvmfs/catalog_mgr_rw.h
Moliholy: Probably my Java-brain betrayed me :)
I'll think about it

cvmfs/catalog_mgr_rw.h
jblomer: Ah, ok

cvmfs/catalog_mgr_rw.h
Moliholy: MakeEmptyDirectoryEntryBase could do the trick, maybe...

cvmfs/catalog_mgr_rw_impl.h
Moliholy: No, because this process is done in postorder, starting by the leaves, which are the lightest nodes.This way if a father gets overflowed it means that the sum of its children's weight is greater than the balance_weight_ (and not the max_weight_!), so each children must have a weight that is lesser than the balance_weight_, which means it cannot be overflowed.

cvmfs/catalog_mgr_rw_impl.h
Moliholy: It would be handled as a normal autogenerated catalog.
In other words, the user should _never_ create .cvmfscatalog files

cvmfs/catalog_mgr_rw_impl.h
jblomer: Ok, please add a comment to the code.

cvmfs/catalog_mgr_rw_impl.h
Moliholy: Hmmmm it is not really propagating weights, it is fixing them. Actually in the python code this function is called fix_weight().
This function is called in the father when one of its children has changed its weight. This phenomenon only happens when one of his children has become a new autogenerated nested catalog, and its weight is now 1 (which represents the sole DirectoryEntry of that directory).
However this is not propagated to the top or the bottom of the tree, but each VirtualNode (only directories, of course) is responsible for calling it when previous operations might have changed the weight of its children (and consequently its own weight).

cvmfs/catalog_mgr_rw_impl.h
jblomer: It would be surely misuse by the user.  I just want to make sure we know what happens if it is misused in this way.  Would the cvmfs_swissknife binary crash?  Or report an error?  (I don't remember anymore.)

cvmfs/catalog_mgr_rw_impl.h
jblomer: Ok, please add this information as a code comment.  I think `FixWeight` is a good name then.

cvmfs/catalog_mgr_rw.h
Moliholy: Yes, I put it protected in case one day somebody would like to create different CatalogBalancers using different techniques, but right now it could perfectly be private.

cvmfs/sync_mediator.cc
Moliholy: This is correct... even though maybe the method is not descriptive enough.
What I mean is that if the WritableCatalogManager IS balanceable, then proceed with the balancing operation.
Maybe the method should be renamed to CanBeBalanced()...

test/unittests/t_catalog.cc
Moliholy: Totally true... I wasn't aware of that, so I thought I modified that by mistake... I will fix it

test/src/596-catalogbalancer/main
Moliholy: And now it should be renamed to CVMFS_AUTOCATALOGS ;)

test/unittests/t_catalog_mgr.cc
Moliholy: They are fixes only. Should I submit them in a separated pull request as well?

jblomer: Yes, you can combine all the fixes into a single pull request.

cvmfs/catalog_mgr_rw_impl.h
Moliholy: I didn't know about that. Thanks for the tip

cvmfs/catalog_mgr_rw_impl.h
Moliholy: Sorry, I meant that the user should never create .cvmfsAUTOcatalog files, not .cvmfscatalog.
During the FSTraversal it would just add the file normally, but if one day that catalog gets underflowed it could be destroyed. If it gets overflowed the .cvmfsautocatalog it would remain there, but the catalog would be rebalanced because it is "supposed" to be an autogenerated catalog.

On the other hand, it is always possible that the user wants that catalog to be controlled by the system, so instead of deleting it he could just place that file and let the system decide in the publishing operation.

cvmfs/swissknife_sync.h
Moliholy: There are other cases like this one in that constructor. Should I change them, or maybe this is another pull request's stuff?

jblomer: Should be all changed.

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,8 +174,72 @@ class WritableCatalogManager : public SimpleCatalogManager {
...
+  struct VirtualNode {
+    vector<VirtualNode> children;
+    unsigned weight;
+    DirectoryEntry dirent;
+    string path;
+    bool is_new_nested_catalog;
+
+    void ExtractChildren(CatalogMgrT *catalog_mgr);
+    void CaltulateWeight();
```
reneme: Typo: `CalculateWeight()`

Moliholy: Wow, I didn't notice o.O
Anyway this method is now renamed to FixWeight :)

test/src/596-catalogbalancer/main
```diff
@@ -0,0 +1,108 @@
...
+
+  start_transaction "$repo"
+  add_files "$repo_dir" 25
+  add_files "$repo_dir/dir100" 100
+  add_files "$repo_dir/dir3" 3
+  add_files "$repo_dir/dir4" 4
+  add_catalog "$repo_dir/dir4"
+  publish_repo "$repo"
+
+  echo "Checking first iteration"
```
reneme: Please also use [`check_catalog_presence()`](https://github.com/cvmfs/cvmfs/blob/devel/test/test_functions#L768) to verify that the catalogs are actually created AND/OR not created.

cvmfs/catalog_mgr_rw_impl.h
Moliholy: Deleted those that were not necessary

test/src/596-catalogbalancer/main
Moliholy: Done

Moliholy:
The PR is ready for another review

cvmfs/catalog_mgr.h
```diff
@@ -139,7 +139,8 @@ template <class CatalogT>
 class AbstractCatalogManager : public SingleCopy {
  public:
   typedef std::vector<CatalogT*> CatalogList;
-
+  typedef CatalogT catalog_t;
```
Moliholy: It was necessary for this:
```
template <class CatalogMgrT>
class CatalogBalancer {
 public:
  typedef typename CatalogMgrT::catalog_t catalog_t;
```

jblomer:
Thanks!  Please remember to split up the PR.

test/unittests/t_catalog_mgr.cc
Moliholy: Never mind, I'm creating Unit tests for the balancer, so this file can be included

Moliholy:
I have just added tests for the catalog balancer

test/src/596-catalogbalancer/main
```diff
@@ -0,0 +1,156 @@
...
+    [ -f "$repo_dir/dir/dir30/.cvmfsautocatalog" ]            || return 10
+  if ! check_catalog_presence "/dir/dir30" "$repo"; then return 101; fi
+  ! [ -f "$repo_dir/dir60/.cvmfscatalog" ]                    || return 11
+  ! [ -f "$repo_dir/dir60/.cvmfsautocatalog" ]                || return 12
+  if check_catalog_presence "/dir60" "$repo"; then return 101; fi
+
+
+  start_transaction "$repo"
+  # now suppose the user creates a '.cvmfsautocatalog' file
+  touch "$repo_dir/dir4/.cvmfsautocatalog"
```
Moliholy: Here there is an extra test that checks what happens if the user creates a .cvmfsautocatalog file.
The systems recognises that catalog as one that can be handled by the balancer, and this one decides to eliminate the catalog

cvmfs/catalog_mgr_rw.h
```diff
@@ -58,16 +58,25 @@ class Statistics;
...
+  friend class CatalogBalancer<WritableCatalogManager>;
   WritableCatalogManager(const shash::Any  &base_hash,
                          const std::string &stratum0,
                          const std::string &dir_temp,
                          upload::Spooler   *spooler,
                          download::DownloadManager *download_manager,
                          uint64_t catalog_entry_warn_threshold,
-                         perf::Statistics *statistics);
+                         perf::Statistics *statistics,
+                         bool is_balanceable,
```
Moliholy: Renamed the parameter to make it more understandable

cvmfs/catalog_mgr_rw.h
```diff
@@ -58,16 +58,25 @@ class Statistics;
...
 namespace catalog {
+template <class CatalogMgrT>
+class CatalogBalancer;
+}
+
+namespace catalog {
 
 class WritableCatalogManager : public SimpleCatalogManager {
  public:
+  friend class CatalogBalancer<WritableCatalogManager>;
```
Moliholy: You need to grant the CatalogBalancer<WritableCatalogManager> permission to merge catalogs and read the thresholds

Moliholy:
Removed some unnecessary changes that will be added back once this PR is approved

cvmfs/catalog_balancer_impl.h
```diff
@@ -0,0 +1,189 @@
...
+    // we directly add a catalog in this node
+    // TODO(molina) apply heuristics here
+    if (heaviest_node != NULL &&
+        heaviest_node->weight >= catalog_mgr_->min_weight_) {
+      // the catalog now generated _cannot_ be overflowed because the tree is
+      // being traversed in postorder, handling the lightest nodes first
+      unsigned max_weight = heaviest_node->weight;
+      AddCatalogMarker(heaviest_node->path);
+      AddCatalog(heaviest_node);
+      virtual_node->weight -= (max_weight - 1);
```
Moliholy: This has changed recently. Even though it does not offer a significant performance improvement, it's always good to have the code optimised.

Moliholy:
I have removed the rest of the changes that do not necessarily belong to this PR, so it's ready for revision again

test/src/596-catalogbalancer/main
```diff
@@ -0,0 +1,156 @@
...
+
+  start_transaction "$repo"
+  add_files "$repo_dir/dir/dir30" 30
+  for i in $(ls "$repo_dir" | grep "file"); do
+    rm "$repo_dir/$i"
+  done
+  publish_repo "$repo"
+
+  echo "Checking second iteration"
+  ! [ -f "$repo_dir/dir3/.cvmfscatalog" ]                     || return 1
```
jblomer: Better: use different return codes (e.g. 10, 11, 12, ... for the second iteration, 20, 21, 22, ... for the third itereration and so on)

test/src/596-catalogbalancer/main
```diff
@@ -0,0 +1,156 @@
+cvmfs_test_name="Catalog balancer"
+cvmfs_test_autofs_on_startup=false
+
+add_files() {
+  local repo_dir=$1
```
jblomer: It's a bit confusing to call this `repo_dir`, too.  It's a directory within the repository, no?

test/src/596-catalogbalancer/main
Moliholy: In the second iteration, for instance, there are 10 error cases, which would increase the number of digits. We would have to add another one perhaps

jblomer: I think there is a limit at 255.  The point is, if possible the return code should uniquely identify the failure location.

test/src/596-catalogbalancer/main
Moliholy: It's the absolute path in the machine where the repository is actually located. It doesn't really make sense to be the root of the repository, because it's obviously the empty string. Anyway it could be renamed to... repo_abs_path? Suggestions are welcome :)

jblomer: Perhaps I don't understand this... further down you call for instance `add_files "$repo_dir/dir100"`.  So inside addfiles, `$repo_dir` is now a sub directory of the full path to the repository, no?

Moliholy: No, it's the absolute path in the system, not in the repository. That function simply adds files in the directory you specify, and is not related at all with any repository, and doesn't understand any logic about them

Moliholy:
When returning the code it only takes the last byte of the integer, which as you said means that codes are between 0 and 255. Basically NUMBER % 256 gives me the actual returned number. I will think another way to do it. Maybe simply 1, 2, 3, 4, 5, 6, 7, 8..., 20...?

jblomer:
Yep, sounds good.

test/unittests/testutil.h
```diff
@@ -523,14 +549,57 @@ class MockCatalogManager : public AbstractCatalogManager<MockCatalog> {
...
+  unsigned GetNumAutogeneratedCatalogs() { return autogenerated_catalogs_; }
+  unsigned GetNumAddedFiles() { return num_added_files_; }
+
+  MockCatalog *FindCatalog(const PathString &path) {
+    map<PathString, MockCatalog*>::iterator it;
+    string best_match = "";
+    unsigned best_match_length = 0;
+    for (it = catalog_map_.begin(); it != catalog_map_.end(); ++it) {
+      string key = it->first.ToString();
+      if (path.ToString().find(key) != string::npos) {
```
jblomer: Shouldn't the condition here be `path.ToString().find(key) == 0`?  In other words: `path` should start with `key`, no?  The current condition is true if `key` is found somewhere in `path`.

test/unittests/testutil.h
```diff
@@ -523,14 +549,57 @@ class MockCatalogManager : public AbstractCatalogManager<MockCatalog> {
...
+        if (length > best_match_length) {
+          best_match_length = length;
+          best_match = key;
+        }
+      }
+    }
+    return catalog_map_[PathString(best_match)];
+  }
+
+  void CreateNestedCatalog(const string &mountpoint) {
```
jblomer: Perhaps a comment would be good here: if this is re-used in a situation where a nested catalog is created manually, the counting gets wrong I guess.

test/unittests/testutil.h
```diff
@@ -523,14 +549,57 @@ class MockCatalogManager : public AbstractCatalogManager<MockCatalog> {
...
+    ++autogenerated_catalogs_;
+  }
+
+
+
+ private:
+  typedef struct {
+    shash::Algorithms GetHashAlgorithm() { return shash::kSha256; }
+  }Spooler;
```
jblomer: Missing blank (`} Spooler;`)

cvmfs/swissknife_sync.h
```diff
@@ -12,6 +12,13 @@
 #include "upload.h"
 
 struct SyncParameters {
+  static const unsigned DEFAULT_MAX_WEIGHT = 100000;
```
jblomer: Please change to match the naming convention for constants, e.g. `kDefaultMaxWeight`

test/unittests/testutil.h
Moliholy: You are totally right. I'll change the condition

test/unittests/testutil.h
Moliholy: The think is that for testing purposes the CreateNestedCatalog method _should_ never be called, because it involves operations that cannot be unit tested. But yes, I agree: a comment would be good

cvmfs/catalog_mgr_rw.h
```diff
@@ -160,6 +180,35 @@ class WritableCatalogManager : public SimpleCatalogManager {
    * Directories don't have extended attributes at this point.
    */
   XattrList empty_xattrs;
+
+  /**
+   * It indicates whether this catalog supports balancing operations
```
jblomer: catalog --> catalog manager

cvmfs/catalog_mgr_rw.cc
```diff
@@ -837,6 +845,36 @@ shash::Any WritableCatalogManager::SnapshotCatalog(WritableCatalog *catalog)
   return hash_catalog;
 }
 
+void WritableCatalogManager::DoBalance() {
+  ReadLock();
```
jblomer: This needs to be a `SyncLock()`

cvmfs/catalog_mgr_rw.cc
```diff
@@ -837,6 +845,36 @@ shash::Any WritableCatalogManager::SnapshotCatalog(WritableCatalog *catalog)
...
+  const CatalogList &catalog_list = GetCatalogs();
+  for (unsigned i = 0; i < catalog_list.size(); ++i) {
```
jblomer: This traverses the catalog tree from top to bottom, right?  Isn't it supposed to traverse the catalog list from bottom to top?

cvmfs/catalog_mgr_impl.h
```diff
@@ -7,17 +7,13 @@
 
 #define __STDC_FORMAT_MACROS
 
-#include "cvmfs_config.h"
```
jblomer: `cvmfs_config.h` should remain.  This file contains constants created by the build system.

cvmfs/catalog_mgr.h
jblomer: Please don't forget to remove this.

cvmfs/catalog.h
```diff
@@ -190,6 +190,12 @@ class Catalog : public SingleCopy {
     return inode_range_.IsInitialized() && initialized_;
   }
   inline bool IsRoot() const { return is_root_; }
+  bool IsAutogenerated() const {
+    DirectoryEntry dirent;
+    return IsInitialized() &&
```
jblomer: Would it be better to assert `IsInitialized()`?

cvmfs/catalog_balancer.h
```diff
@@ -0,0 +1,139 @@
...
+   * For a WritableCatalogManager to be balanced it requires the is_balanced_
+   * flag to be set to true.
+   *
+   * @param catalog the catalog to be balanced, or NULL to balance all catalogs
+   * in the WritableCatalogManager
+   */
+  void Balance(catalog_t *catalog);
+
+ private:
+     /**
```
jblomer: The indentation is a bit flawed.

cvmfs/catalog_balancer.h
```diff
@@ -0,0 +1,139 @@
...
+      if (!IsCatalog() && IsDirectory())
+        ExtractChildren(catalog_mgr);
+    }
+    bool IsDirectory() { return dirent.IsDirectory(); }
+    bool IsCatalog() { return is_new_nested_catalog ||
+        dirent.IsNestedCatalogMountpoint(); }
+  };
+  typedef typename CatalogBalancer<CatalogMgrT>::VirtualNode virtual_node_t;
+
+ private:
```
jblomer: Use `private:` only once

cvmfs/catalog_balancer.h
```diff
@@ -0,0 +1,139 @@
...
+  void PartitionOptimally(VirtualNode *virtual_node);
+  void AddCatalogMarker(std::string path);
+  DirectoryEntryBase MakeEmptyDirectoryEntryBase(std::string name,
+                                                 uid_t uid,
+                                                 gid_t gid);
+  static VirtualNode *MaxChild(VirtualNode *virtual_node);
+  void AddCatalog(VirtualNode *child_node);
+
+ private:
```
jblomer: Use `private:` only once

cvmfs/catalog_balancer_impl.h
```diff
@@ -0,0 +1,189 @@
...
+
+namespace catalog {
+
+template <class CatalogMgrT>
+DirectoryEntryBase
+CatalogBalancer<CatalogMgrT>::MakeEmptyDirectoryEntryBase(
+    string name, uid_t uid, gid_t gid) {
+  shash::Algorithms algorithm = catalog_mgr_->spooler_->GetHashAlgorithm();
+  shash::Any file_hash(algorithm);
+  shash::HashString("", &file_hash);
```
jblomer: That's the wrong hash, cvmfs hashes over the compressed content.

cvmfs/catalog_balancer_impl.h
```diff
@@ -0,0 +1,189 @@
...
+
+  DirectoryEntryBase deb;
+  deb.name_ = NameString(name);
+  deb.mode_ = S_IFREG | S_IRWXU;
```
jblomer: This file should not be executable.

cvmfs/catalog_balancer_impl.h
```diff
@@ -0,0 +1,189 @@
...
+    // catalog's underflow provokes an overflow in the father
+    reverse(catalogs.begin(), catalogs.end());
+    for (unsigned i = 0; i < catalogs.size(); ++i)
+      Balance(catalogs[i]);
+    return;
+  }
+  string catalog_path = catalog->path().ToString();
+  virtual_node_t root_node(catalog_path, catalog_mgr_);
+  root_node.ExtractChildren(catalog_mgr_);
+  // we have just recursively loaded the entire virtual tree!
```
jblomer: This I don't understand.  In order to load the entire virtual tree, `ExtractChildren()` would need to be a recursive function, no?

cvmfs/catalog_balancer_impl.h
```diff
@@ -0,0 +1,189 @@
...
+        &virtual_node->children[i];
+    if (virtual_child->IsDirectory() && !virtual_child->IsCatalog())
+      PartitionOptimally(virtual_child);
+  }
+  virtual_node->FixWeight();
+  while (virtual_node->weight > catalog_mgr_->balance_weight_) {
+    virtual_node_t *heaviest_node =
+        MaxChild(virtual_node);
+    // we directly add a catalog in this node
+    // TODO(molina) apply heuristics here
```
jblomer: heuristics for what?

cvmfs/catalog_balancer_impl.h
```diff
@@ -0,0 +1,189 @@
...
+    if (heaviest_node != NULL &&
+        heaviest_node->weight >= catalog_mgr_->min_weight_) {
+      // the catalog now generated _cannot_ be overflowed because the tree is
+      // being traversed in postorder, handling the lightest nodes first
+      unsigned max_weight = heaviest_node->weight;
+      AddCatalogMarker(heaviest_node->path);
+      AddCatalog(heaviest_node);
+      virtual_node->weight -= (max_weight - 1);
+    } else {
+      // there is no possibility for this directory to be a catalog
```
jblomer: It would make sense to print a warning at this point.  This could be a directory with very many files, for instance, and the user might be interested to know about it.

cvmfs/catalog_balancer_impl.h
```diff
@@ -0,0 +1,189 @@
...
+        max_child = child;
+      }
+    }
+  }
+  return max_child;
+}
+
+template <class CatalogMgrT>
+void CatalogBalancer<CatalogMgrT>::AddCatalog(virtual_node_t *child_node) {
+  if (child_node != NULL) {
```
jblomer: Can this ever happen?  If not, it's better to do `assert(child_node != NULL)`

jblomer:
Quite well progressed!

The integration test case should cover multiple levels (say three) of nested catalogs, too, where overflows and underflows happen in all three levels.

cvmfs/catalog_balancer_impl.h
Moliholy: Totally true. It was changed in the CatalogBalancer, but not in the WritableCatalogManager

cvmfs/catalog_balancer_impl.h
Moliholy: @reneme and I had some ideas to apply here that would require a bit of whiteboard to be explained, but basically is the following:
1) Make the root catalog smaller than the rest, if necessary. We would have to establish a "heuristic" to set the final maximum weight of the root catalog.
2) If a folder is a nested catalog, it's likely that a set of siblings are as well. Say, if /releases/root/5.4.3 is a catalog, then /releases/root/5.x.x are likely to be a catalog. More concretely, we thought that:
```
weight(FOLDER) >= weight(MAX_FOLDER) * c   then FOLDER should be a nested catalog as well,      with 0 < c < 1 for all FOLDER sibling of MAX_FOLDER
```

cvmfs/catalog_balancer_impl.h
Moliholy: Good suggestion. Should that warning be printed in the standard output? Or maybe it would be better in the logs?

cvmfs/catalog_balancer_impl.h
Moliholy: I "should" never be NULL. I'll change that

cvmfs/catalog.h
Moliholy: Hmmm it could be, but I thought it would make sense to leave it this way. A catalog that is not initialised doesn't know if it is autogenerated or not, so in principle we suppose it isn't

cvmfs/catalog_balancer_impl.h
Moliholy: But what is the compressed content of an empty file? Isn't it the empty string?

Moliholy:
PR updated, including three levels of catalogs in the integration test which tests what happens when a catalog is removed due to underflow and provokes an overflow in the father

jblomer:
Thanks!  Can you check once more, I think there are still one or two unhandled comments.

cvmfs/catalog_balancer.h
```diff
@@ -0,0 +1,146 @@
...
+ * catalog gets split in smaller catalogs. One of the new generated catalogs
+ * will be mounted in the original mount point.
+ *
+ * c) The number of entries of the catalog is lesser than min_weight_: the
+ * catalog gets merged with its father (except the root catalog, obviously).
+ */
+template <class CatalogMgrT>
+class CatalogBalancer {
+ public:
+  typedef typename CatalogMgrT::catalog_t catalog_t;
```
Moliholy: I finally used the CatalogMgrT::catalog_t for this, so it shouldn't be removed.
This is simply to make the code clearer, instead of putting so many typenames

cvmfs/catalog_mgr.h
Moliholy: Addressed here https://github.com/cvmfs/cvmfs/pull/1091/files#diff-fa6ec71328a1af0bd1de028407c3a46eR48

Moliholy:
Sorry, I forgot to address one comment. Now I think they are all addressed. I will prepare a separate PR for the MockCatalog changes

jblomer:
Please also have a look in the doc-cvmfs repository.  The table with the server parameters needs to be updated.  We'd also need a new paragraph in the section on Stratum 0s explaining the auto balancing.

Moliholy:
Done in this PR https://github.com/cvmfs/doc-cvmfs/pull/30

PR #1093, opened 2015-09-16
FIX: remove gpg key check (doesn't work with non-root)

reneme:
[eom]

### DIFF
--- a/ci/docker/common.sh
+++ b/ci/docker/common.sh
@@ -7,79 +7,8 @@ die() {
 }
 
 check_yum_environment() {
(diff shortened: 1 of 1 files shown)


jblomer:
Aren't we root in the docker container?

reneme:
Nope. I've [set the user to sftnight](https://github.com/cvmfs/cvmfs/blob/devel/ci/docker/slc6_x86_64/Dockerfile#L27). Building doesn't require root and I figured we should be warned if it accidentally does at some point.

However, the title of this pull request is misleading. The package signatures are checked by yum before installing, just like in any default installation of RedHat. This removes a sanity check of the host machine that wasn't necessary anymore.

PR #1105, opened 2015-09-18
Update: cloc.pl 1.62 -> 1.64

reneme:


jblomer:
Is there something important in this new release?

reneme:
I was hoping it would fix a bug with `--lang-no-ext=` which I didn't bring to work. It didn't fix the problem, but since I already had the new version I just updated it...

PR #1106, opened 2015-09-18
Enable "make check" command to build and run the unit tests

Moliholy:
This Pull Request enables the "make check" feature, which compiles and runs the unit tests (only the fast ones).

The example sequence that compiles and runs _ONLY_ the unit tests would be:
```
mkdir build
cd build
cmake .. && make check
```

If we want to compile only the normal cvmfs sources it's enough with:
```
cmake .. && make
```

If we want to compile the sources and the unit tests, but not to run them:
```
cmake -DBUILD_UNITTESTS=yes .. && make
```

If we want to compile the sources and the unit tests, and run them:
```
cmake .. && make all check
```

If we want to execute the unit tests once they have already been compiled to avoid recompilation:
```
make test ARGS="-V"  # to make it verbose
OR
ctest -V
```
However, if no tests are compiled an error will be raised.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -482,12 +482,11 @@ if (INSTALL_MOUNT_SCRIPTS)
 endif (INSTALL_MOUNT_SCRIPTS)
 
 #
-# compile the unit tests
+# compile the unit tests, if necessary
 #
-if (BUILD_UNITTESTS OR BUILD_UNITTESTS_DEBUG)
-  enable_testing ()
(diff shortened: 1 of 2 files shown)


Moliholy:
Note that it's also possible at any point to compile and run the unit tests with
```
make check
```
independently on the cmake command that has been used.

Moliholy:
Also notice that nothing has changed for the _cvmfs_unittests_debug_ target.

PR #1107, opened 2015-09-21
Secure CVMFS

bbockelm:
This branch is a bit of a prototype that I'd like to share - see CVM-904.  Not really meant for merging, just discussion right now.

This provides HTTPS-based security to protect the contents of a repo.


cvmfs/catalog_mgr.h
```diff
@@ -183,6 +183,7 @@ class AbstractCatalogManager : public SingleCopy {
   uint64_t GetRevision() const;
   bool GetVolatileFlag() const;
   uint64_t GetTTL() const;
+  bool GetVOMSAuthz(std::string&) const;
```
jblomer: Would different nested catalogs contain different VOMS authentication caveats?  Or is it meant to be fixed per repository?

cvmfs/download.h
```diff
@@ -138,6 +145,11 @@ struct JobInfo {
     probe_hosts = false;
     head_request = false;
     follow_redirects = false;
+    pid = -1;
```
jblomer: I wonder if this should be pushed down to the download manager.  Another option would be to attach a string with the credentials to the download job.

jblomer: Ah, curl wants to see a file.  Hm, might be worth patching curl to avoid copying credential files around.

jblomer:
We could have an extra process for fetching VOMS credentials.  This process would be connected by a socket to the fuse module.  If serialized, this credential fetcher process can then use `seteuid()`, which would make it portable to Mac.  Might also add security, if VOMS parsing dies, the fuse module survives.

cvmfs/voms_authz/voms_authz.cc
```diff
@@ -0,0 +1,406 @@
...
+
+    typedef std::map<KeyType, KeyType> PidToSid;
+    PidToSid m_pid_map;
+
+    // Lookup pid's sid and birthdy.
+    // Returns false on failure and sets errno.
+    bool lookup(pid_t pid, KeyType &mykey)
+    {
+        char pidpath[PATH_MAX];
+        if (snprintf(pidpath, PATH_MAX, "/proc/%d/stat", pid) >= PATH_MAX)
```
jblomer: Would it be enough to remember the creation time of /proc/%d?  That could be done in a `stat()` call.

cvmfs/download.h
jblomer: Another option is `CURLOPT_SSL_CTX_DATA` with the drawback that this binds curl to openssl.

cvmfs/voms_authz/voms_authz.cc
bbockelm: I'm still a bit worried about PID reuse here.  ctime has pretty poor granularity - 1s (not that the 1 jiffy in the current is perfect).

Further, I've had folks report to me that ctime doesn't get updated appropriately when the system is under heavy load.  I don't know how much I believe those reports though.

cvmfs/download.h
bbockelm: I'd rather not patch curl, since it prevents folks from using the system version.

*However*, in follow-up work, I did find that curl has a nasty habit of caching HTTP connections and SSL sessions (even if the file changes!).  One performance improvement we can tackle later is see if we can manage session reuse more explicitly to avoid disabling caches.

cvmfs/catalog_mgr.h
bbockelm: I would like this to be fixed in the "tech preview" and gather input from VOs whether they would want different authentication for nested catalogs.

Seems limited value since it's a bit impractical to make the symmetric narrowing of authorization in the HTTPS server.

bbockelm:
Yeah - I also considered a separate credential-fetching process.  It seemed like quite a bit of tricky state (passing FDs and VOMS data, for example) for somewhat marginal gain (Mac support, process separation).

It's something I'd be more amenable to past the 'tech preview' stage.

cvmfs/download.h
jblomer: Ah, that's indeed nasty.  Of course at the moment we do everything we can to enable session reuse.  That requires a careful look into the download manager.  For a start, every curl handle using HTTPS can be encapsulated in a different share handle.  We should use a memory pool of share handles to memory thrashing.  At a later point, we can be smarter and combine easy handles with the same certificate in the same share handle.

cvmfs/voms_authz/voms_authz.cc
jblomer: Hm, perhaps ctime + inode.  I wonder if we can data about the probability of pid reuse.  Or if we can stress a system ourselves and observe what happens.

cvmfs/download.h
bbockelm: Oh - to be clear, for now, I only disable caching specifically for handles that require a client certificate.

"Normal" handles - even non-client-certificate HTTPS - keep HTTP1.1 keepalive and SSL session reuse enabled.

I figure this way we don't adversely affect the 'normal' use case and we have minimal functionality for the secure use case.  I see creating a separate pool of curl handles as a follow-up branch (this branch already touches a dangerously large number of interfaces!).

jblomer:
The separate process should not be complicated.  There is already a blueprint in `monitor.cc` for the watchdog.  I'd prefer to read the credentials only once and from that point on pass and cache memory blocks in cvmfs.  Copying files around largely increases the likelihood of failures.

For curl, `CURLOPT_SSL_CTX_DATA` should give us handle to inject the client certificate directly with `SSL_CTX_use_certificate`.

cvmfs/voms_authz/voms_authz.cc
bbockelm: I don't know much about how proc handles inodes, that's a good idea to check.

However, since proc is not a "real filesystem", I'm not sure how much of a performance hit/hindrance we're looking at by reading out this file.

I know this is about equivalent to the amount of work done when invoking getrusage (glibc's internal implementation opens and parses this file), and we invoke that quite a bit within CMSSW...

It'll be worth measuring.

bbockelm:
I do like the idea of buffering the credentials in memory.  I think we can do that separately from the question about CURLOPT_SSL_CTX_DATA (i.e., we can fallback to files in case we're using the system libcurl).

bbockelm:
Alright @jblomer - by default, credentials are now managed and memory and a separate process manages opening the credential files.  /proc is still used, but the separate process eliminates Linux-specific syscalls; support for Mac should have one less barrier.

jblomer:
I merged the current state into the topic branch CVM-904.  Some modifications where necessary to make everything (in particular unit tests) work with the latest devel branch.  Exception: t_uploaders.cc  For the template foo in there, it's faster if we wait for @reneme to come back for the necessary modifications.

If you have no objections, I'd suggest to close this PR and submit further PRs to CVM-904.

bbockelm:
Yup, that looks good to me.

Is there a Jenkins job for this branch?  I'd be curious if anything else is broken!

jblomer:
In principle yes, at least for the unit tests.  We are working on the build machine end concurrently, so right now it's kind of broken.  Probably back up tomorrow.

PR #1110, opened 2015-09-25
Add support for external data URLs.

bbockelm:
This option allows CVMFS to fetch data files from a location separate
from the metadata files (such as a HTTP data federation).

Does not work with chunked files -- support is planned after the
patches for uncompressed files land.

TODO:
  - Server-side support to set this property has not been added.
  - This is done on a *per-catalog* basis; I'd like to change this
    to be hierarchical (i.e., catalog inherits this setting from
    its parent).

See CVM-907 for more information.

### DIFF
--- a/cvmfs/catalog.cc
+++ b/cvmfs/catalog.cc
@@ -63,6 +63,9 @@ Catalog::Catalog(const PathString &path,
   int retval = pthread_mutex_init(lock_, NULL);
   assert(retval == 0);
 
+  atomic_init32(&external_data_status_);
+  ExternalDataStatus status = Unknown;
+  atomic_write32(&external_data_status_, status);
   database_ = NULL;
   uid_map_ = NULL;
   gid_map_ = NULL;
@@ -228,6 +231,7 @@ bool Catalog::LookupEntry(const shash::Md5 &md5path, const bool expand_symlink,
   bool found = sql_lookup_md5path_->FetchRow();
   if (found && (dirent != NULL)) {
     *dirent = sql_lookup_md5path_->GetDirent(this, expand_symlink);
+    dirent->set_external_data(GetExternalDataLocked());
     FixTransitionPoint(md5path, dirent);
   }
   sql_lookup_md5path_->Reset();
@@ -405,6 +409,39 @@ void Catalog::DropDatabaseFileOwnership() {
   if (NULL != database_) {
     database_->DropFileOwnership();
   }
+}
+
+
+bool Catalog::GetExternalDataLocked() const {
+  bool result;
+  ExternalDataStatus status = static_cast<ExternalDataStatus>(atomic_read32(&external_data_status_));
+  if (status == Present) {
+    result = true;
+  } else if (status == None) {
+    result = false;
+  } else {
+    bool attr = database().GetPropertyDefault<bool>("external_data", false);
+    atomic_write32(&external_data_status_, attr ? Present : None);
+    result = attr;
+  } 
+  return result;
+}
+
+bool Catalog::GetExternalData() const {
+  bool result;
+  ExternalDataStatus status = static_cast<ExternalDataStatus>(atomic_read32(&external_data_status_));
+  if (status == Present) {
+    result = true;
+  } else if (status == None) {
+    result = false;
+  } else {
+    pthread_mutex_lock(lock_);
+    bool attr = database().GetPropertyDefault<bool>("external_data", false);
+    pthread_mutex_unlock(lock_);
+    atomic_write32(&external_data_status_, attr ? Present : None);
+    result = attr;
+  } 
+  return result;
 }
 
 
--- a/cvmfs/catalog.cc
+++ b/cvmfs/catalog.cc
@@ -231,7 +231,7 @@ bool Catalog::LookupEntry(const shash::Md5 &md5path, const bool expand_symlink,
   bool found = sql_lookup_md5path_->FetchRow();
   if (found && (dirent != NULL)) {
     *dirent = sql_lookup_md5path_->GetDirent(this, expand_symlink);
-    dirent->set_external_data(GetExternalDataLocked());
+    dirent->set_external_data(GetExternalDataLocked() == Present);
     FixTransitionPoint(md5path, dirent);
   }
   sql_lookup_md5path_->Reset();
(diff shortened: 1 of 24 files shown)


jblomer:
Many thanks!

Please submit to branch CVM-907.  There are two supporting commits in this branch.  Commit 9d07918792426702a212de560879b561505b2b94 allows for multiple download managers to exist concurrently (more to it later).  Commit f4171234b13ea448589820c1524ee52eae141bb4 reserves bit 7 in the flags of directory entries for external files and it adds statistics counters for external files (number, total size).  The second commit bumps the catalog schema revision to 3, a change that is transparent to clients.  The server tools automatically migrate catalogs from previous schemas.  From the data structure point of view, this allows now for mixed catalogs with external and native files.

We need to think about what to do with `cvmfs_swissknife pull` (replication).  I'd suggest to rename the `SqlAllChunks` query to `SqlAllNativeChunks` and to exclude the external files (and chunks) from it.  I'll submit something to the CVM-907 branch.

Regarding the external data status, I don't think we need to add propagation logic to the client.  The server tools can simply add the flag to all nested catalogs.

As for the downloading, let's use two download managers, one for the native cvmfs CDN (stratum 1 + proxies) and one for the external location.  Even though we'll start with catalogs and data on the same server, if it works well we'll soon want to use different network parameters (timeouts etc.).  And there is already too much if-then-else logic in the download manager.

This PR should have a few additions:
  - An integration test that creates a repository with external files and checks if the client opens it as such (magic extended attributes set etc.).  The integration tests are shell scripts in the test/src directory.  As of test number 500 the server side tests start.  You can have a look in the first server tests and also in the 585-xattr server test for blueprints.
  - `cvmfs_swissknife check` should detect if entries are marked external but the `external_data` property is not set
  - `cvmfs_server info` should show if a repository is flagged as external

Please have a look at `python cpplint.py cvmfs/*`.  I think there are a few style violations.  Otherwise I'll add some minor comments directly in the code.

cvmfs/catalog.cc
```diff
@@ -63,6 +63,9 @@ Catalog::Catalog(const PathString &path,
   int retval = pthread_mutex_init(lock_, NULL);
   assert(retval == 0);
 
+  atomic_init32(&external_data_status_);
```
jblomer: This is always set in a locked context, isn't it?  So `external_data_status_` can just be of type `ExternalDataStatus`

cvmfs/catalog.cc
```diff
@@ -408,6 +414,44 @@ void Catalog::DropDatabaseFileOwnership() {
 }
 
 
+Catalog::ExternalDataStatus Catalog::GetExternalDataLocked() const {
```
jblomer: In the rest of the code, we use `Locked` and `Unlocked` suffixes in reverse meaning, where `Unlocked` means "this routine doesn't acquire a lock".

cvmfs/catalog.cc
```diff
@@ -408,6 +414,44 @@ void Catalog::DropDatabaseFileOwnership() {
...
+    }
+    atomic_write32(&external_data_status_, status);
+  } 
+  return status;
+}
+
+Catalog::ExternalDataStatus Catalog::GetExternalDataUnlocked() const {
+  ExternalDataStatus status = static_cast<ExternalDataStatus>(atomic_read32(&external_data_status_));
+  if (status == Unknown) {
+    pthread_mutex_lock(lock_);
```
jblomer: Currently we sometimes have `pthread_mutex_lock` directly in the code, sometimes we use `MutexLockGuard` from util_concurrency.h.  New code should only use the `MutexLockGuard`.

cvmfs/catalog.h
```diff
@@ -247,6 +248,15 @@ class Catalog : public SingleCopy {
   void FixTransitionPoint(const shash::Md5 &md5path,
                           DirectoryEntry *dirent) const;
 
+  enum ExternalDataStatus {
```
jblomer: Please add some documentation on the meaning of the constants.  The constant names should start with a `k`, like `kUnknown`.

cvmfs/util.h
```diff
@@ -38,6 +38,12 @@ const int kDefaultFileMode = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH;
 const int kDefaultDirMode = S_IXUSR | S_IWUSR | S_IRUSR |
                             S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH;
 
+enum CatalogProperty {
```
jblomer: This is very neat.  The tri-state property type should be part of sql.(cc|h), I think.  @reneme, thoughts?

cvmfs/sync_item.h
```diff
@@ -59,6 +59,7 @@ class SyncItem {
   inline bool IsSymlink()       const { return IsType(kItemSymlink);          }
   inline bool WasSymlink()      const { return WasType(kItemSymlink);         }
   inline bool IsNew()           const { return WasType(kItemNew);             }
+  CatalogProperty IsExternalData() const { return external_data_;             }
```
jblomer: Perhaps better `HasExternalData()`

cvmfs/sync_item.cc
```diff
@@ -10,11 +10,77 @@
 
 using namespace std;  // NOLINT
 
+
+static bool ProcessCatalogProperties(const std::string &scratch_path, bool &external_data) {
+  char buf[128];
+  external_data = false;
+  ssize_t retval;
+  if (-1 != (retval = lgetxattr(scratch_path.c_str(), "user.external_data", buf, 127))) {
```
jblomer: `lgetxattr` is encapsulated in platform.h as `platform_lgetxattr`.  In platform.h (platform_linux.h, platform_osx.h) there are a number of routines that work differently on Mac and Linux.

bbockelm:
- With respect to rebasing the PR on top of CVM-907: I'd be happy to do that, but it'll require me to close this one.  Can you copy/paste your above comment when I do that?
- For cpplint: It would be quite useful if you enabled Travis-CI integration on this repo to enforce items like this.  Travis-CI can then check the simple stuff (build failures, make check, cpplint) so the code reviewer doesn't have to bother with it.

I don't see anything in the above comment disagreeable.  I'll work to rebase and start on the suggested items.

cvmfs/catalog.cc
bbockelm: It's always set in a locked context, but often read from an unlocked context.  You need to keep it an atomic to prevent partial reads.

bbockelm:
Closing in favor of #1115, which is based off CVM-907.

PR #1111, opened 2015-09-25
RFC: Improved Database Reading by using Exclusive Locking

reneme:
Competing against José's Java implementation I found this little tweak: If an SQLite database is opened read-only, one can set `PRAGMA locking_mode=EXCLUSIVE`. However, I am not entirely sure about the implications - therefore RFC. :-)

Regarding the [SQLite documentation](http://sqlite.org/pragma.html#pragma_locking_mode) this implies that no two processes can access the SQLite file at the same time anymore. In `NORMAL` mode, SQLite acquires and releases a file-lock for every read operation. Now: Do we rely on accessing a single SQLite database file read-only in multiple processes concurrently?

Mini-Benchmark:
* traversing all catalog entries in **boss.cern.ch** (~1.6 million entries)
* only considering the actual traversal (no catalog loading, initialisation, ...)

With `locking_mode=NORMAL` this takes (3.09s, 3.11s, 3.02s, 3.15s): **3.09s**
With `locking_mode=EXCLUSIVE` this takes (2.49s, 2.65s, 2.50s, 2.49s): **2.53s**

### DIFF
--- a/cvmfs/sql_impl.h
+++ b/cvmfs/sql_impl.h
@@ -168,7 +168,8 @@ bool Database<DerivedT>::Configure() {
   // unexpected open read-write file descriptors in the cache directory like
   // etilqs_<number>.
   if (!read_write_) {
-    return Sql(sqlite_db() , "PRAGMA temp_store=2;").Execute();
+    return Sql(sqlite_db() , "PRAGMA temp_store=2;").Execute() &&
+           Sql(sqlite_db() , "PRAGMA locking_mode=EXCLUSIVE;").Execute();
   }
   return true;
 }


Moliholy:
I can confirm these numbers running in my machine for both Java and C++

Moliholy:
Also take a look at this parameter: http://sqlite.org/pragma.html#pragma_synchronous
```
On the other hand, some operations are as much as 50 or more times faster with synchronous OFF
```
Sounds interesting...

jblomer:
Do we use r/o databases outside the client?  The latest devel branch comes with a custom SQlite virtual file system implementation where all locking calls are noops.

jblomer:
synchronous=OFF helps for writing only, afaik.  For the server we should activate it.  We anyway have the cvmfs transaction around everything so we don't need the sqlite transactional safety.  As well, we can go for exclusive locking.  We never have shared access to an sqlite db.

Moliholy:
I have a doubt concerning database improvements. Please take a look at this: https://github.com/reneme/cvmfs/blob/rfc-improved_db_reading/cvmfs/catalog.cc#L437

I guess that's not a pre-compiled SQL statement (and if I am wrong please tell me). The same way it happens in other places. Would it make sense to substitute these non-compiled SQL statements for a protected static member that is compiled only once and used where needed? Or maybe it wouldn't make a difference?
That's what I am using in the java implementation at least (https://github.com/Moliholy/cvmfs-java/blob/devel/src/main/java/com/molina/cvmfs/catalog/Catalog.java#L66), with a significant improvement in performance.

PR #1113, opened 2015-09-28
Support for "grafted" files

bbockelm:
This PR adds support for 'grafted' files - files that are added to the repository without being physically present on the scratch filesystem.

For a given file $FOO in a directory, we denote a grafted file by the presence of .cvmfsgraft-$FOO in the same directory.  The .cvmfsgraft- file is line-oriented with key-value pairs.  Currently, two keys are recognized:

size=<integer size>
checksum=<checksum value>

The checksum must be a valid CVMFS checksum, following CVMFS checksumming conventions (such as having suffixes and being applied to the compressed contents).

If grafting is used, it is the responsibility of the repository maintainer to stage the files into the repository.

See CVM-908 for more information.

### DIFF
--- a/cvmfs/sync_item.cc
+++ b/cvmfs/sync_item.cc
@@ -16,6 +16,9 @@ namespace publish {
 SyncItem::SyncItem() :
   union_engine_(NULL),
   whiteout_(false),
+  valid_graft_(false),
+  graft_marker_present_(false),
+  size_(-1),
   scratch_type_(static_cast<SyncItemType>(0)),
   rdonly_type_(static_cast<SyncItemType>(0))
 {
@@ -27,12 +30,16 @@ SyncItem::SyncItem(const string       &relative_parent_path,
                    const SyncItemType  entry_type) :
   union_engine_(union_engine),
   whiteout_(false),
+  valid_graft_(false),
+  graft_marker_present_(false),
   relative_parent_path_(relative_parent_path),
   filename_(filename),
+  size_(-1),
   scratch_type_(entry_type),
   rdonly_type_(kItemUnknown)
 {
   content_hash_.algorithm = shash::kAny;
+  CheckGraft();
 }
 
 
@@ -150,7 +157,7 @@ catalog::DirectoryEntryBase SyncItem::CreateBasicCatalogDirent() const {
   dirent.mode_           = this->GetUnionStat().st_mode;
   dirent.uid_            = this->GetUnionStat().st_uid;
   dirent.gid_            = this->GetUnionStat().st_gid;
-  dirent.size_           = this->GetUnionStat().st_size;
+  dirent.size_           = size_ > -1 ? size_ : this->GetUnionStat().st_size;
   dirent.mtime_          = this->GetUnionStat().st_mtime;
   dirent.checksum_       = this->GetContentHash();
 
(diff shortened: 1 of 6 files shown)


bbockelm:
@jblomer - can I get you to review this?

cvmfs/sync_item.h
```diff
@@ -153,8 +162,11 @@ class SyncItem {
   mutable EntryStat scratch_stat_;
 
   bool whiteout_;
+  bool valid_graft_;
+  bool graft_marker_present_;
   std::string relative_parent_path_;
   std::string filename_;
+  ssize_t size_;
```
jblomer: perhaps better `grafted_size_` to clarify this is only used in the context of grafted files.

cvmfs/sync_item.cc
```diff
@@ -185,4 +192,81 @@ std::string SyncItem::GetScratchPath() const {
...
+  if (fp == NULL) {
+    if (errno != ENOENT) {
+      LogCvmfs(kLogFsTraversal, kLogWarning, "Unable to open graft file (%s): %s (errno=%d)",
+               graftfile.c_str(), strerror(errno), errno);
+    }
+    return;
+  }
+  graft_marker_present_ = true;
+  size_t len = 0;
+  ssize_t read;
```
jblomer: We have a few utility functions in util.cc that should be more handy here.  In particular, `GetLineFile()` and `SplitString()` should make this paragraph shorter.

cvmfs/sync_item.cc
```diff
@@ -185,4 +192,81 @@ std::string SyncItem::GetScratchPath() const {
...
+      }
+    }
+    if (line[read-1] == '\n') {line[read-1] = '\0';}
+    char *value;
+    if (line == (value = strcasestr(line, "size="))) {
+      value += 5;
+      size_t input_len = strlen(value);
+      char *endptr=NULL;
+      errno = 0;
+      long long mysize = strtoll(value, &endptr, 10);
```
jblomer: There is also `String2Uint64()` in util.cc

test/src/597-graft/main
```diff
@@ -0,0 +1,48 @@
+
```
jblomer: A few more things to test would be
 - a grafted file in a sub directory / nested catalog
 - graft markers for symlinks, directory entries (should perhaps result in a warning)
 - remove a grafted file

jblomer:
Except for minor things and a few style violations looks fine to me.  Any other comments, @reneme?

A principal issue with this approach could be that it is easy to "migrate" a grafted file to a native file, by touching it in succeeding publish operations.  How should we deal with this?

test/src/597-graft/main
bbockelm: What's the best way to test for a warning?

bbockelm:
Ok, I think I covered the cases requested in the unit tests!

test/src/597-graft/main
```diff
@@ -0,0 +1,81 @@
...
+cvmfs_test_name="Grafting"
+cvmfs_test_autofs_on_startup=false
+
+verify_grafting() {
+  testfname=$1
+  local file1=$(cat /cvmfs/$CVMFS_TEST_REPO/file1)
+  local file2=$(cat $testfname)
+  if [ "x$file1 " == "x" ]; then
```
reneme: is the space behind `$file1` intentional? This should never be `true`, no?

test/src/597-graft/main
reneme: usually I send the output of the publishing operation into a text file and then `grep` for the warning from there...

cvmfs/sync_item.cc
```diff
@@ -185,4 +195,80 @@ std::string SyncItem::GetScratchPath() const {
   return union_engine_->scratch_path() + relative_path;
 }
 
+
+std::string SyncItem::GetGraftPath() const {
```
reneme: `::GetGraftMarkerPath()` would be more precise.

bbockelm:
@reneme - anything else?

jblomer:
Merged with small adjustments.

PR #1114, opened 2015-09-28
FIX: test_libcvmfs initialization options

Moliholy:
It fixes some initialization options:
- pubkey is not in /etc/cvmfs/keys/cern.ch.pub but in /etc/cvmfs/keys/cern.ch/cern.ch.pub
- I guess proxies=DIRECT is the default option, but it didn't work for me. Instead, proxies=http://ca-proxy.cern.ch:3128 worked

### DIFF
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -1390,7 +1390,8 @@ void DownloadManager::FiniHeaders() {
 
 void DownloadManager::Init(const unsigned max_pool_handles,
                            const bool use_system_proxy,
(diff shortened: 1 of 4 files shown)


jblomer:
The pubkey change is fine.  The proxy should not be set to CERN's proxy though.  This should be set when invoking `test_libcvmfs`, like `HTTP_PROXY=... ./test_libcvmfs ...`

Moliholy:
PR updated

PR #1115, opened 2015-09-28
Support for external data files (CVM-907)

bbockelm:
This is the rebased version of #1110, now based on branch CVM-907.

Haven't gotten a chance to address the review comments - just rebased and made sure things still compile.

### DIFF
--- a/cvmfs/catalog.cc
+++ b/cvmfs/catalog.cc
@@ -234,7 +234,7 @@ bool Catalog::LookupEntry(const shash::Md5 &md5path, const bool expand_symlink,
   bool found = sql_lookup_md5path_->FetchRow();
   if (found && (dirent != NULL)) {
     *dirent = sql_lookup_md5path_->GetDirent(this, expand_symlink);
-    dirent->set_is_external_file(GetExternalDataLocked());
+    dirent->set_is_external_file(GetExternalDataLocked() == Present);
     FixTransitionPoint(md5path, dirent);
   }
   sql_lookup_md5path_->Reset();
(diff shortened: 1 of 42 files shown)


bbockelm:
@jblomer - I added support (I think) for 'swissknife info' in 2d79118 -- however, I can't get that command to work on any mounted repository (seems it expects .cvmfspublished is in the repository itself?), even the standard CERN ones.

Is there a trick to make it work?

bbockelm:
@jblomer - would it be better to have the Fetcher object manage both download managers or have multiple fetchers?

jblomer:
As discussed in #1110 the original comment pasted:

Please submit to branch CVM-907. There are two supporting commits in this branch. Commit 9d07918 allows for multiple download managers to exist concurrently (more to it later). Commit f417123 reserves bit 7 in the flags of directory entries for external files and it adds statistics counters for external files (number, total size). The second commit bumps the catalog schema revision to 3, a change that is transparent to clients. The server tools automatically migrate catalogs from previous schemas. From the data structure point of view, this allows now for mixed catalogs with external and native files.

We need to think about what to do with `cvmfs_swissknife pull` (replication). I'd suggest to rename the `SqlAllChunks` query to `SqlAllNativeChunks` and to exclude the external files (and chunks) from it. I'll submit something to the CVM-907 branch.

Regarding the external data status, I don't think we need to add propagation logic to the client. The server tools can simply add the flag to all nested catalogs.

As for the downloading, let's use two download managers, one for the native cvmfs CDN (stratum 1 + proxies) and one for the external location. Even though we'll start with catalogs and data on the same server, if it works well we'll soon want to use different network parameters (timeouts etc.). And there is already too much if-then-else logic in the download manager.

This PR should have a few additions:

 - An integration test that creates a repository with external files and checks if the client opens it as such (magic extended attributes set etc.). The integration tests are shell scripts in the test/src directory. As of test number 500 the server side tests start. You can have a look in the first server tests and also in the 585-xattr server test for blueprints.
  - `cvmfs_swissknife check` should detect if entries are marked external but the external_data property is not set
  - `cvmfs_server info` should show if a repository is flagged as external

Please have a look at python cpplint.py cvmfs/*. I think there are a few style violations. Otherwise I'll add some minor comments directly in the code.

jblomer:
@bbockelm: a second fetcher object should work fine.  In the same way the download manager got a name in order to distinguish the statistics counters, the `Fetcher` constructor would also need to get a name, which is then used in `statistics->Register`.  That's not very nice. CVM-913 is supposed to take care of this in a nicer way.

bbockelm:
Alright, integration test added.

Do you consider having a second fetcher object a blocker for merging?  I'd prefer to do that in a follow-up.

bbockelm:
Refactored quite a bit to make a second fetcher object.

I think that's the last of the items from the code review.

jblomer:
I'll work on asap, still a bit of a backlog here

jblomer:
Please send this pull request to destination branch CVM-907.  This removes my commits from the diff viewer.

cvmfs/sync_item.cc
```diff
@@ -10,11 +10,86 @@
 
 using namespace std;  // NOLINT
 
+
+static bool ProcessCatalogProperties(const std::string &scratch_path,
+                                     bool &external_data) {  // NOLINT
+  char buf[128];
+  external_data = false;
+  ssize_t retval;
+  if (-1 != (retval = platform_lgetxattr(scratch_path.c_str(),
```
jblomer: I wonder if the extended attribute is necessary at all.  Can't we just look inside the catalog marker if the file size is non-zero?

cvmfs/catalog.cc
```diff
@@ -408,6 +415,48 @@ void Catalog::DropDatabaseFileOwnership() {
...
+    } else {
+      status = parent_ ? parent_->GetExternalDataLocked() : \
+                         kExternalUnspecified;
+    }
+    atomic_write32(&external_data_status_, status);
+  }
+  return status;
+}
+
+Catalog::ExternalDataStatus Catalog::GetExternalDataLocked() const {
```
reneme: That's code duplication with `Catalog::GetExternalDataUnlocked`... Could that be DRY-ed up?

jblomer:
Closing this since it has been merged into CVM-907.  Please push the diff between devel and CVM-907 to the review board.

PR #1116, opened 2015-09-29
Detect valgrind at runtime and do not fail in this case.

bbockelm:
This will only work if valgrind.h is available on the build system
(and will compile cleanly if it is not).


jblomer:
Cool!

PR #1121, opened 2015-09-30
FIX: Fedora 22 Build

reneme:


jblomer:
Yep, please just add a line to the changelog of the spec file.  We can also mention that support for fedora < 21 dropped.

reneme:
✓

PR #1125, opened 2015-10-01
Use IntegerMap for CVMFS_{U/G}ID_MAP

jblomer:
Remove the parsing of UID/GID maps from the options manager and uses the `IntegerMap` class developed for catalog migration.

NOTE: The semantics of `Map()` has been changed: instead of mapping values without explicit rule to zero, they are now mapped to themselves.  The `Map()` method wasn't used so far.

### DIFF
--- a/cvmfs/uid_map.h
+++ b/cvmfs/uid_map.h
@@ -113,10 +113,10 @@ class IntegerMap {
       : k;
   }
 
-  bool IsEmpty() const { 
-    return map_.size() == 0 && !has_default_value_;   
+  bool IsEmpty() const {
+    return map_.size() == 0 && !has_default_value_;
   }
-    
+
   bool   IsValid()    const { return valid_;             }
   bool   HasDefault() const { return has_default_value_; }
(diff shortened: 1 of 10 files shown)


cvmfs/uid_map.h
```diff
@@ -110,7 +110,11 @@ class IntegerMap {
 
     return (HasDefault())
       ? default_value_
-      : T(0);
+      : k;
+  }
+
+  bool IsEmpty() const {
+    return map_.size() == 0 && !has_default_value_;
```
reneme: Mhh, a map without rules but only a default value is still "empty", no? Since `::Map()` is now a NOOP for "empty" maps without default value, I would propose to add two methods instead:

* `::IsEmpty()` for any map with 0 rules
* `::HasEffect()` that returns `false` in the above case

jblomer: [Yep](https://github.com/cvmfs/cvmfs/commit/06d6a6f6df1727a92dcd8bed95f252be55271350)

PR #1129, opened 2015-10-02
FIX: systemd might run multiple times

reneme:
Fedora 22 has multiple `systemd` processes running. Hence, the check for `systemd` in `cvmfs_server` cannot simply do `$(pidof systemd) -eq 1` since this might have returned multiple PIDs. This introduces a `minpidof()` helper to figure out the lowest running PID that can then be checked for `1`.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -78,7 +78,7 @@ minpidof() {
   pidof $1 | tr " " "\n" | sort | head -n1
 }
 SERVICE_BIN="false"
-if ! pidof systemd > /dev/null 2>&1 || [ $(pidof systemd) -ne 1 ]; then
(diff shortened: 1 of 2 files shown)


reneme:
Note: added the same fix to `test_functions` ... Effing code duplications!

bbockelm:
Wouldn't it be more straightforward to do something like this:

```
[ "xsystemd" == "x`cat /proc/1/stat | awk '{print $2;}' | tr -d '()'`" ]
```

?

jblomer:
@bbockelm you might need special permissions to access /proc/1/stat, no?

bbockelm:
Nah - checking /proc/$PID/stat is precisely how 'ps' works.

PR #1137, opened 2015-10-05
FIX: OverlayFS

reneme:
This is a first attempt to get *overlayfs* [fully supported](https://sft.its.cern.ch/jira/browse/CVM-835) in CernVM-FS. Currently I am focussing on the *overlayfs* flavour that is shipped with Fedora 22 (Kernel 4.1.8-200.fc22.x86_64), however the relevant differences to CentOS 7.1 seem to very minimal (except that Fedora's *overlayfs* works properly). Furthermore Ubuntu's version of *overlayfs* behaves different as well. Support that is left for future work, since Ubuntu still provides *AUFS* as an alternative.

*Overlayfs* uses a special character device with a trusted extend attribute (i.e. 'trusted.overlay.opaque') as whiteouts and opaque markers. The former is easily recognised by this proposed patch, the latter requires the `cvmfs_swissknife` binary to obtain **CAP_SYS_ADMIN**.


CMakeLists.txt
```diff
@@ -400,6 +400,7 @@ endif (NOT MACOSX)
 
 find_package (OpenSSL REQUIRED)
 find_package (UUID REQUIRED)
+find_package (LibCAP REQUIRED)
```
jblomer: It should be sufficient to only look for it when server / unit tests are built

jblomer: Hm, I guess this is also Linux only, right?

cvmfs/sync_union.cc
```diff
@@ -377,35 +458,39 @@ bool SyncUnionOverlayfs::ReadlinkEquals(string const &path,
  * @param[in] name of the attribute for which to compare the value
```
jblomer: I think the comments need to be adjusted to the new signature.

jblomer:
Very nice!  Just a little concern on the build system integration.

reneme:
Good catches! Fixed in an [additional Pull Request](https://github.com/cvmfs/cvmfs/pull/1143) to avoid rebase stunts.

jblomer:
Closing since superseded by #1143

PR #1138, opened 2015-10-05
FIX: Integration Tests for OverlayFS and Fedora

reneme:
This fixes several minor things to make the integration test set compatible to both OverlayFS and Fedora 22.

### DIFF
--- a/test/test_functions
+++ b/test/test_functions
@@ -1306,7 +1306,7 @@ display_timestamp() {
 # @param destination_directory
 cp_bin() {
   local destination_directory="$1"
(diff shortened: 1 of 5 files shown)


test/test_functions
```diff
@@ -1286,17 +1300,30 @@ display_timestamp() {
   date --date="$timestamp"
 }
 
+# this copies all top-level files from source to destination ignoring symlinks
+# Note: Using the content of /bin or /usr/bin as integration test guinea pig was
```
jblomer: Please elaborate a bit what the problem is/was.

PR #1139, opened 2015-10-05
Add Compression algorithm options (CVM-906)

djw8605:
This pull request will add bits to the `flags` field in the catalog to specify the compression mechanism on a per file basis.  Currently, only zlib (the default) and none (no compression) are supported.

The compression mechanism is specified on the command line to the mkfs command.  It is written into the `server.conf` file in order to be used transparently in future publish's.

A repository can be a mix of compressed and non-compressed files, since compression is stored in the `flags` for each file.

## TODO
- [x] Abstract the compression in chunk.cc.  Writing with zlib, even with 0 compression, still prepends the data with a zlib header which will fail a transaction size check.
- [ ] Some files are manually added to the repo such as the manifest and history file.  They should follow the repository's default compression mechanism.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -287,12 +287,12 @@ if (BUILD_CVMFS)
   set_target_properties (cvmfs_fuse_debug PROPERTIES VERSION ${CernVM-FS_VERSION_STRING})
 
   # link the stuff (*_LIBRARIES are dynamic link libraries *_archive are static link libraries ... one of them will be empty for each dependency)
-  set (CVMFS_FUSE_LINK_LIBRARIES ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${LEVELDB_LIBRARIES} ${OPENSSL_LIBRARIES} ${FUSE_LIBRARIES} ${LIBFUSE_ARCHIVE} ${SQLITE3_ARCHIVE} ${LIBCURL_ARCHIVE} ${PACPARSER_ARCHIVE} ${LEVELDB_ARCHIVE} ${CARES_ARCHIVE} ${ZLIB_ARCHIVE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} pthread dl)
+  set (CVMFS_FUSE_LINK_LIBRARIES ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${LEVELDB_LIBRARIES} ${OPENSSL_LIBRARIES} ${FUSE_LIBRARIES} ${LIBFUSE_ARCHIVE} ${SQLITE3_ARCHIVE} ${LIBCURL_ARCHIVE} ${PACPARSER_ARCHIVE} ${LEVELDB_ARCHIVE} ${CARES_ARCHIVE} ${ZLIB_ARCHIVE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} ${TBB_LIBRARIES} pthread dl)
 
   target_link_libraries (cvmfs2           ${CVMFS_LOADER_LIBS} ${OPENSSL_LIBRARIES} ${LIBFUSE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} pthread dl)
   target_link_libraries (cvmfs_fuse_debug ${CVMFS2_DEBUG_LIBS} ${CVMFS_FUSE_LINK_LIBRARIES})
   target_link_libraries (cvmfs_fuse       ${CVMFS2_LIBS} ${CVMFS_FUSE_LINK_LIBRARIES})
-  target_link_libraries (cvmfs_fsck       ${CVMFS_FSCK_LIBS} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${ZLIB_ARCHIVE} ${SHA2_ARCHIVE} pthread)
+  target_link_libraries (cvmfs_fsck       ${CVMFS_FSCK_LIBS} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${ZLIB_ARCHIVE} ${SHA2_ARCHIVE} ${TBB_LIBRARIES} pthread)
 
 endif (BUILD_CVMFS)
 
@@ -350,7 +350,7 @@ if (BUILD_LIBCVMFS)
   add_dependencies (libcvmfs cvmfs_only)
 
   add_executable( test_libcvmfs ${TEST_LIBCVMFS_SOURCES} )
-  target_link_libraries( test_libcvmfs ${CMAKE_CURRENT_BINARY_DIR}/libcvmfs.a ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${RT_LIBRARY} ${UUID_LIBRARIES} pthread dl )
+  target_link_libraries( test_libcvmfs ${CMAKE_CURRENT_BINARY_DIR}/libcvmfs.a ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${RT_LIBRARY} ${UUID_LIBRARIES} ${TBB_LIBRARIES} pthread dl )
   add_dependencies (test_libcvmfs libcvmfs)
 
 endif (BUILD_LIBCVMFS)
@@ -382,6 +382,9 @@ if (BUILD_SERVER)
 
   if (TBB_PRIVATE_LIB)
     add_dependencies (cvmfs_swissknife libtbb)
+    add_dependencies (cvmfs_fsck libtbb)
+    add_dependencies (cvmfs_fuse libtbb)
+    add_dependencies (test_libcvmfs libtbb)
   endif (TBB_PRIVATE_LIB)
 
   add_dependencies (cvmfs_swissknife libsha2)
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -287,12 +287,12 @@ if (BUILD_CVMFS)
   set_target_properties (cvmfs_fuse_debug PROPERTIES VERSION ${CernVM-FS_VERSION_STRING})
 
   # link the stuff (*_LIBRARIES are dynamic link libraries *_archive are static link libraries ... one of them will be empty for each dependency)
-  set (CVMFS_FUSE_LINK_LIBRARIES ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${LEVELDB_LIBRARIES} ${OPENSSL_LIBRARIES} ${FUSE_LIBRARIES} ${LIBFUSE_ARCHIVE} ${SQLITE3_ARCHIVE} ${LIBCURL_ARCHIVE} ${PACPARSER_ARCHIVE} ${LEVELDB_ARCHIVE} ${CARES_ARCHIVE} ${ZLIB_ARCHIVE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} ${TBB_LIBRARIES} pthread dl)
+  set (CVMFS_FUSE_LINK_LIBRARIES ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${LEVELDB_LIBRARIES} ${OPENSSL_LIBRARIES} ${FUSE_LIBRARIES} ${LIBFUSE_ARCHIVE} ${SQLITE3_ARCHIVE} ${LIBCURL_ARCHIVE} ${PACPARSER_ARCHIVE} ${LEVELDB_ARCHIVE} ${CARES_ARCHIVE} ${ZLIB_ARCHIVE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} pthread dl)
 
   target_link_libraries (cvmfs2           ${CVMFS_LOADER_LIBS} ${OPENSSL_LIBRARIES} ${LIBFUSE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} pthread dl)
   target_link_libraries (cvmfs_fuse_debug ${CVMFS2_DEBUG_LIBS} ${CVMFS_FUSE_LINK_LIBRARIES})
   target_link_libraries (cvmfs_fuse       ${CVMFS2_LIBS} ${CVMFS_FUSE_LINK_LIBRARIES})
-  target_link_libraries (cvmfs_fsck       ${CVMFS_FSCK_LIBS} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${ZLIB_ARCHIVE} ${SHA2_ARCHIVE} ${TBB_LIBRARIES} pthread)
+  target_link_libraries (cvmfs_fsck       ${CVMFS_FSCK_LIBS} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${ZLIB_ARCHIVE} ${SHA2_ARCHIVE} pthread)
 
 endif (BUILD_CVMFS)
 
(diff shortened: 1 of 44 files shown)


jblomer:
Thanks!  I'll work on asap, still a bit of a backlog here.

djw8605:
No problem.  It's still not 100% complete, just started the pull request to see the direction.

cvmfs/file_processing/chunk.cc
```diff
@@ -63,8 +63,14 @@ void Chunk::Initialize() {
...
   zlib_context_.next_in  = Z_NULL;
   zlib_context_.avail_in = 0;
-  const int zlib_retval = deflateInit(&zlib_context_, Z_DEFAULT_COMPRESSION);
-  assert(zlib_retval == 0);
+  if (compression_alg_ == zlib::kZlibDefault) {
+    const int zlib_retval = deflateInit(&zlib_context_, Z_DEFAULT_COMPRESSION);
+    assert(zlib_retval == 0);
+  } else if (compression_alg_ == zlib::kNoCompression) {
+    // Initialize the zlib library to no compression
+    const int zlib_retval = deflateInit(&zlib_context_, 0);
```
jblomer: This might still prepend the zlib header, no?

cvmfs/compression.h
```diff
@@ -36,6 +36,14 @@ enum StreamStates {
   kStreamEnd,
 };
 
+// Do not change order of algorithms.  Used as flags in the catalog
+enum Algorithms {
+  kZlibDefault = 0,
+  kNoCompression,
+  kAny,
```
jblomer: Perhaps better `kUnknown`

cvmfs/file_processing/chunk.cc
djw8605: Yes, it does.  This was an incomplete patch set.  I'm working (and debugging) another patch set now that should make this much more modular.  I'm getting some sort of error though here: https://github.com/djw8605/cvmfs/blob/debug_compress/cvmfs/compression.cc#L816

jblomer:
I think this already in very good shape!

One thing that is missing is the naming of files in the cache so that `cvmfs_fsck` knows which compression algorithm to apply.

And we would integration testing.

@reneme more comments?

cvmfs/file_processing/chunk.cc
jblomer: If we can help, @reneme knows best about the `ZlibCompressor` class

djw8605: The `ZlibCompressor` class is of my creation.  I'm still debugging it.  It works fine for <1 chunk.  But when >1 chunk, it creates something that is unreadable by the downloader.

jblomer: Ah, sorry @reneme.  I knew it's not mine so I assumed it's yours :)

cvmfs/cache.cc
```diff
@@ -636,8 +636,10 @@ int64_t PosixCacheManager::Write(const void *buf, uint64_t size, void *txn) {
   Transaction *transaction = reinterpret_cast<Transaction *>(txn);
 
   if (transaction->expected_size != kSizeUnknown) {
-    if (transaction->size + size > transaction->expected_size)
+    if (transaction->size + size > transaction->expected_size) {
```
reneme: Why this nested `if`? Wouldn't a simple `&&` be enough?

cvmfs/compression.h
```diff
@@ -36,6 +37,55 @@ enum StreamStates {
...
+  * Abstract Compression class which is inhereited by implementations of 
+  * compression engines such as zlib...
+  */
+class Compressor {
+  public:
+    Compressor() {};
+    virtual ~Compressor() {};
+    virtual int Deflate(upload::CharBuffer &outbuf, size_t &outbufsize, 
+            const unsigned char* inbuf, const size_t inbufsize, 
+            const bool flush) { return 0; };
```
reneme: This can safely be pure virtual, or does a concrete `Compressor` without a `::Deflate()` make sense?

cvmfs/download.h
```diff
@@ -115,7 +115,7 @@ struct Counters {
  */
 struct JobInfo {
   const std::string *url;
-  bool compressed;
+  //bool compressed;
```
reneme: please remove instead of comment out

cvmfs/download.h
```diff
@@ -129,12 +129,13 @@ struct JobInfo {
...
   cvmfs::Sink *destination_sink;
   const shash::Any *expected_hash;
+  zlib::Algorithms compression_alg;
   const std::string *extra_info;
 
   // Default initialization of fields
   void Init() {
     url = NULL;
-    compressed = false;
+    //compressed = false;
```
reneme: please remove.

cvmfs/download.h
```diff
@@ -165,7 +167,11 @@ struct JobInfo {
   {
     Init();
     url = u;
-    compressed = c;
+    //compressed = c;
+    if (c)
```
reneme: ternary operator? 
```C++
compression_alg = c ? zlib::kZlibDefault : zlib::kNoCompression
```
Or even better, make it part of the `Init()` function to avoid code duplication.

cvmfs/file_processing/chunk.cc
```diff
@@ -57,14 +57,12 @@ void Chunk::Initialize() {
...
-  zlib_context_.next_in  = Z_NULL;
-  zlib_context_.avail_in = 0;
-  const int zlib_retval = deflateInit(&zlib_context_, Z_DEFAULT_COMPRESSION);
-  assert(zlib_retval == 0);
+  
+  if (compression_alg_ == zlib::kZlibDefault) {
+    compressor_ = new zlib::ZlibCompressor();
+  } else if (compression_alg_ == zlib::kNoCompression) {
+    compressor_ = new zlib::EchoCompressor();
+  }
```
reneme: What about a factory method in the abstract `Compressor` base class to factor out this distinction:
```C++
compressor_ = zlib::Compressor::Construct(compression_alg_);
```
There is even a helper template (`PolymorphicConstruction<>`) in util.h for this.

cvmfs/file_processing/chunk.h
```diff
@@ -157,6 +166,11 @@ class Chunk {
    */
   size_t                   bytes_written_;
   tbb::atomic<size_t>      compressed_size_;  ///< size of the compressed data
+  
+  /**
+   * Compressor
+   */
+   zlib::Compressor              *compressor_;
```
reneme: util.h has a `UniquePtr<>` implementation instead of a bare pointer.

cvmfs/file_processing/chunk.h
```diff
@@ -93,7 +102,7 @@ class Chunk {
 
   shash::ContextPtr& content_hash_context() { return content_hash_context_; }
   const shash::Any&  content_hash() const { return content_hash_; }
-  z_stream&          zlib_context() { return zlib_context_; }
+  zlib::Compressor*   get_compressor() { return compressor_; }
```
reneme: Please rename to `compressor()` to match the [coding style guidelines](https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Function_Names).
Potentially: `UniquePtr<>::weak_ref()` to return a bare pointer here.

cvmfs/file_processing/chunk.h
```diff
@@ -33,16 +34,24 @@ class File;
...
     current_deflate_buffer_(NULL), bytes_written_(0)
   {
     Initialize();
   }
+  ~Chunk() { 
+    if (!IsFullyProcessed()) 
+      Finalize(); 
+      
+    if (compressor_)
+      delete compressor_;
```
reneme: Remove in favor of `UniquePtr<>`.

cvmfs/file_processing/chunk.cc
```diff
@@ -90,7 +84,9 @@ void Chunk::Finalize() {
   if (deferred_write_) {
     FlushDeferredWrites();
   }
-
+  
+  delete compressor_;
+  compressor_ = 0;
```
reneme: Please remove in favour of `UniquePtr<>`.

cvmfs/file_processing/processor.cc
```diff
@@ -75,55 +75,39 @@ tbb::task* ChunkProcessingTask::execute() {
 void ChunkProcessingTask::Crunch(const unsigned char  *data,
                                  const size_t          bytes,
                                  const bool            finalize) {
```
reneme: This is perhaps the hottest method regarding the parallel file processing of the backend. I believe it's worth measuring how much we hurt performance with several `memcpy()` calls here and in `Compressor::Deflate()` as well as the virtual function call to `::Deflate()`.

cvmfs/swissknife_history.cc
```diff
@@ -40,6 +40,7 @@ void CommandTag::InsertCommonParameters(ParameterList *r) {
   r->push_back(Parameter::Optional(
     'e', "hash algorithm to use (default SHA1)"));
   r->push_back(Parameter::Switch('L', "follow HTTP redirects"));
+  r->push_back(Parameter::Switch('Z', "Compression Algorithm"));
```
reneme: This is supposed to be an optional parameter, no? `Parameter::Switch()` doesn't assume a string parameter after `-Z`. Probably you want to use `Parameter::Optional()` here.

cvmfs/swissknife_history.cc
```diff
@@ -78,6 +79,11 @@ CommandTag::Environment* CommandTag::InitializeEnvironment(
     (args.find('f') == args.end())
     ? ""
     : *args.find('f')->second;
+  
+  const zlib::Algorithms compression_alg =
+    (args.find('Z') == args.end())
+    ? zlib::kZlibDefault
+    : zlib::ParseCompressAlgorithm(*args.find('Z')->second);
```
reneme: This might not work, see above.

reneme:
One general remark of consistent naming: I've seen several versions of class member names for this:

* `Chunk::compression_alg_`
* `SyncItem::GetCompressionAlg()`
* `zlib::ParseCompressAlgorithm()`
* `SpoolerDefinition::compression_algorithm`
* `DirectoryEntryBase::compression_algorithm()`
* `SqlDirent::RetrieveCompressAlgorithm()`

Let's just call it 'compression algorithm' (+ maybe some verb) consistently.

cvmfs/CMakeLists.txt
```diff
@@ -287,12 +287,12 @@ if (BUILD_CVMFS)
   set_target_properties (cvmfs_fuse_debug PROPERTIES VERSION ${CernVM-FS_VERSION_STRING})
 
   # link the stuff (*_LIBRARIES are dynamic link libraries *_archive are static link libraries ... one of them will be empty for each dependency)
-  set (CVMFS_FUSE_LINK_LIBRARIES ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${LEVELDB_LIBRARIES} ${OPENSSL_LIBRARIES} ${FUSE_LIBRARIES} ${LIBFUSE_ARCHIVE} ${SQLITE3_ARCHIVE} ${LIBCURL_ARCHIVE} ${PACPARSER_ARCHIVE} ${LEVELDB_ARCHIVE} ${CARES_ARCHIVE} ${ZLIB_ARCHIVE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} pthread dl)
+  set (CVMFS_FUSE_LINK_LIBRARIES ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${LEVELDB_LIBRARIES} ${OPENSSL_LIBRARIES} ${FUSE_LIBRARIES} ${LIBFUSE_ARCHIVE} ${SQLITE3_ARCHIVE} ${LIBCURL_ARCHIVE} ${PACPARSER_ARCHIVE} ${LEVELDB_ARCHIVE} ${CARES_ARCHIVE} ${ZLIB_ARCHIVE} ${RT_LIBRARY} ${UUID_LIBRARIES} ${SHA2_ARCHIVE} ${TBB_LIBRARIES} pthread dl)
```
reneme: Why do we need TBB here and in `cvmfs_fsck`?

cvmfs/compression.cc
```diff
@@ -779,4 +787,120 @@ bool DecompressMem2Mem(const void *buf, const int64_t size,
...
+  
+  do {
+    
+    // Create a new buffer, big enough to hold the old big_buf, and whhatever
+    // else needs to be deflated
+    size_t needed_space = DeflateBound(stream_.avail_in);
+    upload::CharBuffer* tmp_buf = new upload::CharBuffer(big_buf->used_bytes() + needed_space);
+    memcpy(tmp_buf->free_space_ptr(), big_buf->ptr(), big_buf->used_bytes());
+    tmp_buf->SetUsedBytes(big_buf->used_bytes());
+    delete big_buf;
```
reneme: Not sure if this is correct. `big_buf` was just initialised as empty, how can it have any `used_bytes()` to be copied into `tmp_buf`?

cvmfs/compression.cc
```diff
@@ -779,4 +787,120 @@ bool DecompressMem2Mem(const void *buf, const int64_t size,
...
+    
+    big_buf->SetUsedBytes(big_buf->used_bytes() + (output_space - stream_.avail_out));
+    
+  } while(!(flush_int == Z_NO_FLUSH && retcode == Z_OK && stream_.avail_in == 0) && !(flush_int == Z_FINISH  && retcode == Z_STREAM_END));
+  
+  //outbuf = *big_buf;
+  outbuf.Allocate(big_buf->used());
+  memcpy(outbuf.free_space_ptr(), big_buf->ptr(), big_buf->used());
+  outbuf.SetUsedBytes(big_buf->used());
+  delete big_buf;
```
reneme: As said somewhere else: This is worth measuring, because this creates many short-lived buffer objects only to `memcpy()` things between them. Its potentially hurting performance quite a bit.

cvmfs/cache.cc
djw8605: I'm not entirely sure why it's nested.  It was nested before my modification, but I will modify it.

cvmfs/CMakeLists.txt
djw8605: Yes, it is needed for the compression.cc file.  It requires the CharBuffer, which requires the tbb:scalable buffer allocator.

-Derek

cvmfs/compression.cc
djw8605: During the first iteration, you are correct.  The big_buf will have 0 used bytes.  But, in the next iteration it could have many used bytes.

-Derek

cvmfs/CMakeLists.txt
jblomer: It looks to me that this is getting a bit too complicated.  We would like to avoid TBB dependency for the client.  What is wanted in a first approximation is optionally not to compress files on the server side.  There should be a simple way to get there.

djw8605: Agreed.  I will re-write the compression so it doesn't use the CharBuffer class, which is requiring the TBB libraries.

cvmfs/compression.cc
reneme: Ah, I see. Sorry.
I like the idea of the `Compressor` in the sense of a strategy pattern. However, I believe it should be implemented more efficiently regarding the `memcpy()`s and virtual function calls in the `ChunkProcessingTask` which is quite performance critical code.

Is it actually necessary to support many different compression algorithms in the foreseeable future? If not, I would remove the `Compressor` entirely and just skip the compression code in `ChunkProcessingTask`. If we need to support additional compression algorithms soon, we should look into a `Compressor`-implementation based on templates to benefit from inlining and avoid any temporary buffer creation.

cvmfs/CMakeLists.txt
djw8605: The compression was re-written to eliminate the memcpy's and to remove dependency on the TBB library.  I have removed the TBB library dependency from the `CMakeLists.txt` file. This is done in commit: 6ba468c

djw8605:
@jblomer Do you have any suggestions on a naming convention for the files to indicate compression method?

jblomer:
Looking at my keyboard, I am thinking of a suffix `@<digit>` where different compression algorithms get different digits.  It would use symbols that have no special bash meaning and it can be quite easily separated from the hash encoded in the file name.

djw8605:
@jblomer Regarding the additional suffix @<digit> for fsck.  I am concerned about the number of changes that are required in order to make this viable.  There are many locations where only the hash is given in order to download a chunk, such as swissknife_pull.  In this case, changes need to be made to `swissknife_pull`, catalog, and an updated SQL query in `catalog_sql.h` in order to extract the compression flag for the files.  There are also other locations such as the the `object_fetcher`.

In summary, I'm concerned about the number of change necessary to enable the compression testing for fsck.  It appears that we are changing the key to which every file is access from just the hash, to (hash, compression algorithm).

jblomer:
Hm, @reneme had a good idea: in a first approximation, we probably get by with calculating both the hash of the compressed file and the hash of the uncompressed file in `cvmfs_fsck`.  Should we ever add more compression algorithms we can still switch to a different naming scheme.

jblomer:
If we go for this approach it would be good to have a few numbers on how much this slows down `cvmfs_fsck`.  My gut feeling is "not much", but we should run it on a cache directory with a few gigabytes of data twice to get an educated idea about the penalty for cold file system buffers and warm file system buffers.

djw8605:
I've pushed a commit, 0f2a568, which will modify the `cvmfs_fsck` to check for uncompressed files.  The order of operations are:
1. Compress and hash the cached file.
2. If the hashes match the expected hash, then continue to the next file.
3. If the hashes do not match, hash the uncompressed file
4. If the hashes match now, continue to the next file.
5. If the hashes do not match, begin correcting or reporting the problem.

Therefore, if the file is expected to be compressed, it will take no longer than before (compress & hash).  If the file is expected to be uncompressed, then it will compress & hash + hash the file.  If the file is corrupted, it will need to compress & hash + hash.  In the worst case, we are adding another hashing of the chunk.

reneme:
In fact, compressing a data stream is *a lot* more CPU intense than hashing (out of the blue at least a factor 10). Perhaps we would not lose too much throughput with 'concurrently' hashing the uncompressed data while compressing/hashing it. I could even imagine that this would be quicker on average, due to file system and CPU caches. Of course, this highly depends on the fraction of uncompressed data files in the cache and would be up for measurement.

djw8605:
The processing of the chunks is already happening within a parallel worker thread.  So the compression / hashing is already parallelized at the file level.  I'm worried that any further parallelization may significantly hurt readability of the code, as well as overburden the CPU.

jblomer:
I pushed this PR with a couple of minor modifications and fixes for style violations to CVM-906 branch. Please use the style checker before commits.  I'd also suggest to configure the editor to remove trailing white spaces. A few things left to do:

 - Unit tests for the `Compressor` class
 - Integration test for repositories with uncompressed data and repositories that switch during their lifetime between uncompressed and compressed publish operations.  Should also include invocations od `cvmfs_fsck` and `cvmfs_scrub`.
 - For the `DirectoryEntry`: `compression_algorithm_` should default to zlib and only be set explicitly where it makes sense (e.g. not for directories as in catalog_mgr_rw.cc for the root entry)
 - The `kUnknown` algorithm should be removed.  I'd just `abort()` when parsing the hash algorithm fails
 - Add a comment to what the `DeflateBound` function does
 - In `cvmfs_read()`: the hash algorithm should be taken from the ChunkList.  The read function should not need to request a `DirectoryEntry`.
 - The changes to download.cc/h seem to be a preparation for the possible future use of more compression algorithms.  I don't think that's necessary yet, we can leave these files untouched.
 - The `tag` command in `cvmfs_swissknife` does not need to know about the compression algorithm.  In general, only the data files should be uploaded uncompressed (if wanted).  File catalogs, certificate, and history database should always be zlib compressed since we don't have a means to specify the compression algorithm as we have with file catalogs for the data. 

Notes to me:
 - The revision of libcvmfs needs to be increased
 - The file catalog revision needs to be increased, can be done together with the changes in CVM-907
 - For hotpatching the client, the FileChunk list needs to be migrated as the memory layout changed

jblomer:
Closing, discussion has been taken to review board.

PR #1140, opened 2015-10-06
Add libcvmfs integration tests (597)

Moliholy:
This test creates the typical integration test structure and compiles a binary whose source is included in the same folder. This binary actually tests if the repositories contain the required structure, and it is compiled against libcvmfs.a


jblomer:
Very good!

PR #1141, opened 2015-10-06
Allow Parrot execution of integration tests

Moliholy:
This enables run.sh to execute tests through Parrot. It doesn't mean all tests can now be executed with Parrot, but that they "could" if they don't use fuse (or sudo, in any case).

An idea to identify the tests that can run with parrot would be to manually include a variable, say CVMFS_PARROT_ALLOWED=true, in those tests that can actually run it, and set it to false by default.

Note: when running with Parrot the variable PARROT_ENABLED is set to TRUE

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -74,7 +74,11 @@ report_failure() {
   local workdir=$2
   echo $message
   if [ x$workdir != x ]; then
-    sudo cp $CVMFS_TEST_SYSLOG_TARGET $workdir
+    if [ x$PARROT_ENABLED = "xTRUE" ]; then
+      cp $CVMFS_TEST_SYSLOG_TARGET $workdir
+    else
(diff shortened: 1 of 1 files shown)


jblomer:
Could it be that commits are missing from this PR?

Moliholy:
No, in principle it's fine. It simply allows run.sh to execute tests with Parrot. Before it was impossible due to the sudo's. Now what we should do is to agree in how we want to organise the way integration tests run under Parrot, and I simply suggested a way, but that's something we should talk about.
Independently on the solution we finally choose, all the changes in this PR are necessary to avoid crashes at the beginning when running with Parrot

jblomer:
Ok, so there are more PRs to come, right?

Moliholy:
Yes, once we talk and discuss the preferred way to do it

PR #1142, opened 2015-10-07
Feature: OverlayFS Ignores Hardlinks

reneme:
This breaks up any hardlinks found during a publishing with OverlayFS. If a file with linkcount > 1 is detected, a warning is printed and the hardlink is masked away. Masked hardlink entries are written as ordinary entries into the catalogs. Hence, after the publishing, there are no hardlink relationships between files anymore. In case of OverlayFS the entire hardlink handling procedure in `SyncMediator` is disabled.

We find this to be the better alternative than trying to preserve hardlinks, because on a consecutive publish, OverlayFS doesn't obey their semantics anyhow. Note however, that the publishing stops with an error message when encountering a file inside CVMFS that already has a linkcount > 1 to ensure catalog consistency.

*Note:* This Pull Request is based on [FIX: OverlayFS](https://github.com/cvmfs/cvmfs/pull/1137).


cvmfs/sync_union.cc
```diff
@@ -203,98 +281,36 @@ void SyncUnionOverlayfs::ProcessFile(SyncItem *entry) {
...
+                                      "inconsistencies. Aborting..." ,
+             entry->GetUnionPath().c_str(), entry->GetRdOnlyLinkcount());
+    abort();
+  }
+}
+
+void SyncUnionOverlayfs::MaskFileHardlinks(SyncItem *entry) const {
+  assert(entry->IsRegularFile() || entry->IsSymlink());
+  if (entry->GetUnionLinkcount() > 1) {
+    LogCvmfs(kLogPublish, kLogStderr, "Warning: Found file with linkcount > 1 "
```
reneme: This might actually benefit from a silencing switch, since it might (and will) show up in production.

PR #202, opened 2013-07-08
Add utilities for CRC32 checksumming

bbockelm:
This pull request adds various utilities for calculating CRC32-based checksums.  See pull request #201 for an explanation of the ideas behind the checksum approach.  This includes a small test utility to calculate a CRC32-based checksum for an arbitrary file in a CVMFS cache.

The idea here is this will serve as the base of a lightweight fsck utility and, long-term, runtime verification of data files.

PR #1144, opened 2015-10-08
Add utilities for CRC32 checksumming

bbockelm:
This pull request adds various utilities for calculating CRC32-based checksums. See pull request #201 for an explanation of the ideas behind the checksum approach. This includes a small test utility to calculate a CRC32-based checksum for an arbitrary file in a CVMFS cache.

The idea here is this will serve as the base of a lightweight fsck utility and, long-term, runtime verification of data files.

This is a rebase of PR #202


bbockelm:
@jblomer - are you guys in a place where you can start thinking about this again?

I'm starting to work in the caching layer, and this would be a natural thing to pick up right now.

jblomer:
Closing, expired.  With all the new cache stuff, we might want to restart from where we are now.

PR #1147, opened 2015-10-08
FIX: Hardlink Check for Whiteouts

reneme:
`SyncUnionOverlayfs::ProcessFile()` needs to check if the passed `SyncItem` is a whiteout before doing any hardlink masking on it.


reneme:
Please wait... might be pre-mature.

PR #1149, opened 2015-10-08
Refactor: SyncItem Factory Method

reneme:
This fixes a [code smell](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/sync_item.h#L64) that was already marked as such since July. Namely, a `SyncItem` has methods `IsWhiteout()` and `IsOpaqueDirectory()` that depend on `SyncUnion` to set these properties for each and every `SyncItem` object. Before they might not return the right value. Now, *OverlayFS* added another such property - namely `HasHardlinks()`. Unfortunately also `SyncMediator` was *silently* creating `SyncItem` objects where non of those properties where actually set correctly by `SyncUnion` which broke the OverlayFS implementation.

Naturally only `SyncUnion` can figure out those properties of the `SyncItem` objects, as they depend on the union file system implementation. Therefore this refactors `SyncItem`, `SyncUnion` and `SyncMediator` in a way, that `SyncItems` can only be generated through the factory method `SyncUnion::CreateSyncItem()` which makes sure that all of the above mentioned properties are set before handing the `SyncItem` object out to client code.

In particular `SyncMediator` now needs to go through `SyncUnion` and cannot *silently* generate `SyncItem` objects anymore. This refactoring allows to elegantly fix the hardlink handling in OverlayFS - see next Pull Request.

### DIFF
--- a/cvmfs/sync_item.h
+++ b/cvmfs/sync_item.h
@@ -52,14 +52,11 @@ class SyncItem {
   inline bool IsSymlink()         const { return IsType(kItemSymlink);         }
   inline bool WasSymlink()        const { return WasType(kItemSymlink);        }
   inline bool IsNew()             const { return WasType(kItemNew);            }
+  inline bool IsCharacterDevice() const { return IsType(kItemCharacterDevice); }
 
   inline bool IsWhiteout()        const { return whiteout_;                    }
   inline bool IsCatalogMarker()   const { return filename_ == ".cvmfscatalog"; }
   inline bool IsOpaqueDirectory() const { return IsDirectory() && opaque_;     }
-
-  inline bool IsCharacterDevice() const {
-    return scratch_type_ == kItemCharacterDevice;
-  }
 
   inline shash::Any GetContentHash() const { return content_hash_; }
   inline void SetContentHash(const shash::Any &hash) { content_hash_ = hash; }
@@ -157,6 +154,7 @@ class SyncItem {
       if (S_ISDIR(stat.st_mode)) return kItemDir;
       if (S_ISREG(stat.st_mode)) return kItemFile;
       if (S_ISLNK(stat.st_mode)) return kItemSymlink;
+      if (S_ISCHR(stat.st_mode)) return kItemCharacterDevice;
       return kItemUnknown;
     }
 
--- a/cvmfs/sync_item.h
+++ b/cvmfs/sync_item.h
@@ -39,22 +39,7 @@ class SyncUnion;
  * details.
  */
 class SyncItem {
- private:
-  /**
-   * create a new SyncItem
-   * Note: SyncItems cannot be created by any using code. SyncUnion will take
-   *       care of their creating through a factory method to make sure they
-   *       are initialised correctly (whiteout, hardlink handling, ...)
-   *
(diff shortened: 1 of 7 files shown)


cvmfs/sync_item.h
```diff
@@ -39,34 +39,38 @@ enum SyncItemType {
  * details.
  */
 class SyncItem {
```
jblomer: The usual order is a bit mixed up: we normally have friend classes first, then public, protected, private

cvmfs/sync_item.h
```diff
@@ -169,6 +174,7 @@ class SyncItem {
   mutable EntryStat scratch_stat_;
 
   bool whiteout_;
```
jblomer: Actually, since we are anyway at this class, it would be helpful to add comments to the private members and briefly remind the reader of the union file systems concepts they indicate.

cvmfs/sync_union.h
```diff
@@ -233,12 +252,12 @@ class SyncUnionOverlayfs : public SyncUnion {
                            const std::string &filename);
   void ProcessCharacterDevice(const std::string &parent_dir,
                               const std::string &filename);
-  std::string UnwindWhiteoutFilename(const std::string &filename) const;
+  std::string UnwindWhiteoutFilename(const SyncItem &entry) const;
   std::set<std::string> GetIgnoreFilenames() const;
-  virtual void ProcessFile(SyncItem *entry);
+  virtual void ProcessFile(SyncItem &entry);
```
jblomer: Is this a actually a `const SyncItem &entry`?  Otherwise the pointer version should stay to indicate that `entry` can be changed inside `ProcessFile`.

jblomer:
The factory method makes a lot of sense I think.  Before we merge this, please also cross-check if this conflicts with grafting of files in CVM-908.  If it conflicts, we should cherry pick this commit into CVM-908 branch and make necessary adjustments.

cvmfs/sync_item.h
reneme: Agreed and fixed.

cvmfs/sync_item.h
reneme: Added some additional documentation. Union file system concepts are discussed in more detail in `SyncUnion` which is referenced in `SyncItem` several times. I believe the attentive audience should find that. :-)

cvmfs/sync_union.h
reneme: Ah! That is the reason, this was the only pointer parameter. Anyway, fixing this in the [pull request on top](https://github.com/cvmfs/cvmfs/pull/1150), to avoid a rebase-rage.

reneme: [Done](https://github.com/reneme/cvmfs/commit/d258d78a77005a992cae6f2c4ec7a5bf0e9c56b7) even though I consider this nitpicking. Especially [here](https://github.com/reneme/cvmfs/commit/d258d78a77005a992cae6f2c4ec7a5bf0e9c56b7#diff-4d6b1fda58c9782181cc19b60ba1d3ebL298), where it takes a significant thinking-overhead to figure out that actually the `*entry` is a `const&`-parameter and therefore de-referenced, while the method right underneath is actually changing `entry`.

PR #1151, opened 2015-10-12
FIX: Acquire CAP_SYS_ADMIN only if necessary

reneme:
This handles the need for `CAP_SYS_ADMIN` for `cvmfs_swissknife` on certain OverlayFS implementations. OverlayFS after v22 uses trusted extended file attributes to mark whiteouts and opaque directories. Hence, `cvmfs_swissknife` needs `CAP_SYS_ADMIN` to identify those.

While `cvmfs_server mkfs` the need for `CAP_SYS_ADMIN` is checked (using OverlayFS >= v22) and it is granted to `cvmfs_swissknife` if necessary. Furthermore `cvmfs_swissknife` only tries to acquire this capability if it actually interacts with an OverlayFS >= v22.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -527,8 +527,7 @@ get_available_union_fs() {
   fi
 }
 
-# gets the system file system name of the overlayfs
-# apparently this differs from version to version... sigh!
+# gets the system's file system name of OverlayFS (overlay vs. overlayfs)
 get_overlayfs_name() {
   check_overlayfs || die "overlayfs is not installed!"
   if /sbin/modprobe -q overlayfs || [ -d /sys/module/overlayfs ]; then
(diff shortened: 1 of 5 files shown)


reneme:
please don't merge yet... turn out that also OverlayFS < v22 uses those extended attributes :-(

reneme:
Repaired... OverlayFS requires `CAP_SYS_ADMIN`, also in older versions. :-( Anyway, there is a potential problem with this implementation: When upgrading the CVMFS version, RPM install will potentially silently remove the capabilities. Needs investigation...

jblomer:
Yeah, it's somewhat messy. Regarding packaging, @traylenator do you have an opinion?  With overlayfs, `cvmfs_swissknife` needs to get the `CAP_SYS_ADMIN` capability.  But only then, not for aufs, not for stratum 1s.  Should we give the file capability anyway?  Or fix it dynamically as we do now?  Or install two copies, one with, one without capabilities?

traylenator:
I'd say don't do it by default. strat 0s already have quite a few manual steps based on particular deployment scenario. From my point of view this is easy for me to automate.

jblomer:
Thanks!

PR #1155, opened 2015-10-12
FIX: SLC6 Vagrant Provision

reneme:
Two things:

1. update packages before installing AUFS enabled kernel
2. add `libcap-devel` to the game


jblomer:
rpm and deb packages would need an update, too.

PR #1150, opened 2015-10-08
FIX: Hardlink Check and Preprocessing in OverlayFS

reneme:
This fixes a problem with hardlink handling in OverlayFS where certain `SyncItem`s where skipped due to the *silent* creation of `SyncItem` objects inside `SyncMediator`.

*Note:* This depends on the pull request [Refactor: SyncItem Factory Method](https://github.com/cvmfs/cvmfs/pull/1149).

PR #1156, opened 2015-10-12
FIX: CAP_IS_SUPPORTED is not available on RHEL < 7

reneme:
Turns out, that the convenience macro `CAP_IS_SUPPORTED()` is not available in `<sys/capability.h>` on older Linux versions. We'll just skip this sanity-check on these platforms and live with the less descriptive error message if the capability actually is not supported - which is highly unlikely anyways.

**Note:** This builds on [Refactor: SyncItem Factory Method](https://github.com/cvmfs/cvmfs/pull/1149) and [FIX: Hardlink Check and Preprocessing in OverlayFS](https://github.com/cvmfs/cvmfs/pull/1150).

### DIFF
--- a/cvmfs/platform_linux.h
+++ b/cvmfs/platform_linux.h
@@ -8,6 +8,7 @@
 #define CVMFS_PLATFORM_LINUX_H_
 
 #include <sys/types.h>  // contains ssize_t needed inside <attr/xattr.h>
+#include <sys/xattr.h>
 #include <attr/xattr.h>  // NOLINT(build/include_alpha)
 #include <dirent.h>
 #include <errno.h>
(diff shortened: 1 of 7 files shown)


reneme:
> I think failing would be better than praying...

Definitely. I've had a look how the macro expands, however it cannot be easily back ported as it uses a function that is also not available in RHEL 6.

Anyway, what do you mean with failing would be better? I've just `#ifdef`-ed a sanity check that might give you a nice error message. On RHEL it would likely fail with something more obscure a couple of lines lower...

jblomer:
We should run the integration test suite at this point.

jblomer:
Please also merge with CVM-908

PR #1158, opened 2015-10-13
FIX: Package Dependencies for libcap

reneme:
Both for debian and RHEL.


jblomer:
Yep, please add a changelog entry in the rpm.

reneme:
Ah! Bloody changelog. Fixed.
In fact there is also a change log in the debian incarnation which seems to be rather orphaned. I'll leave that for another PR...

PR #1164, opened 2015-10-13
TEST: Parrot for Integration Tests

Moliholy:
Little pair-programming session with @reneme.

We've managed to run 'some' integration tests through parrot as a proof of concept. Just run:

``` bash
$# parrot_run bash
$# ./run.sh /tmp/parrot.log src/004-davinci
```

However, the test case needed some tuning:
- mask some `sudo`s from `run.sh`
- add a couple of helpers into `test_functions` to mask them in parrot
- mask some calls to `cvmfs_talk`

Say: Not everything is tested just the same way as with FUSE, but the Davinci test passes successfully. As well as 002 and 003 which are pretty much minimal, though.

More to come...


jblomer:
Nice!

I think one problem with running everything in a parrot sandbox is that the configuration can never change and the

jblomer:
sorry, wrong button

jblomer:
Can you change it in such a way that the `run.sh` script takes an additional parameter that determines if the individual test cases are run under parrot or not.  So we could start with a clean parrot environment for every test case.

jblomer:
@Moliholy any news on this?

Moliholy:
It could be perfectly done, but I don't really see the point. I mean, the actual tests are all of them executed in a subshell, so there wouldn't be pollution of the environment; with or without parrot

jblomer:
If they all run in the same parrot sandbox, the cvmfs configuration for all of them is fixed.  No chance to to set different network parameters, different cache parameters, and so on.  But that's the whole point of the integration tests.

Moliholy:
But would we be interested in running certain tests under certain parameters _only_ because they are running under parrot? Even setting that parameter to run every test individually under parrot, is there an infrastructure already built to specify for each test different cvmfs configurations?

jblomer:
We absolutely do want the possibility to set different parameters for different tests.  One way of doing it is to add extra parameter files in the tests directory, like parrot_config.  If this file exists, the test would be ready to run under parrot and the configuration specifics can be picked up from there.

jblomer:
Any progress on this?

Moliholy:
Not at the moment. I'll soon get my hands dirty again with this

Moliholy:
Done. Now if specifying the "-p" option when running run.sh it runs under Parrot. At least the test number four is executed with no problems

jblomer:
Closing, expired.

PR #1168, opened 2015-10-16
FIX/RFC: Do not Mount CVMFS Stratum0's on system boot

reneme:
Apparently there are various different scenarios how mounting both `/var/spool/cvmfs/$repo/rdonly` and `/cvmfs/$repo` on stratum0s can fail during system boot. Usually this results in the boot process being interrupted.

For example:
* `/var/spool/cvmfs` not mounted yet
* Apache not started yet
* mixed up order of mounting `.../rdonly` and `/cvmfs/...`
* ... probably more

I couldn't really wrap my head around what exactly happens in the different scenarios, so this is more of an RFC-Pull Request. Anyway, by just adding **noauto** to the CVMFS `/etc/fstab` entries this issue would be easily (and reliably) fixable. Mounting of both `.../rdonly` and `/cvmfs/...` would happen (pretty much transparently) on the first `cvmfs_server` invocation through the auto-repair-mountpoint feature of CVMFS.

We should make `CVMFS_AUTO_REPAIR_MOUNTPOINT` the default in the next release anyway, as it proved to work rather nicely.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -440,6 +440,7 @@ least CernVM-FS $creator to manipulate this repository."
   #
   #   2.1.20+ -> 2.2.0-1 (2.2.0-0 was a server pre-release and needs migration)
   #     -> new (mandatory) parameters in client.conf (Stratum 0)
+  #     -> adjustments in /etc/fstab
   #
   # Note: I tried to make this code as verbose as possible
   #
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -2143,8 +2143,8 @@ setup_and_mount_new_repository() {
       ofs_workdir=""
     fi
     cat >> /etc/fstab << EOF
-cvmfs2#$name $rdonly_dir fuse allow_other,config=/etc/cvmfs/repositories.d/${name}/client.conf:${CVMFS_SPOOL_DIR}/client.local,cvmfs_suid 0 0 # added by CernVM-FS for $name
-${overlayfs_name}_$name /cvmfs/$name $overlayfs_name upperdir=${scratch_dir},lowerdir=${rdonly_dir}$ofs_workdir,ro$selinux_context 0 0 # added by CernVM-FS for $name
+cvmfs2#$name $rdonly_dir fuse allow_other,config=/etc/cvmfs/repositories.d/${name}/client.conf:${CVMFS_SPOOL_DIR}/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for $name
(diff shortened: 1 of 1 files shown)


DrDaveD:
I'm afraid that would lead to a lot of confusion when people either don't see /cvmfs/<repo> after a reboot and/or see the auto repairing going on for the first transaction.  I haven't seen cases of boot processes being interrupted.  It seems to me that it would be better to really figure out what's going wrong with the boot process.  There could be a script run late in the boot process that does a cvmfs_server operation, but the question would be does that help.  It's also pretty O.S.-specific.

jblomer:
Adding a boot script gets tricky with the different init systems (SysV, systemd, upstart).  Since we have seen the failures René describes in production, it looks to me that this workaround improves the current situation.

Of course we should document the change.  System administrators could still add post-boot scripts to mount.

jblomer:
@reneme: I think we should go for it.  Can you please add the migration as well.

reneme:
Rebased to latest devel and added the `/etc/fstab` migration to fulfil [CVM-925](https://sft.its.cern.ch/jira/browse/CVM-925). Please have a look over the [migration code](https://github.com/reneme/cvmfs/commit/ecaddfbfb4d2dd6a0765ebfd4d17ede0b3f20226), since it naturally is quite delicate.

jblomer:
I see that `CVMFS_AUTO_REPAIR_MOUNTPOINT=yes` is the default.  Do we need upgrade for this, too?

reneme:
Mhh, both are valid configurations. Not sure if we should let the administrator decide or just change the config ...

jblomer:
Since you mention it in the pull request: turning auto mounting on boot off means that `CVMFS_AUTO_REPAIR_MOUNTPOINT` needs to be on, otherwise the repo wouldn't mount after a reboot, no?

reneme:
That's a very valid point. Good catch...

reneme:
in fact... probably we should even remove the possibility to switch this off, no?

jblomer:
Hm, I wouldn't go that far. Let's have one transitional release where it can still be turned off manually.

reneme:
sounds like a good compromise.... new Pull Request is coming...

DrDaveD:
As a system administrator I have now run into this case and it is quite surprising and a bit disturbing that the repos are not mounted after boot.  Also, is there a very simple instruction we can tell people to consistently use to mount all cvmfs repos in an /etc/rc*.d script?  I mean, it seems like there ought to be a simple shell command we can tell people to run, something like "cvmfs_server mount -a" or something like that.
The simplest code I can think of is
```
grep "CernVM-FS" /etc/fstab | awk '{print $2}' | while read M; do mount $M; done
```

PR #1172, opened 2015-10-19
Feature: Print OverlayFS Kernel Deadlock Warning

reneme:
As stated before, OverlayFS can cause the same kernel deadlock [as we've seen with AUFS](https://sft.its.cern.ch/jira/browse/CVM-651). Therefore, this adapts the sanity-check code in `cvmfs_server` to also warn on vulnerable OverlayFS configurations. Furthermore this adapts the integration test case 564 to also check the proper warning functionality with OverlayFS.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -1469,10 +1469,15 @@ health_check() {
   fi
 
   # check configuration for harzardous AUFS setup
-  if [ $quiet -eq 0 ] && find_hazardous_aufs_config $name; then
-    echo "WARNING: AUFS config is vulnerable to kernel deadlocks. Please store "
-    echo "         the scratch space and the cvmfs client cache on different "
-    echo "         partitions. (Silence this with CVMFS_AUFS_WARNING=false)"
+  if [ $quiet -eq 0 ] && find_hazardous_union_file_system_config $name; then
+    echo "WARNING: Union filesystem config is vulnerable to kernel deadlocks."
+    echo "         Please store the scratch space and the CVMFS client cache on"
+    echo "         different partitions."
+    echo ""
+    echo "         This can be silenced with either"
+    echo "             CVMFS_AUFS_WARNING=false"
+    echo "         or  CVMFS_OVERLAYFS_WARNING=false"
+    echo ""
     echo " Scratch Space: ${CVMFS_SPOOL_DIR}/scratch"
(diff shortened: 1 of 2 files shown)


jblomer:
Would this problem go away when the link/unlink magic is used?  You can try by always using the `alien_cache_on_nfs_` code path in `PosixCacheManager::Rename`.  If this works, we'd make this configurable and turned on by default for the cvmfs client backing the server.

reneme:
Yep, in fact that seems to work when commenting out the code-path calling `rename()` like so:

```C++
int PosixCacheManager::Rename(const char *oldpath, const char *newpath) {
  int result;
  // if (!alien_cache_on_nfs_) {
  //   result = rename(oldpath, newpath);
  //   if (result < 0)
  //     return -errno;
  //   return 0;
  // }

  result = link(oldpath, newpath);
  if (result < 0) {
    if (errno == EEXIST)
      LogCvmfs(kLogCache, kLogDebug, "%s already existed, ignoring", newpath);
    else
      return -errno;
  }
  result = unlink(oldpath);
  if (result < 0)
    return -errno;
  return 0;
}
```

reneme:
Well, at least on the OverlayFS in Fedora 22, I'll test on AUFS and Ubuntu and let you know.

reneme:
Cool! That looks like a doable way. Tried successfully on Fedora 22 (upstream kernel OverlayFS), Ubuntu 14.04 (both AUFS and OverlayFS) and SLC 6 (with Kernel 2.6.32-431.23.3 that didn't include our custom patch yet). All successful.

jblomer:
Very well, one problem solved!  So how should we call the parameter? `CVMFS_WORKAROUND_RENAME`?

reneme:
why flag it as a 'workaround'? What about `CVMFS_SERVER_CACHE_MODE`? That'd be more generic and could potentially trigger other 'workaround'-like things down the road?

jblomer:
Just thought because it's dual purpose: it's also used for alien caches on NFS.

jblomer:
Oh, but this other purpose doesn't need a parameter.  Yep, sounds good then.

jblomer:
I'll take care of it.

reneme:
thx!

reneme:
Closing this one, in favor of a workaround and complete removal of the kernel deadlock warning from `cvmfs_server`.

PR #1174, opened 2015-10-19
FIX: Stratum1 URL Handling

reneme:
Catalin just found a bug in the Stratum1 URL handling when creating a fresh replica on S3, like so:

```bash
cvmfs_server add-replica -w http://path/to/S3 -u S3,/tmp,test.cern.ch@/etc/path/to/s3.conf http://path/to/stratum0 /etc/path/to/pubkey.pub
```

TL;DR: When a custom upstream string (i.e. `-u S3,...`) is specified, the Stratum1 base-URL provided with `-w` is not properly added to the configuration file. The problem is a convoluted if-statement which I fixed and refactored a bit in this Pull Request.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -2737,17 +2737,23 @@ add_replica() {
   fi
 
   # upstream generation (defaults to local upstream)
-  local stratum1=""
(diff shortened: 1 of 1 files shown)


reneme:
please don't merge yet... I'm working on a follow up.

PR #1148, opened 2015-10-08
Cvmfs preload

Moliholy:
This PR adds the necessary code to create a cvmfs_preload binary that fetches the necessary files respecting a "relaxed" dirtab, which includes not only the actual positive paths but their parents

PR #1177, opened 2015-10-20
Preload packaging

Moliholy:
This PR ships the cvmfs_preload binary in a self-contained shell script (cvmfs_preload.sh) which includes that binary plus its dependencies as shared libraries. Concretely the dependencies are:
- tbb_cvmfs
- tbbmalloc_cvmfs
- openssl + crypto
- a few openssl's dependencies

Tested in a new fresh centos and ubuntu docker containers


jblomer:
Couple of things:

The `cvmfs_preload` shell script should be what gets installed on the system.  No need for an .sh ending.  When I run it, I get

```
# cvmfs/cvmfs_preload.sh 
cvmfs_preload
libtbbmalloc_cvmfs.so.2
libtbb_cvmfs.so.2
libssl.so
libcrypto.so
libgssapi_krb5.so
libkrb5.so
libk5crypto.so
libkrb5support.so
libkeyutils.so
Not enough arguments: 1

Usage:
...
```

So I guess the untaring should not be verbose.  `LD_LIBRARY_PATH` set once to the temp directory allows you to avoid listing all the libraries in `LD_PRELOAD`.

How did you decide which library to include and which one to take from the system?

Moliholy:
Basically two of them:
- TBB and TBBmalloc because of what was mentioned in https://github.com/cvmfs/cvmfs/pull/1148
- OpenSSL because it would not be likely to have it in some systems. This one, however, has a lot of dependencies and sub-dependencies that are packaged as well.

The rest are assumed to be in the system. Just to recall, here's a list of them:
````
	linux-vdso.so.1 =>  (0x00007ffe58a84000)
	libssl.so.10 => /usr/lib64/libssl.so.10 (0x00000039db600000)
	libcrypto.so.10 => /usr/lib64/libcrypto.so.10 (0x00000039daa00000)
	libtbb_cvmfs.so.2 => /usr/lib64/libtbb_cvmfs.so.2 (0x000000301c600000)
	libtbbmalloc_cvmfs.so.2 => /usr/lib64/libtbbmalloc_cvmfs.so.2 (0x000000301ca00000)
	librt.so.1 => /lib64/librt.so.1 (0x00000039d3200000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00000039d2a00000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00000039d2e00000)
	libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x00000039d5200000)
	libm.so.6 => /lib64/libm.so.6 (0x00000039d3600000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00000039d4e00000)
	libc.so.6 => /lib64/libc.so.6 (0x00000039d2600000)
	libgssapi_krb5.so.2 => /lib64/libgssapi_krb5.so.2 (0x00000039da600000)
	libkrb5.so.3 => /lib64/libkrb5.so.3 (0x00000039d9600000)
	libcom_err.so.2 => /lib64/libcom_err.so.2 (0x00000039d7e00000)
	libk5crypto.so.3 => /lib64/libk5crypto.so.3 (0x00000039d8e00000)
	libz.so.1 => /lib64/libz.so.1 (0x00000039d3a00000)
	/lib64/ld-linux-x86-64.so.2 (0x00000039d2200000)
	libkrb5support.so.0 => /lib64/libkrb5support.so.0 (0x00000039d9200000)
	libkeyutils.so.1 => /lib64/libkeyutils.so.1 (0x00000039d9e00000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00000039d4600000)
	libselinux.so.1 => /lib64/libselinux.so.1 (0x00000039d3e00000)
```

Moliholy:
If you think one of these libraries might not be in a concrete system I can quickly add it

Moliholy:
In my case the find_library command is unable to locate linux-vdso, gcc_s and stdc++. Compiling in SLC6. Any clues?

Moliholy:
cvmfs_preload built using a different shell script.

cvmfs/CMakeLists.txt
```diff
@@ -623,8 +634,8 @@ endif (BUILD_SERVER)
 
 if (BUILD_PRELOADER)
   install(
-    TARGETS      cvmfs_preload
-    RUNTIME
+    FILES      ${CMAKE_CURRENT_BINARY_DIR}/cvmfs_preload
```
Moliholy: Just saying, maybe it's not a good idea to install this binary?

jblomer: I agree, it's debatable.  Let's leave it as it is for the time being.

Moliholy: There is an integration test already that uses the installed version, by the way

PR #1179, opened 2015-10-21
Refactor: Catalog Migration only Commits Changed Catalogs

reneme:
This changes `cvmfs_swissknife migrate` to upload nested catalogs only if they have been changed during the migration step. The rationale being that OverlayFS will need a catalog migration when a previous AUFS-based repository is imported.

To figure out if a catalog database has been modified, I've added `Database<>::GetModifiedRowCount()` along with a unit test case that figures out how many databases rows have been changed since the SQLite database has been opened.

Furthermore, the migration code of `cvmfs_swissknife` needs to step over unmodified catalogs which required an adaption in the concurrent processing. In detail, I replaced two individual `Future<>`s in `PendingCatalog` by a new one (cf. `Future<bool> was_modified`) that acts both as a modification flag and a signal for finished processing.

### DIFF
--- a/cvmfs/sql.h
+++ b/cvmfs/sql.h
@@ -124,6 +124,13 @@ class Database : SingleCopy {
   bool                read_write()      const { return read_write_;           }
 
   /**
+   * Provides the number of rows modified by INSERT, UPDATE or DELETE statements
+   * that have been run against this database since it was opened.
+   * @return  number of rows modified by all executed manipulating statements
+   */
+  unsigned GetModifiedRowCount() const;
+
+  /**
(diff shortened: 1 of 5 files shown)


reneme:
amended with some minor fixes found during integration testing.

jblomer:
Perhaps I'm paranoid but can you add an integration test that has changes only in a nested catalog?  I guess that would be part of a follow-up PR?

reneme:
> Perhaps I'm paranoid but can you add an integration test that has changes only in a nested catalog? I guess that would be part of a follow-up PR?

I'm just as paranoid... working on opening pull request right now. ;-)

PR #1180, opened 2015-10-21
add CVMFS_SERVER_CACHE_MODE parameter, currently works around rename …

jblomer:
…in the cvmfs cache

From the client side, that should be all. @reneme can you take over from here?


reneme:
Looks good to me. I'll integrate it into the server code tomorrow. If the entire test suite on a deadlock-vulnerable configuration doesn't trigger a deadlock we should (hopefully) be fine with removing the workaround code.

PR #1185, opened 2015-10-22
FIX: Enable Server Cache Mode (link/unlink)

reneme:
This enables the `CVMFS_SERVER_CACHE_MODE` for Stratum0 repositories. I wonder if we should anyway keep the kernel deadlock warnings in place, or just hope that people never fiddle with this parameter...


reneme:
Remark: Migration code comes in another pull request.

jblomer:
I think we can take out the warnings.  There are lots of other options to screw up by modifying the client options...

PR #1192, opened 2015-10-22
FIX: Use lsof instead of fuser for Busy Mountpoints

reneme:
This replaces `fuser` by `lsof` to print a list of processes occupying a mountpoint. The rationale being that `fuser` (for some reason) doesn't print usable information on OverlayFS (kernel 4.2.0), i.e. it claims that the mountpoint is occupied by the kernel itself. O.o

This fixes integration test case 570 for the bleeding edge OverlayFS on Ubuntu 15.10.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -94,13 +94,6 @@ is_systemd() {
   [ x"$SERVICE_BIN" = x"false" ]
 }
 
-# Find the fuser binary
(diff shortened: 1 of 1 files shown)


DrDaveD:
Be sure to include lsof as a required dependency in the packaging, I don't think it is always there.

reneme:
Actually we've been depending on `lsof` before. But I'll double-check in the packages, thanks.

PR #1199, opened 2015-10-26
Packaging: cvmfs-devel on Ubuntu (+pkgmap update)

reneme:
This adds the `cvmfs-devel` package to the debian build and updates the `pkgmap` files accordingly.


jblomer:
should we rather call it cvmfs-dev on debian to stay consistent with other packages?

reneme:
> should we rather call it cvmfs-dev on debian to stay consistent with other packages?

[Sounds reasonable](https://github.com/reneme/cvmfs/commit/56fc53cee67d70939def6545d3324af3c46f9126).

PR #1205, opened 2015-11-02
Add support for SHA3-256 and SHAKE128 with 160 output bits

jblomer:
Adds two new cryptographic hashes: SHA3-256 and SHAKE-128, both based on the [Keccak code package](https://github.com/gvanas/KeccakCodePackage) from the Keccak designers.  SHA3-256 is a drop-in replacement for SHA256 (according to NIST) but faster.  In the longer run, I'd make SHAKE-128 with 160 output bits the default.  This provides 80 security bits where due to various attacks I'm not exactly sure how many security bits we still get from SHA-1.  Having only 160 output bits has the advantage of not bloating the file catalogs.

**Note**: while working on the unit tests for Keccak, I discovered that the sha2 implementation we used so far was broken.  There seemed to be an overflow of some kind, hashing extremely long messages gave unexpected outputs.  This PR switches to the sha2 implementation from mbedtls, which from all I can tell through testing works correctly.  Luckily, the SHA256 support was not yet officially released.

**Note**: this adds the `swissknife hash` command which simplifies the `cvmfs_server` script a little: less parsing of openssl outputs.


cvmfs/cvmfs_server
```diff
@@ -1523,24 +1523,11 @@ create_whitelist() {
   echo `date -u "+%Y%m%d%H%M%S"` > ${whitelist}.unsigned
   echo "E`date -u --date='+30 days' "+%Y%m%d%H%M%S"`" >> ${whitelist}.unsigned
   echo "N$name" >> ${whitelist}.unsigned
```
reneme: There is a very similar implementation of this functionality [in the test functions](https://github.com/cvmfs/cvmfs/blob/devel/test/test_functions#L1206). Perhaps it's a good idea to change this piece of code as well to avoid confusion down the road.

reneme:
Potentially a bit of a mean question: if SHA2 was not released by us yet, do we actually need SHA2 or  could we just rely solely on SHA3 right away?

cvmfs/hash.h
```diff
@@ -54,7 +57,7 @@ const char kSuffixCertificate  = 'X';
  * Corresponds to Algorithms.  "Any" is the maximum of all the other
  * digest sizes.
  */
-const unsigned kDigestSizes[] = {16, 20, 20, 32, 32};
+const unsigned kDigestSizes[] = {16, 20, 20, 32, 32, 20, 32};
```
reneme: I have to feeling that this gets a bit out of hand. With seven supported hash algorithms I think this starts to reach the boundary of unsupportability. Hard to say on the spot what would be a better solution, I'll have a look tomorrow.

cvmfs/swissknife_hash.cc
```diff
@@ -0,0 +1,36 @@
...
+#include <sys/types.h>
+#include <sys/uio.h>
+#include <unistd.h>
+
+#include <cstdio>
+
+#include "hash.h"
+#include "logging.h"
+
+/* compress or decompress from stdin to stdout */
```
reneme: Copy and paste bug here?

cvmfs/swissknife_hash.cc
```diff
@@ -0,0 +1,36 @@
...
+int swissknife::CommandHash::Main(const swissknife::ArgumentList &args) {
+  bool fingerprint = (args.find('f') != args.end());
+  shash::Any hash(shash::ParseHashAlgorithm(*args.find('a')->second));
+
+  shash::ContextPtr context_ptr(hash.algorithm);
+  context_ptr.buffer = smalloc(context_ptr.size);
+
+  unsigned char buf[4096];
```
reneme: The Magic number should be a constant!

cvmfs/hash.h
```diff
@@ -62,14 +65,15 @@ const unsigned kMaxDigestSize = 32;
  * for backwards compatibility MD5 and SHA-1 have none.
  */
 extern const char *kAlgorithmIds[];
-// in hash.cc: const char *kAlgorithmIds[] = {"", "", "-rmd160", "-sha256", ""};
-const unsigned kAlgorithmIdSizes[] = {0, 0, 7, 7, 0};
-const unsigned kMaxAlgorithmIdentifierSize = 7;
+// in hash.cc: const char *kAlgorithmIds[] = {"", "", "-rmd160", "-sha256", ...
+const unsigned kAlgorithmIdSizes[] = {0, 0, 7, 7, 5, 9, 0};
+const unsigned kMaxAlgorithmIdentifierSize = 9;
```
reneme: C++11/14: not exactly sure about that, but could that be determined with constexpr in C++14 at compile time?

jblomer:
I though about removing sha3-256 and sha256.  We would save 12 bytes for every shash::Any object.  On the other hand, now that we technically can do it, I think it's better to give the user a choice on the trade-off of security and performance.  It's hard to decide how trustworthy sha3 really is since it is rather new.

jblomer:
Actually, it might also be helpful for grafting of files to support the most common hash functions.  This way, files don't need to be rehashed.

cvmfs/cvmfs_server
jblomer: [yep](https://github.com/cvmfs/cvmfs/commit/ac5e75640ca360cb820c46a950fb3c9542d97c26)

cvmfs/swissknife_hash.cc
jblomer: [yep](https://github.com/cvmfs/cvmfs/commit/12fe35c8be0e1c6aa47bd48e38b60a385e1ba9de)

cvmfs/cvmfs_server
jblomer: [yep](https://github.com/cvmfs/cvmfs/commit/ac5e75640ca360cb820c46a950fb3c9542d97c26)

PR #1227, opened 2015-11-04
Merge with latest devel, send certificate always to the native location (in addition to the alt path)

jblomer:
This is a mini pull request, mostly to test the workflow.

@bbockelm two items on my list
- The manifest class should hand out native and alternative paths
- There is probably no need for an additional binary cvmfs_cred_fetcher but this can be the cvmfs2 binary with magic calling conventions (`__CRED_FETCHER__`).  We do the same for the quota manager.


jblomer:
Ah, regarding the workflow: I would leave pull requests to the CVM-904 branch open and wait for comments.

cvmfs/swissknife_sign.cc
```diff
@@ -141,9 +141,12 @@ int swissknife::CommandSign::Main(const swissknife::ArgumentList &args) {
       spooler->RegisterListener(&CommandSign::CertificateUploadCallback, this);
 
     // Safe certificate (and wait for the upload through a Future)
+    spooler->ProcessCertificate(certificate, "");
     std::string alt_path = manifest->alt_catalog_path();
-    if (alt_path.size()) {alt_path += "cert";}
-    spooler->ProcessCertificate(certificate, alt_path);
+    if (alt_path.size()) {
+      alt_path += "cert";
+      spooler->ProcessCertificate(certificate, alt_path);
```
bbockelm: Are we sure it's safe to call "ProcessCertificate" twice?

bbockelm:
Switching to the alternate calling conventions sounds fine to me too.

cvmfs/swissknife_sign.cc
jblomer: Should be fine, `ProcessCertificate` is a convenience wrapper for `ProcessFile` that adds the 'X' suffix to the file name.

bbockelm:
@jblomer - can you merge this?

PR #1229, opened 2015-11-08
Allow to install CernVM-FS outside of /usr

reneme:
This allows to install all artefacts of the CernVM-FS client into other locations than `/usr`, `/etc` and `/sbin` as a first step to conform with SIP on OS X El Capitan. Furthermore `brew` requires to install without `sudo` into `/usr/local`.

With that and `CMAKE_INSTALL_PREFIX=/usr/local` I can create private mounts of LHC repositories on El Capitan.

Note: This build on top of [FIX: OS X El Capitan Build](https://github.com/cvmfs/cvmfs/pull/1228).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,20 +28,6 @@ set (CernVM-FS_VERSION_STRING "${CernVM-FS_VERSION_MAJOR}.${CernVM-FS_VERSION_MI
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
 
 #
-# Set install prefix to /usr.  Cvmfs is not relocatable.
-#
-if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-  SET(CMAKE_INSTALL_PREFIX
-    "/usr" CACHE PATH "/usr install prefix" FORCE
-  )
-else (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-  if (NOT ${CMAKE_INSTALL_PREFIX} STREQUAL "/usr")
-    Message("Warning: CernVM-FS is not relotable and expects to be installed under /usr")
-  endif ()
-endif (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-
-
-#
 # detect the operating system and the distribution we are compiling on
 #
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -28,20 +28,6 @@ set (CernVM-FS_VERSION_STRING "${CernVM-FS_VERSION_MAJOR}.${CernVM-FS_VERSION_MI
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
 
 #
-# Set install prefix to /usr.  Cvmfs is not relocatable.
```
jblomer: At this point, shouldn't we just remove the `else` branch?  So, the default should still be /usr (perhaps with a different default on OS X)

jblomer:
`cvmfs_config chksetup` looks for installed files and needs to be adapted, too.  It might also be necessary to look into the search paths for loading libcvmfs_fuse in `loader.cc`

jblomer:
Actually, only on El Capitan and later should the default install path point outside /usr.  Otherwise Mac users might update an installation in /usr with one in /usr/local which leads to funny results.

reneme:
This pull request was anyway meant as a preview... I'm just 'making it work' right now. Anyway: I believe, we should make the install location configurable both in the build/install scripts and the production code. Then we are flexible how and where to install it. Setting the defaults depending on the platforms makes sense, of course.

I'm wondering how to properly find the configuration files that might reside in either `/etc` or `/usr/local/etc`. We could make this implicit, by searching for config files based on the binary install location. What do you think?

Also we need to carefully check for other absolute paths. So there is `loader.cc`, `cvmfs_config` and (I suppose) `cvmfs_talk`? More 'obvious' locations to check?

Perhaps let's discuss this via Skype tomorrow?

reneme:
Making `mount -t cvmfs atlas.cern.ch /cvmfs/atlas.cern.ch` work with caveats:

* `sudo mkdir -p  /Library/Filesystems/cvmfs.fs/Contents/Resources`
* `sudo ln -s /usr/local/sbin/mount_cvmfs /Library/Filesystems/cvmfs.fs/Contents/Resources/mount_cvmfs`
* `sudo ln -s /usr/local/etc/cvmfs /etc/cvmfs`

reneme:
Fixing caveat `sudo ln -s /usr/local/etc/cvmfs /etc/cvmfs` by doing this opportunistically in `cvmfs_config setup`.

PR #1232, opened 2015-11-09
Provide tool to create grafts

bbockelm:
This provides `cvmfs_swissknife graft`, a simple tool for calculating graft files.

### DIFF
--- a/cvmfs/swissknife_graft.cc
+++ b/cvmfs/swissknife_graft.cc
@@ -4,146 +4,91 @@
  * Process a set of input files and create appropriate graft files.
  */
 
+#include "swissknife_graft.h"
+
 #include <fcntl.h>
 #include <unistd.h>
-#include <sys/stat.h>
 
 #include <cstdio>
-#include <iostream>
 #include <vector>
-#include <sstream>
 
 #include "compression.h"
 #include "fs_traversal.h"
 #include "hash.h"
-
-#include "swissknife_graft.h"
-
-#define CHUNK_SIZE 16*1024
-
-int swissknife::CommandGraft::ChecksumFd(int fd, shash::Any &id, off_t &processed_size) {  // NOLINT
-  shash::ContextPtr hash_context(id.algorithm);
-  hash_context.buffer = alloca(hash_context.size);
-  shash::Init(hash_context);
-
-  z_stream strm;
-  zlib::CompressInit(&strm);
-  zlib::StreamStates retval;
-
-  std::vector<unsigned char> buf; buf.reserve(CHUNK_SIZE);
-  off_t cksum_size = 0;
-  bool eof;
-
-  do {
-    ssize_t nbytes = read(fd, &buf[0], CHUNK_SIZE);
-    if (nbytes < 0) {
-      if (errno == EINTR) {continue;}
-      zlib::CompressFini(&strm);
-      return nbytes;
-    }
-    cksum_size += nbytes;
-    eof = nbytes == 0;
-    retval = zlib::CompressZStream2Null(&buf[0], nbytes, eof, &strm,
-                                        &hash_context);
-    if (retval == zlib::kStreamDataError) {
-      zlib::CompressFini(&strm);
-      errno = EINVAL;
-      return -1;
-    }
-  } while (!eof);
-
-  zlib::CompressFini(&strm);
-  if (retval != zlib::kStreamEnd) {
-    errno = EINVAL;
-    return -1;
-  }
-  shash::Final(hash_context, &id);
-  processed_size = cksum_size;
-  return 0;
-}
-
-
-static void splitpath(const std::string &path,
-                      std::string &out_dirname,
-                      std::string &out_fname) {  // NOLINT
-  size_t dir_sep = path.rfind('/');
-  if (dir_sep != std::string::npos) {
-    out_dirname = path.substr(0, dir_sep+1);
-    out_fname = path.substr(dir_sep+1);
-  } else {
-    out_dirname = ".";
-    out_fname = path;
-  }
-}
+#include "platform.h"
 
 
 bool swissknife::CommandGraft::DirCallback(const std::string &relative_path,
                                            const std::string &dir_name) {
-  if (!m_output_file.size()) {return true;}
-  std::string full_path = m_output_file + "/" + 
-                          (relative_path.size() ? relative_path : ".") + "/" + dir_name;
-  if (-1 == mkdir(full_path.c_str(), 0755)) {
-    if (errno == EEXIST) {
-      struct stat sbuf;
-      if ((stat(full_path.c_str(), &sbuf) == 0) && (S_ISDIR(sbuf.st_mode))) {
-        return true;
-      }
-    }
+  if (!m_output_file_.size()) {return true;}
+  std::string full_output_path = m_output_file_ + "/" +
+                          (relative_path.size() ? relative_path : ".") +
+                          "/" + dir_name;
+  std::string full_input_path = m_input_file_ + "/" +
+                                (relative_path.size() ? relative_path : ".") +
+                                "/" + dir_name;
+  platform_stat64 sbuf;
+  if (-1 == platform_stat(full_input_path.c_str(), &sbuf)) {
     return false;
   }
-  return true;
+  return MkdirDeep(full_output_path, sbuf.st_mode);
 }
 
 
 void swissknife::CommandGraft::FileCallback(const std::string &relative_path,
                                             const std::string &file_name) {
-  std::string full_input_path = m_input_file + "/" +
-                                (relative_path.size() ? relative_path : ".") + "/" + file_name;
+  std::string full_input_path = m_input_file_ + "/" +
+                                (relative_path.size() ? relative_path : ".") +
+                                "/" + file_name;
   std::string full_output_path;
-  if (m_output_file.size()) {
-    full_output_path = m_output_file + "/" +
-                       (relative_path.size() ? relative_path : ".") + "/" + file_name;
+  if (m_output_file_.size()) {
+    full_output_path = m_output_file_ + "/" +
+                       (relative_path.size() ? relative_path : ".") +
+                       "/" + file_name;
   }
   Publish(full_input_path, full_output_path, false, false);
 }
 
 
-int swissknife::CommandGraft::Recurse(const std::string &input_file, const std::string &output_file) {
-  m_output_file = output_file;
-  m_input_file = input_file;
+int swissknife::CommandGraft::Recurse(const std::string &input_file,
+                                      const std::string &output_file) {
+  m_output_file_ = output_file;
+  m_input_file_ = input_file;
 
   FileSystemTraversal<CommandGraft> traverser(this, input_file, true);
   traverser.fn_new_file       = &CommandGraft::FileCallback;
   traverser.fn_new_dir_prefix = &CommandGraft::DirCallback;
   traverser.Recurse(input_file);
-  return 1;
+  return 0;
 }
 
 
 int swissknife::CommandGraft::Main(const swissknife::ArgumentList &args) {
   const std::string &input_file = *args.find('i')->second;
-  const std::string output_file = (args.find('o') == args.end()) ? "" : *(args.find('o')->second);
-  m_verbose = args.find('v') != args.end();
+  const std::string output_file =
+      (args.find('o') == args.end()) ? "" : *(args.find('o')->second);
+  m_verbose_ = args.find('v') != args.end();
 
-  struct stat sbuf;
+  platform_stat64 sbuf;
   bool output_file_is_dir = output_file.size() &&
-                            (0 == stat(output_file.c_str(), &sbuf)) &&
+                            (0 == platform_stat(output_file.c_str(), &sbuf)) &&
                             S_ISDIR(sbuf.st_mode);
   if (output_file_is_dir && (input_file == "-")) {
-    std::cerr << "Output file (" << output_file << "): Is a directory" << std::endl;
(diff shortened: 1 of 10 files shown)


cvmfs/swissknife_graft.h
```diff
@@ -0,0 +1,51 @@
...
+  int Recurse(const std::string &input_file, const std::string &output_file);
+  int ChecksumFd(int fd, shash::Any &id, off_t &processed_size);
+
+ private:
+  void FileCallback(const std::string &relative_path,
+                    const std::string &file_name);
+  bool DirCallback(const std::string &relative_path,
+                   const std::string &dir_name);
+
+  std::string m_output_file;
```
jblomer: Naming of member variables with a trailing underscore, like `output_file_`

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+ *
+ * Process a set of input files and create appropriate graft files.
+ */
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
+
+#include <cstdio>
+#include <iostream>
```
jblomer: Please avoid `iostream` and `sstream`.  I think it can be replace throughout the code by string concatenation and the `Stringify...` functions from util.cc

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+#include <vector>
+#include <sstream>
+
+#include "compression.h"
+#include "fs_traversal.h"
+#include "hash.h"
+
+#include "swissknife_graft.h"
```
jblomer: should be at the top of includes

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+
+#define CHUNK_SIZE 16*1024
```
jblomer: This would be a chunk for every 16kB, right?  I think that's very small.  Native files chunk as of a few MB.

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+  return 1;
+}
+
+
+int swissknife::CommandGraft::Main(const swissknife::ArgumentList &args) {
+  const std::string &input_file = *args.find('i')->second;
+  const std::string output_file = (args.find('o') == args.end()) ? "" : *(args.find('o')->second);
+  m_verbose = args.find('v') != args.end();
+
+  struct stat sbuf;
```
jblomer: `struct stat` should be replace by `struct platform_stat64` from platform.h

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+  bool output_file_is_dir = output_file.size() &&
+                            (0 == stat(output_file.c_str(), &sbuf)) &&
```
jblomer: --> `platform_stat()`

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+                            S_ISDIR(sbuf.st_mode);
+  if (output_file_is_dir && (input_file == "-")) {
+    std::cerr << "Output file (" << output_file << "): Is a directory" << std::endl;
+    return 1;
+  }
+
+  if (input_file != "-") {
+    bool input_file_is_dir = (0 == stat(input_file.c_str(), &sbuf)) &&
```
jblomer: --> `platform_stat()`

cvmfs/swissknife_graft.h
```diff
@@ -0,0 +1,51 @@
...
+  ParameterList GetParams() {
+    ParameterList r;
+    r.push_back(Parameter::Mandatory('i', "Input file to process ('-' for reading"
+                                       " from stdin)"));
+    r.push_back(Parameter::Optional('o', "Output location for graft file"));
+    r.push_back(Parameter::Switch('v', "Verbose output"));
+    return r;
+  }
+
+  int Main(const ArgumentList &args);
```
jblomer: Everything except `Main()` can be private, I think

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+  } else {
+    fd = open(input_file.c_str(), O_RDONLY, 0755);
+    if (fd < 0) {
+      std::stringstream ss; ss << "Unable to open input file (" << input_file << ")";
+      perror(ss.str().c_str());
+      return 1;
+    }
+  }
+
+  shash::Any hash(shash::kSha1);
```
jblomer: It would be nice to have an option to let the user select the hash algorithm

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+  if (output_file.size() && m_verbose) {
+    std::cout << "Grafting " << input_file << " to " << output_file << std::endl;
+  } else if (!output_file.size()) {
+    std::cout << "Grafting " << input_file << std::endl;
+  }
+  int fd;
+  if (input_file_is_stdin) {
+    fd = 0;
+  } else {
+    fd = open(input_file.c_str(), O_RDONLY, 0755);
```
jblomer: With `O_RDONLY`, I thought a mode does not need to be provided

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+    errno = EINVAL;
+    return -1;
+  }
+  shash::Final(hash_context, &id);
+  processed_size = cksum_size;
+  return 0;
+}
+
+
+static void splitpath(const std::string &path,
```
jblomer: There is `GetFileName()` and `GetParentName()` available in util.cc.

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+    std::string dirname, fname;
+    std::string graft_fname;
+    if (output_file_is_dir) {
+      splitpath(input_file, dirname, fname);
+      graft_fname =  output_file + "/.cvmfsgraft-" + fname;
+    } else {
+      splitpath(output_file, dirname, fname);
+      graft_fname = dirname + "/.cvmfsgraft-" + fname;
+    }
+    fd = open(graft_fname.c_str(), O_CREAT|O_TRUNC|O_WRONLY, 0755);
```
jblomer: should be 0644 as a file mode.

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+    if (fd < 0) {
+      if ((errno == EISDIR) && !input_file_is_stdin) {
```
jblomer: I'm not quite sure in which situation this condition can be true

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+        splitpath(input_file, dirname, fname);
+        graft_fname = output_file + "/.cvmfsgraft-" + fname;
+        fd = open(graft_fname.c_str(), O_CREAT|O_TRUNC|O_WRONLY, 0755);
```
jblomer: 0755 --> 0644

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+  } else {
+    fd = 1;
+  }
+  std::stringstream ss;
+  ss << "size=" << processed_size << std::endl;
+  ss << "checksum=" << hash.ToString(true) << std::endl;
+  std::string graft_contents = ss.str();
+  size_t nbytes = graft_contents.size();
+  const char *buf = graft_contents.c_str();
+  while (nbytes) {
```
jblomer: Using `fdopen()`, this can be transformed in `CopyMem2File()` from compression.h

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+  // Create and truncate the output file.
+  std::string output_fname;
+  if (output_file_is_dir) {
+    std::string dirname, fname;
+    splitpath(input_file, dirname, fname);
+    output_fname = output_file + "/" + fname;
+  } else {
+    output_fname = output_file;
+  }
+  fd = open(output_fname.c_str(), O_CREAT|O_TRUNC|O_WRONLY, 0755);
```
jblomer: Perhaps file mode and file owner should be taken from the input file.

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+
+int swissknife::CommandGraft::Recurse(const std::string &input_file, const std::string &output_file) {
+  m_output_file = output_file;
+  m_input_file = input_file;
+
+  FileSystemTraversal<CommandGraft> traverser(this, input_file, true);
+  traverser.fn_new_file       = &CommandGraft::FileCallback;
+  traverser.fn_new_dir_prefix = &CommandGraft::DirCallback;
+  traverser.Recurse(input_file);
+  return 1;
```
jblomer: This way `Recurse()` would always return an error code.  Is this intended?

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+  }
+}
+
+
+bool swissknife::CommandGraft::DirCallback(const std::string &relative_path,
+                                           const std::string &dir_name) {
+  if (!m_output_file.size()) {return true;}
+  std::string full_path = m_output_file + "/" + 
+                          (relative_path.size() ? relative_path : ".") + "/" + dir_name;
+  if (-1 == mkdir(full_path.c_str(), 0755)) {
```
jblomer: Might be better to take the mode and owner from the graft source

jblomer: There is `MkdirDeep()` in util.h

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+    if (errno == EEXIST) {
+      struct stat sbuf;
+      if ((stat(full_path.c_str(), &sbuf) == 0) && (S_ISDIR(sbuf.st_mode))) {
```
jblomer: --> `platform_stat()`

cvmfs/swissknife_graft.cc
```diff
@@ -0,0 +1,249 @@
...
+
+#include "compression.h"
+#include "fs_traversal.h"
+#include "hash.h"
+
+#include "swissknife_graft.h"
+
+#define CHUNK_SIZE 16*1024
+
+int swissknife::CommandGraft::ChecksumFd(int fd, shash::Any &id, off_t &processed_size) {  // NOLINT
```
jblomer: `id` and `processed_size` should be pointers

jblomer: Could this be replaced by `CompressFd2Null()` from compression.h?

jblomer: Ah, I see that the processed size is needed as well.  I think it would be nicer to extend `CompressFd2Null()` with an `uint64_t` output pointer that defaults to `NULL`.

jblomer:
Many thanks!

Besides the small comments, I think we'd need
  - rebased to the latest devel
  - integration test
  - Output should be handled by `LogCvmfs()`

cvmfs/swissknife_graft.cc
bbockelm: I'd like to hold off on that until we get the uncompressed files in (and do both compression and hash flag at once).

cvmfs/swissknife_graft.cc
bbockelm: I'd like to have a single function which pulls out both parts of the path.  I'll move this to `util.cc`.

cvmfs/swissknife_graft.cc
bbockelm: Yes - this is no longer a necessary branch after a refactor.  Good catch.

cvmfs/swissknife_graft.cc
bbockelm: I'm not comfortable with `CopyMem2File` as is because I cannot determine whether it is safe under EINTR.

cvmfs/swissknife_graft.cc
bbockelm: Good idea, I'll add that.

cvmfs/swissknife_graft.cc
jblomer: You mentioned this point several times. I think we largely ignored `EINTR` in the code so far.  This might have been a mistake.  Is `EINTR` something that actually happens in practice?

In any case, I'd prefer fixing `CopyMem2File` and other spots.  We could, for instance, add in util.cc `SafeRead` and `SafeWrite` wrappers for `read` and `write`.  Just not sure how to unit test this.

cvmfs/swissknife_graft.cc
bbockelm: Yeah, we can move this with some small tweaks to `CompressFd2Null()`

bbockelm: Oh - and `CompressFd2Null()` has a nasty bug: `result` is initialized to -1 (for the error case) but the function return is a `bool`.

Hence, it'll return `true` on either success or failure.  I'll fix that too.

jblomer: Thank!  I guess I need to add a few unit tests...

bbockelm:
I rebased to latest devel and tackled most of the code review comments.

Working out improving output handling and integration test.

cvmfs/swissknife_graft.h
bbockelm: Fixed.

bbockelm:
Alright, I think I've addressed all the comments.

jblomer:
Replaced by #1265

PR #1237, opened 2015-11-10
FIX: OS X El Capitan

reneme:
This (technically) fixes support for OS X El Capitan with CernVM-FS being installed into `/usr/local` to coexist with Apple's [SIP](https://support.apple.com/en-us/HT204899). For that, some parts of the CernVM-FS client need to become slightly relocatable.

Fixed issues:
* `cvmfs_config` script is aware of it's installation location and takes it into account when looking for other CernVM-FS components
* internal platform abstraction can determine operating system version (OS X version on mac vs. kernel version on Linux)
* mount helper additionally looks into `/usr/local/bin` for `cvmfs2` binary when running on OS X >= 10.11
* `cvmfs_config` can initially load `config.sh` both from `/etc/cvmfs/` and `/usr/local/etc/cvmfs` for bootstrapping the system with `cvmfs_config setup`
* `cvmfs_config setup` symlinks both `/usr/local/etc/cvmfs` and `/usr/local/sbin/mount_cvmfs` into appropriate places on Mac OS X >= 10.11
* `cvmfs_config` checks for non-emptiness of `CVMFS_USER` as an additonal sanity check

*Note:* this builds on top of [Allowing to Install CVMFS outside of /usr](https://github.com/cvmfs/cvmfs/pull/1236).

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -7,6 +7,8 @@ SBIN_BASE="$( [ x"$INSTALL_BASE" = x"/usr" ] && echo "/sbin" || echo "${INSTALL_
 
 if [ -f /etc/cvmfs/config.sh ]; then
   . /etc/cvmfs/config.sh
+elif [ -f ${INSTALL_BASE}/etc/cvmfs/config.sh ]; then
+  . ${INSTALL_BASE}/etc/cvmfs/config.sh
 else
   echo "/etc/cvmfs/config.sh missing"
   exit 1
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -192,6 +192,15 @@ cvmfs_setup() {
   mkdir -p /var/lib/cvmfs
   chown cvmfs:cvmfs /var/lib/cvmfs
 
+  # symlink configuration into /etc on OS X (>= 10.11)
+  if [ x"$sys_arch" = x"Darwin" ] && \
+     compare_versions $(sw_vers -productVersion) -ge "10.11"; then
+    if [ ! -e /etc/cvmfs           ] && \
+       [   -e /usr/local/etc/cvmfs ]; then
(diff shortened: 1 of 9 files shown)


reneme:
*PS:* This supersedes a [previous pull request](https://github.com/cvmfs/cvmfs/pull/1229) to disentangle build system changes and OS X El Capitan fixes.

cvmfs/cvmfs_config
```diff
@@ -245,6 +312,12 @@ cvmfs_chksetup() {
     fi
   done
 
+  # Check that CVMFS_USER is set
+  if [ x"$CVMFS_USER" = x"" ]; then
```
jblomer: Shouldn't this be handled by the `required_list` of variables?

reneme: In principle yes. However, I was getting unclear follow-up errors when `CVMFS_USER` was not set but used before the `required_list` was actually checked.

jblomer: I see.  Can you recover what went wrong?  Perhaps we can just move the check for required parameters to an earlier place.

reneme: I've been running the script without having `/usr/local/etc/cvmfs` symlinked into `/etc/cvmfs` and added this check for convenience. Its not strictly necessary but I find it doesn't hurt either.

jblomer: I agree, it doesn't hurt.  I'm just wondering if we will remember after some time why it is there.  Wouldn't the more accurate thing to test be if <ETCDIR>/cvmfs/default.conf exists?

reneme: Not sure. The root cause for this addition is that `$CVMFS_USER` was used before it was checked for existence. Personally I find this check more understandable than checking for `default.conf` where I have no detailed idea what is inside there.

jblomer: Ah, now I understand.  It is just just a few lines down the script.  Thanks!

PR #1249, opened 2015-11-23
FIX T_Util.RfcTimestamp

Moliholy:
Now the comparison is made within a maximum of two seconds between both values


jblomer:
Please have a second careful look at this.  I don't think it's correct.

Moliholy:
True, the value is seconds, not milliseconds. Now it should be fixed

PR #1252, opened 2015-11-24
add cvmfs_rsync to cvmfs-server package

DrDaveD:
As described in [CVM-814](https://sft.its.cern.ch/jira/browse/CVM-814).

Requires that rsync be installed on the test machines.   How do I ensure that?

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -636,6 +636,12 @@ if (BUILD_SERVER)
     PERMISSIONS    OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ
   )
 
+  install (
+    FILES      cvmfs_rsync
+    DESTINATION    bin
+    PERMISSIONS    OWNER_READ OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
+  )
+
   install(
     FILES      webapi/cvmfs-api.wsgi
     DESTINATION    "/var/www/wsgi-scripts"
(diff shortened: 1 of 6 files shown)


jblomer:
Thanks!

It looks mostly good to me.  There should be a new dependency on `rsync` for the cvmfs-server packages.  I think rsync needs to be installed on the integration test VMs, too (under test/ cloud_testing/platforms/...).

For the integration test, please transfer the ownership of files to the test user (sftnight) after rsyncing.  I run the tests as root in a VM, in which case it fails.  Furthermore, the integration tests currently fails at

    [ $(get_catalog_count $repo_name) -eq 8 ] || return 101

@reneme any other comments?

DrDaveD:
> There should be a new dependency on rsync for the cvmfs-server packages.

Ok I added that.

> I think rsync needs to be installed on the integration test VMs, too (under test/cloud_testing/platforms/...).

If it is a dependency of the cvmfs-server package, won't it automatically be there?

> For the integration test, please transfer the ownership of files to the test user (sftnight) after rsyncing. I run the tests as root in a VM, in which case it fails.

Ah, sorry, I hadn't realized that it was run as root.  I thought it was sudo'ing to root when needed.  I now made it sudo -u $CVMFS_TEST_USER whenever writing into the repo.

> Furthermore, the integration tests currently fails 

It works for me now running as root.  If it still fails for you, please send me the whole log output.

DrDaveD:
It occurred to me that an algorithm based on .cvmfsdirtab would not find all .cvmfsautocatalog files, so I thought of instead using cvmfs_server list-catalogs.  That turns out to in fact be a lot simpler as well.

I also updated the tests to mix up the use of trailing slash or not on the source and destination directories, and to also check for .cvmfscatalog files getting recreated after getting erroneously removed.

PR #1256, opened 2015-11-27
Packaging: OS X Installer Can Upgrade CVMFS

reneme:
This should make the new OS X installer clever enough to do a clean transition between CernVM-FS 2.1.x and CernVM-FS 2.2.0. Please review carefully. :-)

Testing should also be done, but will be somewhat manual I fear.


reneme:
moved the `/usr`->`/usr/local` migration code to `cvmfs_server setup`. It's only run on Mac OS of course.

reneme:
ready for review... When built on OS X 10.10, the OpenSSL libraries on 10.11 are properly picked up.

cvmfs/cvmfs_config
```diff
@@ -1624,6 +1627,131 @@ runas() {
...
+      echo "found and quarantined legacy configuration file '$conf_file'"
+    fi
+  done
+
+  # rescue all configuration files that have been added by the user
+  local cwd="$(pwd)"
+  cd "$old_conf_dir" # remove $old_conf_dir prefix from `find` output
+  for conf_file in $(find . -type f); do
+    echo "found and rescued user added configuration file '$conf_file'"
+    mv -f $conf_file ${new_conf_dir}/${conf_file}
```
jblomer: Here it might be necessary to create the base directory in `${new_conf_dir}` first

reneme: good catch! Thanks.

reneme:
Introduced the recommended changes both in the installer's conclusion-screen and the setup script.

PR #1261, opened 2015-12-04
FIX: Remove Use of `sudo` in `cvmfs_server`

reneme:
This erases the use of `sudo` from `cvmfs_server` as requested in [CVM-245](https://sft.its.cern.ch/jira/browse/CVM-245). Impersonating a specific user is now done either by `runuser` or (if this is not available) using `su`. Turns out that older versions of Ubuntu don't provide `runuser`.

Furthermore I found an orphaned un-config command that tried to remove `/etc/sudoers` configuration that was actually not done anymore since the rise of `cvmfs_suid_helper`.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -832,7 +832,8 @@ is_owner_or_root() {
 # a specific CVMFS repository.
 #
 # @param name   the name of the repository whose owner should be impersonated
-# @return       a shell invocation to impersonate $CVMFS_USER
+# @return       a shell invocation to impersonate $CVMFS_USER via stdout or
+#               exit code 1 if user couldn't be impersonated
 get_user_shell() {
   local name=$1
   local shell_cmd=""
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -1707,7 +1707,7 @@ mangle_s3_cvmfs_url() {
 
 # lowers restrictions of hardlink creation if needed
 # allows AUFS to properly whiteout files without root privileges
-# Note: this function requires sudo
+# Note: this function requires a privileged user
 lower_hardlink_restrictions() {
   if [ -f /proc/sys/kernel/yama/protected_nonaccess_hardlinks ] && \
      [ $(cat /proc/sys/kernel/yama/protected_nonaccess_hardlinks) -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


cvmfs/cvmfs_server
```diff
@@ -825,14 +832,25 @@ is_owner_or_root() {
...
 
   load_repo_config $name
-  local user_shell="sh -c"
-  [ $(whoami) != $CVMFS_USER ] && user_shell="sudo -E -H -u $CVMFS_USER sh -c"
-  echo "$user_shell"
+  if [ x"$(whoami)" = x"$CVMFS_USER" ]; then
+    shell_cmd="sh -c"
+  elif is_root; then
+    if [ $HAS_RUNUSER -ne 0 ]; then
+      shell_cmd="runuser -l $CVMFS_USER -c"
```
jblomer: Do we need to worry about "login shell" or not?  So far, we preserve root's environment expect for `$HOME`.  This is different from what `runuser -l` does, no?

reneme: Good catch! Should be `runuser -m` in fact. [fixed](https://github.com/reneme/cvmfs/commit/26308b126acd27a9afe2956a50c577df31abd88f)

cvmfs/cvmfs_server
```diff
@@ -825,14 +832,25 @@ is_owner_or_root() {
...
 
   load_repo_config $name
-  local user_shell="sh -c"
-  [ $(whoami) != $CVMFS_USER ] && user_shell="sudo -E -H -u $CVMFS_USER sh -c"
-  echo "$user_shell"
+  if [ x"$(whoami)" = x"$CVMFS_USER" ]; then
+    shell_cmd="sh -c"
+  elif is_root; then
+    if [ $HAS_RUNUSER -ne 0 ]; then
+      shell_cmd="runuser -m $CVMFS_USER -c"
```
jblomer: So as far as I can see there is no `runuser` option that corresponds to `sudo -H` (use `$HOME` of `$CVMFS_USER`).  But we can probably ignore this...?

cvmfs/cvmfs_server
```diff
@@ -825,14 +832,25 @@ is_owner_or_root() {
...
+    else
+      shell_cmd="su $CVMFS_USER -c"
```
jblomer: Also `su -m`?

cvmfs/cvmfs_server
reneme: We are not writing into the user's `$HOME` anyway. As long as not downstream tool depends on it, it should be fine. Integration tests will show, but I think we are rather safe. The only things that are called with that wrapper are `cvmfs_swissknife`, `rm`, `touch` and `date`.

cvmfs/cvmfs_server
reneme: Apparently not. Citing the [man page](http://man7.org/linux/man-pages/man1/su.1.html):
```
       For backward compatibility, su defaults to not change the current
       directory and to only set the environment variables HOME and SHELL
       (plus USER and LOGNAME if the target user is not root).  It is
       recommended to always use the --login option (instead of its shortcut
       -) to avoid side effects caused by mixing environments.
```

PR #1269, opened 2015-12-07
Feature: Skip Apache Configuration/Reload if Requested

reneme:
This allows to skip the Apache configuration when running `cvmfs_server mkfs`, `import` or `add-replica`. All of those commands can be called with `-p` for that purpose. An integration test is added that uses a minimal python webserver as a backend instead of Apache.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -263,18 +263,19 @@ Supported Commands:
                 [-m replicable] [-f union filesystem type] [-v volatile content]
                 [-g disable auto tags] [-a hash algorithm (default: SHA-1)]
                 [-z enable garbage collection] [-s S3 config file]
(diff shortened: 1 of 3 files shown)


cvmfs/cvmfs_server
```diff
@@ -1965,17 +1977,18 @@ CVMFS_AUTO_GC=true
 EOF
   fi
 
-  # make sure that the config file does not exist, yet
-  remove_apache_config_file "cvmfs.${name}.conf" || true
+  if [ $configure_apache -eq 1 ]; then
+    # make sure that the config file does not exist, yet
+    remove_apache_config_file "cvmfs.${name}.conf" || true
```
jblomer: unrelated to this diff, but it might be preferable to rename existing config files rather than to delete them.

PR #1271, opened 2015-12-07
Don't use localhost as default stratum 1 (CVM-892)

jblomer:
As a result, if `CVMFS_SERVER_URL` is empty, the error from the downloader is "malformed url" instead of "connection problem".  This leads to immediate failure as it is supposed to be (CVM-892).


traylenator:
Makes sense.

PR #1275, opened 2015-12-08
FIX: use timezone variable in T_Util.RfcTimestamp

Moliholy:
The (I hope) last fix for this test. Navigating through the documentation and the own time.h file I've seen there is a variable called timezone that represents the seconds of difference with respect to GMT, which is exactly what was left in the test.

Tested on the ARM fedora machine that was having problems with this issue, and in a standard SLC6. According to http://stackoverflow.com/a/5157134/2352087 this should be a platform-independent solution.

### DIFF
--- a/test/unittests/t_util.cc
+++ b/test/unittests/t_util.cc
@@ -792,7 +792,7 @@ TEST_F(T_Util, RfcTimestamp) {
   time_t time1 = time(NULL);
   string str = RfcTimestamp();
   strptime(str.c_str(), format, &tm);
-  time_t time2 = mktime(&tm);
+  time_t time2 = mktime(&tm) - timezone;
   EXPECT_GT(2, time2 - time1);
   setlocale(LC_TIME, curr_locale);
 }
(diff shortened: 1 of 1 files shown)


jblomer:
What about daylight saving?

test/unittests/t_util.cc
```diff
@@ -792,7 +792,10 @@ TEST_F(T_Util, RfcTimestamp) {
   time_t time1 = time(NULL);
   string str = RfcTimestamp();
   strptime(str.c_str(), format, &tm);
-  time_t time2 = mktime(&tm);
+  time_t time2 = mktime(&tm) - timezone;
+  if (tm.tm_isdst > 0) {
```
Moliholy: I had no issues with this while testing it, but just in case if it is in daylight saving time (i.e., during summer) it will subtract one hour

jblomer: Sorry, I wasn't clear in my comment.  I am not sure if extra handling of daylight saving is necessary.  But it might.  So did you test if it is necessary?

Moliholy: I tested it in a SLC6 setting different dates (summer and winter) and in none of the cases it enters in this new 'if', so in principle it looks like it's not necessary, but perhaps depending on the platform it could...

PR #1278, opened 2015-12-08
Feature: Create Fresh Repository Signing Keys on `cvmfs_server import`

reneme:
This implements [a feature request](https://sft.its.cern.ch/jira/browse/CVM-865) from @DrDaveD. It allows to create a fresh repository signing key and an associated certificate on `cvmfs_server import` like so:

```bash
cvmfs_server import -r -t atlas.cern.ch
```

Both switches are required for that to work `-t` is enabling the fresh creation of repository signing key and certificate while `-r` will rewrite the repository's whitelist. Note that `-r` can also be used alone (it was already implemented), for example if an expired whitelist needs to be imported. Both cases require the repository's masterkey naturally.

I didn't write a completely new integration test but extended 591 for that purpose.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -275,7 +275,7 @@ Supported Commands:
                 [-l import legacy repo (2.0.x)] [-s show migration statistics]
                 [-f union filesystem type] [-c file ownership (UID:GID)]
                 [-k path to keys] [-g chown backend] [-r recreate whitelist]
-                [-p no apache config]
+                [-p no apache config] [-t recreate repo key and certificate]
                 <fully qualified repository name>
                 Imports an old CernVM-FS repository into a fresh 2.1.x repo
   publish       [-p pause for tweaks] [-n manual revision number] [-v verbose]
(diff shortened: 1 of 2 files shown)


cvmfs/cvmfs_server
```diff
@@ -3023,6 +3037,19 @@ import() {
...
+  # check availability of repository signing key and certificate
+  local keys="$public_key"
+  if [ $recreate_repo_key -eq 0 ]; then
+    if [ ! -f ${keys_location}/${private_key} ] || \
+       [ ! -f ${keys_location}/${certificate} ]; then
+      die "repository signing key or certificate not found (use -t maybe?)"
+    fi
+    keys="$keys $private_key $certificate"
+  else
+    [ $recreate_whitelist -ne 0 ] || die "using -t implies whitelist recreation (use -r maybe?)"
```
DrDaveD: Why not just let -t imply -r, rather than failing if it isn't used?

reneme: I want to make that explicit so that people make a conscious decision about altering their whitelist.

PR #1281, opened 2015-12-09
Add `cvmfs_config killall` command

jblomer:
This is supposed to get out of stuck situations, e.g. when `cvmfs_config reload` hangs.  It kills all `cvmfs_config` and `cvmfs2` processes and clears /var/run/cvmfs/

@traylenator @DrDaveD: I'd appreciate a second look on the bash code.

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -134,7 +134,7 @@ cvmfs_config_usage() {
  echo "  reload [-c | <repository>]"
  echo "  umount"
  echo "  wipecache"
- echo "  killall"
+ echo "  killall [-l(azy unmount)] [-k(ill cvmfs user sessions)]"
  echo "  bugreport"
 }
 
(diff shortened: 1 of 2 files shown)


DrDaveD:
Overall it looks pretty good to me.  My comments are on the umounting code.   First, might you want to share code with cvmfs_umount?  There seems to be duplication.  Second, it might be good for umount to have an option to kill all current users of a filesystem, and for killall to use it.  You could try fuser -k, although in my experience that's not always reliable.  Another thing that I have sometimes found useful is umount -l to force a unmount and let the processes die when they're ready.

jblomer:
Many thanks!  Code duplication is removed.  Killing of user sessions has been recorded for the release after [CVM-953](https://sft.its.cern.ch/jira/browse/CVM-953).  Regarding `umount -l`, I left this out because it can be a pain with cvmfs repositories.  The fuse processes block on a POSIX lock, so if a process from a lazily unmounted repository is not removed, the next attempt to mount it hangs mysteriously (until someone looks into syslog).

PR #1286, opened 2015-12-10
Feature: Informational REST-"API" on CernVM-FS Server

reneme:
This pull request is work in progress and meant to facilitate the collaboration on this issue. Related tickets are [CVM-860](https://sft.its.cern.ch/jira/browse/CVM-860) and [CVM-862](https://sft.its.cern.ch/jira/browse/CVM-862). ~~Currently this contains some refactoring change sets of an [underlying pull request](https://github.com/cvmfs/cvmfs/pull/1285) but I will `rebase && push -f` it as soon as that one goes upstream.~~

For the moment, this adds apache configuration (feedback welcome) to expose `/cvmfs/info` when either `cvmfs_server mkfs` or `cvmfs_server add-replica` is run. Later we will need to add a `cvmfs_server migrate` step to retrofit this on existing installations.

Furthermore I create some stub files, namely `repositories` (for stratum0), `replicas` (for stratum1) and `maintainer` (for general information about the CernVM-FS server). Note that I introduced a sub-directory to accomodate the schema revision of this "API".

I'll keep this pull request up to date as I go along. Any feedback more than welcome.


reneme:
I would call this a 'minimal working version'. What it does is dropping JSON files (that are currently actually malformed) whenever a new repository/replica is created or gets destroyed. For each repository object I only list the name and an URI pointing to another JSON outlet that is left for TODO.

reneme:
now producing valid JSON in one single file. Looks like so:

```json
{
  "schema"       : 1,
  "repositories" : [
    {
      "name"  : "atlas-nightlies.cern.ch",
      "url"   : "/cvmfs/atlas-nightlies.cern.ch"
    },
    {
      "name"  : "test.cern.ch",
      "url"   : "/cvmfs/test.cern.ch"
    },
    {
      "name"  : "test.local",
      "url"   : "/cvmfs/test.local"
    }
  ],
  "replicas" : [
    {
      "name"  : "test.local",
      "alias" : "test2.replica ",
      "url"   : "/cvmfs/test2.replica"
    }
  ]
}
```

cvmfs/cvmfs_server
```diff
@@ -2026,6 +2027,38 @@ Alias /cvmfs/${name} ${storage_dir}
...
+    FileETag None
+
+    ExpiresActive On
+    ExpiresDefault "access plus 3 days"
+    ExpiresByType text/html "access plus 5 minutes"
+    ExpiresByType application/x-cvmfs "access plus 2 minutes"
+</Directory>
+EOF
+}
+
```
DrDaveD: Wouldn't it be good to have a common function for create_apache_config_for_repository and create_apache_config_for_global_info?

cvmfs/cvmfs_server
```diff
@@ -2336,6 +2370,96 @@ has_file_descriptors_on_mount_point() {
...
+
+_write_info_file() {
+  local info_file=$(_global_info_path)
+  local tmp_file=$(mktemp)
+
+  cat - > $tmp_file
+  chmod 0644 $tmp_file
+  mv -f $tmp_file $info_file
+  rm -f $tmp_file
```
DrDaveD: Is there any need to do rm after mv?

cvmfs/cvmfs_server
reneme: Given they are more or less the same, yes.

cvmfs/cvmfs_server
reneme: :-) just to be sure... fixing that, thanks!

cvmfs/cvmfs_server
reneme: For the moment this whole patch here is more the "make it work" approach. Just wanted to prototype that a bit, to have some base for discussion. I'm relying about on the apache experts to come up with a reasonable config here. :)

DrDaveD: The current apache config used for cvmfs repositories is not ideal but it works, so I figured that the easiest thing would be to just use the same one, changing only the minimum necessary.  That's why I made a create_apache_config_core() function in my compat/update_repositories_info script, assuming something like that would be used for both purposes.

reneme: I've taken your `create_apache_config_core()` as an inspiration to actually [DRY up the config creation](https://github.com/reneme/cvmfs/commit/b334daa56cafd1f0f13832ae5a33e927336c9f84) of `mkfs` and `add-replica` which was indeed a code duplication. However, I think it makes sense to keep the `/cvmfs/info` config separate and optimise it for its purpose, no?

reneme:
updated with [an additional resource](https://github.com/reneme/cvmfs/commit/ed04309207197d166dc06686df60655cabd31087) `/cvmfs/info/meta` containing custom (read: not auto-generated) meta information.

reneme:
Running `cvmfs_server update-info` will open `/cvmfs/info/meta` in an editor to allow for changes. Furthermore, it will try to use `jq` to verify the JSON validity after editing. Additionally `update-info` can be used as an migration-path for existent installations as it creates the `/cvmfs/info` HTTP resource opportunistically.

cvmfs/cvmfs_server
DrDaveD: Well a lot of it would be in common anyway.  The only parts that would be different would be the expiration parameters, but cvmfs-servermon has to override them anyway (with Cache-control: max-age header in the request, as discussed in the [ticket](https://sft.its.cern.ch/jira/browse/CVM-860?focusedCommentId=68835&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-68835)) at least when reading per-repo .cvmfs* files so it might as well do the same here.  I think it would be easiest/best to just leave the apache config the same.

reneme:
Update: allow for an instant retry when JSON was found to be invalid.

reneme:
@DrDaveD: I kind of agree. Just wanted to put it to the discussion. But you are right, that information can be obtained to sufficient precision automatically. Removing...

reneme:
changed *deb* and *rpm* packaging accordingly and added a second (somewhat trivial) test case for the /info/meta file.

reneme:
From my side this is maturing, apart from the following things:
- [ ] Apache configuration (completely shared or tuned accordingly)
- [ ] Agreement on data records (anything else that is needed?)
- [ ] S3 is currently not supported (is that a problem right now?)

reneme:
Unified the Apache configuration for both repositories and info-endpoint. If S3 is not an issue, this is ready from my side.

PR #1288, opened 2015-12-14
Repositories with alternative certificate and catalog path

jblomer:
Not quite ready yet but close.  While working on this prerequisite for CVM-904, I realized that a fixed path for the symlink (like .cvmfsroot) has a race condition: the time between updating the symlink and updating .cvmfspublished.  The symlink must contain the hash.  Of course, we accumulate symlinks this way but this is a second order problem which can be easily solved.


jblomer:
Ready to merge from my side.  

Eventually, we probably want to have an alternative path for the history database, too.  We also might want to add support for alternative paths into the replication code.  But I think we don't need to care about this right now.

@bbockelm please let me know if you have any comments.

cvmfs/manifest.cc
```diff
@@ -114,7 +118,8 @@ string Manifest::ExportString() const {
     "R" + root_path_.ToString() + "\n" +
     "D" + StringifyInt(ttl_) + "\n" +
     "S" + StringifyInt(revision_) + "\n" +
-    "G" + StringifyBool(garbage_collectable_) + "\n";
+    "G" + StringifyBool(garbage_collectable_) + "\n"
```
reneme: Knit picking: That is missing a `+` in the end of the line. Doesn't matter for compilation, but maybe for the sake of consistency?

cvmfs/upload_local.cc
```diff
@@ -215,7 +236,24 @@ void LocalUploader::FinalizeStreamedUpload(UploadStreamHandle  *handle,
...
+               "(errno=%d)", final_path.c_str(), dest_path.c_str(),
+               strerror(errno), errno);
+      atomic_inc32(&copy_errors_);
+      Respond(handle->commit_callback, UploaderResults(cpy_errno));
+      return;
+    }
+  }
+
   const CallbackTN *callback = handle->commit_callback;
+  // TODO(jblomer): do we leak memory if the function is left earlier?
```
reneme: we do! But if it is left early, the using code bails out with an `assert()`...

reneme: ... TODO is reasonable though. ;-)

cvmfs/swissknife_sync.cc
```diff
@@ -595,7 +600,8 @@ int swissknife::CommandSync::Main(const swissknife::ArgumentList &args) {
   sync->Traverse();
 
   LogCvmfs(kLogCvmfs, kLogStdout, "Exporting repository manifest");
-  UniquePtr<manifest::Manifest> manifest(mediator.Commit());
+  const bool use_alt_path = params.voms_authz.size();
```
reneme: please! 0 and 1 are somewhat boolean, but why not `! params.voms_authz.empty()`?

cvmfs/manifest.cc
jblomer: Yep, `+` is missing

cvmfs/swissknife_sync.cc
jblomer: Yep, `empty()` is better.  Please stay positive in these reviews.

jblomer:
Superseded by #1291

PR #1291, opened 2015-12-15
Cvm 904 prereq

reneme:
This is a refactoring of [this pull request](https://github.com/cvmfs/cvmfs/pull/1288) that minimises the changes in the CVMFS server internals.

It doesn't touch the server's file processing at all. Instead it introduces `Spooler::PlaceBootstrappingShortcut()` that can be used after the file processing is done as usual. Specifically it is done just before the new `Manifest` is published in `cvmfs_swissknife sign`.

Note: That is still a bit work in progress. I need to add a unit test case and an S3 implementation.


jblomer:
Much better!

reneme:
> Much better!

thanks! But I wasn't done with that one. :-) Well, sending the rest separately. ...

PR #1292, opened 2015-12-15
TEST: Some Unit Tests as a Follow-up

reneme:
**Note:** Work in progress... doesn't work for S3 yet.

This is a follow up for the prematurely merged #1291. Adding some unit tests.


reneme:
Can go in now.

PR #1293, opened 2015-12-15
Allow for configuration of dns timeout and retry (CVM-875)

jblomer:
Introduces `CVMFS_DNS_RETRIES` and `CVMFS_DNS_TIMEOUT` (which are only affective for resolving proxies).  Contains also a fix to the download manager when a DNS name cannot be resolved.  Multiple fixes to the DNS timeout integration test.


DrDaveD:
Don't they also apply to servers when proxy is DIRECT?

jblomer:
The cvmfs custom name resolver is currently only used to resolve proxy names.  That should cover the grid use case as host names are resolved by the proxy.

For direct connections, curl uses c-ares with some reasonable defaults.

PR #1294, opened 2015-12-15
Add ipv6 support for geo api

DrDaveD:


test/src/611-geoservice/main
```diff
@@ -0,0 +1,97 @@
...
+}
+
+
+cvmfs_run_test() {
+  logfile=$1
+
+  echo "checking if both IPv4 and IPv6 addresses are available"
+  local ipv4addr="`cvmfs_getaddr -4`"
+  local ipv6addr="`cvmfs_getaddr -6`"
+  [ -n "$ipv4addr" ] && [ -n "$ipv6addr" ] || return 1
```
jblomer: The test fails on non dual stack machines.  I think this is a bit strict, the test should rather quit with a warning.

DrDaveD: I thought all machines at CERN were dual stack, even though most of them don't have IPv6 addresses listed in DNS.  I'd really like to make sure that the IPv6 code is exercised.  Ah, I guess maybe you're talking about some non-scientific linux ports, is that right?  As long as some of your test machines have IPv6, then I will change it to skip the IPv6 tests with a warning if an IPv6 address is not available.

DrDaveD:
Ok, please try again.  The geoservice test now skips IPv6 tests if no IPv6 address is available.  I also put in cleanup of the replica repo that it adds, and quoted the $result variable in case it contains an error message.

jblomer:
Sorry, I wasn't clear.  You can actually return a warning from the test case with

    return $CVMFS_GENERAL_WARNING

which might be best if IPv6 support is missing.

DrDaveD:
The only other two cases of CVMFS_GENERAL_WARNING that I see instead set the variable to a value of "1" and continue.  Is that want you want me to do?

jblomer:
Ah, right, it is `CVMFS_GENERAL_WARNING_FLAG=1` which internally changes the return value of the test case to `CVMFS_GENERAL_WARNING` if the test succeeds.

DrDaveD:
Right.  I changed the geoservice test to now set CVMFS_GENERAL_WARNING_FLAG=1 if there's no IPv6 address.

PR #1298, opened 2015-12-16
TEST: Regression in File Chunking

reneme:
Found that totally by chance while testing a different thing. Apparently file chunking fails with file sizes that are exactly a power of two. I expect something like an off-by-one error or similar. Possibly also in conjunction with the `CVMFS_MAX_CHUNK_SIZE`. I didn't look into it yet, this is more a TODO regression test.


reneme:
updated with a regression unit test... Fix follows tomorrow...

PR #1300, opened 2015-12-16
Fix mac compatibility problems when running the unit tests

Moliholy:
When running on mac there are certain commands that have to be modified in the test infrastructure. With these changes in principle we can run the following tests:

000, 001, 002, 003, 008 and 009.

I am currently testing others, so more changes might come.


cvmfs/cvmfs_config
```diff
@@ -352,8 +356,6 @@ cvmfs_chksetup() {
     fi
     # Check that /etc/auto.cvmfs is referenced in auto.master
     # Check that /etc/auto.cvmfs is executable on Linux
-    local global_mount_dir
-    global_mount_dir=$CVMFS_MOUNT_DIR
```
jblomer: Why is this being removed?

test/test_functions
```diff
@@ -955,7 +963,11 @@ load_repo_config() {
 
 
 get_millisecond_epoch() {
-  echo $(( $(date +%s%N) / 1000000 ))
+  if running_on_osx; then
+    echo $(( $(date +%s) * 1000 ))
```
Moliholy: I know this is an ugly workaround, but this function is only used to get the milliseconds spent running the test, so it's not crucial.
This is because the '%N' is not supported on mac, and there is no equivalent

test/test_functions
```diff
@@ -303,6 +303,14 @@ cvmfs_mount() {
   if [ "x$CVMFS_TEST_DEBUGLOG" != "x" ]; then
     sudo sh -c "echo \"CVMFS_DEBUGLOG=$CVMFS_TEST_DEBUGLOG\" >> /etc/cvmfs/default.local" || return 100
   fi
+
+  if running_on_osx; then
+    for i in $(echo "$repositories" | tr "," " "); do
+      mkdir -p "/cvmfs/$i"
```
jblomer: This should also take care of the "config repository".  Something like

    eval $(cvmfs_showconfig | grep ^CVMFS_CONFIG_REPOSITORY=)
    if [ "x$CVMFS_CONFIG_REPOSITORY" != "x" ]; then
      mkdir -p /cvmfs/$CVMFS_CONFIG_REPOSITORY
      ...

cvmfs/cvmfs_config
Moliholy: It's copied a few lines to the top.
global_mount_dir was only set if running on Linux, otherwise it remains empty, which causes a failure when being evaluated in the final condition.

Moliholy:
Now it checks the showconfig.

test/test_functions
```diff
@@ -303,6 +305,15 @@ cvmfs_mount() {
   if [ "x$CVMFS_TEST_DEBUGLOG" != "x" ]; then
     sudo sh -c "echo \"CVMFS_DEBUGLOG=$CVMFS_TEST_DEBUGLOG\" >> /etc/cvmfs/default.local" || return 100
   fi
+
+  eval $(cvmfs_config showconfig | grep -e ^CVMFS_CONFIG_REPOSITORY=)
```
jblomer: We should move this into the OS X if block because it is quite a lengthy operation.  Note that `CVMFS_CONFIG_REPOSITORY` is optional.  It might be set but it doesn't need to be.

Moliholy:
That eval is now inside the OS X block

test/test_functions
```diff
@@ -303,6 +305,17 @@ cvmfs_mount() {
   if [ "x$CVMFS_TEST_DEBUGLOG" != "x" ]; then
     sudo sh -c "echo \"CVMFS_DEBUGLOG=$CVMFS_TEST_DEBUGLOG\" >> /etc/cvmfs/default.local" || return 100
   fi
+
+  if running_on_osx; then
+    eval $(cvmfs_config showconfig | grep -e ^CVMFS_CONFIG_REPOSITORY=)
+    if [ "x$CVMFS_CONFIG_REPOSITORY" != "x" ]; then
```
jblomer: Keep in mind that `CVMFS_CONFIG_REPOSITORY` is optional.  But there is always something in `$repositories`

Moliholy: It should be fixed now

test/test_functions
```diff
@@ -303,12 +305,31 @@ cvmfs_mount() {
   if [ "x$CVMFS_TEST_DEBUGLOG" != "x" ]; then
     sudo sh -c "echo \"CVMFS_DEBUGLOG=$CVMFS_TEST_DEBUGLOG\" >> /etc/cvmfs/default.local" || return 100
   fi
+
+  if running_on_osx; then
+      for i in $(echo "$repositories" | tr "," " "); do
+        eval $(cvmfs_config showconfig $i | grep -e ^CVMFS_CONFIG_REPOSITORY=)
```
jblomer: Let's discuss this spot later.

Moliholy:
New changes added to fix some integration tests on mac

test/src/010-du/main
```diff
@@ -6,8 +6,9 @@ cvmfs_run_test() {
 
   cvmfs_mount sft.cern.ch "CVMFS_MAX_RETRIES=3" || return 1
 
-  sudo du -ch --max-depth=3 /cvmfs/sft.cern.ch || return 2
-  sudo sh -c "echo 3 > /proc/sys/vm/drop_caches"
+  # this used to be --max-depth=3 instead of -d 3, but it had to be changed for compatibility with mac
+  sudo du -ch -d 3 /cvmfs/sft.cern.ch || return 2
```
jblomer: The problem is that Linux doesn't recognize the `-d` parameter...

test/test_functions
```diff
@@ -331,7 +354,7 @@ cvmfs_umount() {
     fi
 
     local timeout=5
-    while cat /proc/mounts | grep -q " /cvmfs/$r "; do
+    while mount | grep -q " /cvmfs/$r "; do
```
jblomer: should be `mount` only on OS X, otherwise `cat /proc/mounts`

test/src/010-du/main
jblomer: @Moliholy please remember to look at this.

test/test_functions
jblomer: @Moliholy please remember to look at this.

Moliholy:
PR updated:

- Various fixes in the code so that the integration tests run on mac
- Added various scripts and others modified to make their execution possible through the automatic project in Jenkins
- This PR uses the changes in https://github.com/cernvm/ci-scripts/pull/2

test/src/010-du/main
Moliholy: Fixed

test/test_functions
Moliholy: Fixed

PR #1303, opened 2015-12-17
Change geoapi to use passed-in proxy name as priority source address

DrDaveD:
As documented in https://sft.its.cern.ch/jira/browse/CVM-722


jblomer:
We should probably also update the documentation and recommend a name caching daemon on the stratum 1 host.

DrDaveD:
I put documentation change in a [pull request](https://github.com/cvmfs/doc-cvmfs/pull/32)

PR #1308, opened 2015-12-20
Merge devel into CVM-904

bbockelm:
As we have recently merged some CVM-904 pre-requisite work into devel, let's merge devel back into CVM-904.


bbockelm:
@jblomer - f94874f contains a critical bugfix that I noticed when running the integration test.

Given that commit, integration tests work again.

PR #1309, opened 2015-12-21
Add support for client contexts.

bbockelm:
When doing secure CVMFS, any time we fetch a remote object, we must
understand what context the fetch occurs in - who requests the
fetch operation?

Having this information will allow us, within a FUSE process, grab
the underlying credential corresponding to the current context.

This is not immediately useful in itself, but helps to decrease the size of CVM-904.


jblomer:
Please have a look at #1316 for this functionality.

PR #1313, opened 2015-12-22
First draft of cvmfs_cp

bbockelm:
cvmfs_cp is a small cp-like CLI for CVMFS.  It allows you to copy
files in and out of a CVMFS repo, regardless of how the host is
setup.

Beyond immediate value in testing independently of the host config,
this would allow "light" users of CVMFS to utilize CVMFS's capabilities
even on sites where it is not mounted.

@jblomer @DrDaveD - the is the basic version.  I plan on further extending it with the various repo and global opts automatically filled in when possible (similar to how parrot_run works).

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -97,6 +97,10 @@ set (TEST_LIBCVMFS_SOURCES
   test_libcvmfs.cc
 )
 
+set (CVMFS_CP_SOURCES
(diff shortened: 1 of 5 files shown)


dthain:
+1 for this tool, we would like to use it for accessing container images without having a full blown mount

jblomer:
See https://github.com/bbockelm/cvmfs-cp

PR #1315, opened 2015-12-24
add cvmfs_snapshots command to cvmfs-server package

DrDaveD:
See https://sft.its.cern.ch/jira/browse/CVM-813


jblomer:
@reneme: can this be done as `cvmfs_server snapshot *`?

reneme:
I don't see any obvious road blocks. I believe it makes sense to put that under `cvmfs_server snapshot *` instead of adding a wrapper script.

What about the introduced configuration variable `CVMFS_REPLICA_ACTIVE`? Should that also abort specific `cvmfs_server snapshot` operations?

DrDaveD:
I had thought about doing it as cvmfs_server snapshot -a, which would be better than '*' since the latter is expanded by the shell.   It can be done that way if you prefer, but on the other hand I don't think it would make it any simpler; there's practically nothing in the cvmfs_server infrastructure that it can take advantage of.  I thought that by having it in a separate script it would be a little easier to modify later.  For instance I was thinking I'd change it (this week yet) to re-order the repositories to do it from oldest snapshot to newest (based on the .cvmfs_last_snapshot file).  Perhaps at sometime in the future we'd like to introduce more parallelism.  Having it as a separate script also makes it a little more obvious what it does in case somebody wants to replace it with their own script that does something slightly differently.

reneme:
What bugs me with the additional script is mainly UX concerns. Right now *everything* is done through `cvmfs_server` - for functionality supported by us, of course.

I agree that `cvmfs_server` becomes too big and is hard to maintain already. We should re-think the internal structure of `cvmfs_server` but that should not happen by factoring functionality into many different scripts. Having `cvmfs_server snapshot` and adding `cvmfs_snapshots` is confusing in my opinion.

jblomer:
@DrDaveD: can you modify the PR to make the functionality available as `cvmfs_server snapshot -a`?  I'd find that much nicer than an additional binary.

DrDaveD:
OK

DrDaveD:
Rather than undoing changes in this branch I'm going to create a new branch.  Closing this pull request.

PR #1316, opened 2015-12-27
Modified: support for client context

jblomer:
This PR is supposed to replace #1309.  Instead of passing the context through the call stack, I'd prefer to store the information in thread local storage at the beginning of a file system callback.  It can then be retrieved by the fetcher when needed.

There is a risk of bugs where mistakenly someone else's context is used but we have only very few, well-known file system callbacks.  So we should be able to get this right.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -60,6 +60,7 @@ set (CVMFS_CLIENT_SOURCES
   sqlitevfs.cc sqlitevfs.h
   fetch.cc fetch.h
   sink.h
+  clientctx.cc clientctx.h
 )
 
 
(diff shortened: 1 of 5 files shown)


bbockelm:
Looks good to me.  Feel free to merge when you're ready and I'll make the adjustments for CVM-904.

bbockelm:
@jblomer - working with this, I noticed that initialization of the ClientCtx::instance_ singleton is not thread-safe.  Is that on purpose?

jblomer:
@bbockelm: ah, yes, I forgot to mention it.  I think it doesn't matter.  The instance can be retrieved during `Init()` in cvmfs.cc and the cleanup routine should be called in `Fini()`.

PR #1321, opened 2015-12-28
Add `cvmfs_talk cleanup rate` command (CVM-270)

jblomer:
Let the quota manager keep track of cleanup events over time.  

Adds helper to query monotonic time with second resolution.  Monotonic time is platform-specific.

Introduces the new `Recorder` class.  The recorder counts events in certain fixed interval bins and for a certain fixed maximum time window in the past.  Implemented as a ring buffer.  Multiple recorders can be put together with more coarse-grained interval bins as the time window in the past increases.  This way, we can keep track of the last 4 days with ~1.5kB memory that is allocated as a block.

### DIFF
--- a/cvmfs/statistics.cc
+++ b/cvmfs/statistics.cc
@@ -4,8 +4,10 @@
 
 #include "statistics.h"
 
+#include <algorithm>
 #include <cassert>
 
+#include "platform.h"
 #include "smalloc.h"
 #include "util.h"
 #include "util_concurrency.h"
@@ -32,6 +34,7 @@ std::string Counter::PrintRatio(Counter divider) {
 
 
 //-----------------------------------------------------------------------------
+
 
 Counter *Statistics::Lookup(const std::string &name) {
   MutexLockGuard lock_guard(lock_);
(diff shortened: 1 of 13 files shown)


cvmfs/statistics.cc
```diff
@@ -94,6 +97,114 @@ Statistics::~Statistics() {
...
+/**
+ * If necessary, capacity_s is extended to be a multiple of resolution_s
+ */
+Recorder::Recorder(uint32_t resolution_s, uint32_t capacity_s)
+  : last_timestamp_(0)
+  , capacity_s_(capacity_s)
+  , resolution_s_(resolution_s)
+{
+  assert((resolution_s > 0) && (capacity_s > resolution_s));
+  int has_remainder = (capacity_s_ % resolution_s_) != 0;  // 0 or 1
```
reneme: `bool` instead of `// 0 or 1`?

cvmfs/statistics.h
```diff
@@ -81,6 +82,70 @@ class Statistics {
...
+
+  /**
+   * Size of the bins for the tick counters.
+   */
+  uint32_t resolution_s_;
+
+  /**
+   * Shorthand for bins_.size(), constant during lifetime of the recorder.
+   */
+  unsigned no_bins_;
```
reneme: why not mark it `const` and init in the constructor's member initialisation then?

jblomer: Hm, in present form that would look a bit ugly since the final capacity is only known after a bit of calculation, which would then move to the member initialization, too.

reneme: I've seen that. But it's a simple devision of the constructor's parameters. I find that a reasonable price to pay in order to make this information duplication as safe as possible.

jblomer: It's a simple division if the capacity is a multiple of the resolution.  Otherwise it needs to be expanded to be one.  Not a lot of computation but 3 lines in the constructor body.  How would it look like as member initialization?

reneme: Overlooked that. I'd do it with a static helper function in the class which would also hide the details of this expansion. Though I agree, It's a bit a matter of taste.

```C++
Recorder::Recorder(uint32_t resolution_s, uint32_t capacity_s)
  : last_timestamp_(0)
  , capacity_s_(ExpandCapacityAsNeeded(capacity_s, resolution_s)),
  , resolution_s_(resolution_s)
  , no_bins_(capacity_s_ / resolution_s_)
```

reneme: However, you could make all but `last_timestamp_` const members which is actually nice to have.

jblomer: Just tried, non-static const members break the default assignment operator (needed in `MultiRecorder`).  Changing `int` to `bool` thing, for the rest I'll leave it as is.

PR #1326, opened 2015-12-29
Infrastructure to fetch credentials and check VOMS credentials.

bbockelm:
This is the core of the VOMS work; added to the build but not integrated into the FUSE process.

Does the following:
- Fetch and cache credentials from the user process
- Enable curl handles to use cached credentials.
- Implements callouts to libvomsapi to parse credentials.
- Implements authz logic (given an authorization, checks to see if the current credential is acceptable).


jblomer:
Merged with whitespace fixes and additional TODO comments

PR #1328, opened 2015-12-30
Have FUSE process report the currently-used pubkey.

bbockelm:
Allow the signature manager to track the pubkey (either from list
of keys or the certificate's pubkey) last used to verify data, and
expose this information as an extended attribute.

With this, we can deduce all the critical information (host and
pubkey) necessary to access a mounted repo via libcvmfs (where
the access may happen on a different worker node).

This is the only CVMFS-side necessary for a larger project: integrating CVMFS within HTCondor.  The goal is to have the schedd instruct the worker node to mount and configure a specific repository for a given job (without needing any configuration files on the worker node beforehand); the schedd should also be able to specify the repo version on the worker node should match the submit node.  If the worker node is unable to actually mount a repository (for example, if it's really a pilot running as non-root), then the user should be able to specify files to copy into the scratch directory (and have them moved via `cvmfs_cp` instead of actually staging from the schedd itself).  This way, HTCondor users can always rely on CVMFS regardless of its availability on the worker node.


jblomer:
@bbockelm 

That would be very useful!

My only concern is on `cvmfs_cp`.  I'm a bit reluctant to maintain a set of command line utilities in parallel to existing fuse module and the parrot utilities.  I'm also afraid that `cvmfs_cp` might start as an innocent utility that will quickly consume us when it gets extended to more and more use cases.

Couldn't we just use `parrot_cp` or the parrot sandbox?

bbockelm:
Hi,

`parrot_cp` is fairly heavyweight compared to `cvmfs_cp`: it pulls in a lot of extra baggage - a pretty heavy dependency tree - unrelated to CVMFS.  Further, I think `parrot_cp` is good evidence there's not too-deep a rabbit hole here: there is no `parrot_ls`, `parrot_mkdir`, etc.

That said, if we ship `libcvmfs` in a more usable format (include static library in the `-devel` RPM, ship it as a dynamic library), we could move `cvmfs_cp` into a separate repo.  It just seems to be a fairly small utility to be a standalone package.

That's said, let's move the discussion to the `cvmfs_cp` PR?

Brian

jblomer:
I pushed a [merged version with minor modifications](https://github.com/cvmfs/cvmfs/tree/bbockelm-pubkey_xattr) to the upstream repository.  I'm not sure if it does what you expect though: if the local cache is up to date there won't be a verification of the whitelist and the last use public key remains empty.

bbockelm:
How disruptive would it be to re-validate on startup?  Seems like it would be fairly low-cost, CPU-wise, right?

jblomer:
That's a possibility.  It would require some reordering in the verification of manifest and whitelist.  Wouldn't be simpler to print out all public keys and to use all of them on the worker nodes?

bbockelm:
What's the worst-case scenario for number of public keys?  If it's less than 5 in all our known cases, I think that would work.

jblomer:
It's less than 5 in all known cases.

bbockelm:
Alright, I'll rework the branch to just return all keys.  Is it better to keep these rendered as PEM in memory or should I create it dynamically on request?

jblomer:
I think it's easier to render them on request because the vector with `RSA *` structs is already there.

bbockelm:
Ok, sounds good.  Working on it.

bbockelm:
If I write out a pubkey file with multiple keys, will the CVMFS client load all the pubkeys?  Or will I need to setup a directory structure on the remote side?

jblomer:
I never tried files with concatenated public keys.  I think it only works with a directory with .pub files.

bbockelm:
Alright - it shouldn't really matter, HTCondor could generate either.

bbockelm:
Ok, I think this one is ready to go.

cvmfs/signature.cc
```diff
@@ -290,6 +290,47 @@ bool SignatureManager::LoadPublicRsaKeys(const string &path_list) {
...
+std::string SignatureManager::GetActivePubkeys() {
+  std::string pubkeys;
+  for (std::vector<RSA *>::const_iterator it = public_keys_.begin();
+       it != public_keys_.end();
+       it++) {
+    pubkeys += GenerateKeyText(*it);
+  }
+  if (certificate_) {
+    // Increases the refcount of the rsa object.
+    RSA *rsa = EVP_PKEY_get1_RSA(X509_get_pubkey(certificate_));
```
jblomer: The public key of the certificate shouldn't matter, actually.  It's not used for the whitelist verification.

jblomer:
Ok, please query the attribute somewhere in the integration tests.

bbockelm:
Alright - added an integration test for this attribute to verify it responds and is correct.

cvmfs/signature.cc
```diff
@@ -574,6 +609,10 @@ bool SignatureManager::Verify(const unsigned char *buffer,
     )
   {
     result = true;
+    RSA *rsa = EVP_PKEY_get1_RSA(X509_get_pubkey(certificate_));
```
jblomer: This is not necessary anymore, is it?

PR #1333, opened 2016-01-02
Fix unittests and cpplint.

bbockelm:


jblomer:
Ok, as for the function signatures, we usually use const (references) for input parameters followed by pointers (no references) for output parameters.  I'll change it in the merge process.

bbockelm:
Yeah - I had been meaning to bring that up.

Non-const references have an important role to play in C++ which I think you're overlooking.  They provide three important pieces of information about the API:
- That the function does not accept NULL as valid input.
- Indication of object lifetime and ownership: indicates that the function does not take ownership of the object, the caller will keep the object alive throughout the function call, and that the callee cannot reference the object after the function exit.

While I find it irritating that you can't determine from text of the function call whether an argument is const, non-const, or copied (although this can be determined by looking at the signature), I find myself wasting far more time trying to deduce object lifetimes and ownerships (which cannot be deduced by looking at the signature).

jblomer:
We'll think about it.  Perhaps we find other means (like naming conventions) to clarify the ownership and lifetime.  At some point we decided to follow the [Google rules|https://google.github.io/styleguide/cppguide.html#Reference_Arguments]. Changing parameter passing now for some parts would add to the inconsistency of the code base.

bbockelm:
Yeah - I'll try to remember to use pointers: just goes against my grain.

This gets easier than in C++11, of course, as they bring in smart pointers to help define object lifetime information.

jblomer:
Ah, this might help in some cases: there is a `UniquePtr` in util.cc

PR #1337, opened 2016-01-03
CVM-904: devel merge

bbockelm:
Merge devel after another chunk of the CVM-904 branch was pulled in.


jblomer:
This has the wrong destination branch, no?

bbockelm:
Rats, it does.  I'll fix it real quick.

PR #1338, opened 2016-01-03
CVM-904: Secure CVMFS

bbockelm:
This PR finishes off the secure CVMFS work:
- Enforce VOMS authz attribute requirements for the FUSE process.
- Set ClientCtx as necessary in the FUSE process.  This allows us to pass user information to the download layer.
- If HTTPS is used and a ClientCtx is available, download files with the client certificate.

Includes an integration test.

Note that merge conflicts will exist until #1337 is merged.


bbockelm:
Whoops, PR #1340 is now the correct thing to merge first.

jblomer:
@bbockelm some adjustments added.

PR #1339, opened 2016-01-03
Add support for passing compression algorithm when creating graft files.

bbockelm:
Grafting does not go through the spooler process, yet compression
algorithm is only known post-spooling.  This passes the compression
algorithm to the sync stage, allowing us to set the graft's properties
correctly.

Without this, all grafted files have the default zlib compression bit set.

### DIFF
--- a/cvmfs/catalog_balancer_impl.h
+++ b/cvmfs/catalog_balancer_impl.h
@@ -55,8 +55,10 @@ void CatalogBalancer<CatalogMgrT>::AddCatalogMarker(string path) {
       MakeEmptyDirectoryEntryBase(".cvmfsautocatalog", parent.uid(),
                                            parent.gid());
   string relative_path = path.substr(1);
-  catalog_mgr_->AddFile(cvmfscatalog, xattr, false, relative_path);
-  catalog_mgr_->AddFile(cvmfsautocatalog, xattr, false, relative_path);
+  catalog_mgr_->AddFile(cvmfscatalog, xattr, false, zlib::kZlibDefault,
+                        relative_path);
+  catalog_mgr_->AddFile(cvmfsautocatalog, xattr, false, zlib::kZlibDefault,
+                        relative_path);
 }
 
 template <class CatalogMgrT>
--- a/cvmfs/catalog_mgr_rw.cc
+++ b/cvmfs/catalog_mgr_rw.cc
@@ -336,6 +336,7 @@ void WritableCatalogManager::AddFile(
   const DirectoryEntry  &entry,
   const XattrList       &xattrs,
         bool             external_data,
+        zlib::Algorithms compression_algorithm,
   const std::string     &parent_directory)
 {
   const string parent_path = MakeRelativePath(parent_directory);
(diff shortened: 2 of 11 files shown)


jblomer:
If I understand correctly, this would result in passing a `DirectoryEntryBase` and a `zlib::Algorithms` to the catalog manager.  The compression algorithm in the directory entry can contradict the explicitly passed algorithm.

Since the spooler is anyway not relevant for grafted files, I think it would make more sense to add support for a `compression=` field in the graft files (which probably should default to none).

bbockelm:
I considered the `compession=` field approach also, but didn't go there because graft files would have a different behavior than the repository default.  That is, if you create a repo with a default of uncompressed, I really didn't want graft files to have a different default from "real" files.

A similar thing occurs with `external_data` (I think you pointed it out in the other commit).

The actual reason why I do this is because the `DirectoryEntryBase` object is `const`; I could create a copy of it and pass it to the catalog manager.  However, the catalog manager turns around and creates yet another copy of `DirectoryEntryBase`.  Since I didn't want three copies of the same object to exist, I added to the call arguments.

An alternate approach would be to eliminate the external and compression properties from the call arguments and make the `DirectoryEntryBase` non-const.

jblomer:
Yeah, that might get confusing.

I think the problem was to postpone setting of these attributes in the first place.  The union file system engine can ask the sync mediator about external/non-external and the compression algorithm when it creates the sync item.  That would simplify the pipeline.  Should I look into it?

bbockelm:
I'm not 100% sure I follow you - but certainly feel free to look into simplifying things!

Since the external_data argument is already there for `AddFile`, would it be best to merge this and #1354, then do the simplification?  It's probably an equivalent amount of work to remove 2 arguments versus 1 argument.  This would avoid causing merge conflicts.

jblomer:
Ok, I'm looking into it.

jblomer:
All right, the push for closing #1354 should address this as well.  I hope I didn't screw up, but the integration tests still run.  Now the sync items and `DirectoryBaseEntries` immediately get the right setting for external data and compression algorithm.

Looking at compressed/uncompressed files, I wonder if we still need the "external" marker in the catalog property table.  We can add a parameter CVMFS_EXTERNAL_DATA=yes to /etc/cvmfs/repositories.d/$reponame/server.conf (like we do for the compression algorithm) and use it to pass a switch to `cvmfs_swissknife sync`.

bbockelm:
Hi,

For publishing, we indeed ended up with slightly different approach for what became similar client features (recall that originally, external-data was a catalog property in the client while compression was a per-file flag).  I think the *best* of both worlds would be:
1.  The configuration used for `cvmfs_server mkfs` are encoded as configuration parameters in `/etc/cvmfs/repositories.d/$reponame/server.conf`.  This sets the default for all future publications
2.  A command line switch to `cvmfs_server publish` overrides the default setting for all files in a given publication.

(1) would be a change to external data.  (2) would be a change to compression.

Separately: Now that you mention it, this particular case isn't covered by the integration tests.  Now that we advertise the compression status of the file, it should be easy to test.

Brian

jblomer:
Yep, sounds good to me!  I'll have a look.

bbockelm:
Actually - this case is covered in the integration test code added in #1358.  I merged `devel` into that branch and verified everything works.

PR #1341, opened 2016-01-03
Expose the compression algorithm used as an xattr.

bbockelm:
It's useful information to know - but more importantly, it makes
writing integration tests involving publishing uncompressed files
easier.


jblomer:
Thanks, please add a line to the compression integration test that queries the attribute.

bbockelm:
@jblomer - done.  I didn't see a simple way to test for zlib compression in integration test `617-uncompressed`, so I added that portion into `597-graft`.

PR #1342, opened 2016-01-04
CVM-904: Get and set VOMS authz info

bbockelm:
Expose the currently loaded authz as an xattr (`authz`).

When publishing, allow a user to write `.cvmfsauthz` in the root of the repository; replace current authz list with the contents of that file.

Note that I purposely didn't include an integration test here: it'll be a lot easier to include this in the `600-securecvmfs` integration test in CVM-904 (once it is merged).  I did test a simple set / get by hand.


cvmfs/cvmfs.cc
```diff
@@ -1628,6 +1628,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     } else {
       attribute_value = "DIRECT";
     }
+  } else if (attr == "user.authz") {
+    bool has_authz = catalog_manager_->GetVOMSAuthz(&attribute_value);
+    if (!has_authz) {
+      fuse_reply_err(req, ENOATTR);
```
jblomer: Function should return immediately here

cvmfs/sync_mediator.cc
```diff
@@ -626,7 +626,34 @@ void SyncMediator::AddFile(const SyncItem &entry) {
...
+      }
+      close(fd);
+      catalog_manager_->SetVOMSAuthz(new_authz);
+    } else {
+      LogCvmfs(kLogPublish, kLogStderr, "Encountered a authz file (%s) outside"
+               " the root directory; authz can only be set at top-level.",
+               entry.GetRelativePath().c_str());
+      abort();
+    }
+  }else if (entry.HasGraftMarker()) {
```
jblomer: nit: missing whitespace

cvmfs/util.cc
```diff
@@ -1546,6 +1546,49 @@ bool SafeWrite(int fd, const void *buf, size_t nbyte) {
...
+
+/**
+ * Pull file contents into a string
+ */
+bool SafeReadToString(int fd, std::string *final_result) {
+  if (!final_result) {return false;}
+
+  std::string tmp_result;
+  static const int buf_size = 4096;
+  std::vector<char> buf; buf.reserve(buf_size);
```
jblomer: I'd prefer this on the stack with `alloca`

jblomer:
This is indeed only a small patch.  Please add unit tests for `SafeRead`, `SafeReadToString` and the additional piece of integration test to `600-securecvmfs`

jblomer:
On a second thought, we don't really need the marker file, do we?  We can set the new VOMS attribute with a parameter to `cvmfs_server publish` and we can query it through the extended attribute.

cvmfs/util.cc
bbockelm: Ok, I can fix that.  4k isn't too horrible for stack allocations

bbockelm:
Actually, I tried the CLI approach: since multiple authz's can be specified, constructing the command line got pretty tedious.

Hm - although we could specify an authz file on the CLI (equivalent to the contents of the marker file)?  Which would you prefer?

jblomer:
I'd prefer the CLI with a file then, one less special file to take care of during sync.

bbockelm:
Sounds good... give me a few minutes to refactor.

bbockelm:
Ok, I think I got all the review comments fixed.

cvmfs/cvmfs.cc
jblomer: I think this is still missing

cvmfs/util.cc
jblomer: This seems to be missing.

cvmfs/cvmfs.cc
bbockelm: Fixed now.

PR #1345, opened 2016-01-04
FIX: Don't rely on VOMS Availability

reneme:
This fixes the `CMake` configuration process when `voms-devel` is not installed.


jblomer:
@bbockelm just adding you so that you're informed of the change.  Please let me know if you follow the PRs anyway.

PR #1348, opened 2016-01-04
FIX: Build on ARMv7 (due to size_t type mismatch)

reneme:
On ARMv7 (32 bit) `size_t` is an `unsigned int` whose pointer cannot be (implicitly) casted to `uint64_t` (aka. `unsigned long long int`). Making this explicit in one of the unit tests, should repair the build.


reneme:
Amendment: `DecompressMem2Mem()` expects a pointer to `uint64_t` but it get passed a `size_t*`.

PR #1349, opened 2016-01-04
FIX: linking (order)

jblomer:


reneme:
For reference: works on the Ubuntu1404 docker image.

PR #1354, opened 2016-01-04
CVM-935: Chunking support for external data

bbockelm:
This adds chunking support for external data files.

External data files are files that are stored in their entirety on a separate host where the LFN is based on the file's pathname.  In order to support chunking of these files, we need the fetcher / downloader to expose the ability to retrieve a byte range from the external file.

Includes updates to the external data integration test.


cvmfs/catalog_mgr_rw.cc
```diff
@@ -358,6 +358,7 @@ void WritableCatalogManager::AddFile(
 void WritableCatalogManager::AddChunkedFile(
   const DirectoryEntryBase  &entry,
   const XattrList           &xattrs,
+  const bool                 external_data,
```
jblomer: This information should already be contained in `entry`

bbockelm: Yeah - it's the same explanation as in #1339.

PR #1358, opened 2016-01-04
CVM-934: Support for chunks when grafting

bbockelm:
This adds support for specifying grafts when grafting.

*Note*: This needs to go in after #1339 in order for the integration test to succeed (and when #1339 is in I'll need to do some merge conflict fixes too) because the integration test requires compression to work with grafting.  So, please don't merge immediately.


jblomer:
@bbockelm: There should be one more integration test covering the full chain of the expected use case: graft an uncompressed, external, large file including information on partial checksums and download it on the client side with HTTP byte ranges.

PR #1359, opened 2016-01-04
CVM-964: Uncompressed graft support for swissknife graft command

bbockelm:
This adds `cvmfs_swissknife graft` CLI support for calculation of graft files of uncompressed files.


jblomer:
@bbockelm: I changed the default compression algorithm to `none` in the graft utility, assuming that most people will use it for uncompressed files.  I also added the optional `-a` parameter to specify the hash algorithm.

PR #1363, opened 2016-01-05
Add cvmfs_server snapshot -a option

DrDaveD:
Refer to pr #1315 and [CVM-813](https://sft.its.cern.ch/jira/browse/CVM-813)

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -338,6 +338,10 @@ Supported Commands:
   snapshot      [-t fail if other snapshot is in progress]
                 <fully qualified name>
                 Synchronize a Stratum 1 replica with the Stratum 0 source
+  snapshot -a   [-s use separate logs in /var/log/cvmfs for each repository]
+                [-n do not warn if /etc/logrotate.d/cvmfs does not exist]
+                [-i skip repositories that have not run initial snapshot]
+                Do snapshot on all active replica repositories
   migrate       <fully qualified name>
                 Migrates a repository to the current version of CernVM-FS
   list          List available repositories
@@ -4915,8 +4919,10 @@ __snapshot_failed() {
   to_syslog_for_repo $alias_name "failed to snapshot from $CVMFS_STRATUM0"
 }
 
-snapshot() {
-  local alias_names
+__do_snapshot() {
+  local alias_names="$1"
+  local abort_on_conflict=$2
+  local alias_name
   local name
   local user
   local spool_dir
@@ -4927,27 +4933,7 @@ snapshot() {
   local timeout
   local retries
   local retcode=0
-  local abort_on_conflict=0
   local gc_timespan=0
-
-  OPTIND=1
-  while getopts "t" option; do
-    case $option in
-      t)
-        abort_on_conflict=1
-      ;;
-      ?)
-        shift $(($OPTIND-2))
-        usage "Command snapshot: Unrecognized option: $1"
-      ;;
-    esac
-  done
-  shift $(($OPTIND-1))
-
-  # get repository names
-  check_parameter_count_for_multiple_repositories $#
-  alias_names=$(get_or_guess_multiple_repository_names $@)
-  check_multiple_repository_existence "$alias_names"
 
   for alias_name in $alias_names; do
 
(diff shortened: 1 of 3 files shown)


cvmfs/cvmfs_server
```diff
@@ -5077,6 +5063,176 @@ snapshot() {
...
+    echo "${snapshot_time}:${repo}"
+
+  done|sort -n|cut -d: -f2)"
+
+  for repo in $repos; do 
+    if [ $separate_logs == 1 ]; then
+      log=/var/log/cvmfs/$repo.log
+    fi
+
+    (
```
bbockelm: Just a thought - if you use `flock` here with the following semantics:
```
       (
         flock -s 200
         # ... commands executed under lock ...
       ) 200>/var/lock/mylockfile
```
(see `man 1 flock`), then you'd be able to eliminate all the custom lockfile management code for snapshots - POSIX locks are guaranteed by the kernel.

cvmfs/cvmfs_server
```diff
@@ -5077,6 +5063,176 @@ snapshot() {
...
+  if [ $do_all == 1 ]; then
+    [ $# == 0 ] || die "no non-option parameters expected with -a"
+
+    # ignore if there's a -t option, it's always implied with -a
+
+    __do_all_snapshots $allopts
+
+    # always return success because this is used from cron and we
+    #  don't want cron sending an email every time something fails
+    # errors will be in the log
```
bbockelm: I don't like this approach.  If admins didn't want failure notifications, they can add:
```
MAILTO=
```
in the entry in `cron.d`.

I would propose if any repo fails, then print a message containing all failed repos and return non-zero.

bbockelm:
With `snapshot -a`, can we also ship a default cronjob?  Other than differing in frequency, is there any reason why a single cronjob wouldn't work for almost every CVMFS Stratum-1?

(Similar to fetch-crl-cron, have it disabled by default and allow it to be managed via sysvinit service / systemd unit)

cvmfs/cvmfs_server
DrDaveD: This function doesn't need any lock files, so I'm not sure what you're suggesting.

bbockelm: I was just thinking out loud: `__do_snapshot` needs lockfiles and the hand-rolled, non-POSIX lockfile implementation has been finicky at best.

cvmfs/cvmfs_server
DrDaveD: Even if MAILTO is set it's a problem because it will send a message every 15 minutes if any of the snapshots have a problem. which is fairly common.  This is the way that the current cron entry that all the stratum 1s use works.   If somebody wants to look for errors they should either have their own monitor looking for ERROR in the log files or use the new cvmfs-servermon.

cvmfs/cvmfs_server
DrDaveD: Ah, well __do_snapshot is the old code.   Locking there was discussed in [CVM-810](https://sft.its.cern.ch/jira/browse/CVM-810)

DrDaveD:
The cron job for most is going to be to just run "cvmfs_server snapshot -a" but I don't think it should get installed by default.  Different stratum 1s run the cron on different minutes of the hour, plus many of them will want to do things slightly differently.  For instance RAL & FNAL use my cvmfsha-stratum1 package which runs something different from cron, and CERN alone among the big stratum 1s runs the cron as non-root.  Also some will want to use different extra options with cvmfs_server snapshot -a.

cvmfs/cvmfs_server
bbockelm: :) Yeah - note I wasn't happy with the resolution of CVM-810 either!

Anyhow, don't consider this a blocker - but it would be an improvement if you wanted to tackle it.

bbockelm:
I think there's a difference between "installed by default" and "enabled by default".  Again, look at the `fetch-crl-cron` for inspiration.  The cron job is installed by default, but is a no-op until you do `service fetch-crl-cron` start (which basically just touches a lockfile on disk.

IIRC, `fetch-crl-cron` also gives you the ability to drop a configuration file (probably `/etc/sysconfig/fetch-crl`) that allows you to customize the options.  I think that could also have a useful equivalent here.

DrDaveD:
Ok that's a nice friendly way to do it but I'm not convinced it's enough value added to justify the time to implement.  fetch-crl is a different case, lots of people need to install it on a lot of nodes, where this affects a relatively small number of people who up to this point were entirely on their own.  It's also trickier because we wouldn't want to install it until the first replica is created.

cvmfs/cvmfs_server
```diff
@@ -5077,6 +5063,176 @@ snapshot() {
...
+  shift $(($OPTIND-1))
+
+  if [ ! -d /var/log/cvmfs ]; then
+    if ! mkdir /var/log/cvmfs 2>/dev/null; then
+      die "/var/log/cvmfs does not exist and could not create it"
+    fi
+  fi
+  [ -w /var/log/cvmfs ] || die "cannot write to /var/log/cvmfs"
+
+  if [ $logrotate_nowarn == 0 ] && [ ! -f /etc/logrotate.d/cvmfs ]; then
```
jblomer: Isn't `==` bash specific, i.e. shouldn't this be `-eq`?

jblomer:
Looks good to me, @reneme please take another look.

PR #1364, opened 2016-01-05
Sorted external hosts using CVMFS Stratum-1s.

bbockelm:
The external data hosts are assumed to not be running the GeoAPI
service.  Instead, use the CVMFS Stratum-1 servers listed in CVMFS_URL
to sort the external data hostnames with the GeoAPI.

@DrDaveD - this is the change we discussed earlier today.  Can you take a look at it?

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -2593,7 +2593,8 @@ static int Init(const loader::LoaderExports *loader_exports) {
   cvmfs::external_download_manager_ = new download::DownloadManager();
   cvmfs::external_download_manager_->Init(cvmfs::kDefaultNumConnections, false,
       cvmfs::statistics_, "download-external");
-  cvmfs::external_download_manager_->SetHostChain(external_host.size() ?
+  
+  cvmfs::external_download_manager_->SetHostChain(!external_host.empty() ?
                                                   external_host : hostname);
   if ((dns_timeout_ms != download::DownloadManager::kDnsDefaultTimeoutMs) ||
       (dns_retries != download::DownloadManager::kDnsDefaultRetries))
@@ -2644,9 +2645,16 @@ static int Init(const loader::LoaderExports *loader_exports) {
   g_external_download_ready = true;
   if (use_geo_api) {
     std::vector<std::string> host_chain;
-    cvmfs::external_download_manager_->GetHostInfo(&host_chain, NULL, NULL);
-    cvmfs::download_manager_->GeoSortServers(&host_chain);
-    cvmfs::external_download_manager_->SetHostChain(host_chain);
+    // If no external host was specified, reuse the geo API ordering
+    // of the regular download manager.
+    if (external_host.empty()) {
+      cvmfs::download_manager_->GetHostInfo(&host_chain, NULL, NULL);
+      cvmfs::external_download_manager_->SetHostChain(host_chain);
+    } else {
+      cvmfs::external_download_manager_->GetHostInfo(&host_chain, NULL, NULL);
+      cvmfs::download_manager_->GeoSortServers(&host_chain);
+      cvmfs::external_download_manager_->SetHostChain(host_chain);
+    }
   }
 
 
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -2643,7 +2643,10 @@ static int Init(const loader::LoaderExports *loader_exports) {
     download::DownloadManager::kSetProxyBoth);
   g_external_download_ready = true;
   if (use_geo_api) {
-    cvmfs::external_download_manager_->ProbeGeo();
+    std::vector<std::string> host_chain;
+    cvmfs::external_download_manager_->GetHostInfo(&host_chain, NULL, NULL);
+    cvmfs::download_manager_->GeoSortServers(&host_chain);
+    cvmfs::external_download_manager_->SetHostChain(host_chain);
   }
 
 
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -1717,25 +1717,7 @@ void DownloadManager::GetTimeout(unsigned *seconds_proxy,
  * removes the host list.
  */
 void DownloadManager::SetHostChain(const string &host_list) {
-  pthread_mutex_lock(lock_options_);
-  opt_timestamp_backup_host_ = 0;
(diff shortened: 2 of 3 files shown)


cvmfs/download.cc
```diff
@@ -1739,6 +1739,30 @@ void DownloadManager::SetHostChain(const string &host_list) {
...
+  }
+
+  opt_host_chain_ = new vector<string>(host_list);
+  opt_host_chain_rtt_ =
+    new vector<int>(opt_host_chain_->size(), kProbeUnprobed);
+  // LogCvmfs(kLogDownload, kLogSyslog, "using host %s",
+  //          (*opt_host_chain_)[0].c_str());
+  pthread_mutex_unlock(lock_options_);
+}
+
```
DrDaveD: The two SetHostChain() functions are almost identical, can't one be implemented by calling the other?

DrDaveD:
Overall it looks reasonable to me.  I haven't tested it though, I hope you have.  In fact the 054-geoapi test should probably be expanded to test sorting of external servers.

bbockelm:
Yup - it tests fine locally.  Looking at 054-geoapi, is there really any way to run it outside CERN?

DrDaveD:
It was limited to CERN until I [fixed](https://github.com/cvmfs/cvmfs/commit/37c0d431da817f33251f6a8f6288fee052f3f302) it 3 weeks ago.  Now it runs for me at Fermilab and should run elsewhere.  I see that it skips one check outside of CERN, return code 20. 

I just ran 054-geoapi on a fermicloud machine (without this pull request included) and it got almost all the way through but failed on the last test, 39.  Ah, that is because it's not at CERN after all.  I'm not sure how I missed it before.  It's comparing the order from a DIRECT proxy to an externally available proxy at CERN; that stays the same order at CERN but opposite order at FNAL.  I'll submit a patch tomorrow to have 054-geoapi do all checks correctly at CERN or FNAL and skip 20 and 39 elsewhere, like is done in the 614-geoservice order checks.

bbockelm:
If you're fixing up the tests, can you also map unl.edu properly?  Should be the same ordering as FNAL.

jblomer:
@bbockelm @DrDaveD Would we actually see multiple external host URLs?  I though for data there is something like a global redirector which acts as a central entry point.

DrDaveD:
> Would we actually see multiple external host URLs? I though for data there is something like a global redirector which acts as a central entry point.

I asked Brian that yesterday and he said that the redirector was upstream of the external URL servers; those servers still need to be geosorted on the client.

cvmfs/download.cc
```diff
@@ -1955,6 +1979,87 @@ void DownloadManager::ProbeHosts() {
...
+  }
+
+  std::vector<std::string> host_chain;
+  GetHostInfo(&host_chain, NULL, NULL);
+
+  std::vector<std::string> server_dns_names;
+  server_dns_names.reserve(servers->size());
+  for (unsigned i = 0; i < servers->size(); ++i) {
+    std::string host = dns::ExtractHost((*servers)[i]);
+    server_dns_names.push_back(host.empty() ? (*servers)[i] : host);
```
jblomer: Currently merging, I'm wondering if `host.empty() ?` is necessary.  Can we just push back the result from `ExtractHost`?

cvmfs/cvmfs.cc
```diff
@@ -2643,7 +2643,10 @@ static int Init(const loader::LoaderExports *loader_exports) {
     download::DownloadManager::kSetProxyBoth);
   g_external_download_ready = true;
   if (use_geo_api) {
-    cvmfs::external_download_manager_->ProbeGeo();
+    std::vector<std::string> host_chain;
```
jblomer: This runs the Geo-API twice even if there are no dedicated external servers.

bbockelm:
Yup - OSG runs a network of about 5 large-scale proxy servers ("StashCache" - each has 10TB storage, 10Gbps) that are meant to serve as a front for the data federation of smaller VOs (in some cases, the "federation" may be in reality be a single NFS server).

So we do expect non-trivial list of EXTERNAL_URLs in this case.

cvmfs/cvmfs.cc
bbockelm: It shouldn't.  There's an early-exit in the case where the sorting is trivial (less than 2 servers).

cvmfs/download.cc
bbockelm: Yeah, that's an easy fix.

cvmfs/download.cc
bbockelm: It is necessary as `ProbeGeo` calls with a list of hostnames but the initialization of `external_download_manager_` calls with a list of URLs.

cvmfs/cvmfs.cc
jblomer: Ah, I think the problem is that if external is the same as the normal urls, there is no need to do the sorting twice.  I.e., if `CVMFS_EXTERNAL_URL`, the external download manager should just apply the already sorted list from the normal one.

cvmfs/download.cc
```diff
@@ -1955,6 +1961,87 @@ void DownloadManager::ProbeHosts() {
...
+  }
+  std::string host_list = JoinStrings(server_dns_names, ",");
+
+  // Protect against concurrent access to prng_
```
jblomer: There is a duplicate of this block in `ProbeGeo()`, which I think should be removed.

cvmfs/download.cc
```diff
@@ -1955,6 +1961,87 @@ void DownloadManager::ProbeHosts() {
...
+        break;
+      }
+    } else {
+      LogCvmfs(kLogDownload, kLogDebug | kLogSyslogWarn,
+               "GeoAPI request %s failed with error %d [%s]",
+               url.c_str(), result, Code2Ascii(result));
+    }
+  }
+  if (!success) {
+    LogCvmfs(kLogDownload, kLogDebug | kLogSyslogWarn,
```
jblomer: There is a duplicate of this log in `ProbeGeo()`, which I think should be removed.

cvmfs/cvmfs.cc
bbockelm: Ah, ok.  That's an easy fix then.

bbockelm:
Ok, I think I got all the review comments fixed.

PR #1372, opened 2016-01-06
FIX: Minor improvements to `cvmfs_server repo-metainfo`

reneme:
This fixes some minor problems with `cvmfs_server repo-metainfo`:

* add missing bash completion (along with `cvmfs_server update-info`)
* typo in `edit_json_unTil_valid()`
* default JSON skeleton wasn't valid JSON

Furthermore I'd like to suggest two things:

1. rename `cvmfs_server repo-metainfo` to `cvmfs_server update-repo-info` to match its name to `cvmfs_server update-info` and show that both are related
2. create the dummy meta info by default on `cvmfs_server mkfs`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,7 +3,7 @@
   * Print error message at the end of a failing `cvmfs_server check` (CVM-958)
   * Use patched pacparser 1.3.5 for IPv6 support (CVM-903)
   * Add support for VOMS authentication in fuse module (CVM-904)
-  * Add `cvmfs_server repo-metainfo` command (CVM-804)
+  * Add `cvmfs_server update-repoinfo` command (CVM-804)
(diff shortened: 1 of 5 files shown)


jblomer:
Sounds good!  Can you add these two points?

reneme:
Renamed to `cvmfs_server update-repoinfo`, some usage string cosmetics for `cvmfs_server`, fixed a leaked temporary file, some typos and added a default repo info at `cvmfs_server mkfs`.

HOWEVER: The pointer to the repoinfo file is lost when the repository is published after it has been updated... working on that...

reneme:
In principle that can go in.... it will break the 618 unit test, as I added some regression test for the issue described above. But since the fix will be a bigger and more common change, I'd put that into a separate pull request.

PR #1380, opened 2016-01-08
Refactor: Manifest, Download- and SignatureManager handing in `cvmfs_swissknife`

reneme:
This aims at centralising the handling of `DownloadManager` and `SignatureManager` in `cvmfs_swissknife`'s `Command` classes while getting rid of some global state in the meanwhile.  Additionally there are now helper functions for `Manifest` downloading in `cvmfs_swissknife`. It's not perfect yet, but arguably the right direction.

Furthermore I factored out the `main()` function of `cvmfs_swissknife` into its own compilation unit to allow `cvmfs_preload` to benefit from the same helper methods.

This is quite a change set, but most of it is just API adaptions.


cvmfs/swissknife_main.cc
```diff
@@ -0,0 +1,173 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include <string>
+
+#include "cvmfs_config.h"
```
jblomer: should be before `#include <string>`

PR #1381, opened 2016-01-08
FIX: Handle Manifest Committing as Updates

reneme:
**Note:** This builds on [Refactor: Manifest, Download- and SignatureManager handing in `cvmfs_swissknife`](https://github.com/cvmfs/cvmfs/pull/1380).

This changes the handling strategy of the `Manifest` (i.e. `.cvmfspublished) by using the previous `.cvmfspublished` as an update basis, instead of recreating it from scratch for every `cvmfs_swissknife sync`. The advantage being, that "passive" manifest fields are just carried over and don't need to be recreated every time. It makes the addition of those fields less error prone (see [here](https://github.com/cvmfs/cvmfs/pull/1372) for example).

### DIFF
--- a/cvmfs/swissknife_sync.cc
+++ b/cvmfs/swissknife_sync.cc
@@ -589,6 +589,19 @@ int swissknife::CommandSync::Main(const swissknife::ArgumentList &args) {
   if (!this->InitDownloadManager(follow_redirects)) {
     return 3;
   }
+
+  if (!this->InitSignatureManager(params.public_keys, params.trusted_certs)) {
+    return 3;
+  }
+
+  UniquePtr<manifest::Manifest> manifest;
+  manifest = this->FetchRemoteManifest(params.stratum0,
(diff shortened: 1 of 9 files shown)


cvmfs/swissknife_sync.cc
```diff
@@ -582,6 +588,19 @@ int swissknife::CommandSync::Main(const swissknife::ArgumentList &args) {
...
+  if (!this->InitSignatureManager(params.public_keys, params.trusted_certs)) {
+    return 3;
+  }
+
+  UniquePtr<manifest::Manifest> manifest;
+  manifest = this->FetchRemoteManifest(params.stratum0,
+                                       params.repo_name,
+                                       params.base_hash);
+  if (!manifest) {
+    return 3;
```
jblomer: I expect that this can happen and would be the first point where a network glitch shows (e.g. with the S3 backend).  There should be an error message.

PR #1382, opened 2016-01-08
Feature: Add CernVM-FS Version String to Repository Manifests

reneme:
This adds the CernVM-FS version string in the `.cvmfspublished` manifest for the rationale described in [CVM-817](https://sft.its.cern.ch/jira/browse/CVM-817). The field name will be `V` and is set during manifest signing.

**Note:** This contains [Refactor: Manifest, Download- and SignatureManager handing in `cvmfs_swissknife`](https://github.com/cvmfs/cvmfs/pull/1380/files) and [FIX: Handle Manifest Committing as Updates](https://github.com/cvmfs/cvmfs/pull/1381).

### DIFF
--- a/cvmfs/manifest.cc
+++ b/cvmfs/manifest.cc
@@ -39,6 +39,7 @@ Manifest *Manifest::Load(const map<char, string> &content) {
   // Required keys
   shash::Any catalog_hash;
   shash::Md5 root_path;
+  std::string creator_version;
   uint32_t ttl;
   uint64_t revision;
 
@@ -56,6 +57,10 @@ Manifest *Manifest::Load(const map<char, string> &content) {
   if ((iter = content.find('S')) == content.end())
     return NULL;
   revision = String2Uint64(iter->second);
+  if ((iter = content.find('V')) == content.end())
+    return NULL;
+  creator_version = iter->second;
+
 
   // Optional keys
   uint64_t catalog_size = 0;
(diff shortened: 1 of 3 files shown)


jblomer:
While we don't necessarily need the version string in the manifest, we do want to have it as a property in the catalog.  The information in the manifest should then be a plain export from what's in the catalog.

Also, I think that updating the creator version with `VERSION` on every commit changes its semantics.  It would then contain the cvmfs version that committed the last update, not the version that was used to initially create the catalog.

reneme:
Well, in the catalogs we already have the catalog revision for that purpose. I can add the CernVM-FS version there too, but what would be the added value?

Maybe 'creator version' is the wrong name here: The idea is, that future versions of CernVM-FS would refuse to update a repository whose latest revision has been published with an incompatible older version (or take automatic migration measures before).

jblomer:
I see. I'm not yet fully convinced that this will work.  As far as catalog migration is concerned, indeed the catalog revision already stored should be sufficient.  As for the release manager machine and its configuration, I think it's wrong to store its version in the manifest because the details of the release manager machine are not part of the repository.  This would lead to confusion when we in the future have multiple release manager machines (possibly with different software versions) working on the same repo. Let's discuss tomorrow.

reneme:
I see three different things, that need to be versioned:

* The release manager machine infrastructure (`server.conf` - CVMFS_CREATOR_VERSION)
* The catalog revisions (in the individual catalogs)
* The repository storage revision (`.cvmfspublished` - `V` field)

Even with multiple release manager machines accessing the same backend storage, we'll probably need all three. Even though, I agree that in this case the CernVM-FS version number might not be the best fit but rather an independent storage revision number...

jblomer:
Ok, let's link to these useful notes in the ticket, move the target for 2.3.0 and free the `V` key in the manifest for the time being.

PR #1384, opened 2016-01-08
FIX: Missing `chmod` in `test_functions` Messes up Environment

reneme:
Making `/etc/cvmfs/.../server.conf` only readable by root blocks things like `cvmfs_server list`.


jblomer:
We are sure that the server.conf doesn't contain any credentials, e.g. for the storage backend, right?

jblomer:
oh, test_functions.  Never mind.

PR #1392, opened 2016-01-12
Fix CURL flags

bbockelm:
When a file descriptor is ready for _writing_ (POLLOUT), it was incorrectly being marked as ready for _reading_ in `libcurl`.  This caused busy wait loops in the download thread, where `libcurl` would try to read from a specified descriptor and keep getting EAGAIN.

In testing, this gave a respectable 10% speedup in downloads.


jblomer:
Cool!  Thanks!  How did you spot this?

bbockelm:
I noticed this while `strace`'ing the cvmfs2 process while doing performance studies.  The raw IO performance is still behind where I want it - may be an item for 2.3.x.

jblomer:
That is an interesting topic for 2.3.x.  I'm pretty sure there is room for improvement since cvmfs was not yet optimized for high throughput.  There might be some architectural limits, one of them could be the fuse layer.

PR #1400, opened 2016-01-13
Feature: Non-Interactive `cvmfs_server update-repoinfo`

reneme:
This adds `cvmfs_server update-repoinfo -f` that allows for non-interactive updates of the repository's JSON meta data. Simply pass a JSON file path with the `-f`. If `cvmfs_server` finds `jq` in the system, it uses it to validate the passed in JSON file. Furthermore this updates integration test 618 to also use this code path.


reneme:
Appending 618 for testing Stratum1 pulling of meta info and fixing it for S3 backends.

PR #1404, opened 2016-01-14
Avoid busy-wait when downloading files

bbockelm:
CVMFS currently wakes up once a millisecond to check if libcurl has any ongoing transfers which have timed out.  As the granularity of our timeouts is in seconds, this is overkill: we can back off and save system CPU time.

libcurl actually provides a callback interface where it will inform CVMFS when the next upcoming library timeout will occur.  However, due to the proximity of the release, I simply increase the granularity.

### DIFF
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -477,7 +477,12 @@ void *DownloadManager::MainDownload(void *data) {
   while (true) {
     int timeout;
     if (still_running) {
-      timeout = 1;
+      // Specify a timeout for polling in ms; this allows us to return
+      // to libcurl once a second so it can look for internal operations
+      // which timed out.  libcurl has a more elaborate mechanism
+      // (CURLMOPT_TIMERFUNCTION) that would inform us of the next potential
+      // timeout.  TODO(bbockelm) we should switch to that in the future.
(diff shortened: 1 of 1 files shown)


jblomer:
1 second is too coarse-grained I think (timeouts can then be off by 1s).  How about 50ms or 100ms?

bbockelm:
100ms sounds fine to me: the difference between 1000 and 100ms is likely minimal.  1ms is just too small.

jblomer:
@bbockelm: According to our continuous benchmarks, at the time of the change the performance of the downloader seriously degraded.  We'll change it back and see what happens.

Moliholy:
Here's an screenshot of the download time in the last 90 days:
screen shot 2016-01-21 at 11 10 38

bbockelm:
Hi,

That's fine: it's not worth delaying 2.2.0 over.

Unfortunately, if this causes any impact of download times, I suspect the poll interval of 1ms is actually hiding a different bug.  Perhaps one of the file descriptors is missing from the poll set?

Regardless, in 2.3.0, we can tackle the more thorough fix mentioned above (utilizing `libcurl`'s callback handler to get the appropriate poll timeout interval).

Brian

PR #1414, opened 2016-01-15
FIX: Integration Test 610

reneme:
This fixes integration test 610 (Alternative catalog path) by separating the creation of `.cvmfsalt-` bootstrapping shortcut symlinks from the `CVMFS_VOMS_AUTHZ` server parameter.

Furthermore it does a small refactoring in `cvmfs_server mkfs` moving the creation of  `CVMFS_VOMS_AUTHZ` to `create_config_files_for_new_repository`. Rationale is, that *all* configuration parameters for new repositories are done there centrally.


reneme:
Note: rebased and `push -f`-ed to resolve merge conflict in `cvmfs_server`.

PR #1420, opened 2016-01-15
Avoid using uninitialized output size variable

bbockelm:
Make sure we initialize `avail_out`; if the memory happened to be 0, this would trigger an infinite loop in `swissknife graft`.

@reneme


reneme:
confirmed to work for me. Thanks!

PR #1425, opened 2016-01-18
FIX: Cache Expiry HTTP Headers

reneme:
This integrates the fix provided by @DrDaveD and described in [CVM-974](https://sft.its.cern.ch/jira/browse/CVM-974) along with a regression test.

RFC: Do you think it is okay to recreate the apache config file from scratch during migration from 2.1.20 instead of doing the `sed`-stunt I pushed in here? Or might that cause issues during operation?

**Note:** Putting this here for discussion. Please don't merge yet, I need to iterate on the config migration step and test it.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -2076,7 +2076,6 @@ Alias /cvmfs/${name} ${storage_dir}
 
     ExpiresActive On
     ExpiresDefault "access plus 3 days"
-    ExpiresByType text/html "access plus 5 minutes"
     ExpiresByType application/x-cvmfs "access plus 2 minutes"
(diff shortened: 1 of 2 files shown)


DrDaveD:
I don't think it should be a problem to overwrite those config files.  People are not supposed to edit them, as the starting comment says.  Don't create them if they don't exist though, some admins remove them.

reneme:
This now slightly adapts the apache configuration to deal with the JSON files in `/cvmfs/info/v1/*.json`. Ready for merging... However, I need to have a look over the migration code again - details in [CVM-950](https://sft.its.cern.ch/jira/browse/CVM-950).

PR #1438, opened 2016-01-20
Fix for cvmfs_config killall

jblomer:
Fixes the usage string and the killing of cvmfs2 processes after a hotpatch from a previous version.


Moliholy:
Precisely one minute ago @reneme and I had a discussion about this, and now it's fixed. The problem on mac OS X is that it doesn't support the -q (quiet) option, and also passing a full path. Soon I will submit that PR and we can take a look

PR #1439, opened 2016-01-20
Mac integration tests 2

Moliholy:
Continuation of https://github.com/cvmfs/cvmfs/pull/1300

Rest of the tests for mac OS X fixed. Not all of them are included, of course, but I'd say a good number of them are.

I know right now there are merge conflicts, but there are certain fixes that might deserve special attention, like https://github.com/cvmfs/cvmfs/pull/1438


cvmfs/cvmfs_config
```diff
@@ -1144,7 +1155,7 @@ cvmfs_killall() {
   echo "OK"
 
   echo -n "Terminating cvmfs2 processes... "
-  killall -q -KILL ${INSTALL_BASE}/bin/cvmfs2
```
Moliholy: @jblomer this is a merge conflict. Is this solution compatible with the one submitted in https://github.com/cvmfs/cvmfs/pull/1438?

test/cloud_testing/platforms/osx_x86_64_test.sh
```diff
@@ -1,34 +1,68 @@
...
+                          src/058-keysdir                         \
+                          src/059-fallbackproxy                   \
+                          src/060-hidexattrs                      \
+                          src/062-loadtag                         \
+                          src/063-uidmap                          \
+                          src/064-fsck                            \
+                          src/065-http-400                        \
+                          src/066-killall                         \
+                          src/067-wpad                            \
+                          src/068-rocache
```
reneme: Now that there are more tests *enabled* than *disabled* I would change this list to _exclude_ incompatible tests actually. First of all, because it's shorter and second, because it automatically runs new tests by default. Like:

```bash
./run.sh $logfile -x src/004-davinci
                     src/005-asetup
                     src/006-buildkernel
                     src/007-testjobs
                     src/008-default_domain
                     src/016-dnsunreachable
                     src/017-dnstimeout
                     src/024-reload-during-asetup
                     src/039-reloadalarm
                     src/040-aliencache
                     src/045-oasis
                     src/052-roundrobindns
                     src/055-ownership
                     src/056-lowspeedlimit
                     src/057-parallelmakecache
                     src/061-systemdnokill
                     --
                     src/0*
```

Furthermore we should look into running the tests as a non-privileged user.

test/src/060-hidexattrs/main
```diff
@@ -1,11 +1,20 @@
...
 cvmfs_test_name="Hide extended attributes"
 
+list_xattrs() {
+  local repo=$1
+  if running_on_osx; then
+    xattr $repo
+  else
+    attr -ql $repo
+  fi
+}
```
reneme: Perhaps move `list_xattrs()` next to `get_xattr()` into `test_functions`?

test/src/063-uidmap/main
```diff
@@ -4,15 +4,15 @@ cvmfs_test_name="Uid/Gid Mapping of files and directories"
...
+  local uid_map=$(pwd)/uid.map
+  local gid_map=$(pwd)/gid.map
 
   echo "* 1337" > "$uid_map"
   echo "* 1338" > "$gid_map"
   chmod 0444 "$uid_map" "$gid_map"
 
   cvmfs_mount grid.cern.ch "CVMFS_CLAIM_OWNERSHIP=no" \
-    "CVMFS_UID_MAP=$uid_map" "CVMFS_GID_MAP=$gid_map"
+    "CVMFS_UID_MAP=$uid_map" "CVMFS_GID_MAP=$gid_map" "CVMFS_DEBUGLOG=/tmp/cvmfs-test.log"
```
reneme: `CVMFS_DEBUGLOG` was for debugging only and doesn't need to be in the vanilla test.

Moliholy: Yep, I forgot to remove it

test/cloud_testing/platforms/osx_x86_64_test.sh
Moliholy: Yes, that's a pending issue I'll deal with very soon

Moliholy: On the other hand: that would imply that new tests will be automatically added to the mac list. Is that what we'd like, or would we prefer to have these kind of tests under strict control?

Moliholy:
Also notice that the test 052-roundrobindns could also be fixed to be included. Anyway there's no need to do it in this PR.

test/cloud_testing/platforms/osx_x86_64_test.sh
reneme: That's definitely what we want! When new tests are automatically added on the Mac, they either just work (which is good) or fail and scream with a red button in Jenkins. Then we can react (either by fixing it or by blacklisting the test). This way we cannot possibly forget to deal with the mac platform.

Moliholy: That totally makes sense. It's now changed :)

test/test_functions
```diff
@@ -1983,7 +2026,7 @@ purge_disk_cache() {
 try_automount() {
```
reneme: `try_automount()` is a misleading name here because the function either tries to *auto mount* or mounts manually. However when reading `try_automount()` I would expect it to "try to automount" or do nothing on failure. What about renaming it to: `try_automount_or_manual_mount()`?

Moliholy: Hmmm that's not exactly what happens, I think. On Mac OS X there is no automount, so we have to manually mount it to cause **the most similar effect**. Of course this is not a real automount, and nothing guarantees that (for linux or mac) it's going to be either "mounted" or "automounted", depending on the OS. In both cases if it fails for whatever reason it does nothing, which is exactly what it's supposed to do.

So I see that the function is "trying to trigger the automount" cheating a bit on mac, but essentially doing that.

test/src/023-reload_safe_path_traversal/main
```diff
@@ -20,13 +20,23 @@ cvmfs_run_test() {
   local reload_cmd
 
   # mount sft repository
-  cvmfs_mount sft.cern.ch \
-    "CVMFS_KCACHE_TIMEOUT=5" \
-    "CVMFS_MAX_RETRIES=3" || return 1
+  if running_on_osx; then
+    cvmfs_mount sft.cern.ch \
+      "CVMFS_KCACHE_TIMEOUT=5" \
+      "CVMFS_MAX_RETRIES=3" || return 11
```
jblomer: The if branch and the else branch are the same except for the return value.  Is this intended?

test/src/023-reload_safe_path_traversal/main
```diff
@@ -20,13 +20,23 @@ cvmfs_run_test() {
...
+    cvmfs_mount sft.cern.ch \
+      "CVMFS_KCACHE_TIMEOUT=5" \
+      "CVMFS_MAX_RETRIES=3" || return 1
+  fi
 
   # decrease the kernel cache timeout to safe some time
   sudo cvmfs_config umount || return 4
-  sudo cvmfs_config probe  || return 5
+  if running_on_osx; then
+    cvmfs_mount_direct sft.cern.ch
```
jblomer: there should be an `|| return 5` at the end

test/src/041-rocache/main
```diff
@@ -1,37 +1,74 @@
 
 cvmfs_test_name="Read-only cache"
 
+lazy_umount() {
+  local dir=$1
+  if running_on_osx; then
+    sudo umount "$dir"
```
jblomer: This is not the "lazy unmount" on OS X, is it?

Moliholy: There is no "lazy" umount option in OS X, so it simply unmounts it.

jblomer: Well, there is `diskutil umount force $mountpoint`...

jblomer: Ah, this function isn't used at all on OS X, right?  Then you can just remove the OS X branch in this function.

Moliholy: It is used in the cleaup() function for both linux and OSX in this very file, but it can't be a lazy unmount, so instead it's a normal one

Moliholy: True, I didn't realize of it

PR #1453, opened 2016-01-23
Build with VOMS on relevant platforms.

bbockelm:
We should build with VOMS on platforms where VOMS should be available (definitely RHEL and Fedora; not sure about SUSE).  Otherwise, all the secure CVMFS stuff is disabled.

*Note*: I don't know much about the build environment for the nightlies; something else in the package might need to be adjusted to make sure `voms-devel` is available.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,5 @@
 2.2.0:
+  * Add VOMS as a build dependency on platforms where it is available.
   * Detect missing 'http:// proxy prefix in chksetup (CVM-979)
   * Remove sudo dependency from Linux packages
   * Fix race when reloading at the same time as evicting data from the cache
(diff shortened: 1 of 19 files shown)


jblomer:
Please add a line to the changelog and modify the `Dockerfile` files in `ci/docker/<platform>` so that voms-devel gets installed by yum/dnf.

jblomer:
It probably makes sense to have the same capabilities on all the platforms.  I think voms-devel exists almost everywhere.  Let's first try to activate it unconditionally, also in the deb package.

jblomer:
After a quick check: voms-devel is available for all EL (including EL4).  Also for Ubuntu 12.04 (and later, I guess), where it is called `voms-dev`.  I didn't find it in the standard SuSE or openSuSE repos.  SuSE is popular on supercomputers though, so I think we should have a solution for that for cvmfs 2.3.

bbockelm:
It's not for this ticket, but you probably want to switch yum-based platforms to `yum builddep`.  Given a SRPM, this automatically resolves and installs the build dependencies.

http://linux.die.net/man/1/yum-builddep

Much better than listing them one-by-one.

jblomer:
Thanks!  I meant actually the changelog that is part of the rpm (the cvmfs changelog entry can stay I think).

PR #1462, opened 2016-01-26
FIX: Tonight's Integration Tests

reneme:
Download GeoIP database from a copy located at CERN, as download directly from MaxMind failed transiently. Fix: Several minor things for the hotpatching test.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -34,7 +34,7 @@ CVMFS_UPDATEGEO_HOUR=10 # First hour of day for update, 0-23, default 10am
 CVMFS_UPDATEGEO_MINDAYS=25 # Minimum days between update attempts
 CVMFS_UPDATEGEO_MAXDAYS=100 # Maximum days old before considering it an error
 
-CVMFS_UPDATEGEO_URLBASE="http://geolite.maxmind.com/download/geoip/database"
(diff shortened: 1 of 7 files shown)


cvmfs/cvmfs_server
```diff
@@ -34,7 +34,7 @@ CVMFS_UPDATEGEO_HOUR=10 # First hour of day for update, 0-23, default 10am
 CVMFS_UPDATEGEO_MINDAYS=25 # Minimum days between update attempts
 CVMFS_UPDATEGEO_MAXDAYS=100 # Maximum days old before considering it an error
 
-CVMFS_UPDATEGEO_URLBASE="http://geolite.maxmind.com/download/geoip/database"
+CVMFS_UPDATEGEO_URLBASE="${CVMFS_UPDATEGEO_URLBASE:=http://geolite.maxmind.com/download/geoip/database}"
 CVMFS_UPDATEGEO_URLBASE6="${CVMFS_UPDATEGEO_URLBASE}/GeoLiteCityv6-beta"
```
DrDaveD: This could have been overridden in /etc/cvmfs/cvmfs_server_hooks.sh instead, which is the documented way.

Also, ":=" already assigns a default value to the variable being substituted as a side effect; it is more proper to use ":-".

reneme: Thanks for the valid comments. Since I use the same variable (i.e. `CVMFS_UPDATEGEO_URLBASE`) the ":=" shouldn't do any harm here. Though I wasn't aware of this side effect.

Anyway, I'll have a look if I can use `cvmfs_server_hooks.sh` for the same purpose.

PR #1464, opened 2016-01-27
Fix: remove sha256 and sha3-256 but keep sha3-shake128

jblomer:
It turns out that a combination of two things can ruin the hotpatch from 2.1.20 to 2.2.0:
  - A race during the reload can result in old quota manager process staying alive
  - Due to the increased length of sha256 and sha3-256, the memory layout of the data sent between fuse processes and the quota manager changed.  So reloaded fuse modules send data that are garbage to old quota managers.

As for the race, this is an issue of already installed versions so we can't possibly fix it with newer versions.  The fix itself requires redoing most of the cvmfs initialization and is postponed to a later release.

Instead we decided to drop SHA256 and SHA3-256.  We still keep the support for the SHA3 variant SHAKE128 with 160 output bits.  SHAKE128 with variable number of output bits is standardized by NIST. We use it as an unbroken SHA-1, i.e. having the full 80 security bits.  Additional advantage is that all memory structures holding content hashes are back from 32 bytes to 20 bytes.

Drawbacks: 
  - We can't graft SHA256 and SHA3-256.
  - It was possible to create SHA256 catalogs with nightly builds of cvmfs in the last few months.  We never supported or announced this.  Such catalogs would be garbage for the 2.2.0 release.  Using an appropriate nightly build, such repositories could be turned back to SHA1 and would then be usable by the release.

Nothing prevents us from adding SHA-2 and SHA3-256 to a later release.

@DrDaveD @bbockelm please note this relatively big last-minute change.


reneme:
Looks good to me on first glance. I'd add a `TODO` in the updated unit test code with a pointer to the Git commit hash that removed the SHA3 related tests... You'll be happy to have it at some point, I guess. :-)

jblomer:
[Good point](https://github.com/cvmfs/cvmfs/commit/8c7b5cd66e37ad086ceae8d6a41fd88854869d85)

PR #1475, opened 2016-01-28
Travis integration

Moliholy:
This PR:
- Hides the complexity of .travis.yml by calling a custom script that passes the cpplint, builds and (if on linux) runs the unit tests.
- Fixes a small error that makes cpplint fail so that when this PR is approved it directly successes.

An example of the results can be found here: https://travis-ci.org/Moliholy/cvmfs/jobs/105384291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,15 +32,7 @@ addons:
       - lldb
 
 before_script:
-  - mkdir -p build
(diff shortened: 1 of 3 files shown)


Moliholy:
Well, actually an example of the results will be found in travis after this PR is analyzed :-)

PR #1476, opened 2016-01-28
FIX setting time on Ubuntu 15.10 in integration test

jblomer:


test/src/052-roundrobindns/main
```diff
@@ -180,9 +180,15 @@ cleanup_dns_test() {
 cvmfs_run_test() {
   logfile=$1
 
+  # Necessary on Ubuntu 15.10
+  sudo systemctl stop systemd-timesyncd
```
reneme: doesn't that choke on non-systemd systems?

jblomer: it should just print an error and continue, no?

PR #1487, opened 2016-01-29
Fix coverity-detected bug

jblomer:


cvmfs/cvmfs.cc
```diff
@@ -1900,7 +1903,6 @@ bool Pin(const string &path) {
         return false;
       }
       cache_manager_->Close(fd);
-      return true;
```
DrDaveD: Maybe coverity called it a bug, but it doesn't change functionality at all as far as I can tell.  Does it?  It just eliminates an unnecessary line of code.

reneme: This is a scope of a for-loop. So it does change functionality, because the loop is always executed once at max.

DrDaveD: Oh, I see.  I looked at the source and thought I had traced it back to an if, my mistake.

PR #1488, opened 2016-01-29
Various high-priority Coverity fixes

Moliholy:


cvmfs/s3fanout.cc
```diff
@@ -877,6 +877,7 @@ S3FanoutManager::S3FanoutManager() {
   opt_backoff_max_ms_ = 0;
   opt_ipv4_only_ = false;
 
+  available_jobs_ = 0;
```
jblomer: This is a pointer and should be initialized to `NULL`

cvmfs/swissknife_pull.cc
```diff
@@ -536,6 +536,7 @@ int swissknife::CommandPull::Main(const swissknife::ArgumentList &args) {
 
   if (!this->InitSignatureManager(master_keys, trusted_certs)) {
     LogCvmfs(kLogCvmfs, kLogStderr, "failed to initalize CVMFS signatures");
+    free(workers);
```
jblomer: Here I think a better fix is to initialize `workers` first to NULL and to move the actual memory allocation for `workers` down to where it is needed.

cvmfs/util.cc
```diff
@@ -755,12 +755,13 @@ bool GetUidOf(const std::string &username, uid_t *uid, gid_t *main_gid) {
...
   char buf[16*1024];
   struct group grp;
-  struct group *result;
-  getgrnam_r(groupname.c_str(), &grp, buf, sizeof(buf), &result);
-  if (result == NULL)
-    return false;
-  *gid = result->gr_gid;
-  return true;
+  struct group *result = NULL;
+  if (getgrnam_r(groupname.c_str(), &grp, buf, sizeof(buf), &result) == 0 &&
```
jblomer: Seems to be a false positive of Coverity, afaik.  The documentation of the function claims that `result` will be set to `NULL` if the group isn't found or if there is an error.  Initializing `result` is fine, the rest of the changes make the code unnecessarily more difficult to read, I find.

Moliholy: According to the documentation:

_On success, getgrnam_r() and getgrgid_r() return zero, and set *result to grp. If no matching group record was found, these functions return 0 and store NULL in *result. In case of error, an error number is returned, and NULL is stored in *result._

Which means you are right, but we do not exhaustively check everything. On the other hand, the code is actually more readable...
I guess there is no problem if we gain more readability in exchange, so I'd better change it back

cvmfs/swissknife_pull.cc
```diff
@@ -571,6 +569,10 @@ int swissknife::CommandPull::Main(const swissknife::ArgumentList &args) {
   download_manager()->SetRetryParameters(retries, timeout, 3*timeout);
   download_manager()->Spawn();
 
+  // initializing the threads now in case of cleanup
```
jblomer: Please have a second look on the comment.  I don't quite understand it.

Moliholy: Depending on when the threads are initialized they might have to be deleted (cleaned up). The idea, as it was mentioned before, is to delay as much as possible the creation of the threads so that it avoids these points (error cases mostly) where the function exits, because otherwise it'd be mandatory to put the corresponding _delete[] workers_. That being said, there is no special reason for this comment and it could be perfectly deleted.

jblomer: Ok, I find this comment confusing.  Please rephrase or remove.

jblomer:
I think this conflicts with @reneme's refactoring of the signing utility.  Can you recheck?

cvmfs/swissknife_pull.cc
Moliholy: Fixed

Moliholy:
It did conflict when merging with the devel branch, but I took @reneme's changes in the sole conflict point. However, I guess that's why all his commits appear in the history of this PR

jblomer:
I'm pretty sure that if I merge this, old stuff will come to live again (compare `swissknife_sign` in its current version.)

jblomer:
Superseded by #1593

PR #1490, opened 2016-02-01
FIX: Coverity Issues

reneme:
Deals with several coverity alerts. Most of them are leaked FDs and/or memory in error cases. Nothing particularly severe.

### DIFF
--- a/cvmfs/clientctx.h
+++ b/cvmfs/clientctx.h
@@ -58,9 +58,9 @@ class ClientCtxGuard {
  public:
   ClientCtxGuard(uid_t uid, gid_t gid, pid_t pid)
     : set_on_construction_(false)
-    , old_uid_(0)
-    , old_gid_(0)
-    , old_pid_(0)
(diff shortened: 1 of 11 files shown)


cvmfs/clientctx.h
```diff
@@ -56,9 +56,13 @@ class ClientCtx {
  */
 class ClientCtxGuard {
  public:
-  ClientCtxGuard(uid_t uid, gid_t gid, pid_t pid) :
-      set_on_construction_(false) {
-      // Implementation guarantees old_ctx is not null.
+  ClientCtxGuard(uid_t uid, gid_t gid, pid_t pid)
+    : set_on_construction_(false)
+    , old_uid_(0)
```
jblomer: uid, gid, and pid should be initialized to -1 to indicate that they are invalid.

reneme: Done.

reneme:
Rebased to `devel` (solving several conflicts).

PR #1492, opened 2016-02-01
FIX: Last Second Fixes

reneme:
Fixes the integration test runs on Fedora (in particular 22) by starting Apache up-front. Furthermore this updates the SELinux sanity check in `cvmfs_config chksetup` to skip on EL5 and work on Fedora.

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -140,8 +140,8 @@ cvmfs_config_usage() {
 }
 
 has_selinux() {
-  which sestatus   > /dev/null 2>&1 && \
   which getenforce > /dev/null 2>&1 && \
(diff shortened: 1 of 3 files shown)


cvmfs/cvmfs_config
```diff
@@ -417,7 +422,8 @@ cvmfs_chksetup() {
   fi
 
   # Check SELinux label of cache directory
-  if [ -f /selinux/enforce ] && [ $(cat /selinux/enforce) -ne 0 ]; then
+  if has_selinux && which semodule > /dev/null 2>&1 && \
+     semodule --list-modules 2> /dev/null | grep -q cvmfs; then
```
reneme: The old check didn't work on Fedora (because the `/selinux` directory is not available there).

PR #1495, opened 2016-02-03
T util fix

Moliholy:
- Fixes many coverity issues in t_util mainly, but also others.
- Creates a UniquePtr that support _void*_ to release memory allocated via malloc/calloc


jblomer:
How does this PR relates to the previous ones?  Looks like it's built #1488?

Moliholy:
Yes, this branch keeps working on the coverity fixes, but they are both independent ones. I'll fix that

Moliholy:
Now they should be totally independent PRs

cvmfs/util.h
```diff
@@ -279,32 +279,66 @@ struct hash_murmur {
...
+ protected:
+  void Free() {
+    static_cast<DerivedT*>(this)->Free();
+  }
   T *ref_;
 };
 
+template <class T>
+class UniquePtr : public UniquePtrBase<T, UniquePtr<T> > {
+ private:
```
jblomer: Order should be `public`, `protected`, `private`

cvmfs/util.h
```diff
@@ -279,32 +279,66 @@ struct hash_murmur {
...
+  typedef UniquePtrBase<T, UniquePtr<T> > BaseT;
+ public:
+  friend class UniquePtrBase<T, UniquePtr<T> >;
```
jblomer: `friend class` declarations come before `public`

cvmfs/util.h
```diff
@@ -279,32 +279,66 @@ struct hash_murmur {
...
+  using BaseT::operator=;
```
jblomer: Why is this necessary?  The assignment operator is already public in the base class.

cvmfs/util.h
Moliholy: I know, but it has to be defined in advance concretely in this case. Otherwise the _BaseT::_ three lines below won't work

test/unittests/t_util.cc
```diff
@@ -424,7 +425,8 @@ TEST_F(T_Util, MakePipe) {
   void *buffer_output = scalloc(100, sizeof(char));
   MakePipe(fd);
   write(fd[1], to_write.c_str(), to_write.length());
-  read(fd[0], buffer_output, to_write.length());
+  int bytes_read = read(fd[0], buffer_output, to_write.length());
+  EXPECT_EQ(bytes_read, to_write.length());
```
jblomer: This compares signed and unsigned integers, I think.

test/unittests/t_util.cc
```diff
@@ -437,7 +439,8 @@ TEST_F(T_Util, WritePipe) {
   void *buffer_output = scalloc(20, sizeof(char));
   MakePipe(fd);
   WritePipe(fd[1], to_write.c_str(), to_write.length());
-  read(fd[0], buffer_output, to_write.length());
+  int bytes_read = read(fd[0], buffer_output, to_write.length());
+  EXPECT_EQ(bytes_read, to_write.length());
```
jblomer: Same here

cvmfs/util.h
Moliholy: Concretely the _operator=_ is by default both overwritten and inherited. The first one implies that the second one is _hidden_, so that _ussing BaseT::operator=_ tells the compiler not to try to create another operator= for the derived class which would hide the inherited one.

jblomer: I see, thanks.

test/unittests/t_util.cc
Moliholy: read() returns a ssize_t ---> __darwin_ssize_t ---> long, at least on mac

jblomer: Yep, that's the signed one.  And `to_write.length()` is unsigned, isn't it?

test/unittests/t_util.cc
Moliholy: I think it's a long

test/unittests/t_util.cc
Moliholy: Ah sorry, you mean in the comparison. Yes, string::length() returns an unsigned int. I'll fix that with a static_cast

Moliholy: Yes, you are right. I think it should now be fixed. In any case for the tests we run there shouldn't be a problem because we deal with very small values, always positive.

test/unittests/t_util.cc
```diff
@@ -437,7 +439,8 @@ TEST_F(T_Util, WritePipe) {
   void *buffer_output = scalloc(20, sizeof(char));
   MakePipe(fd);
   WritePipe(fd[1], to_write.c_str(), to_write.length());
-  read(fd[0], buffer_output, to_write.length());
+  ssize_t bytes_read = read(fd[0], buffer_output, to_write.length());
+  EXPECT_EQ(static_cast<size_t>(bytes_read), to_write.length());
```
jblomer: There are a couple of more spots in this PR where `EXPECT_EQ` compares signed and unsigned ints.

Moliholy: Should be solved now

PR #1496, opened 2016-02-03
Disable VOMS checks.

bbockelm:
Unfortunately, libvomsapi doesn't do all the validation we need
for this to function correctly.  For now, we disable VOMS checks and
always assume they fail - we'll fix this up in the next devel series.

This does allow HTTPS to be used for URLs as we blindly pass the
client credential to the remote host and don't need to validate it
in the client.

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -505,13 +505,13 @@ static bool CheckVoms(const fuse_ctx &fctx) {
   // Get VOMS information, if any.  If VOMS authz is present and VOMS is
   // not compiled in, then deny authorization.
   if ((fctx.uid != 0) && voms_authz_->size()) {
-#ifdef VOMS_AUTHZ
-    return CheckVOMSAuthz(&fctx, *voms_authz_);
-#else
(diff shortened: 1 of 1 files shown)


DrDaveD:
How could somebody get a valid VOMS extension without the underlying proxy being valid?

I also don't understand how any validation in the client can be trusted, since anybody could run an older client and still read what's in a catalog.

In any case, I think this should be able to be disabled in 2.2.1 without there being any rush to put out that new release; meanwhile affected repos can be removed from the stratum 1.

bbockelm:
Unfortunately, it's simple enough to get a valid VOMS extension -- you just need the public part of someone's proxy.  It would be strange to give this to random people, but it is the _public_ portion.

The validation is relevant because processes can access files in the cache they would otherwise be unable to access.  Older clients wouldn't be able to download the files to the cache; newer clients would be able to download to the cache for valid users (then inadvertently let invalid users access the cache).

I have the first draft of a fix to the validation issue (see #1497); it's not rocket science, but definitely needs more time to mature than we have for 2.2.x.

PR #1497, opened 2016-02-04
Secure CVMFS v2

bbockelm:
This PR fixes a deficiency of the original secure CVMFS pull request.  The certificate handling library originally used (libvomsapic) only validates the VOMS extensions - but doesn't look at the credentials!

This loads Globus libraries at runtime; to the best of my knowledge, Globus has the most widely tested and deployed proxy validation routines.  The loaded Globus libraries are used to validate the proxy chain; OpenSSL is used to verify the private and public key match.

The Globus API is unfortunately fairly complex: while the PR looks large, it is mostly loading symbols from various libraries.

Tests to come.


bbockelm:
@jblomer - everything is here but the updated integration test.  Unfortunately, I've run out of time tonight.

Given the tight schedule (and turn-around necessary for the new dependencies on the build hosts), I'd advise to consider merging ASAP and I can do the integration test in a follow-up.

bbockelm:
Alright, I think this one is ready to go.

bbockelm:
@jblomer - I think this now has all the pieces we discussed.

Particularly, all the interaction with the user environment - fetching and parsing of the certificate - now happens in the child process.  The parent process receives the DN and the serialized VOMS information from the child.

Further, the executable for the child process can be set to be an arbitrary path.  This would allow us to ship fixups to the credential retrieving / parsing / validating code at a higher cadence than upstream.

jblomer:
@bbockelm: based on this version, I pushed a first patch set to the branch fix-secure-cvmfs.  It's still under development.  The `voms_authz` things eventually will just become `authz` things.  The `AuthzSessionCache` (now `AuthzSessionManager`) stores "membership" and "token" for sessions, which it gets from the credentials fetcher.  The fetcher can also set ttl values for positive/negative caching.

The "membership" is an arbitrary string that is compared to the requirement from the cvmfs root catalog.  The token is a BLOB of the credentials (if it is a certificate) that can be passed to the download manager.

As I said, more to come.

jblomer:
Basic infrastructure is there.  Since we might have other authz helpers in the future, I moved away from the dark art of passing file descriptors between processes and moved more to a CGI model: input from stdin, output to stdout.  This also allows for simple integration with other languages.  There is a simple request-reply communication.  Messages a 4 byte version, a 4 byte length and then JSON text.  That allows us to evolve the protocol in the future.

Now, I'm moving on to the curl part and to moving the existing authorization code into an umbrella for the modified cvmfs-helper protocol.

bbockelm:
Just peaked a the `fix-secure-cvmfs` branch.  Everything looks good so far.  It would be perhaps easier to compare things if you could merge in some of the more rote Globus manipulation code - it would really shrink the diff's between the two.

jblomer:
All right, minus cosmetics and packaging I think we are back on track.  At least the integration test is back working.

jblomer:
Superseded by #1576

PR #1498, opened 2016-02-04
File descriptor wrapper class

Moliholy:
- Creates FileDescriptor and Socket utility classes
- It helps closing the file descriptors when the container object is destroyed, so shielding the resource management under eventual error cases / programmer mistakes
- Extra functionality to read, write, accept connections and listen to connections


Moliholy:
Operator= changed to Assign() method in order to improve readability

jblomer:
Superseded by #1593

PR #1505, opened 2016-02-11
RFC: Opportunistically Retry umount with Force

reneme:
This is a request for comment!

Recently on **lhcbdev.cern.ch** and a while ago in **geant4.cern.ch** we've been seeing release manager machines in an inconsistent state. The transaction is committed successfully but uounting of `/cvmfs/lhcbdev.cern.ch` fails with 'device busy'. This locks up the release manager machine in an inconsistent state requiring to do the following:

```bash
cvmfs_suid_helper rw_umount lhcbdev.cern.ch
cvmfs_suid_helper rdonly_umount lhcbdev.cern.ch
cvmfs_server abort -f lhcbdev.cern.ch
```

The installed version there is CernVM-FS 2.1.20 but the code path in question has not been touched significantly since then, as far as I can see. I can't tell yet what exactly locks up the mountpoint, unfortunately. Also, there are checks for open file descriptors right before the remount happens (using `lsof`). There is a potential race between the check and the actual action of course, but it should be rather short. 

This proposes to apply the 'forced remount' not only if the file descriptor check detected something before, but also if the 'graceful remount' failed. Anything I might be missing out?

Side note: This fix might actually also work around the OverlayFS `/proc/<PID>/fd` bug in 4.x kernels to a certain extend.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -1133,14 +1133,22 @@ close_transaction() {
   local tx_lock="${CVMFS_SPOOL_DIR}/in_transaction"
   local tmp_dir="${CVMFS_SPOOL_DIR}/tmp"
 
+  # if not explicitly asked, try if umounting works without force
+  if [ $use_fd_fallback -eq 0 ]; then
+    if ! cvmfs_suid_helper rw_umount     $name || \
+       ! cvmfs_suid_helper rdonly_umount $name; then
+      use_fd_fallback=1
+    fi
+  fi
+
+  # if explicitly asked for or the normal umount failed we apply more force
   if [ $use_fd_fallback -ne 0 ]; then
     cvmfs_suid_helper rw_lazy_umount     $name
     cvmfs_suid_helper kill_cvmfs         $name
     cvmfs_suid_helper rdonly_lazy_umount $name
-  else
-    cvmfs_suid_helper rw_umount $name
-    cvmfs_suid_helper rdonly_umount $name
   fi
+
+  # continue with the remounting
   cvmfs_suid_helper clear_scratch $name
   [ ! -z "$tmp_dir" ] && rm -fR "${tmp_dir}"/*
   cvmfs_suid_helper rdonly_mount $name > /dev/null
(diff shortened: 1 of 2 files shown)


jblomer:
I think the logic is fine.  I'm only wondering if we print a warning before forcefully killing the file system stack.

reneme:
This is actually the fallback of the fallback of the fallback...
If there are file descriptors detected right away on `cvmfs_server publish`, the user will get a warning and a request for confirmation. If confirmed, the publishing will start and later forcefully remount (fallback 1).

On the other hand, if file descriptors are detected only after the actual publishing has been finished (which might have been running for a substantial time), already since 2.1.20 (I believe) it falls back to force without further asking questions (fallback 2). 

Now, this will also fall back to force if the usual `umount` failed for any other reason (*new* fallback 3).

jblomer:
I see. I think we should log the forcful remount to syslog.  We can also use `wall` to inform interactive users and wait for say 1 minute before the disruption.

reneme:
There you go. By default this will now contact everybody through `wall`, wait for one minute and apply force afterwards. One can disable this warning (and waiting) with `CVMFS_FORCE_REMOUNT_WARNING=false` in `server.conf`. The syslog message already contained a note when force was used like so:
```txt
(test.cern.ch) closed transaction (using force)
```

DrDaveD:
Guys, I don't really like the idea of using wall.  That kind of assumes a CERN model of a separate VM for every repository.  At fermilab we have one release manager machine with 16 repositories and growing.  I'd hate for an issue with one repository to confuse all the other people logged in to the machine.
Including @bbockelm in the discussion.

reneme:
I understand your concern. You think its not enough to allow this warning to be switched off in the server's configuration (cf. `CVMFS_FORCE_REMOUNT_WARNING=false`)?

As a side note: also at CERN we are now sharing VMs for several small repositories.

DrDaveD:
When I first glanced at the code it looked to me like that was only removing the grace period, but I see now that it turns it off completely.  I guess it would be OK for me to recommend to the Fermilab operators to set that option.  Maybe it would help if the initial broadcasted message included a line saying how to disable the warning.

PR #1508, opened 2016-02-12
Remove BUILD_LIBCVMFS restriction on AArch64

davidlt:
The issue was resolved in binutils 2.26 release and the fix was
backported to RHELSA DP 7.2 and CentOS 7.2.1511.

Resolves errors while building cvmfs-universal.spec on aarch64.

Signed-off-by: David Abdurachmanov <David.Abdurachmanov@cern.ch>


jblomer:
Thanks!

PR #1509, opened 2016-02-12
DELETION: docker-related files moved to ci-scripts

Moliholy:
Docker-related files are now in ci-scripts (https://github.com/cernvm/ci-scripts/pull/3), so they can now be removed in this repository.

In any case, this PR could be perfectly delayed until we are 100% sure the tests are working in order not to pollute the git history


Moliholy:
Checking travis it seems the container collapsed this time for the pair mac OS X - gcc, but definitely there shouldn't be any problem because this PR only deals with bash scripts, which is not built nor executed

PR #1511, opened 2016-02-15
Feature: Maintaining a Reference Database

reneme:
This adds a release manager machine specific reference log. It tracks root-hashes whenever they are added to the CAS. In detail these are *certificates*, *root catalogs*, *history databases* and *metainfo files*. This information is useful for a full-featured garbage collection.

The changes in this pull request are the first step, namely adding hashes to the `Reflog` when a new manifest gets signed (`cvmfs_server publish`, `cvmfs_server tag`, ...). Soon we will also do similar updates for `cvmfs_server snapshot`. Furthermore the GC needs to be adapted to use the information provided by the `Reflog`.


reneme:
This now passes the integration test 577 and 579 as well as the unit test `T_GarbageCollector.FindAndSweepOrphanedNamedSnapshot` which test for garbage collecting orphaned revisions ([CVM-758](https://sft.its.cern.ch/jira/browse/CVM-758)). However, it is quite preliminary code still.

reneme:
Removed `CatalogTraversal<>::TraversePruned()` along with its related unit test cases and adapted the `GarbageCollector<>` to rely solely on `Reflog` to find condemned revisions.

That simplifies the `GarbageCollector<>` on the one hand and better decouples `GarbageCollector<>` and `CatalogTraversal<>`. In any case `CatalogTraversal<>::TraversePruned()` was questionable design from the beginning, as it made `CatalogTraversal<>` stateful.

Furthermore this allows to remove the "recycle_bin" table from the `History` database which was a poor design decision as well. Unfortunately I need to check first if that would be a non-backward compatible change somehow. For example if a 2.2.1 Stratum1 pulls from a 2.3.0 Stratum0 and assumes "recycle_bin" to be there for garbage collection.

reneme:
There is more things to be done, but technically it works now. I suggest to review and merge this to avoid growing the Pull Request even more. @jblomer: perhaps we should to sit together for that review, since it is quite a huge change set.

jblomer:
Cool!  @reneme: I'll first have a look myself and then pass by.

@Moliholy: Can you disable the Mac builds for travis?  I think like this they are a little bit useless.

Moliholy:
@jblomer it looks like mac machines are _so_ slow that they do not produce output on time (10 minutes). I'll fix that soon so that we can use them

cvmfs/compression.cc
```diff
@@ -42,6 +42,22 @@ static bool CopyFile2File(FILE *fsrc, FILE *fdest) {
...
 
+bool CopyPath2File(const std::string &src, FILE *fdest) {
+  int retval = -1;
+  struct stat info;
+
+  FILE *fsrc = fopen(src.c_str(), "r");
+  if (!fsrc) goto file_copy_final;
+
+  if (!CopyFile2File(fsrc, fdest)) goto file_copy_final;
+  retval = fstat(fileno(fsrc), &info);
```
jblomer: should be a `platform_fstat`

cvmfs/swissknife_sign.h
```diff
@@ -8,6 +8,7 @@
 #include <string>
 
 #include "hash.h"
+#include "reflog.h"
```
jblomer: I think this can go to the .cc file

cvmfs/reflog.h
```diff
@@ -0,0 +1,73 @@
...
+ public:
+  static Reflog* Open(const std::string &database_path);
+  static Reflog* Create(const std::string &database_path,
+                        const std::string &repo_name);
+
+  std::string CloseAndReturnDatabaseFile() {
+    return database_->CloseAndReturnDatabaseFile();
+  }
+
+ public:
```
jblomer: This can be removed.

cvmfs/reflog.cc
```diff
@@ -0,0 +1,182 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include "reflog.h"
+
+using namespace manifest;  // NOLINT
```
jblomer: This can be `namespace manifest {`

cvmfs/reflog.cc
```diff
@@ -0,0 +1,182 @@
...
+
+Reflog* Reflog::Open(const std::string &database_path) {
+  Reflog *reflog = new Reflog();
+  if (NULL == reflog || !reflog->OpenDatabase(database_path)) {
```
jblomer: reflog can never be `NULL` at this point, afaics.

cvmfs/swissknife_pull.cc
```diff
@@ -710,12 +743,31 @@ int swissknife::CommandPull::Main(const swissknife::ArgumentList &args) {
...
         meta_info.data());
       StoreBuffer(info, meta_info.size(), meta_info_hash, true);
+      if (reflog && !reflog->AddMetainfo(meta_info_hash)) {
+        LogCvmfs(kLogCvmfs, kLogStderr, "Failed to add metainfo to Reflog.");
+        goto fini;
+      }
     }
+
+    // upload Reflog database
+    reflog->CommitTransaction();
```
jblomer: Would this be `NULL` when preloading the cache?

jblomer:
Looks very fine to me!  I think I'd like to pass by to get a better understanding of the gc classes.

cvmfs/compression.cc
reneme: there is other `fstat()` just next to this function... I'll update all of them.

cvmfs/reflog.cc
reneme: well, in case the `new Reflog()` fails due to oom?

cvmfs/swissknife_pull.cc
reneme: good catch!

reneme:
I've just added fixes for the review suggestions. Let's have a look over the GC mechanics on the whiteboard maybe.

PR #1513, opened 2016-02-17
FIX: rename bucket when there is only one per account

Moliholy:


cvmfs/upload_s3.cc
```diff
@@ -276,6 +276,10 @@ int S3Uploader::GetKeysAndBucket(const std::string &filename,
  */
 std::string S3Uploader::GetBucketName(unsigned int use_bucket) const {
   std::stringstream ss;
+  if (number_of_buckets_ == 1) {
```
reneme: Please better check for both running numbers. I.e. :+1: 
```C++
if (number_of_buckets_ == 1 && keys_.size() == 1) { ...
```

Moliholy: Fixed

PR #1516, opened 2016-02-18
New feature - Http2

Moliholy:
This PR is not meant to be merged yet. It's just to check the addition of a new library to the system. This PR:
- Updates libcurl to the latest version
- Adds nghttp2 library as a dependency of curl
- Modifies curl build to actually build its new dependency and enable the http2 feature

Notice that nghttp2 is not part of the standard build, but it's actually built during the cmake step. Perhaps this could be optimized, so suggestions are more than welcome.


Moliholy:
Just for the record:
- libcurl http2 tutorial: https://curl.haxx.se/libcurl/c/http2-download.html
- How to download apache 2.4.17 (which first supports http2): http://tecadmin.net/enable-http2-in-apache/

cvmfs/cvmfs_server
```diff
@@ -2055,6 +2055,13 @@ Alias /cvmfs/${name} ${storage_dir}
     ExpiresByType application/x-cvmfs "access plus 2 minutes"
     ExpiresByType application/json    "access plus 2 minutes"
 </Directory>
+
+# Load http2 module
+LoadModule http2_module modules/mod_http2.so
```
jblomer: Will Apache still load this config even if the module is missing?

Moliholy: it won't, but this module should be available if using a sufficiently new apache version, which (I guess) is intended to happen. In any case it would be possible to check the installed modules in advance, and only if the module is present then it could be loaded

Moliholy: Fixed

Moliholy:
Finally I found the way to successfully compile the new library in all the platforms

cvmfs/CMakeLists.txt
```diff
@@ -365,7 +365,7 @@ if (BUILD_CVMFS)
 
   # link the stuff (*_LIBRARIES are dynamic link libraries *_archive are static link libraries ... one of them will be empty for each dependency)
   set (CVMFS_FUSE_LINK_LIBRARIES ${SQLITE3_LIBRARY} ${CARES_LIBRARIES}
-         ${CURL_LIBRARIES} ${LIBCURL_ARCHIVE} ${PACPARSER_LIBRARIES}
+         ${CURL_LIBRARIES} ${LIBCURL_ARCHIVE} ${NGHTTP2_ARCHIVE} ${PACPARSER_LIBRARIES}
```
jblomer: The nghttp2 library needs to be before the curl library/archive (like cares).  Otherwise linkers might not be able to resolve the symbols.

cvmfs/CMakeLists.txt
```diff
@@ -390,9 +390,10 @@ if (BUILD_LIBCVMFS)
   set (LIBCVMFS_LIBS ${LIBCVMFS_LIBS} ${CMAKE_CURRENT_BINARY_DIR}/libcvmfs_only.a)
 
   if (LIBCURL_BUILTIN)
-    add_dependencies (cvmfs_only libcares libcurl)
+    add_dependencies (cvmfs_only libcares libnghttp2 libcurl)
     set (LIBCVMFS_LIBS ${CARES_ARCHIVE} ${LIBCVMFS_LIBS} )
-    set (LIBCVMFS_LIBS ${LIBCURL_ARCHIVE} ${LIBCVMFS_LIBS} )
+    set (LIBCVMFS_LIBS ${LIBCVMFS_LIBS} )
```
jblomer: Can this be removed?

cvmfs/CMakeLists.txt
jblomer: Sorry, that's not true.  Needs to be after libcurl, so it's correct.

cvmfs/CMakeLists.txt
```diff
@@ -437,7 +438,7 @@ if (BUILD_LIBCVMFS)
   add_dependencies (libcvmfs cvmfs_only)
 
   add_executable( test_libcvmfs ${TEST_LIBCVMFS_SOURCES} )
-  target_link_libraries( test_libcvmfs ${CMAKE_CURRENT_BINARY_DIR}/libcvmfs.a ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${RT_LIBRARY} ${UUID_LIBRARIES} pthread dl )
+  target_link_libraries( test_libcvmfs ${CMAKE_CURRENT_BINARY_DIR}/libcvmfs.a ${SQLITE3_LIBRARY} ${CARES_LIBRARIES} ${NGHTTP2_ARCHIVE} ${CURL_LIBRARIES} ${PACPARSER_LIBRARIES} ${ZLIB_LIBRARIES} ${OPENSSL_LIBRARIES} ${RT_LIBRARY} ${UUID_LIBRARIES} pthread dl )
```
jblomer: cares & nghttp2 after libcurl

cvmfs/CMakeLists.txt
```diff
@@ -477,17 +478,17 @@ if (BUILD_SERVER)
 
   # link the stuff (*_LIBRARIES are dynamic link libraries)
   target_link_libraries (cvmfs_swissknife  ${CVMFS_SWISSKNIFE_LIBS}
-                         ${SQLITE3_LIBRARY}  ${CURL_LIBRARIES} ${LIBCURL_ARCHIVE}
-                         ${CARES_LIBRARIES} ${CARES_ARCHIVE} ${TBB_LIBRARIES}
-                         ${ZLIB_LIBRARIES} ${ZLIB_ARCHIVE} ${OPENSSL_LIBRARIES}
-                         ${SQLITE3_ARCHIVE} ${RT_LIBRARY} ${VJSON_ARCHIVE}
-                         ${SHA3_ARCHIVE} ${CAP_LIBRARIES} pthread dl)
+                         ${SQLITE3_LIBRARY}  ${CURL_LIBRARIES} ${NGHTTP2_ARCHIVE} ${LIBCURL_ARCHIVE}
+                         ${CARES_LIBRARIES} ${CARES_ARCHIVE} ${NGHTTP2_ARCHIVE}
```
jblomer: Here nghttp2 is listed twice

cvmfs/CMakeLists.txt
```diff
@@ -512,11 +513,11 @@ if (BUILD_SERVER)
       set (TBB_DEBUG_LIBRARIES ${TBB_LIBRARIES})
     endif (NOT TBB_DEBUG_LIBRARIES)
     target_link_libraries (cvmfs_swissknife_debug  ${CVMFS_SWISSKNIFE_LIBS}
-                           ${SQLITE3_LIBRARY}  ${CURL_LIBRARIES} ${LIBCURL_ARCHIVE}
-                           ${CARES_LIBRARIES} ${CARES_ARCHIVE} ${TBB_DEBUG_LIBRARIES}
-                           ${ZLIB_LIBRARIES} ${ZLIB_ARCHIVE} ${OPENSSL_LIBRARIES}
-                           ${SQLITE3_ARCHIVE} ${RT_LIBRARY} ${VJSON_ARCHIVE}
-                           ${SHA3_ARCHIVE} ${CAP_LIBRARIES} pthread dl)
+                           ${SQLITE3_LIBRARY} ${NGHTTP2_ARCHIVE} ${CURL_LIBRARIES} ${LIBCURL_ARCHIVE}
+                           ${CARES_LIBRARIES} ${CARES_ARCHIVE} ${NGHTTP2_ARCHIVE}
```
jblomer: nghttp2 listed twice

cvmfs/download.cc
```diff
@@ -177,25 +177,27 @@ static size_t CallbackCurlHeader(void *ptr, size_t size, size_t nmemb,
   //          header_line.c_str());
 
   // Check http status codes
-  if (HasPrefix(header_line, "HTTP/1.", false)) {
-    if (header_line.length() < 10)
+  if (HasPrefix(header_line, "HTTP/", false)) {
+    if (header_line.length() < 12) {
```
jblomer: Just curious: what exactly is the header string for http2, i.e. why is the line 2 bytes longer?

cvmfs/download.cc
```diff
@@ -231,9 +231,13 @@ static size_t CallbackCurlHeader(void *ptr, size_t size, size_t nmemb,
   if ((info->destination == kDestinationMem) &&
       HasPrefix(header_line, "CONTENT-LENGTH:", true))
   {
-    char *tmp = reinterpret_cast<char *>(alloca(num_bytes+1));
     uint64_t length = 0;
-    sscanf(header_line.c_str(), "%s %"PRIu64, tmp, &length);
+    size_t pos = header_line.find(":") + 1;
+    assert(header_line.length() > pos);
```
jblomer: That's dangerous: a malicious web server can force cvmfs into a crash.

cvmfs/download.cc
```diff
@@ -231,9 +231,13 @@ static size_t CallbackCurlHeader(void *ptr, size_t size, size_t nmemb,
...
+    if (header_line[pos] == ' ') {
+      ++pos;
+    }
+    sscanf(&header_line[pos], "%"PRIu64, &length);
```
jblomer: What happens if header_line has no more characters after the space?

cvmfs/CMakeLists.txt
Moliholy: Yes, it is actually a mistake. I'll put all the curl-related libraries in a single line

externals/nghttp2/src/makeHook.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+make clean
+make
+make install
```
jblomer: Why `make install`?

cvmfs/CMakeLists.txt
Moliholy: In the original CMakeList.txt CARES_LIBRARIES appeared before, so I preferred not to change it, at least in principle. I'll do it now

jblomer:
The changes in s3fanout, swissknife_sign, swissknife_pull, and uti come from a different branch.  Please submit a rebased clean patch set.

cvmfs/download.cc
Moliholy: "HTTP/1.1" -> 8 characters
"HTTP/2"    -> 6 characters
So two less. That's basically the reason. The rest is (should be?) equal

cvmfs/download.cc
Moliholy: And what would a good solution be? Should we simply return 0 or num_bytes, ignoring the rest?

cvmfs/download.cc
jblomer: Hm, I still don't understand.  If the new header is 2 bytes less, why did the test change to test for two more bytes?

cvmfs/download.cc
Moliholy: It's not clear. One could think it will most probably try to get the next 64 bits of a piece of memory not reserved by the program, which can potentially make it crash. However, when reading the documentation (http://www.cplusplus.com/reference/cstdio/sscanf/) it claims it returns EOF on failure when reading.

Shall the size be checked in advance?

externals/nghttp2/src/makeHook.sh
Moliholy: It's been configured to install in the same directory where it's being built in order to be able to use the CURL's "--with-nghttp2=<path>" building option. That groups up all the generated artifacts in a lib/ directory which is required to build it.

cvmfs/download.cc
reneme: Actually the HTTP/2 header is `HTTP/2.0 [...]`, didn't we figure this out while implementing this patch? The old check was simply wrong...

`HTTP/1.1 200` is 12 characters to begin with. But looking into the history of the code a few lines below, it used to check only the first digit of the HTTP status code. When it was extended to check the second and third digit as well, this sanity check was just not updated.

Moliholy: Oh, you are right... what we should check is if the line is 10 bytes long for HTTP/1.1 and 8 (not 12!) for HTTP/2. Instead of subtracting I added by mistake.
In that case I think both could check a length of 10, in case after some time a "HTTP/2.1", for instance, appears.

cvmfs/download.cc
```diff
@@ -177,25 +177,27 @@ static size_t CallbackCurlHeader(void *ptr, size_t size, size_t nmemb,
...
-    if (header_line.length() < 10)
+  if (HasPrefix(header_line, "HTTP/", false)) {
+    if (header_line.length() < 12) {
       return 0;
+    }
 
     unsigned i;
     for (i = 8; (i < header_line.length()) && (header_line[i] == ' '); ++i) {}
 
     // TODO(jblomer): consolidate the code
```
reneme: This TODO can go, as the patch-set below "consolidate[s] the code"... It now checks the parsed HTTP status code rather than doing funny offset magic to check individual status code positions.

Moliholy: Fixed

cvmfs/download.cc
reneme: Huh? No... it is okay how it is now all for `HTTP/1.0`, `HTTP/1.1` and `HTTP/2.0`! Note that H2 also sends the minor version number in the header and there must be a minimum of **12** characters for a correct first header line including an HTTP status code:

```
example:
HTTP/2.0 200

 1: H
 2: T
 3: T
 4: P
 5: /
 6: 2
 7: .
 8: 0
 9:  
10: 2
11: 0
12: 0
```

Moliholy:
I know, I know, I just rebased it

reneme:
> I know, I know, I just rebased it

Alright. Just wanted to make sure. :-)

cvmfs/download.cc
jblomer: @Moliholy please add a code comment that shows the expected header line as @reneme pointed out

cvmfs/download.cc
jblomer: As mentioned in the [libcurl documentation](https://curl.haxx.se/libcurl/c/CURLOPT_HEADERFUNCTION.html), returning zero will indicate an error to libcurl.  So yes, that's what we should do if the header can't be parsed.

cvmfs/download.cc
jblomer: Yes, just make sure that the header is long enough to contain at least one digit for the content length.  Then it should be fine.

externals/nghttp2/src/makeHook.sh
jblomer: Hm, I still think it's unnecessary.  Just after `make` there should already be a copy of libnghttp2.a somewhere in the tree.

jblomer:
HTTP2 efforts parked for the time being.  Cleanups moved to #1705.

PR #1517, opened 2016-02-23
Refactor: Rename catalog::Sql to catalog::SqlCatalog

reneme:
This fixes a severe code smell. Namely we've had two implementations of `class Sql`, one residing in the namespace `sqlite` and one (that actually directly inherited from the first) in namespace `catalog`. This renames the latter one to `catalog::SqlCatalog` for clarity and consistency with `history::SqlHistory` and `manifest::SqlReflog` (to come [here](https://github.com/cvmfs/cvmfs/pull/1511)).

### DIFF
--- a/cvmfs/catalog.cc
+++ b/cvmfs/catalog.cc
@@ -170,7 +170,7 @@ bool Catalog::OpenDatabase(const string &db_path) {
   }
 
   // Find out the maximum row id of this database file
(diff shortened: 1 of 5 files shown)


jblomer:
Travis complains, otherwise I think it's fine.

reneme:
Fixed.

PR #1518, opened 2016-02-23
Feature: Lazy Initialisation of SQL statements

reneme:
**Work in progress...**

This allows derived `Sql` wrapper classes to opt for lazy initialisation of their prepared statements. We hope to safe memory and CPU cycles this way. For example each catalog opened for writing is preparing 17 SQL statements even though it might only ever need a small subset of them.

Furthermore this changes some low-hanging fruits in *history_sql.cc* to utilise the new `Sql::DeferredInit()` for demoing the API.

More adaptions in other `Sql` subclasses to come...

### DIFF
--- a/cvmfs/catalog_sql.cc
+++ b/cvmfs/catalog_sql.cc
@@ -506,13 +506,13 @@ bool SqlDirentWrite::BindDirentFields(const int hash_idx,
 
 
 SqlListContentHashes::SqlListContentHashes(const CatalogDatabase &database) {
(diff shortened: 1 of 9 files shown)


reneme:
This should be ready for review. After all the `Catalog` SQL stuff wasn't so hard to convert. Naturally that pretty much screams for an integration test run.

reneme:
Rebased on latest `devel` and added `Reflog` SQL statement patch.

PR #1519, opened 2016-02-24
FIX: lack of output on travis

Moliholy:
Fixes an error when running on travis, which concretely complains on mac machines because there is no output after ten minutes. This way there will be output every 30 seconds


Moliholy:
Perhaps you'd like to have more dots in the output. What about printing every 10 seconds instead of 30?

PR #1520, opened 2016-02-25
Increase the number of dots in travis

Moliholy:
It's considerably ugly not to see more dots printed in the output. The activity and the "Done" are pretty close. Printing one every 10 seconds instead of 30 would make it prettier


Moliholy:
Considerably better now :)
screen shot 2016-02-25 at 16 40 58

PR #1524, opened 2016-03-02
Fixes and improvements in the memory management

jblomer:
- Fixes a few small memory leaks during reload of the libcvmfs_fuse library
- Update sqlite library
- Add a dedicated memory manager for sqlite.  This ensures that sqlite's memory needs don't interfere with other mallocs and significantly reduces memory fragmentation
- Revert too pessimistic setting of fuse page cache
- Remove a few unused fields from `DirectoryEntry`
- Remove unused lookup of directory entries by inode
- Avoid unloading the libcvmfs_fuse library under valgrind for better debugging (symbol resolution)
- Add per-database memory statistics to `cvmfs_talk internal affairs`
- Add `cvmfs_talk detach nested` to emulate what happens when a new root catalog comes in

### DIFF
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -187,6 +187,7 @@ class AbstractCatalogManager : public SingleCopy {
   bool GetVOMSAuthz(std::string *authz) const;
   int GetNumCatalogs() const;
   std::string PrintHierarchy() const;
+  std::string PrintAllMemStatistics() const;
 
   /**
    * Get the inode number of the root DirectoryEntry
@@ -299,6 +300,7 @@ class AbstractCatalogManager : public SingleCopy {
   // Catalog *Inode2Catalog(const inode_t inode);
   std::string PrintHierarchyRecursively(const CatalogT *catalog,
                                         const int level) const;
+  std::string PrintMemStatsRecursively(const CatalogT *catalog) const;
 
   InodeRange AcquireInodes(uint64_t size);
   void ReleaseInodes(const InodeRange chunk);
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -37,7 +37,8 @@ const unsigned kSqliteMemPerThread = 1*1024*1024;
  */
 enum LookupOptions {
   kLookupSole        = 0x01,
-  kLookupRawSymlink  = 0x02,
+  // kLookupFull        = 0x02  not used anymore
+  kLookupRawSymlink  = 0x10,
 };
 
 
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -37,8 +37,7 @@ const unsigned kSqliteMemPerThread = 1*1024*1024;
  */
 enum LookupOptions {
   kLookupSole        = 0x01,
-  kLookupFull        = 0x02,
-  kLookupRawSymlink  = 0x10,
+  kLookupRawSymlink  = 0x02,
 };
 
 
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -959,8 +959,14 @@ int main(int argc, char *argv[]) {
   session = NULL;
   mount_options = NULL;
 
-  dlclose(library_handle_);
-  library_handle_ = NULL;
+#ifdef HAS_VALGRIND_HEADERS
+  if (!RUNNING_ON_VALGRIND) {
+#endif
+    dlclose(library_handle_);
+    library_handle_ = NULL;
+#ifdef HAS_VALGRIND_HEADERS
+  }
+#endif
 
   LogCvmfs(kLogCvmfs, kLogSyslog, "CernVM-FS: unmounted %s (%s)",
            mount_point_->c_str(), repository_name_->c_str());
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -960,6 +960,8 @@ int main(int argc, char *argv[]) {
   mount_options = NULL;
 
 #ifdef HAS_VALGRIND_HEADERS
+  // If the libcvmfs_fuse library is unloaded, valgrind can't resolve the 
+  // symbols anymore.  We skip under valgrind.
   if (!RUNNING_ON_VALGRIND) {
 #endif
     dlclose(library_handle_);
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -442,6 +442,25 @@ static void SetFuseOperations(struct fuse_lowlevel_ops *loader_operations) {
 }
 
 
+static void *OpenLibrary(const string &path) {
+  return dlopen(path.c_str(), RTLD_NOW | RTLD_LOCAL);
+}
+
+
+static void CloseLibrary() {
+#ifdef HAS_VALGRIND_HEADERS
+  // If the libcvmfs_fuse library is unloaded, valgrind can't resolve the
+  // symbols anymore.  We skip under valgrind.
+  if (!RUNNING_ON_VALGRIND) {
+#endif
+    dlclose(library_handle_);
+    library_handle_ = NULL;
+#ifdef HAS_VALGRIND_HEADERS
+  }
+#endif
+}
+
+
 static CvmfsExports *LoadLibrary(const bool debug_mode,
                                  LoaderExports *loader_exports)
 {
@@ -464,7 +483,7 @@ static CvmfsExports *LoadLibrary(const bool debug_mode,
   vector<string>::const_iterator i    = library_paths.begin();
   vector<string>::const_iterator iend = library_paths.end();
   for (; i != iend; ++i) {  // TODO(rmeusel): C++11 range based for
-    library_handle_ = dlopen((*i).c_str(), RTLD_NOW | RTLD_LOCAL);
+    library_handle_ = OpenLibrary(*i);
     if (library_handle_ != NULL) {
       break;
     }
@@ -492,19 +511,6 @@ static CvmfsExports *LoadLibrary(const bool debug_mode,
   }
 
   return *exports_ptr;
-}
-
-static void CloseLibrary() {
-#ifdef HAS_VALGRIND_HEADERS
-  // If the libcvmfs_fuse library is unloaded, valgrind can't resolve the
-  // symbols anymore.  We skip under valgrind.
-  if (!RUNNING_ON_VALGRIND) {
-#endif
-    dlclose(library_handle_);
-    library_handle_ = NULL;
-#ifdef HAS_VALGRIND_HEADERS
-  }
-#endif
 }
 
 
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -494,6 +494,19 @@ static CvmfsExports *LoadLibrary(const bool debug_mode,
   return *exports_ptr;
 }
 
+static void CloseLibrary() {
+#ifdef HAS_VALGRIND_HEADERS
+  // If the libcvmfs_fuse library is unloaded, valgrind can't resolve the
+  // symbols anymore.  We skip under valgrind.
+  if (!RUNNING_ON_VALGRIND) {
+#endif
+    dlclose(library_handle_);
+    library_handle_ = NULL;
+#ifdef HAS_VALGRIND_HEADERS
+  }
+#endif
+}
+
 
 Failures Reload(const int fd_progress, const bool stop_and_go) {
   int retval;
@@ -517,8 +530,7 @@ Failures Reload(const int fd_progress, const bool stop_and_go) {
 
   SendMsg2Socket(fd_progress, "Unloading Fuse module\n");
   cvmfs_exports_->fnFini();
-  dlclose(library_handle_);
-  library_handle_ = NULL;
+  CloseLibrary();
 
   if (stop_and_go) {
     CreateFile(*socket_path_ + ".paused", 0600);
@@ -959,16 +971,7 @@ int main(int argc, char *argv[]) {
   session = NULL;
   mount_options = NULL;
 
-#ifdef HAS_VALGRIND_HEADERS
-  // If the libcvmfs_fuse library is unloaded, valgrind can't resolve the 
-  // symbols anymore.  We skip under valgrind.
-  if (!RUNNING_ON_VALGRIND) {
-#endif
-    dlclose(library_handle_);
-    library_handle_ = NULL;
-#ifdef HAS_VALGRIND_HEADERS
-  }
-#endif
+  CloseLibrary();
 
   LogCvmfs(kLogCvmfs, kLogSyslog, "CernVM-FS: unmounted %s (%s)",
            mount_point_->c_str(), repository_name_->c_str());
--- a/cvmfs/smalloc.h
+++ b/cvmfs/smalloc.h
@@ -73,6 +73,26 @@ static inline void __attribute__((used)) smunmap(void *mem) {
   assert((retval == 0) && "Invalid umnmap");
 }
 
+
+/**
+ * Used when the caller remembers the size, so that it can call sxunmap later.
+ */
+static inline void * __attribute__((used)) sxmmap(size_t size) {
+  void *mem = mmap(NULL, size, PROT_READ | PROT_WRITE,
+                   MAP_PRIVATE | PLATFORM_MAP_ANONYMOUS, -1, 0);
+  assert((mem != MAP_FAILED) && "Out Of Memory");
+  return mem;
+}
+
+/**
+ * Free memory acquired by sxmmap.
+ */
+static inline void __attribute__((used)) sxunmap(void *mem, size_t size) {
+  int retval = munmap(mem, size);
+  assert((retval == 0) && "Invalid umnmap");
+}
+
+
 #ifdef CVMFS_NAMESPACE_GUARD
 }  // namespace CVMFS_NAMESPACE_GUARD
 #endif
--- a/cvmfs/smalloc.h
+++ b/cvmfs/smalloc.h
@@ -74,6 +74,25 @@ static inline void __attribute__((used)) smunmap(void *mem) {
 }
 
 
+/**
+ * Used when the caller remembers the size, so that it can call sxunmap later.
+ */
+static inline void * __attribute__((used)) sxmmap(size_t size) {
+  void *mem = mmap(NULL, size, PROT_READ | PROT_WRITE,
+                   MAP_PRIVATE | PLATFORM_MAP_ANONYMOUS, -1, 0);
+  assert((mem != MAP_FAILED) && "Out Of Memory");
+  return mem;
+}
+
+/**
+ * Free memory acquired by sxmmap.
+ */
+static inline void __attribute__((used)) sxunmap(void *mem, size_t size) {
+  int retval = munmap(mem, size);
+  assert((retval == 0) && "Invalid umnmap");
+}
+
+
 #ifdef CVMFS_NAMESPACE_GUARD
 }  // namespace CVMFS_NAMESPACE_GUARD
 #endif
--- a/cvmfs/smalloc.h
+++ b/cvmfs/smalloc.h
@@ -53,10 +53,16 @@ static inline void * __attribute__((used)) smmap(size_t size) {
   assert(size > 0);
   assert(size < std::numeric_limits<size_t>::max() - 4096);
 
+  const int anonymous_fd = -1;
+  const off_t offset = 0;
   size_t pages = ((size + 2*sizeof(size_t))+4095)/4096;  // round to full page
-  unsigned char *mem =
-    static_cast<unsigned char *>(mmap(NULL, pages*4096, PROT_READ | PROT_WRITE,
-                                 MAP_PRIVATE | PLATFORM_MAP_ANONYMOUS, -1, 0));
+  unsigned char *mem = static_cast<unsigned char *>(
+    mmap(NULL,
+         pages*4096,
+         PROT_READ | PROT_WRITE,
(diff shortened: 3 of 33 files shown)


cvmfs/catalog_mgr.h
```diff
@@ -37,8 +37,7 @@ const unsigned kSqliteMemPerThread = 1*1024*1024;
  */
 enum LookupOptions {
   kLookupSole        = 0x01,
-  kLookupFull        = 0x02,
-  kLookupRawSymlink  = 0x10,
+  kLookupRawSymlink  = 0x02,
```
reneme: Maybe stick to the `0x10` definition for `kLookupRawSymlink`? Just a gut feeling...

cvmfs/sqlitemem.h
```diff
@@ -0,0 +1,291 @@
...
+     * block.
+     */
+    AvailBlockCtl *head_avail_;
+    AvailBlockCtl *rover_;  ///< The free block where the next search starts.
+    uint32_t no_reserved_;
+  };
+
+
+  /**
+   * A continues chunk of memory from which fixed-sized chunks are given as
```
reneme: A *continuous* ?

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+
+using namespace std;  // NOLINT
+
+namespace sqlite {
+
+uint32_t MemoryManager::MallocArena::GetSize(void *ptr) const {
+  assert(Contains(ptr));
+
+  ReservedBlockCtl *block_ctl = reinterpret_cast<ReservedBlockCtl *>(
+    reinterpret_cast<char *>(ptr) - sizeof(ReservedBlockCtl));
```
reneme: Perhaps factor that out into an inlinable method like `GetReservedBlockCtl(void *ptr)`?

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  // Update the list of available blocks
+  if (remaining_size == 0) {
+    // Remove free block p from the list of available blocks
+    q->link_next = p->link_next;
+    q->GetNextPtr(arena_)->link_prev = q->ConvertToLink(arena_);
+  } else {
+    p->size = remaining_size;
+    void *upper_tag =
+      reinterpret_cast<char *>(p) + remaining_size - sizeof(AvailBlockTag);
+    new (upper_tag) AvailBlockTag(remaining_size);
```
reneme: If I understand correctly, this is shrinking `AvailBlockCtl p` from the back, to accommodate the requested memory. Why not have a `AvailBlockCtl::ShrinkTo(remaining_size)` method do this?

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+    p = q->GetNextPtr(arena_);
+    if (p->size >= total_size)
+      break;
+    if (p == head_avail_) {
+      if (wrapped)
+        return NULL;
+      wrapped = true;
+    }
+    q = p;
+  } while (true);
```
reneme: I would put that loop into a method `FindFreeBlock()` returning either the block or NULL

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+    p->size = remaining_size;
+    void *upper_tag =
+      reinterpret_cast<char *>(p) + remaining_size - sizeof(AvailBlockTag);
+    new (upper_tag) AvailBlockTag(remaining_size);
+  }
+  rover_ = q->GetNextPtr(arena_);
+
+  // Place the new allocation, which also sets the block type tag at the end
+  char *new_block = reinterpret_cast<char *>(p) + remaining_size;
+  new (new_block) ReservedBlockCtl(total_size);
```
reneme: After `FindFreeBlock()` the rest of the method would then be `ReserveBlock()`. I claim that way it would be a lot easier to understand the subparts, given the amount of pointer arithmetic in that method.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+ * the list of available blocks.
+ */
+void MemoryManager::MallocArena::Free(void *ptr) {
+  assert(Contains(ptr));
+
+  no_reserved_--;
+
+  ReservedBlockCtl *block_ctl = reinterpret_cast<ReservedBlockCtl *>(
+    reinterpret_cast<char *>(ptr) - sizeof(ReservedBlockCtl));
+  char prior_tag = *(reinterpret_cast<char *>(block_ctl) - 1);
```
reneme: This is getting the tag from the previous block, isn't it? What happens if we are dealing with the first block in the arena?

jblomer: This is why we have a fake first block and a fake last block in the arena (see constructor).  This is what I meant with the comment "Every allocated block has a predecessor and a successor in the arena."

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  assert((prior_tag == kTagAvail) || (prior_tag == kTagReserved));
+  assert(block_ctl->size <= 0);
+
+  int32_t new_size = -(block_ctl->size);
```
reneme: `ReservedBlockCtl` "knows" that it stores negative values... Hence, `ReservedBlockCtl::size()` should do this 'conversion' not the using code here.

cvmfs/sqlitemem.h
```diff
@@ -0,0 +1,291 @@
...
+      }
+      AvailBlockCtl *GetPrevPtr(char *base) {
+        return reinterpret_cast<AvailBlockCtl *>(base + link_prev);
+      }
+      int32_t ConvertToLink(char *base) {
+        return reinterpret_cast<char *>(this) - base;
+      }
+      int32_t size;  // always positive
+      int32_t link_next;
+      int32_t link_prev;
```
reneme: are those pointers?

jblomer: They are offsets.  Because the arena is small, 4 bytes suffice instead of 8 bytes for pointers.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+    assert(prior_size > 0);
+    new_size += prior_size;
+    new_avail = reinterpret_cast<AvailBlockCtl *>(
+      reinterpret_cast<char *>(block_ctl) - prior_size);
+    AvailBlockCtl *next = new_avail->GetNextPtr(arena_);
+    AvailBlockCtl *prev = new_avail->GetPrevPtr(arena_);
+    prev->link_next = new_avail->link_next;
+    next->link_prev = new_avail->link_prev;
+    if (rover_ == new_avail)
+      rover_ = head_avail_;
```
reneme: Perhaps `new_avail->MergeWithPrevious()`?

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  }
+
+  int32_t succ_size = *reinterpret_cast<int32_t *>(
+    reinterpret_cast<char *>(new_avail) + new_size);
```
reneme: This assumes that both `ReservedBlockCtl` and `AvailBlockCtl` have their size as the first member.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  if (succ_size >= 0) {
```
reneme: And that uses the fact that `ReservedBlockCtl` stores its negative sizes, to figure out if the next block is reserved or not?

cvmfs/sqlitemem.cc
jblomer: yes.

cvmfs/sqlitemem.cc
jblomer: yes.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+    // Merge with succeeding block and remove the block from the list
+    AvailBlockCtl *succ_avail = reinterpret_cast<AvailBlockCtl *>(
+      reinterpret_cast<char *>(new_avail) + new_size);
+    AvailBlockCtl *next = succ_avail->GetNextPtr(arena_);
+    AvailBlockCtl *prev = succ_avail->GetPrevPtr(arena_);
+    prev->link_next = succ_avail->link_next;
+    next->link_prev = succ_avail->link_prev;
+    new_size += succ_size;
+    if (rover_ == succ_avail)
+      rover_ = head_avail_;
```
reneme: Giving it a second thought, probably it's better to have `MallocArena::MergeFreeBlocks(blk_a, blk_b);` rather then `AvailBlockCtl::MergeWith...()` as this potentially needs to touch `MallocArena::rover_`.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  new (new_block) ReservedBlockCtl(total_size);
+  return new_block + sizeof(ReservedBlockCtl);
+}
+
+
+/**
+ * The arena starts with a pointer to this followed by the AvailBlockCtl of
+ * head_avail_, followed by a reserved tag to prevent it from being merged,
+ * followed by a free block spanning the arena until the end tag.  The end tag
+ * is a single negative int, which mimics another reserved block.
```
reneme: Ah! So this prevents you from merging with the 0th or (n+1)th block...

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+ */
+MemoryManager::MallocArena::MallocArena()
+  : arena_(reinterpret_cast<char *>(sxmmap_align(kArenaSize)))
+  , head_avail_(reinterpret_cast<AvailBlockCtl *>(arena_ + sizeof(void *)))
+  , rover_(head_avail_)
+  , no_reserved_(0)
+{
+  int32_t usable_size = kArenaSize -
+    (sizeof(void *) + sizeof(AvailBlockCtl) + 1 + sizeof(int32_t));
```
reneme: Scratch notes:
`sizeof(void*)` ==> `this` pointer
`sizeof(AvailBlockCtl)` ==> first and only AvailBlockCtl at init time
`1` ==> the fake `ReservedTag` to not merge 'virtual' successor block
`sizeof(int32_t)` ==> magic number at the end of the arena?

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  *reinterpret_cast<MallocArena **>(arena_) = this;
+  AvailBlockCtl *free_block =
+    new (arena_ + sizeof(void *) + sizeof(AvailBlockCtl) + 1) AvailBlockCtl();
```
reneme: what is the `+1` here?

jblomer: That's the `kTagReserved`

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  free_block->size = usable_size;
+  free_block->link_next = free_block->link_prev =
+    head_avail_->ConvertToLink(arena_);
+
+  head_avail_->size = 0;
+  head_avail_->link_next = head_avail_->link_prev =
+    free_block->ConvertToLink(arena_);
+
+  // Prevent succeeding blocks from merging
+  *(reinterpret_cast<char *>(free_block) - 1) = kTagReserved;
```
reneme: Ah! There are actually two `AvailBlockCtl`s! Only that the constructor of the one at `head_avail_` is never called! Furthermore `head_avail_` is only a sentinel that serves as a list-head, right?

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,489 @@
...
+  // Final tag: reserved block marker
+  *reinterpret_cast<int32_t *>(arena_ + kArenaSize - sizeof(int32_t)) = -1;
```
reneme: shouldn't that be `... = kTagReserved`?

Ah! No... it is again exploiting the fact that `ReservedBlockCtl` stores its `size < 0`.

cvmfs/sqlitemem.cc
jblomer: Correct.

cvmfs/loader.cc
```diff
@@ -517,8 +530,7 @@ Failures Reload(const int fd_progress, const bool stop_and_go) {
 
   SendMsg2Socket(fd_progress, "Unloading Fuse module\n");
   cvmfs_exports_->fnFini();
-  dlclose(library_handle_);
-  library_handle_ = NULL;
+  CloseLibrary();
```
reneme: Perhaps wrap `dlopen()` into `OpenLibrary()` for consistency?

cvmfs/smalloc.h
```diff
@@ -73,6 +73,44 @@ static inline void __attribute__((used)) smunmap(void *mem) {
   assert((retval == 0) && "Invalid umnmap");
 }
 
+
+/**
+ * Used when the caller remembers the size, so that it can call sxunmap later.
+ */
+static inline void * __attribute__((used)) sxmmap(size_t size) {
+  void *mem = mmap(NULL, size, PROT_READ | PROT_WRITE,
+                   MAP_PRIVATE | PLATFORM_MAP_ANONYMOUS, -1, 0);
```
reneme:
```C++
const int anonymous_fd = -1;
const off_t offset     =  0;
void *mem = mmap(..., anonymous_fd, offset);
```
Poor-mans parameter naming for readability.

cvmfs/smalloc.h
```diff
@@ -73,6 +73,44 @@ static inline void __attribute__((used)) smunmap(void *mem) {
...
+  uintptr_t head = size - (uintptr_t(mem) % size);
+  sxunmap(mem, head);
+  mem += head;
+  uintptr_t tail = size - head;
+  if (tail > 0)
+    sxunmap(mem + size, tail);
+  return mem;
+}
+
+
```
reneme: Perhaps wrap this `mmap()` into a thin class? Provides you with RAII and keeps the `*mem` and `size` contained for `sxunmap()`.

reneme: Also the class could have a private constructor and static factory methods for a simple `mmap()` and an aligned `mmap()`.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,508 @@
...
+
+  assigned_ = true;
+}
+
+
+/**
+ * Needs to be the first operation on an opened sqlite database.  Returns the
+ * location of the buffer.
+ */
+void *MemoryManager::AssignLookasideBuffer(void *db) {
```
reneme: why `void *db` when it `reinterpret_cast<>`s it into an `sqlite3 *` right after entering the method?

cvmfs/sql_impl.h
```diff
@@ -137,6 +137,11 @@ bool Database<DerivedT>::OpenDatabase(const int flags) {
     return false;
   }
 
+  if (!read_write_ && MemoryManager::HasInstance()) {
+    database_.lookaside_buffer =
+      MemoryManager::GetInstance()->AssignLookasideBuffer(sqlite_db());
+  }
+
```
reneme: this rather fits into the `Database<>::Configure()` method.

cvmfs/sqlitemem.h
```diff
@@ -0,0 +1,303 @@
...
+ * It is implemented as a singleton.  GetInstance() will reserve memory blocks,
+ * AssignGlobalArenas will set the global sqlite configuration and
+ * CleanupInstance frees the memory blocks.  These three routines are not
+ * thread-safe.  AssignGlobalArenas has to be called as first sqlite operation,
+ * CleanupInstance has to be called after sqlite3_shutdown.  If the singleton
+ * is alive, read-only sqlite databases will automatically use it for their
+ * lookaside buffers (see sql.h).  Assignment of lookaside buffers is
+ * thread-safe.
+ */
+class MemoryManager {
```
reneme: `MemoryManager` implies that this is a general purpose manager while it is actually tailored for *sqlite*. `SqliteMemoryManager` maybe?

cvmfs/sqlitemem.h
```diff
@@ -0,0 +1,303 @@
...
+    if (instance_ == NULL)
+      instance_ = new MemoryManager();
+    return instance_;
+  }
+  static void CleanupInstance();
+  static bool HasInstance() { return instance_ != NULL; }
+  ~MemoryManager();
+
+  void AssignGlobalArenas();
+  void *AssignLookasideBuffer(void *db);
```
reneme: should probably be `sqlite_db *db` instead of `void *`, no?

cvmfs/sqlitemem.h
```diff
@@ -0,0 +1,303 @@
...
+    /**
+     * Number of bitmap ints for 128 buffers.
+     */
+    static const unsigned kNoBitmaps = kBuffersPerArena / (sizeof(int) * 8);
+
+    LookasideBufferArena();
+    ~LookasideBufferArena();
+
+    void *GetBuffer();
+    void PutBuffer(void *buffer);
```
reneme: `PutBuffer` (as well as `PutLookasideBuffer` and `PutMemory`) is freeing the memory chunk pointed to by `void*`, right? I find the naming unclear. Why not `Reserve***()` and  `Free***()`?

jblomer: Hm, it is a well-known naming convention from the kernel.  And it makes short names...

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,508 @@
...
+}
+
+
+/**
+ * Sqlite ensures that ptr != NULL and new_size > 0.
+ */
+void *MemoryManager::xRealloc(void *ptr, int new_size) {
+  void *new_ptr = xMalloc(new_size);
+  memcpy(new_ptr, ptr, new_size);
+  xFree(ptr);
```
reneme: I wonder if it would make sense to implement a `MemoryManager::ReallocMemory()`? If the next memory region is free by chance, this could just extend the allocated region. Questionable how likely this is, though.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,508 @@
...
+ * Sqlite ensures that ptr != NULL.
+ */
+int MemoryManager::xSize(void *ptr) {
+  return instance_->GetMemorySize(ptr);
+}
+
+
+int MemoryManager::xRoundup(int size) {
+  return size;
+}
```
reneme: A comment would be nice. I guess, we do not need to ensure certain block sizes, hence `xRoundup()` just returns the parameter.

cvmfs/sqlitemem.cc
jblomer: Indeed there is room for optimization.  But looking at the sqlite statistics, realloc is anyway a very rare operation so I decided it's not worth any effort.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,508 @@
...
+
+
+/**
+ * Opens new arenas as necessary.
+ */
+void *MemoryManager::GetMemory(int size) {
+  void *p = malloc_arenas_[idx_last_arena_]->Malloc(size);
+  if (p != NULL)
+    return p;
+  unsigned N = malloc_arenas_.size();
```
reneme: Knit picking: newlines to separate code blocks.

1. Try last successful arena
[newline]
2. Try all other arenas
[newline]
3. Create new arena

cvmfs/sqlitemem.h
```diff
@@ -0,0 +1,303 @@
...
+  MemoryManager();
+
+  void *GetLookasideBuffer();
+  void PutLookasideBuffer(void *buffer);
+
+  void *GetMemory(int size);
+  void PutMemory(void *ptr);
+  int GetMemorySize(void *ptr);
+
+  pthread_mutex_t *lock_;
```
reneme: this can be a class member, no? saves you from the burden of `delete lock_`.

cvmfs/sqlitemem.cc
```diff
@@ -0,0 +1,508 @@
...
+
+
+/**
+ * Closes empty areas.
+ */
+void MemoryManager::PutMemory(void *ptr) {
+  MallocArena *M = MallocArena::GetMallocArena(ptr);
+  M->Free(ptr);
+  unsigned N = malloc_arenas_.size();
+  if ((N > 1) && M->IsEmpty()) {
```
reneme: quick way out instead of additional indentations:
```C++
if (N <= 1 || !M->IsEmpty())
  return;

...
```

test/unittests/t_sqlitemem.cc
```diff
@@ -0,0 +1,254 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include <gtest/gtest.h>
+
+// #include <malloc.h> for malloc_usable_size
```
reneme: ?

test/unittests/t_sqlitemem.cc
```diff
@@ -0,0 +1,254 @@
...
+using namespace std;  // NOLINT
+
+namespace sqlite {
+
+class T_Sqlitemem : public ::testing::Test {
+ protected:
+  virtual void SetUp() {
+    EXPECT_FALSE(MemoryManager::HasInstance());
+    mem_mgr_ = MemoryManager::GetInstance();
+    EXPECT_TRUE(MemoryManager::HasInstance());
```
reneme: potentially `ASSERT_TRUE` makes more sense here... if the previous call fails there is no point in running the test...

reneme:
Im through with this. Interesting read!

jblomer:
Almost all of the comments are included.  For the time being I'll leave `sxmmap` as a thin wrapper around `mmap` like the other functions in smalloc.h.  If the caller does not want to remember the size, there is still the `smmap` call.

PR #1526, opened 2016-03-04
FIX/RFC: Make Sure that `cvmfs_suid_helper` Runs in Clean Environment

reneme:
That should make sure to run `cvmfs_suid_helper` always in a fresh environment. Amending  `$LD_LIBRARY_PATH` with a path that points into the CVMFS repository being published might make `cvmfs_suid_helper` pick up shared libraries from there. However, `cvmfs_suid_helper` will then not be able to `umount` the repository because it holds open file descriptors on the mount point itself.

That's sort of a quick fix... Better ideas?

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -1092,6 +1092,10 @@ check_lock() {
   __is_valid_lock "${path}" "$ignore_stale"
 }
 
+run_suid_helper() {
+  env -i /usr/bin/cvmfs_suid_helper $@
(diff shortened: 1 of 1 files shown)


DrDaveD:
Hang on, LD_LIBRARY_PATH can't be honored by a setuid-root program.  That would be a security violation.

DrDaveD:
In any case a setuid-root program has to clean its own environment, it can't depend on a calling script to do it.

PR #1527, opened 2016-03-04
FIX: `grep` Treating Whitelist as Binary Data

reneme:
We use `grep` to parse the expiry date of a downloaded whitelist. In some environments `grep` treats the whitelist as binary data (because of the attached binary signature) and produces bogus output.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -1649,9 +1649,9 @@ create_whitelist() {
 
 
 get_expiry_from_string() {
-  local whitelist_string="$1"
+  local whitelist="$1"
 
-  local expires=$(echo "$whitelist_string" | grep -e '^E[0-9]\{14\}$' | tr -d E)
+  local expires=$(echo "$whitelist" | grep --text -e '^E[0-9]\{14\}$' | tr -d E)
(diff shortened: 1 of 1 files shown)


jblomer:
Interesting.  Where did it break?

Moliholy:
Here:
```
1632:  create_whitelist() {
1633:    local name=$1  #   <---- here it breaks
1634:    local user=$2
1635:    local spooler_definition=$3
```bash
It takes the binary content, tries to assign it and fails

**Comment by DrDaveD** [comment]
According to the grep man page, --text is equivalent to --binary-files=text, and the description of --binary-files says it only has an effect if the first few bytes of a file indicate that it contains binary data.  Are you sure that --text makes a difference?

A solution that I use in several scripts is cat -v which turns binary data into ascii data, then make sure the command exits before it gets to the signature.  For example,
```
REVISION="`cat -v /srv/cvmfs/$REPO/.cvmfspublished|sed -n '/^S/{s/^S//p;q;}'`"
```

Moliholy:
We just tried manually modifying the cvmfs_server script and that solved the problem immediately. We'll take a deeper look anyway, just in case

PR #1529, opened 2016-03-08
FIX: `cvmfs_server` Open File Descriptor Check relies on `$PATH` for `lsof`

reneme:
This fixes an issue in `cvmfs_server` which prevents it from checking for open file descriptors on `cvmfs_server publish` and `cvmfs_server abort` if `lsof` cannot be found in the `$PATH`. In particular this is crucial for `cvmfs_server` commands that are called through `cron`, as it sets the `$PATH` to something like `/bin:/usr/bin:.`.

Along with `lsof` there have been a handful of other commands from `/sbin` and `/usr/sbin` that might have suffered the same issue.

Along with the fix there is a regression test that closely resembles the release manager setup of **lhcbdev.cern.ch** where the problem was initially found. See `526*/main` for more details.

*Note:* This builds on top of [Feature: Add `cvmfs_server publish -f` Disabling Interactive Open File Descriptor Dialog](https://github.com/cvmfs/cvmfs/pull/1528).

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -300,7 +300,7 @@ Supported Commands:
   publish         [-p pause for tweaks] [-n manual revision number] [-v verbose]
                   [-a tag name] [-c tag channel] [-m tag description]
                   [-X (force external data) | -N (force native data)]
-                  [-Z compression algorithm]
+                  [-Z compression algorithm] [-f use force remount if necessary]
                   <fully qualified name>
                   Make a new repository snapshot
   rmfs            [-p(reserve) repo data and keys] [-f don't ask again]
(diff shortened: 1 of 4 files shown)


reneme:
updated with `find_sbin()` for a poor-man's `which`.

cvmfs/cvmfs_server
```diff
@@ -54,21 +54,44 @@ publish_after_hook() { :; }
 
 [ -f /etc/cvmfs/cvmfs_server_hooks.sh ] && . /etc/cvmfs/cvmfs_server_hooks.sh
 
+find_sbin() {
+  local bin_name="$1"
+  local bin_path=""
+  for d in /sbin /usr/sbin /usr/local/sbin /bin /usr/bin /usr/local/bin; do
```
jblomer: I'm unsure, should we include /usr/local in this?

reneme: mmh, for the moment we would look for the following tools:
* `a2enmod`
* `apache2`
* `apache2ctl`
* `apachectl`
* `getcap`
* `getenforce`
* `httpd2`
* `lsof`
* `modprobe`
* `pidof`
* `runuser`
* `sestatus`
* `set-up`

Most of it is clearly system stuff which probably wouldn't end up in `/usr/local`. Only for apache I am not so sure...

reneme:
one more fix: `find_sbin()` lookup must be allowed to fail.

PR #1532, opened 2016-03-10
Tweak: Use `posix_fadvice()` in `cvmfs_swissknife sync`

reneme:
This uses [`posix_fadvice()`](http://man7.org/linux/man-pages/man2/posix_fadvise.2.html) to tell the kernel our intentions.

Files opened by `AsyncReader` are flagged with `FADV_WILLNEED` begging the kernel for readahead into kernel caches. Just before closing `AsyncReader` will reflag them to `FADV_DONTNEED` as they can now safely be evicted.

Furthermore `LocalUploader` flags files with `FADV_DONTNEED` as they are write-once read-never and do not need to be caches extensively.

### DIFF
--- a/cvmfs/file_processing/async_reader_impl.h
+++ b/cvmfs/file_processing/async_reader_impl.h
@@ -12,6 +12,7 @@
 #include <cerrno>
 
 #include "../logging.h"
+#include "../util/posix.h"
 
 // TODO(remeusel): remove this... wrong namespace (for testing)
 namespace upload {
--- a/cvmfs/file_processing/async_reader_impl.h
+++ b/cvmfs/file_processing/async_reader_impl.h
@@ -116,12 +116,7 @@ void Reader<FileScrubbingTaskT, FileT>::OpenNewFile(FileT *file) {
   }
   assert(fd > 0);
 
-  const int advice = POSIX_FADV_WILLNEED;
-  const int error_code = posix_fadvise(fd, 0, file->size(), advice);
-  if (error_code != 0) {
-    LogCvmfs(kLogSpooler, kLogVerboseMsg, "fd advice failed for %s (%d - %s)",
-             file->path().c_str(), error_code, strerror(error_code));
-  }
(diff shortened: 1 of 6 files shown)


reneme:
Findings so far: There is a huge difference between `man posix_fadvise` and what is actually happening in the kernel. While [the documentation](http://man7.org/linux/man-pages/man2/posix_fadvise.2.html) suggests `posix_fadvise()` to somewhat set a "cache strategy" for a file, the actual [kernel source](http://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/mm/fadvise.c#n115) reveals that:

1. `POSIX_FADV_NOREUSE` is a **NOOP**... thank you very much!
2. `POSIX_FADV_WILLNEED` is essentially a `readahead()` with all pros and cons
3. `POSIX_FADV_DONTNEED` is evicting pages immediately but won't change the strategy after the invocation: Hence, the pattern should be: 1. `open()`, 2. `read()/write()`, 3. `posix_fadvise()`, `[repeat 2. and 3. ...]`, `close()`.
4. `posix_fadvise()` is conservative for non-page-aligned offsets: it will rather free too little pages rather then purging useful stuff.

So taking all this into account and comparing this pull request against the latest *devel* on a single-core OpenStack machine with 2GiB RAM (~ 1.8G Page Cache), publishing 50.000 30kB files (~ 1.45GiB) I see a speedup of more than 3x (3:30 vs. 11:00 minutes) where the former virtually doesn't read anything from disk and the latter heavily suffers from page cache thrashing.

Bottom line: As long as the entire CernVM-FS change set fits comfortably into the page cache, we see reasonable performance. Beyond that, also this version will suffer from cache trashing.

What needs to be done:

* more measurements
  - in particular on the **test.cern.ch** with *zfs* and more than one core
* have a second look at the page alignments of the `posix_fadvise()` offsets
* make it compile on OS X again :o)

jblomer:
Interesting!

reneme:
rebased to latest master (merging include conflicts due to #1535).

PR #1533, opened 2016-03-10
Make Fence a common class for the loader and the fuse module

jblomer:
The concept of a "Fence" to protect critical regions was used in the loader and in the fuse module with very similar implementations.  Make it a common utility class.


cvmfs/fence.h
```diff
@@ -0,0 +1,97 @@
...
+ public:
+  explicit FenceGuard(Fence *fence) : fence_(fence) {
+    fence_->Enter();
+  }
+  ~FenceGuard() {
+    fence_->Leave();
+  }
+ private:
+  Fence *fence_;
+};
```
reneme: Sorry for looking at it only after you merged...
You could either conform `Fence` to the `Lockable` interface or write a template specialisation of `LockGuard<>`. In both cases one could just do:
```C++
Fence f;
LockGuard lg(f);
```

jblomer: Ok, noted for a follow-up PR.

PR #1535, opened 2016-03-11
Split util in several smaller units

jblomer:
A first steps towards smaller utility units.  In a follow-up, the util_concurrency should be moved in the same way and perhaps a few other units can move into util (like fence, prng).


reneme:
Looks like a reasonable split to me. Perhaps `t_util.cc` could be split accordingly but on the other hand its testing "everything" in `./util/...`.

PR #1538, opened 2016-03-14
Use identical includes in all source directories

jblomer:
Add the cvmfs directory to the default includes. The cvmfs directory becomes the common root for includes so that `#include "..."` is the same no matter where a source file is in the directory hierarchy. This also gets rid of ".." in the include paths.


cvmfs/util/util_string.cc
```diff
@@ -1,530 +0,0 @@
...
-    if (data[pos] == '=')
-      decoded->erase(decoded->length()-1);
-  }
-  return true;
-}
-
-
-#ifdef CVMFS_NAMESPACE_GUARD
-}  // namespace CVMFS_NAMESPACE_GUARD
-#endif
```
reneme: is all that not needed anymore?

reneme:
very reasonable change! Cheers!

cvmfs/util/util_string.cc
jblomer: Ah, sorry, that should have gone into a separate pull request.

This was added by mistake.  The actual file is util/string.cc (this is util/util_string.cc)

PR #1541, opened 2016-03-16
Feature: Asynchronous Cleanup of Scratch Directory

reneme:
This Pull Request is requesting for comments!

After `cvmfs_server publish`-ing a large change set it might take a significant time to clean-out the scratch directory (i.e. `/var/spool/cvmfs/<fqrn>/scratch`). This allows to run this clean-up asynchronously to allow for quicker publishing round-trips. Setting `CVMFS_ASYNC_SCRATCH_CLEANUP=false` disables this and reverts back to the legacy behaviour.

The general idea of this workflow is:

0. create `current` and `wastebin` inside `/var/spool/cvmfs/<fqrn>/scratch`
1. publish as usual (using `.../scratch/current` as union fs scratch directory)
2. umount `/cvmfs/<fqrn>` and `/var/spool/cvmfs/<fqrn>/rdonly`
3. move `.../scratch/current` into `.../wastebin/waste.XXXXXX/current`
4. create a fresh `.../scratch/current`
5. spawn daemon process asynchronously deleting everything in `.../wastebin`
6. mount `/var/spool/cvmfs/<fqrn>/rdonly` and `/cvmfs/<fqrn>`

Note that there might be multiple concurrent `cvmfs_suid_helper clear_scratch_async` but they should not step on each other's feet. Subject to review: Is this actually true?

Theoretically back-to-back transactions should become quicker but this effect might be mitigated due to the background process stressing the file system with the recursive delete.

Furthermore this requires a repository migration step *2.2.x --> 2.3.0* to be developed.

### DIFF
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -541,6 +541,13 @@ least CernVM-FS $creator to manipulate this repository."
     fi
   fi
 
+  if [ "$creator" = "2.2.0-1" ] || [ "$creator" = "2.2.1-1" ]; then
+    if version_greater_or_equal "2.3.0"; then
+      _repo_is_incompatible "$creator" $nokill
+      return $?
+    fi
+  fi
+
   return 0
 }
 
@@ -5643,6 +5650,47 @@ migrate_2_1_20() {
   load_repo_config $name
 }
 
+migrate_2_2_0() {
+  local name=$1
+  local destination_version="2.3.0-1"
+  local server_conf="/etc/cvmfs/repositories.d/${name}/server.conf"
+
+  # get repository information
+  load_repo_config $name
+  [ ! -z $CVMFS_SPOOL_DIR     ] || die "\$CVMFS_SPOOL_DIR is not set"
+  [ ! -z $CVMFS_USER          ] || die "\$CVMFS_USER is not set"
+  [ ! -z $CVMFS_UNION_FS_TYPE ] || die "\$CVMFS_UNION_FS_TYPE is not set"
+
+  echo "--> umount repository"
+  run_suid_helper rw_umount $name
+
+  echo "--> updating scratch directory layout"
+  local scratch_dir="${CVMFS_SPOOL_DIR}/scratch"
+  rm -fR   ${scratch_dir}
+  mkdir -p ${scratch_dir}/current
+  mkdir -p ${scratch_dir}/wastebin
+  chown -R $CVMFS_USER ${scratch_dir}
+
+  echo "--> updating /etc/fstab"
+  local tmp_fstab=$(mktemp)
+  local comment="added by CernVM-FS for ${name}"
+  sed -e "s~^\(.*\)\(${scratch_dir}\)\(.*${comment}\)$~\1\2/current\3~" /etc/fstab > $tmp_fstab
+  cat $tmp_fstab > /etc/fstab
+  rm -f $tmp_fstab
+
+  echo "--> remount repository"
+  run_suid_helper rw_mount $name
+
+  echo "--> updating server.conf"
+  sed -i -e "s/^\(CVMFS_CREATOR_VERSION\)=.*/\1=$destination_version/" $server_conf
+
+  echo "--> ensure binary permission settings"
+  ensure_swissknife_suid $CVMFS_UNION_FS_TYPE
+
+  # update repository information
+  load_repo_config $name
+}
+
 migrate() {
   local names
   local retcode=0
@@ -5706,6 +5754,13 @@ migrate() {
          x"$creator" = x"2.2.0-0" ];
     then
       migrate_2_1_20 $name
+    fi
+
+    if [ x"$creator" = x"2.2.0-1" -o   \
+         x"$creator" = x"2.2.1-1" ] && \
+         is_stratum0 $name;
+    then
+      migrate_2_2_0 $name
     fi
 
   done
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -5672,11 +5672,8 @@ migrate_2_2_0() {
   chown -R $CVMFS_USER ${scratch_dir}
 
   echo "--> updating /etc/fstab"
-  local tmp_fstab=$(mktemp)
   local comment="added by CernVM-FS for ${name}"
-  sed -e "s~^\(.*\)\(${scratch_dir}\)\(.*${comment}\)$~\1\2/current\3~" /etc/fstab > $tmp_fstab
-  cat $tmp_fstab > /etc/fstab
-  rm -f $tmp_fstab
+  sed -i -e "s~^\(.*\)\(${scratch_dir}\)\(.*${comment}\)$~\1\2/current\3~" /etc/fstab
 
   echo "--> remount repository"
   run_suid_helper rw_mount $name
--- a/cvmfs/cvmfs_server
+++ b/cvmfs/cvmfs_server
@@ -2120,7 +2120,7 @@ create_config_files_for_new_repository() {
 
   # other configurations
   local spool_dir="/var/spool/cvmfs/${name}"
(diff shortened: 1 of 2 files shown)


cvmfs/cvmfs_suid_helper.cc
```diff
@@ -104,64 +105,127 @@ static void KillCvmfs(const string &fqrn) {
...
+
+static bool ClearDirectory(const string &path) {
+  ScopedWorkingDirectory swd(path);
+  if (!swd) {
+    return false;
+  }
+
+  bool success = true;
+  ScopedWorkingDirectory::DirectoryEntry dirent;
+  while (success && swd.NextDirectoryEntry(&dirent)) {
```
reneme: This stops the recursive delete as soon as there was an error, which might result in an unfinished cleanup. Whatever is left-over should be taken care of by the following publish. Not sure if we should have more safe guards here?

DrDaveD:
It's a good idea in general.  I do something similar in frontier-squid to asynchronously clear the cache.   I do it all in bash however, which seems a lot simpler than the rather extensive changes to run_suid_helper.  I can give you the details if you're interested.

I don't see anything about catching signals.  In frontier-squid I run the rm with nohup so they'll keep going after a user logs out.  I think that's important.

I suggest reversing the sense of the variable to be CVMFS_SYNC_SCRATCH_CLEANUP so it can default to false and not require a migration step.  I doubt that anybody will have a problem with the default and want to change it.

reneme:
> I do it all in bash however, which seems a lot simpler than the rather extensive changes to run_suid_helper.

Unfortunately we need to do it in `cvmfs_suid_helper` because the scratch area might contain files with skewed permissions (f.e. hidden housekeeping stuff from the union file system). The changes to this code aren't so extensive, its mostly a refactoring of the recursive delete.

> I don't see anything about catching signals.

Frankly I'm not sure, but do I need to do that after a double-fork?

> I suggest reversing the sense of the variable to be CVMFS_SYNC_SCRATCH_CLEANUP so it can default to false and not require a migration step.

That's a good idea. I also thought about making this more transparent. That will require a bit of extra code in `cvmfs_server` but perhaps it is more reasonable than the explicit migration step.

Thanks for the comments.

DrDaveD:
> Unfortunately we need to do it in cvmfs_suid_helper because the scratch area might contain files with skewed permissions (f.e. hidden housekeeping stuff from the union file system). The changes to this code aren't so extensive, its mostly a refactoring of the recursive delete.

I wasn't suggesting to not use cvmfs_suid_helper, but it could be cvmfs_suid_helper &

> Frankly I'm not sure, but do I need to do that after a double-fork?

I'm pretty sure you do, unless you call setsid() to give it a separate process group.  Try it.

reneme:
> I'm pretty sure you do, unless you call setsid() to give it a separate process group. Try it.

It seems to be fine (Ubuntu 15.10) with just this double-fork but anyway I've added `setsid()`.

Furthermore I've added a `cvmfs_server migrate` step for the change in `/var/spool/cvmfs/<repo>/scratch`. We won't easily get around that, I'm afraid.

cvmfs/cvmfs_server
```diff
@@ -5624,6 +5650,47 @@ migrate_2_1_20() {
...
+  # get repository information
+  load_repo_config $name
+  [ ! -z $CVMFS_SPOOL_DIR     ] || die "\$CVMFS_SPOOL_DIR is not set"
+  [ ! -z $CVMFS_USER          ] || die "\$CVMFS_USER is not set"
+  [ ! -z $CVMFS_UNION_FS_TYPE ] || die "\$CVMFS_UNION_FS_TYPE is not set"
+
+  echo "--> umount repository"
+  run_suid_helper rw_umount $name
+
+  echo "--> updating scratch directory layout"
```
jblomer: Can this break if the server package is update while a transaction is open?

cvmfs/cvmfs_server
```diff
@@ -5624,6 +5650,47 @@ migrate_2_1_20() {
...
+  local scratch_dir="${CVMFS_SPOOL_DIR}/scratch"
+  rm -fR   ${scratch_dir}
+  mkdir -p ${scratch_dir}/current
+  mkdir -p ${scratch_dir}/wastebin
+  chown -R $CVMFS_USER ${scratch_dir}
+
+  echo "--> updating /etc/fstab"
+  local tmp_fstab=$(mktemp)
+  local comment="added by CernVM-FS for ${name}"
+  sed -e "s~^\(.*\)\(${scratch_dir}\)\(.*${comment}\)$~\1\2/current\3~" /etc/fstab > $tmp_fstab
```
jblomer: Can we use `sed -i -e` and avoid handling the temp file?

cvmfs/cvmfs_suid_helper.cc
```diff
@@ -104,64 +105,129 @@ static void KillCvmfs(const string &fqrn) {
...
+    assert(cwd == path);
+    return string(cwd);
+  }
+
+  void ChangeDirectory(const string &path) {
+    const int retval = chdir(path.c_str());
+    assert(retval == 0);
+  }
+
+  bool IsMetaEntry(const platform_dirent64 *dirent) {
```
jblomer: Perhaps `IsDotEntry`?

cvmfs/cvmfs_suid_helper.cc
```diff
@@ -104,64 +105,129 @@ static void KillCvmfs(const string &fqrn) {
...
+  const string wastebin = string(kSpoolArea) + "/" + fqrn + "/scratch/wastebin";
+
+  // double-fork to daemonize the process
+  pid_t pid;
+  int statloc;
+  if ((pid = fork()) == 0) {
+    int retval = setsid();
+    assert(retval != -1);
+    if ((pid = fork()) == 0) {
+      close(0); close(1); close(2);
```
jblomer: Inside `CleanupDirectory`, we would still print to `stderr` if things fail, no?

cvmfs/cvmfs_server
reneme: Potentially... We could check for open transactions in `migrate()` but that is going to cause a dead-lock. `cvmfs_server` will prevent you from closing the transaction with `cvmfs_server abort` if the version number doesn't fit.

Anyway: That might be more favourable than deleting the scratch directory which might contain valuable data.

cvmfs/cvmfs_suid_helper.cc
reneme: [aye](https://github.com/cvmfs/cvmfs/pull/1541/commits/3d4610df4aa7da545f05f2eecfdfee1c5dd61d1a).

cvmfs/cvmfs_server
reneme: [mais, bien sûre](https://github.com/cvmfs/cvmfs/pull/1541/commits/f29e4b2f83cde42921a42709cbb8c02d150db98e).

cvmfs/cvmfs_suid_helper.cc
reneme: Mhh, wouldn't that pollute a random shell that potentially is long gone? My hope would be that consecutive runs of this cleanup would free such things. However, some kind of log would probably make sense. Syslog?

cvmfs/cvmfs_server
reneme: I'll submit a patch for that in another pull request.

cvmfs/cvmfs_suid_helper.cc
reneme: [Fixed](https://github.com/cvmfs/cvmfs/pull/1541/commits/4763580be4b970d575b110a6168affc77445c1a0) by code duplication after all. I didn't factor out `Daemonize()` in `util/posix.h` because it calls to `LogCvmfs()` whose dependency I'd like to avoid.

PR #1542, opened 2016-03-17
Optimisation: Create/Remove Nested Catalogs

reneme:
This was on my list since a long time: Disentangling the management of nested catalogs from the processing of `.cvmfscatalog` in `SyncMediator`. This might speed up snapshotting significantly in certain situations.

Up to now, a nested catalog was created as soon as `cvmfs_server publish` hit a `.cvmfscatalog` file, and removed when it found a deleted `.cvmfscatalog` file. Conceptually that is fine, but it might lead to *many* unnecessary catalog updates. 

For example: Imagine a user adds a new directory `release` and marks it a nested catalog (i.e. `release/.cvmfscatalog`), furthermore she adds 150.000 files to this new directory. In the worst case, `cvmfs_server publish` would traverse all 150.000 files, adding them to the father catalog of `release`. Only in the end it would find `.cvmfscatalog` which forced `cvmfs_server publish` to split the father at `release` and copying all 150.000 entries to the new nested catalog.

Similarly this could happen when `cvmfs_server publish` first finds a deleted `.cvmfscatalog` file and copys-up all entries into the father just before removing all of them, because actually the entire directory was deleted.

That patch disentangles the creation of nested catalogs from the processing of `.cvmfscatalog` in `cvmfs_server publish`. That simplifies the code and also allows to do the catalog creation smarter. I.e. a newly created directory containing a `.cvmfspublished` file will immediate result in a new (and empty) catalog that gets filled right afterwards. Similarly a removed directory containing `.cvmfspublished` is deleting an empty catalog and avoids an expensive and useless copy-up.

TODO: In the latter case, it would be even better to just remove the catalog and not even bother deleting all the entries first. But with our main use case directory deletion doesn't happen too often...

Furthermore this adds a test-case to verify the catalog creation/deletion sequencing.

### DIFF
--- a/cvmfs/sync_item.cc
+++ b/cvmfs/sync_item.cc
@@ -49,8 +49,7 @@ SyncItem::SyncItem(const string       &relative_parent_path,
   compression_algorithm_(zlib::kZlibDefault)
 {
   content_hash_.algorithm = shash::kAny;
-  // Note: graft marker for non-regular files are silently ignored
-  if (IsRegularFile()) {CheckGraft();}
+  CheckMarkerFiles();
 }
 
 
@@ -214,6 +213,12 @@ std::string SyncItem::GetScratchPath() const {
   const string relative_path = GetRelativePath().empty() ?
                                "" : "/" + GetRelativePath();
   return union_engine_->scratch_path() + relative_path;
+}
+
+void SyncItem::CheckMarkerFiles() {
+  if (IsRegularFile()) {
+    CheckGraft();
+  }
 }
 
 
--- a/cvmfs/sync_item.cc
+++ b/cvmfs/sync_item.cc
@@ -20,6 +20,7 @@ SyncItem::SyncItem() :
   whiteout_(false),
   opaque_(false),
   masked_hardlink_(false),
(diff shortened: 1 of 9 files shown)


cvmfs/sync_item.cc
```diff
@@ -216,6 +217,18 @@ std::string SyncItem::GetScratchPath() const {
...
+void SyncItem::CheckMarkerFiles() {
+  if (IsRegularFile()) {
+    CheckGraft();
+  } else if (IsDirectory()) {
+    CheckCatalogMarker();
+  }
+}
+
+void SyncItem::CheckCatalogMarker() {
+  has_catalog_marker_ = FileExists(GetUnionPath() + "/.cvmfscatalog");
```
jblomer: We should probably also look for `.cvmfsautocatalog`

reneme: `.cvmfsautocatalog` is not supposed to trigger nested catalog creation. It is dropped by the `CatalogBalancer` during `cvmfs_server publish`. I think should leave it out of those mechanics.

PR #1547, opened 2016-03-23
Feature: Regenerate Reflog from Available Catalog and History Chains

reneme:
This should be the final building block of the `Reflog` business. It introduces a new `cvmfs_swissknife` command (i.e. `recreate_reflog`) that walks as much of the root `Catalog` and `History` chains as possible to reconstruct the `Reflog`.

The use case is two fold: **1st** it allows for an easy transition from the old-style garbage collection to the `Reflog` driven version and **2nd** it serves as a failure recovery mechanism in case the original `Reflog` has been corrupted. We cannot guarantee that a full `Reflog` is reconstructable from the stored `Catalog` and `History` chains but this gives some best-effort.

Furthermore this contains a [minor syslog logging fix](https://github.com/cvmfs/cvmfs/commit/87bdd4aeb3526c4f5395a6b7307d144da0640374) which I was too lazy to extract. Also it adapts integration test 563 "Garbage Collection After Legacy Repository Import" to specifically check for the new feature along with CernVM-FS 2.0.x catalogs.

Note: This builds on top of [FIX: Reflog Update During Snapshotting](https://github.com/cvmfs/cvmfs/pull/1544), [Feature: Reflog cleanup on Garbage Collection](https://github.com/cvmfs/cvmfs/pull/1545) and [Feature: Swissknife Ignores Missing Reflog](https://github.com/cvmfs/cvmfs/pull/1546).

### DIFF
--- a/cvmfs/swissknife_gc.cc
+++ b/cvmfs/swissknife_gc.cc
@@ -126,6 +126,8 @@ int CommandGc::Main(const ArgumentList &args) {
     }
   }
 
+  reflog->BeginTransaction();
+
   GcConfig config;
   config.uploader                = uploader.weak_ref();
   config.keep_history_depth      = revisions;
@@ -152,7 +154,12 @@ int CommandGc::Main(const ArgumentList &args) {
 
   GC collector(config);
   const bool success = collector.Collect();
-  uploader->TearDown();
+
+  if (!success) {
+    LogCvmfs(kLogCvmfs, kLogStderr, "garbage collection failed");
+    uploader->TearDown();
+    return 1;
+  }
 
   if (deletion_log_file != NULL) {
     const int bytes_written = fprintf(deletion_log_file,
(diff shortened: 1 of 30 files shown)


cvmfs/swissknife_gc.cc
```diff
@@ -159,6 +169,19 @@ int CommandGc::Main(const ArgumentList &args) {
...
 
+  reflog->CommitTransaction();
+  const std::string reflog_db = reflog->CloseAndReturnDatabaseFile();
+  uploader->Upload(reflog_db, ".cvmfsreflog");
+  uploader->WaitForUpload();
+  unlink(reflog_db.c_str());
+
+  if (uploader->GetNumberOfErrors() > 0) {
+    LogCvmfs(kLogCvmfs, kLogStderr, "failed to upload updated Reflog "
+                                    "(not critical)");
```
jblomer: consider returning an error case here

reneme: done

PR #1552, opened 2016-04-06
FIX: Remove Usage of SSLv3_client_method() in libcurl

reneme:
This disables SSLv3 in the `libcurl` version we are currently using. As of OpenSSL 1.0.2g SSLv3 is [disabled on Ubuntu](https://launchpad.net/ubuntu/+source/openssl/1.0.2g-1ubuntu1) 15.10 which prevents a successful build of CernVM-FS. The OpenSSL ABI didn't change, hence this doesn't affect our released binaries. This patch will most likely become obsolete when we update our `libcurl` version.


reneme:
yuk! sorry, wrong push!

reneme:
There you go...

PR #1559, opened 2016-04-08
Download manager

Moliholy:
This PR collects all the changes I had time to make in order to modify the download manager to use curl_multi_perform, removing the massive thread spawn and hence optimising the resources. However there is still work to be done.


jblomer:
Cool!  Thanks for the submission!

jblomer:
HTTP/2 efforts parked for the time being.

PR #1561, opened 2016-04-11
FIX: memory corruption during custom sqlite realloc

jblomer:


reneme:
A regression test would be cool. Perhaps allocating a small chunk of data first and then realloc-ing a huge piece of memory would somewhat reliably reproduce that?

PR #1569, opened 2016-04-13
Fix reallocations by small deltas

jblomer:
There is an obvious, even better solution than `std::min(new_size, old_size)`.  If the old size can accomodate the new size, we don't need to do anything.

The new `xRealloc` stress test could trigger the segfault pretty quickly.


cvmfs/sqlitemem.cc
```diff
@@ -339,6 +339,9 @@ void SqliteMemoryManager::xFree(void *ptr) {
  */
 void *SqliteMemoryManager::xRealloc(void *ptr, int new_size) {
   int old_size = xSize(ptr);
+  if (old_size >= new_size)
+    return ptr;
```
reneme: sweet!

PR #1570, opened 2016-04-15
Tuning: Speed up Catalog Updates during Publish

reneme:
This removes inefficiencies from the catalog update code during publishing. Profiling of these code paths revealed that path lookups and nested catalog discovery happened more often than necessary. Especially for large change sets and many touched nested catalogs we see a significant penalty.

In certain workloads one can expect a 25% speedup for the catalog updates as seen in the chart below (blue = before, green = after). This is meta-data updates per seconds over number of inserted directory entries (measured on my dev machine).

screen shot 2016-04-15 at 14 10 17

I'll put a more detailed comments into the patch itself.

### DIFF
--- a/cvmfs/catalog_mgr_rw.cc
+++ b/cvmfs/catalog_mgr_rw.cc
@@ -172,13 +172,15 @@ manifest::Manifest *WritableCatalogManager::CreateRepository(
 
 
 /**
- * Tries to retrieve the catalog containing the given path
- * This method is just a wrapper around the FindCatalog method of
- * AbstractCatalogManager to provide a direct interface returning
- * WritableCatalog classes.
- * @param path the path to look for
- * @param result the retrieved catalog (as a pointer)
- * @return true if catalog was found, false otherwise
+ * Retrieve the catalog containing the given path.
+ * Other than AbstractCatalogManager::FindCatalog() this mounts nested
+ * catalogs if necessary and returns  WritableCatalog objects.
+ * Furthermore it optionally returns the looked-up DirectoryEntry.
+ *
+ * @param path    the path to look for
+ * @param result  the retrieved catalog (as a pointer)
+ * @param dirent  is set to looked up DirectoryEntry for 'path' if non-NULL
+ * @return        true if catalog was found
  */
 bool WritableCatalogManager::FindCatalog(const string     &path,
                                          WritableCatalog **result,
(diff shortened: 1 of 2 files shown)


cvmfs/catalog_mgr_rw.cc
```diff
@@ -172,29 +172,34 @@ manifest::Manifest *WritableCatalogManager::CreateRepository(
...
+ * @param result  the retrieved catalog (as a pointer)
+ * @param dirent  is set to looked up DirectoryEntry for 'path' if non-NULL
+ * @return        true if catalog was found
  */
-bool WritableCatalogManager::FindCatalog(const string &path,
-                                         WritableCatalog **result)
-{
+bool WritableCatalogManager::FindCatalog(const string     &path,
+                                         WritableCatalog **result,
+                                         DirectoryEntry   *dirent) {
```
reneme: This method retrieves the `Catalog` containing `path` and then does a lookup to ensure that `path` is available. Sometimes the `DirectoryEntry` of this lookup is needed in the caller as well. The (optional) `dirent` out-parameter allows the caller to obtain it directly and safe an expensive call to `Lookup()`.

cvmfs/catalog_mgr_rw.cc
```diff
@@ -172,29 +172,34 @@ manifest::Manifest *WritableCatalogManager::CreateRepository(
...
+  bool retval = MountSubtree(ps_path, best_fit, &catalog);
   if (!retval)
     return false;
 
   catalog::DirectoryEntry dummy;
-  bool found = LookupPath(path, kLookupSole, &dummy);
+  if (NULL == dirent) {
+    dirent = &dummy;
+  }
+  bool found = catalog->LookupPath(ps_path, dirent);
```
reneme: At this point `catalog` is guaranteed to be the correct `Catalog`. Therefore we can directly call `catalog->LookupPath()` instead of `CatalogMgr<>::LookupPath()` which would do another costly `AbstractCatalogManager::FindCatalog()`.

jblomer: I believe the new version would return the parent catalog for a transition point, while the old version would return the nested catalog.

jblomer: Hm, or maybe not... I think we should test this.

reneme: This shouldn't make a difference. The returned catalog is the same for sure, since the `FindCatalog()` call wasn't changed. The `DirectoryEntry` was not returned before but looked up again via `Catalog::LookupPath()` (see changes just below).

In fact the logic didn't change at all, it just avoids calling certain lookups more than needed.

PR #1571, opened 2016-04-18
Feature: Concurrent Catalog Committing and Snapshotting

reneme:
This parallelises the catalog committing code that is part of `cvmfs_server publish`. Depending on the size of the published change set and the number of modified catalogs, this can potentially increase the publishing speed significantly.


reneme:
Fixed a race condition when `WritableCatalog::UpdateNestedCatalog()` and `Catalog::FindNested()` were used concurrently. Furthermore added a nested catalog creation stress test and fixed the cpplint errors.

PR #1576, opened 2016-04-21
FIX/Feature: cvmfs client-side authorization

jblomer:
Another round for the authorization.  Apologies for the large number of changes.  @bbockelm Apologies for screwing up `git blame`; all the meat is a copy & paste of your code, it just turned out to be easier to use the editor than git commands.

Principal change is a better separation between the cvmfs client and the authorization helper, including a well-defined interface between the two. The authorization helper executable needs to be defined in `CVMFS_AUTHZ_HELPER`.  Actually we might want to be smarter and define a naming convention to automatically deduct the authorization helper from the authz string in the catalog.  Currently there is of course only one (X.509 proxy) but we might see others in the future. @bbockelm, what do you think?

The new helper `cvmfs_x509_helper` also comes with its own spec file so that it can be developed and deployed independently from cvmfs (e.g. faster).

I'm still wondering if we want to have a dedicated audit log (different from debug log, syslog), to log all authorization relevant events.  In particular successful/failed authorizations, of course.

Regarding the interface: the helper gets spawned by the cvmfs client with stdin/stdout connected.  There is a command/reply style of messages.  Messages have a 4 byte version (`=1`), a 4 byte length, and then a JSON text that needs to contain the top-level struct `cvmfs_authz_v1 { ... }`.  Communication starts with a handshake where the client passes logging parameters to the authz helper.  The client then sends zero or more authorization requests that are answered by a positive or negative permit each that includes the access token for positive permits (in this case: X.509 proxy).  The permits are cached by the client with a TTL that the helper can chose.  On unmount, the client sends a `quit` command to the helper.

The download manager got an abstract interface of a class that can perform per-transfer surgery on CURL handles.  This way, the download manager does not need to know about authz internals.

All in all, I don't think we can make this modification part of a bugfix release with a clear conscience.  Once we are all happy with the code and have it merged, I'd suggest to have a phone call on release plans.  I'd much rather make a 2.3 feature release before Q4/16 and move the remaining tasks to version 2.4.


bbockelm:
Two thoughts:
- I had always planned to provide a prefix of some sort (`x509:/DC=ch/DC=.../CN=Brian` versus `voms:/cms` versus `oauth2:some-great-authz`) in the future.  If we are already doing this much work, we might as well add it now.  I assume we ought to ignore prefixes that are not understood and fall back to the current logic for things starting with `/`?
- I agree it's becoming a bit disingenuous to call this a bugfix (although it certainly does fix a bug...).  On the other hand, I really need something I can deliver to LIGO relatively soon.  I was hoping to start scaling up the use in the summer, meaning we're already late.  Could we branch 2.3 now, mostly consisting of this feature?

DrDaveD:
I'm also fine with renaming what we were expecting to be 2.2.2 to 2.3.0.

bbockelm:
Hm - when running this branch, I see:
```
Apr 21 21:07:58 hcc-briantest7.unl.edu cvmfs2: (ligo.osgstorage.org) starting authz helper /usr/bin/cvmfs_x509_helper
Apr 21 21:07:58 hcc-briantest7.unl.edu cvmfs2: (ligo.osgstorage.org) authz helper uses unknown protocol version 1936287828
Apr 21 21:07:58 hcc-briantest7.unl.edu cvmfs2: (ligo.osgstorage.org) authz helper /usr/bin/cvmfs_x509_helper enters fail state, no more authorization
```
Byteswap issues?

bbockelm:
Oh - and when the helper dies, `wait` isn't called on it: it stays around as a zombie process.

bbockelm:
Ah - for some reason, glibc is reopening stderr as /dev/null; maybe something to do with setuid handling?  Anyhow, disabled this for now.

However, `cvmfs_x509_helper` asserts:
```
(gdb) up 1
#4  0x0000000000404437 in ParseRequest (
    msg="{\"cvmfs_authz_v1\":{\"msgid\":2,\"revision\":0,\"uid\":1221,\"gid\":10058,\"pid\":1955911,\"membership\":\"/DC=es/DC=irisgrid/O=uib/CN=alicia.sintes\n/DC=org/DC=cilogon/C=US/O=LIGO/CN=Alicia Sintes alicia.sintes@lig"...) at /home/cse496/bbockelm/projects/cvmfs/cvmfs/authz/x509_helper.cc:128
128	  assert((json != NULL) && (json->first_child != NULL));
(gdb) p json
$1 = (JSON *) 0x0
(gdb) p msg
$2 = "{\"cvmfs_authz_v1\":{\"msgid\":2,\"revision\":0,\"uid\":1221,\"gid\":10058,\"pid\":1955911,\"membership\":\"/DC=es/DC=irisgrid/O=uib/CN=alicia.sintes\n/DC=org/DC=cilogon/C=US/O=LIGO/CN=Alicia Sintes alicia.sintes@lig"...
(gdb) p err_desc
$3 = 0x40b3e8 "Control characters not allowed in strings"
(gdb) p err_pos
$4 = 0x7fffd2d21e8d "/DC=es/DC=irisgrid/O=uib/CN=alicia.sintes\n/DC=org/DC=cilogon/C=US/O=LIGO/CN=Alicia Sintes alicia.sintes@ligo.org\n/DC=org/DC=cilogon/C=US/O=LIGO/CN=Nelson Christensen nelson.christensen@ligo.org\n/DC=or"...
(gdb) p err_line
$5 = 1
```

Looks like it's not properly escaping the newline `\n` character?

jblomer:
I see, let's base64 the membership string.

jblomer:
As for the zombie: the failure handling is currently very simple.  Whenever something goes wrong, the cvmfs client puts the helper in "fail state" and doesn't do any authz anymore.  It will call `wait` eventually but you need to `cvmfs_config reload` the client (or unmount).

jblomer:
For checking the call context: the client now exports `CVMFS_AUTHZ_HELPER=yes` when starting the authz helper.  This is used on the helper end to determine the call context instead of testing file descriptor 2.

jblomer:
I'm not sure if I got the automatic authz helper discovery right.  The way it works now is that the everything before the first colon is treated as an "authz schema".  Cvmfs then looks for `/usr/libexec/cvmfs/authz/cvmfs_<authz_schema>_helper`.  The helper will receive the authz string with the schema stripped.

cvmfs/authz/authz_curl.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <string>
+
+#include "download.h"
+
+class AuthzSessionManager;
+
+class AuthzAttachment : public download::CredentialsAttachment {
+ public:
+  explicit AuthzAttachment(AuthzSessionManager *sm);
```
bbockelm: GCC generates a warning that there is no virtual destructor.

cvmfs/authz/authz_fetch.h
```diff
@@ -0,0 +1,195 @@
...
+   * The "wire" protocol: 4 byte version, 4 byte length, JSON message.  Must
+   * be the same for cvmfs and helper.
+   */
+  static const uint32_t kProtocolVersion;  // = 1;
+
+  AuthzExternalFetcher(const std::string &fqrn,
+                       const std::string &progname,
+                       const std::string &search_path);
+  AuthzExternalFetcher(const std::string &fqrn, int fd_send, int fd_recv);
+  ~AuthzExternalFetcher();
```
bbockelm: Destructor should be virtual.

bbockelm:
Hm - on further examination, I'm not sure the authz helper discovery approach is right.  Take an illustrative subset of the LIGO repo:
```
/DC=org/DC=cilogon/C=US/O=LIGO/CN=Ka-Lok Lo ka-lok.lo@ligo.org
/DC=org/DC=cilogon/C=US/O=LIGO/CN=Otto Hannuksela otto.hannuksela@ligo.org
osg:/osg/ligo
LIGO:/ligo
```
So, there's a mix of VOMS extensions and X509 DNs (note the VOMS entries require a `:` character).

I was thinking we could upgrade this to something like this:
```
x509:/DC=org/DC=cilogon/C=US/O=LIGO/CN=foo
x509:/DC=org/DC=cilogon/C=US/O=LIGO/CN=bar
x509:osg:/osg/ligo
x509:LIGO:/ligo
```
Which then allows us to add lines of the form:
```
krb5:foo@FNAL.GOV
oauth2:baz
```
However, the current system doesn't really allow mixing auth types inside a single auth list.  I would be OK with maybe backing out b7af054 and saving this for later.

jblomer:
Couldn't we prefix the multi-line authz string once with `x509`?  The colon as a separator is indeed a bit confusing, how about a pipe instead, i.e.

    x509|/DC=org/DC=cilogon/C=US/O=LIGO/CN=foo
    /DC=org/DC=cilogon/C=US/O=LIGO/CN=bar
    osg:/osg/ligo
    LIGO:/ligo

If we later want to mix authz schemes, we can introduce a `multi` scheme with a corresponding helper that takes care of the dispatching.

bbockelm:
The pipe sounds good.

jblomer:
I switched to `%` as a separator, the pipe is a pain given all the bash code in the server.

bbockelm:
Should we default to X509 if nothing is specified?

Or would it be better to have client tools reject updates that don't contain a method?

PR #1591, opened 2016-05-02
Forward CVMFS environment to authz helper.

bbockelm:
Although it sounds like we might eventually take a different tack, I wanted to record this work.

Here's one approach on `CVM-1010`; simply forward the runtime environment and options to the authz helper.


jblomer:
Superseded by #1595

PR #1595, opened 2016-05-04
Fixes and tests for authz module

jblomer:


cvmfs/authz/authz_fetch.cc
```diff
@@ -118,8 +122,16 @@ void AuthzExternalFetcher::ExecHelper() {
...
   char *argv[] = {argv0, NULL};
-  char *envp0 = strdupa("CVMFS_AUTHZ_HELPER=yes");
-  char *envp[] = {envp0, NULL};
+
+  const bool strip_prefix = true;
+  vector<string> authz_env =
+    options_manager_->GetEnvironmentSubset("CVMFS_AUTHZ_", strip_prefix);
+  vector<char *> envp;
+  for (unsigned i = 0; i < authz_env.size(); ++i)
+    envp.push_back(strdupa(authz_env[i].c_str()));
```
bbockelm: Is `strdupa` necessary here?  This should create a string object, which will allocate new memory.  No?

jblomer: I wasn't sure and I thought a little stack alloc won't hurt.  I think I remember that we had problems in the past with using `string.c_str()` directly.  I thought the c string is only temporarily valid.

PR #1615, opened 2016-07-01
Initial implementation of a tiered cache.

bbockelm:
This implementation provides a hierarchical write-through cache. It allows a site to setup a local disk cache in addition to a site-wide alien cache.

I envision this to be useful in two cases:
   1. The performance of the alien cache (or, in the future, object storage) is better than a HTTP proxy setup because the underlying storage system allows several storage servers to peer.
   2. When the repository is a secure CVMFS repo, as secure repos cannot be used with HTTP proxies.

No tests provided at this point - I just want to make sure this is the correct approach, coding-wise (plus push things through Travis-CI).


bbockelm:
Hi @jblomer -

Do you think this work needs to wait for the generic config syntax (https://sft.its.cern.ch/jira/browse/CVM-1053) first?  How is the RAM cache approaching config?

Brian

jblomer:
Hi @bbockelm, I think it would make sense to have the tiered cache manager with final syntax, tests, etc. part of cvmfs main code and to attach Ceph and Hadoop "concrete" cache managers using the (yet to be defined) IPC protocol as an extra process.

jblomer:
Superseded by #1690

PR #1616, opened 2016-07-04
Feature ramcache

trshaffer:
Here's a first draft of the RAM cache feature. I put in unit tests to check the core functionality, so I should probably work on gluing it into the FUSE module next. I think more statistics and logging details would also be a good idea.

Let me know what you think, @jblomer


cvmfs/CMakeLists.txt
```diff
@@ -59,6 +61,7 @@ set (CVMFS_CLIENT_SOURCES
   history_sqlite.cc history_sqlite.h
   json_document.cc json_document.h
   loader.h
+  kvstore.h kvstore.cc
```
jblomer: This was my mistake, I think.  The kvstore files should be before `loader.h` for alphabetical order.

cvmfs/cache_ram.h
```diff
@@ -0,0 +1,191 @@
...
+#include "statistics.h"
+#include "cache.h"
+#include "hash.h"
+#include "util/pointer.h"
+#include "kvstore.h"
+
+namespace cache {
+
+/**
+ * ...
```
jblomer: Here it would be good to have a short description of the purpose of the RamCacheManager and a brief sketch how it is implemented.

cvmfs/cache_ram.h
```diff
@@ -0,0 +1,191 @@
...
+   * volatile entries. If there is still not enough room, try evicting regular entries. If there is *still* not enough
+   * space, give up an return failure. Note that evictions only occur if they will produce enough space for the transaction;
+   * if -ENOSPC is returned, the states of the transaction and cache are unchanged.
+   * @param txn A pointer to space allocated for storing the transaction details
+   * @returns -ENOSPC if the transaction would exceed the size of the cache
+   */
+  virtual int CommitTxn(void *txn);
+
+ private:
+  struct ReadOnlyFd {
```
jblomer: It would be good to explain that the null hash serves as an invalid handle, and perhaps why we don't need to know about the size of the object at this point (I guess out-of-bounds reads either don't happen or they are handled by the kvstore implementation).

cvmfs/cache_ram.h
```diff
@@ -0,0 +1,191 @@
...
+    if ((fd < 0) || (static_cast<unsigned>(fd) >= open_fds_.size()))
+      return false;
+    return open_fds_[fd].handle != invalid_fd_;
+  }
+
+  int AddFd(const ReadOnlyFd &fd);
+  int64_t CommitToKvStore(Transaction *transaction);
+  virtual int DoOpen(const shash::Any &id);
+
+  uint64_t max_size;
```
jblomer: Should end on underscore, i.e. `max_size_`

cvmfs/cache_ram.h
```diff
@@ -0,0 +1,191 @@
...
+  shash::Any invalid_fd_;
```
jblomer: That can probably become a `static const`.  I think it would be better to use the term handle instead of fd, e.g. `kInvalidHandle`.

cvmfs/cache_ram.h
```diff
@@ -0,0 +1,191 @@
...
+  std::vector<ReadOnlyFd> open_fds_;
+  pthread_rwlock_t rwlock_;
+  kvstore::MemoryKvStore pinned_entries_;
```
jblomer: Would it be possible to use a single kvstore for all object types?  Let's discuss.

cvmfs/cache_ram.h
```diff
@@ -0,0 +1,191 @@
...
+  virtual int Dup(int fd);
+
+  /**
+   * No effect for in-memory caches
+   * @param id The hash key
+   * @returns -EBADFD if fd is not valid
+   */
+  virtual int Readahead(int fd);
+
+  /** Get the amount of space to be allocated for a call to @ref StartTxn */
```
jblomer: Is this still Doxygen syntax?  I'm not sure, if it is, it's good to know that one can do one-liners this way.

cvmfs/cache_ram.cc
```diff
@@ -0,0 +1,250 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+#include "cvmfs_config.h"
+#include "cache_ram.h"
+
+#include <errno.h>
+#include <assert.h>
```
jblomer: --> `<cassert>`

cvmfs/cache_ram.cc
```diff
@@ -0,0 +1,250 @@
...
+#include <new>
+#include <string.h>
```
jblomer: --> `<cstring>`

cvmfs/cache_ram.cc
```diff
@@ -0,0 +1,250 @@
...
+#include "smalloc.h"
+#include "util/posix.h"
+#include "util_concurrency.h"
+
+using namespace std;  // NOLINT
+
+namespace cache {
+
+
+int RamCacheManager::AddFd(const ReadOnlyFd &fd) {
```
jblomer: We should probably have an upper, enforced limit for the number of open file descriptors.

cvmfs/cvmfs.cc
```diff
@@ -72,7 +72,7 @@
 #include "authz/authz_session.h"
 #include "auto_umount.h"
 #include "backoff.h"
-#include "cache.h"
```
jblomer: Here it's probably ok to just include `cache.h`

cvmfs/kvstore.h
```diff
@@ -0,0 +1,126 @@
...
+namespace kvstore {
+
+struct MemoryBuffer {
+  void *address;
+  size_t size;
+  unsigned int refcount;
+  cache::CacheManager::ObjectType object_type;
+};
+
+class MemoryKvStore :SingleCopy {
```
jblomer: I guess for the time being, methods don't need to be virtual.

jblomer: There should be a comment for the class, explaining purpose and use.

cvmfs/kvstore.h
```diff
@@ -0,0 +1,126 @@
...
+   * @returns True iff there was an entry at id
+   */
+  virtual bool Delete(const shash::Any &id);
+
+  /**
+   * Delete the oldest entries until the KvStore uses less than the given size
+   * @param size The maximum size to make the KvStore
+   * @returns True iff the shrink succeeds
+   */
+  virtual bool Shrink(size_t size);
```
jblomer: Perhaps better `ShrinkTo` or `ShrinkUntil`

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,121 @@
+#include <errno.h>
+#include <string.h>
```
jblomer: --> `<cstring>`

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,121 @@
...
+#include <assert.h>
```
jblomer: --> `<cassert>`

cvmfs/kvstore.h
```diff
@@ -0,0 +1,126 @@
...
+   * @param id The hash key
+   * @returns True iff the entry is present
+   */
+  virtual bool PopBuffer(const shash::Any &id, MemoryBuffer *buf);
+
+  /**
+   * Get the total space used for data
+   */
+  virtual size_t GetUsed() { return used_bytes; }
+ protected:
```
jblomer: Can be `private`, I guess

cvmfs/kvstore.h
```diff
@@ -0,0 +1,126 @@
...
+  size_t used_bytes;
+  lru::LruCache<shash::Any, MemoryBuffer> Entries;
```
jblomer: --> `entries_`

cvmfs/kvstore.h
```diff
@@ -0,0 +1,126 @@
...
+   * @returns True iff the entry is present
+   */
+  virtual bool PopBuffer(const shash::Any &id, MemoryBuffer *buf);
+
+  /**
+   * Get the total space used for data
+   */
+  virtual size_t GetUsed() { return used_bytes; }
+ protected:
+  size_t used_bytes;
```
jblomer: --> `used_bytes_`

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,121 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
+
+bool MemoryKvStore::Unref(const shash::Any &id) {
+  WriteLockGuard guard(rwlock_);
+  MemoryBuffer mem;
+  if (Entries.Lookup(id, &mem) && mem.refcount > 0) {
```
jblomer: Are we sure we silently want to ignore double unrefs?  It probably is an indicator for a misuse, so we might want to assert, no?

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,121 @@
...
+  } else {
+    return false;
+  }
+}
+
+int64_t MemoryKvStore::Read(const shash::Any &id, void *buf, size_t size, off_t offset) {
+  MemoryBuffer mem;
+  ReadLockGuard guard(rwlock_);
+  if (Entries.Lookup(id, &mem)) {
+    uint64_t copy_size = min(mem.size - offset, size);
```
jblomer: We probably have to catch `offset > mem.size`

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,121 @@
...
+    memcpy(buf, static_cast<char *>(mem.address) + offset, copy_size);
```
jblomer: This might fail for `copy_size == 0`

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,121 @@
...
+    return -ENOENT;
+  }
+}
+
+bool MemoryKvStore::Commit(const shash::Any &id, const kvstore::MemoryBuffer &buf) {
+  WriteLockGuard guard(rwlock_);
+  bool existed = false;
+  MemoryBuffer mem;
+  if (Entries.Lookup(id, &mem)) {
+    Delete(id);
```
jblomer: That would overwrite the existing ref count, right?  I'm not sure if this is what we want.  We might commit something that is already in use and has refcount > 1.

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,121 @@
...
+  } else {
+    return false;
+  }
+}
+
+bool MemoryKvStore::Shrink(size_t size) {
+  WriteLockGuard guard(rwlock_);
+  shash::Any key;
+  MemoryBuffer buf;
+  while (used_bytes > size) {
```
jblomer: Wouldn't we only take objects with refcount 0?

cvmfs/kvstore.h
```diff
@@ -0,0 +1,126 @@
...
+   * @returns A reference count, or -ENOENT if the entry is absent
+   */
+  virtual int64_t GetRefcount(const shash::Any &id);
+
+  /**
+   * Increase the reference count on the entry at id
+   * @param id The hash key
+   * @returns True if the entry exists and was updated
+   */
+  virtual bool Ref(const shash::Any &id);
```
jblomer: Perhaps more clear as `IncRef`.

cvmfs/cache_ram.cc
```diff
@@ -0,0 +1,250 @@
...
+  bool sweep_tail = false;
+  {
+    ReadLockGuard guard(rwlock_);
+    if (!IsValid(fd)) return -EBADFD;
+    file_descriptor = open_fds_[fd];
+    assert(pinned_entries_.GetBuffer(file_descriptor.handle, &buf));
+  }
+
+  WriteLockGuard guard(rwlock_);
+  assert(pinned_entries_.Unref(file_descriptor.handle));
```
jblomer: `assert()` might be optimized away (it isn't in cvmfs, but in general it might).  So it's better to store the result of `Unref` in a variable and test the variable in assert.

cvmfs/cache_ram.cc
```diff
@@ -0,0 +1,250 @@
...
+  return 0;
+}
+
+
+int RamCacheManager::StartTxn(const shash::Any &id, uint64_t size, void *txn) {
+  Transaction *transaction = new (txn) Transaction();
+  transaction->id = id;
+  transaction->pos = 0;
+  transaction->expected_size = size;
+  //TODO realloc() on write for kSizeUnknown?
```
jblomer: Unfortunately yes, this can happen

cvmfs/cache_ram.cc
```diff
@@ -0,0 +1,250 @@
...
+{
+  Transaction *transaction = reinterpret_cast<Transaction *>(txn);
+  transaction->description = description;
+  transaction->object_type = type;
+}
+
+
+int64_t RamCacheManager::Write(const void *buf, uint64_t size, void *txn) {
+  Transaction *transaction = reinterpret_cast<Transaction *>(txn);
+  uint64_t copy_size= min(size, transaction->size - transaction->pos);
```
jblomer: If we new the size upfront, it is an error to write more than the expected size.

test/unittests/t_cache_ram.cc
```diff
@@ -0,0 +1,347 @@
...
+#include "cache_ram.h"
+#include "statistics.h"
+
+using namespace std;  // NOLINT
+using cache::RamCacheManager;
+
+static const unsigned cache_size = 1024;
+static const unsigned alloc_size = 16;
+
+
```
jblomer: I think it makes sense to add a text fixture to do the common test case initialization, e.g. the `RamCacheManager`.  The unit test needs to live in the same name space as the production code in order to make the gtest macro magic work.

test/unittests/t_kvstore.cc
```diff
@@ -0,0 +1,211 @@
...
+#include "hash.h"
+#include "cache.h"
+#include "kvstore.h"
+#include "statistics.h"
+
+using namespace std;  // NOLINT
+
+static const unsigned cache_size = 1024;
+static const size_t malloc_size = 16;
+
```
jblomer: I think it would be better with a test fixture.

jblomer:
Very well!  I'm not sure if we need separate kv stores, in particular the one storing open/pinned files.  I guess this could be done with a single kv store instance and reference counting.  Let's discuss.

bbockelm:
Is it possible to re-submit this request with just the refactoring of the cache / cache_posix layer?  I suspect the header improvements would be separately useful.

jblomer:
That's right, I think.  @trshaffer do you feel comfortable splitting it into two pull requests?

cvmfs/cache_ram.cc
bbockelm: Further, you probably want to keep a "free list" of known invalid handles to avoid a linear scan of the data structure when opening / closing files in the cache.

Something like this:

```
  if (free_fd_.empty()) {  // No holes in map; add new int.
    i = open_fds_.size();
    open_fds_.push_back(fd);
  } else {
    i = free_fd_.back();
    free_fd_.pop_back();
    open_fd_[i] = fd;
  }
  return i;
```

bbockelm: The above would make assignment of a FD to be constant time (well, plus a lock...) with the tradeoff of extra memory usage.

To bound the memory usage, one could bound the maximum size of the `free_fd_` list.  This would cause a sequential scan to refill it if it is empty but `open_fds_` has known vacancies.  You'd also need to track the occupancy of `open_fds_` - but that can also be done in constant time with simple counters.

At that point, it would be convenient to split it out to a separate "fd tracker" object, as I'm facing the exact same issues with my `librados`-based cache.  Once we can uplift the CVMFS code base to C++11, such an object would be a prime candidate for using lock-free, atomics-based methods.

jblomer: Hm, if we do not need to return the smallest free file descriptor, we can use a single array with the first part holding available fds and the second part the currently open ones.  Open and close would then translate into swap operations of elements in the two parts.

If we want the smallest available file descriptor on open, we can take a heap (priority queue) and have log(n) access time, which should be still good enough for practical purposes.

What exactly would you need from an fd tracker in other cache manager implementations?  Just something that manages taken and free numbers?

bbockelm:
Any news on progress toward splitting the PRs?

trshaffer:
@bbockelm the refactored headers just got merged into `devel`

trshaffer:
I think this PR is about in a good place for a merge. @jblomer I'll have a look at that sqlite allocator, and in the meantime, care to review the changes?

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
+ * additional privileges on the host such as filesystem access.
+ * This cache resides in a single process' memory, and does not
+ * support sharing or persistence. This cache backend is a good
+ * choice in highly restrictive environments, e.g. HPC resources,
+ * where it is not possible to make use of a shared/local filesystem.
+ *
+ * To use this cache backend, set @p CVMFS_CACHE_PRIMARY=ram. There
+ * are not many knobs to configure; at present only the size to be
+ * used for cached objects. @p CVMFS_CACHE_RAM_MB sets the cache size
+ * as an absolute value, while @p CVMFS_CACHE_RAM_PERCENT sets the
```
jblomer: What would happen if both are set?

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
+ * size as a percentage of the host memory size.
  */
 class RamCacheManager : public CacheManager {
  public:
+  struct Counters {
+    perf::Counter *n_acquire;
```
jblomer: Can be removed, I think

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
+    perf::Counter *n_getsize;
+    perf::Counter *n_close;
+    perf::Counter *n_pread;
+    perf::Counter *n_dup;
+    perf::Counter *n_readahead;
+    perf::Counter *n_sizeoftxn;
+    perf::Counter *n_starttxn;
```
jblomer: I think that's the same than `n_sizeoftxn` and `n_ctrltxn`, so probably one one of them should survive.

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
+    perf::Counter *n_ctrltxn;
+    perf::Counter *n_write;
+    perf::Counter *n_reset;
+    perf::Counter *n_openfromtxn;
```
jblomer: I think that's the same than `n_committxn` and can be removed.

cvmfs/cache_ram.h
trshaffer: I couldn't think of a reasonable ordering, so I just made MB take precedence. Should initialization fail instead?

jblomer: Good question... how about using one parameter: `CVMFS_CACHE_RAM_SIZE`? If it ends with a percent sign, it's interpreted accordingly.

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
+  virtual ~RamCacheManager() {
+    pthread_rwlock_destroy(&rwlock_);
+  }
   virtual bool AcquireQuotaManager(QuotaManager *quota_mgr);
 
+  /**
+   * Open a new file descriptor into the cache. Note that opening entries effectively pins them in the cache,
+   * so it may be necessary to close unneeded file descriptors to allow eviction to make room in the cache
+   * @param id The hash key
+   * @returns The smallest free integer file descriptor, or -ENOENT if the entry
```
jblomer: We might soften the requirement on the return value to "an integer between 0 and max number of open fds".

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
+#include "statistics.h"
 #include "util/pointer.h"
 
 namespace cache {
 
+// The null hash (hashed output is all null bytes) serves as a marker for
+// an invalid handle
+static const shash::Any kInvalidHandle;
+
+static const unsigned kMaxHandles = 8192;
```
jblomer: We probably would like to have this configurable, but let's move this to another PR.

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
+   */
   virtual int OpenFromTxn(void *txn);
+
+  /**
+   * Free the resources allocated for the pending transaction
+   * @param txn A pointer to space allocated for storing the transaction details
+   */
   virtual int AbortTxn(void *txn);
+
+   /**
```
jblomer: Whitespace: one too many at the start of the line

cvmfs/cache_ram.h
```diff
@@ -13,75 +13,275 @@
...
-    int64_t handle;
-    uint64_t size;
+    ReadOnlyFd() : handle(shash::Any()), pos(0), store(NULL) { }
+    ReadOnlyFd(const shash::Any &h, uint64_t pos)
+      : handle(h)
+      , pos(pos)
+      , store(NULL) { }
+    shash::Any handle;
+    uint64_t pos;
+    kvstore::MemoryKvStore *store;
```
jblomer: Do we need to store the `kvstore` in the `ReadOnlyFd`?  Could we just store it once in the `RamCacheManager` object?

jblomer: Ah, one for volatile files, one for regular files.  Never mind.

cvmfs/cache_ram.cc
```diff
@@ -20,68 +24,114 @@ namespace cache {
...
+    perf::Inc(counters_.n_enfile);
+    return -ENFILE;
+  }
 }
 
-
 bool RamCacheManager::AcquireQuotaManager(QuotaManager *quota_mgr) {
-  return false;
+  perf::Inc(counters_.n_acquire);
+  if (quota_mgr == NULL) {
```
jblomer: Here you can simply `assert(quota_mgr != NULL)`, I think

cvmfs/cache_ram.cc
```diff
@@ -20,68 +24,114 @@ namespace cache {
...
+  assert(rc);
+  open_fds_[fd].handle = kInvalidHandle;
+  LogCvmfs(kLogCache, kLogDebug, "closed fd %d", fd);
+  sweep_tail = (static_cast<unsigned>(fd) == (open_fds_.size() - 1));
 
   if (sweep_tail) {
-    WriteLockGuard guard(rwlock_);
     unsigned last_good_idx = open_fds_.size() - 1;
-    while (open_fds_[last_good_idx].handle < 0)
+    while (open_fds_[last_good_idx].handle != kInvalidHandle)
```
jblomer: This can be an endless loop, I think.

cvmfs/cache_ram.cc
```diff
@@ -92,22 +142,30 @@ int64_t RamCacheManager::Pread(
...
+  if (!IsValid(fd)) {
+    LogCvmfs(kLogCache, kLogDebug, "bad fd %d on Dup", fd);
+    return -EBADF;
+  }
+  assert(open_fds_[fd].store);
+  rc = open_fds_[fd].store->IncRef(open_fds_[fd].handle);
+  assert(rc);
+  LogCvmfs(kLogCache, kLogDebug, "dup fd %d", fd);
+  perf::Inc(counters_.n_dup);
+  return AddFd(open_fds_[fd]);
```
jblomer: I think you'd need to check the return value of `AddFd` and decrease the reference count if it is unsuccessful.

cvmfs/cache_ram.cc
```diff
@@ -116,21 +174,28 @@ int RamCacheManager::Dup(int fd) {
...
+  LogCvmfs(kLogCache, kLogDebug, "new transaction with id %s",
+           id.ToString().c_str());
   Transaction *transaction = new (txn) Transaction();
   transaction->id = id;
   transaction->pos = 0;
   transaction->expected_size = size;
   transaction->size = (size == kSizeUnknown) ? kPageSize : size;
   if (transaction->size) {
-    transaction->buffer = smmap(transaction->size);
+    transaction->buffer = scalloc(1, transaction->size);
```
jblomer: For larger objects, `smmap` might be a good choice because it can be freed more cleanly.  Is there a reason not to use it?

cvmfs/cache_ram.cc
```diff
@@ -141,59 +206,156 @@ void RamCacheManager::CtrlTxn(
...
+      transaction->buffer = realloc(transaction->buffer, transaction->size);
+      if (!transaction->buffer) {
+        LogCvmfs(kLogCache, kLogDebug, "realloc failed: %s", strerror(errno));
+        return -errno;
+      }
+    } else {
+      LogCvmfs(kLogCache, kLogDebug,
+               "attempted to write more than requested (%u>%u)",
+               size, transaction->size);
+      return -ENOSPC;
```
jblomer: I think that's more an `-EIO`, receiving more data than promised in the file catalog.

cvmfs/cache_ram.cc
```diff
@@ -141,59 +206,156 @@ void RamCacheManager::CtrlTxn(
...
+    }
+  }
+
+
+  uint64_t copy_size = min(size, transaction->size - transaction->pos);
```
jblomer: At this point, we already know that the remaining buffer is large enough to host `size` bytes, no?

cvmfs/cache_ram.cc
```diff
@@ -141,59 +206,156 @@ void RamCacheManager::CtrlTxn(
...
 
 int64_t RamCacheManager::CommitToKvStore(Transaction *transaction) {
-  if (transaction->handle >= 0)
-    return transaction->handle;
+  kvstore::MemoryBuffer buf;
+  buf.address = transaction->buffer;
+  kvstore::MemoryKvStore *store;
+  if (transaction->expected_size == kSizeUnknown) {
+    buf.size = transaction->pos;
+    buf.address = realloc(buf.address, buf.size);
```
jblomer: I'm not sure if it actually can happen... but theoretically `buf.size` might be zero, not sure if realloc can handle that.

cvmfs/cache_ram.cc
```diff
@@ -141,59 +206,156 @@ void RamCacheManager::CtrlTxn(
...
+  perf::Inc(counters_.n_committxn);
+  return CommitToKvStore(transaction);
 }
 
 
 int64_t RamCacheManager::CommitToKvStore(Transaction *transaction) {
-  if (transaction->handle >= 0)
-    return transaction->handle;
+  kvstore::MemoryBuffer buf;
+  buf.address = transaction->buffer;
```
jblomer: Oh, nice!  It's not copied anymore.

cvmfs/cache_ram.cc
jblomer: I see the reason further down.  Never mind.  It could be an uninitialized `smalloc` though, no?

cvmfs/cache_ram.cc
```diff
@@ -141,59 +206,156 @@ void RamCacheManager::CtrlTxn(
...
-  return -EIO;
+  Transaction *transaction = reinterpret_cast<Transaction *>(txn);
+
+  if (transaction->pos + size > transaction->size) {
+    if (transaction->expected_size == kSizeUnknown) {
+      perf::Inc(counters_.n_realloc);
+      transaction->size = max(2*transaction->size, size + transaction->pos);
+      LogCvmfs(kLogCache, kLogDebug, "reallocate transaction for %s to %u B",
+               transaction->id.ToString().c_str(), transaction->size);
+      transaction->buffer = realloc(transaction->buffer, transaction->size);
```
jblomer: I think there is an `srealloc` available in `smalloc.h`

cvmfs/cache_ram.cc
jblomer: Have a look for `srealloc`.

cvmfs/cache_ram.cc
```diff
@@ -141,59 +206,156 @@ void RamCacheManager::CtrlTxn(
...
+    buf.refcount = 0;
+  }
+
+  int64_t regular_size = regular_entries_.GetUsed();
+  int64_t volatile_size = volatile_entries_.GetUsed();
+  int64_t overrun = regular_size + volatile_size + buf.size - max_size_;
+
+  if (overrun > 0) {
+    perf::Inc(counters_.n_overrun);
+    volatile_entries_.ShrinkTo(max((int64_t) 0, volatile_size - overrun));
```
jblomer: I think shrinking should be done down to some percent of the overall cache size (if possible).  Otherwise we'd constantly shrink once the cache is full.

cvmfs/kvstore.h
```diff
@@ -0,0 +1,208 @@
...
+
+using namespace std;  // NOLINT
+
+namespace kvstore {
+
+struct MemoryBuffer {
+  void *address;
+  size_t size;
+  unsigned int refcount;
+  cache::CacheManager::ObjectType object_type;
```
jblomer: Since we use two `KvStores` for the two object types, do we still need to store it here?

jblomer:
General comment: we have a tendency to comment class methods where they are implemented as opposed to where they are created.  It can make the header more compact so that you'd have a good chance to see the class interface on a single screen page.  No need to change it in this PR though.

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,233 @@
...
+#include <string.h>
+
+#include <algorithm>
+
+#include "logging.h"
+#include "util_concurrency.h"
+
+using namespace std;  // NOLINT
+
+namespace kvstore {
```
jblomer: I think we should try to reduce the number of namespaces.  If it's not too big of a change, I think we should get rid of `namespace kvstore {`

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,233 @@
...
+int64_t MemoryKvStore::GetSize(const shash::Any &id) {
+  MemoryBuffer mem;
+  perf::Inc(counters_.n_getsize);
+  if (entries_.Lookup(id, &mem)) {
+    // LogCvmfs(kLogKvStore, kLogDebug, "%s is %u B", id.ToString().c_str(),
+    //          mem.size);
+    return mem.size;
+  } else {
+    LogCvmfs(kLogKvStore, kLogDebug,
+             "miss%s on GetSize",
```
jblomer: whitespace

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,233 @@
...
+int64_t MemoryKvStore::Read(
+  const shash::Any &id,
+  void *buf,
+  size_t size,
+  size_t offset
+) {
+  MemoryBuffer mem;
+  perf::Inc(counters_.n_read);
+  ReadLockGuard guard(rwlock_);
+  if (entries_.Lookup(id, &mem)) {
```
jblomer: It's probably easier to read to quit early on the error case.  The actual logic can then be un-indented.

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,233 @@
...
+) {
+  MemoryBuffer mem;
+  perf::Inc(counters_.n_commit);
+  WriteLockGuard guard(rwlock_);
+  LogCvmfs(kLogKvStore, kLogDebug, "commit %s", id.ToString().c_str());
+  if (entries_.Lookup(id, &mem)) {
+    LogCvmfs(kLogKvStore, kLogDebug, "commit overwrites existing entry");
+    used_bytes_ -= mem.size;
+    perf::Xadd(counters_.sz_deleted, mem.size);
+    counters_.sz_size->Set(used_bytes_);
```
jblomer: I would be enough to set the counter in line 166, I guess.

cvmfs/kvstore.cc
```diff
@@ -0,0 +1,233 @@
...
+
+bool MemoryKvStore::Delete(const shash::Any &id) {
+  perf::Inc(counters_.n_delete);
+  WriteLockGuard guard(rwlock_);
+  return DoDelete(id);
+}
+
+bool MemoryKvStore::DoDelete(const shash::Any &id) {
+  MemoryBuffer buf;
+  if (entries_.Lookup(id, &buf)) {
```
jblomer: Here you could also return early for the trivial case.

cvmfs/kvstore.h
```diff
@@ -0,0 +1,208 @@
...
+   */
+  size_t GetUsed() { return used_bytes_; }
+
+ private:
+  size_t used_bytes_;
+  unsigned int entry_count_;
+  unsigned int max_entries_;
+  lru::LruCache<shash::Any, MemoryBuffer> entries_;
+  pthread_rwlock_t rwlock_;
+  bool DoDelete(const shash::Any &id);
```
jblomer: Methods should be declared before member variables

cvmfs/lru.h
```diff
@@ -700,6 +720,66 @@ class LruCache : SingleCopy {
     return counters_;
   }
 
+  /**
+   * Prepares for in-order iteration of the cache entries to perform a filter operatoin.
```
jblomer: typo in "operatoin"

cvmfs/lru.h
```diff
@@ -700,6 +720,66 @@ class LruCache : SingleCopy {
...
+  virtual bool FilterNext() {
+    assert(filter_entry_);
+    filter_entry_ = filter_entry_->next;
+    return !filter_entry_->IsListHead();
+  }
+
+ /**
+  * Delete the current cache list entry
+  */
+  virtual void FilterDelete() {
```
jblomer: I'd think that `FilterDelete()` would/should rather perform like an "FilterDeleteAndNext".  Perhaps I misunderstand.  Let's discuss.

cvmfs/mountpoint.cc
```diff
@@ -175,7 +222,17 @@ bool FileSystem::CreateCache() {
...
-  uuid_cache_ = cvmfs::Uuid::Create(cache_dir_ + "/uuid");
+  switch (cache_mgr_type) {
+  case cache::kPosixCacheManager:
+    uuid_cache_ = cvmfs::Uuid::Create(cache_dir_ + "/uuid");
+    break;
+  case cache::kRamCacheManager:
+    uuid_cache_ = cvmfs::Uuid::Create("");
+    break;
+  case cache::kUnknownCacheManager:
+    uuid_cache_ = NULL;
```
jblomer: We can abort here, for an invalid cache we would have returned earlier.

cvmfs/mountpoint.cc
```diff
@@ -510,56 +567,67 @@ bool FileSystem::SetupNfsMaps() {
...
-    quota_mgr = PosixQuotaManager::Create(
-                  cache_dir_,
-                  quota_limit_,
-                  quota_threshold,
-                  found_previous_crash_);
-    if (quota_mgr == NULL) {
-      boot_error_ = "Failed to initialize lru cache";
-      boot_status_ = loader::kFailQuota;
-      return false;
+bool FileSystem::SetupQuotaMgmt(cache::CacheManagerIds mgr_type) {
```
jblomer: Probably easier to start with an `if {` block for the RAM cache manager with an early return.  The rest of the code can then stay unmodified.

cvmfs/mountpoint.h
```diff
@@ -219,7 +220,7 @@ class FileSystem : SingleCopy, public BootFactory {
   bool LockWorkspace();
   bool SetupCrashGuard();
   bool CreateCache();
-  bool SetupQuotaMgmt();
+  bool SetupQuotaMgmt(cache::CacheManagerIds mgr_type);
```
jblomer: I think we'd want to store the cache manager type as a class member variable of `MountPoint`.

cvmfs/lru.h
```diff
@@ -659,6 +661,24 @@ class LruCache : SingleCopy {
   }
 
   /**
+   * Removes and returns the least recently used cache entry
+   * @param key (out) address at which to write the removed key
+   * @param value (out) address at which to write the removed value
+   * @return true iff an entry was deleted
+   */
+  virtual bool PopOldest(Key *key, Value *value) {
```
jblomer: This method is unused, isn't it?

jblomer:
@trshaffer Very good!  I think we're getting there.  I'll add a few todo items for further pull requests to the JIRA ticket.

cvmfs/cache_ram.cc
trshaffer: You're right, `realloc` to size 0 is equivalent to `free`. The fix I made is correct as long as no one tries to use a buffer of size 0. Should we just make the commit fail for this case so there's no risk of misuse?

cvmfs/kvstore.h
trshaffer: There are more object types defined than kvstores, so I included the field to avoid dropping information. We would also need it for e.g. determining if a pinned entry has any actual references (refcount starts at 1 instead of 0). I don't really make use of it, so if the rest of the code doesn't care about object types, I could remove it.

cvmfs/kvstore.cc
trshaffer: The awkward layout here is so that it's safe to commit or open from transaction. I added some comments to clarify

cvmfs/mountpoint.h
trshaffer: It seems more natural as a member of `Filesystem`. The cache and quota setup is there

cvmfs/cache_ram.cc
jblomer: I think I'd only do the `realloc` and the check for `buf.address` if `buf.size > 0`.

cvmfs/kvstore.h
jblomer: You're right, I think we want to keep it... if only to allow for proper reporting of things in the cache.

cvmfs/mountpoint.h
jblomer: Ah, yes, correct.

PR #1637, opened 2016-07-28
Feature o1 fds

trshaffer:
I changed the routines for opening/closing file descriptors in the RAM cache to O(1) and switched to fixed allocation of fds. An index of used/free fds is necessary, and each `ReadOnlyFd` needs to know its position in the fd index, increasing the memory use by ~2 `int`s/fd. I was also able to remove some unneeded fields from `ReadOnlyFd`.

### DIFF
--- a/cvmfs/cache_ram.cc
+++ b/cvmfs/cache_ram.cc
@@ -52,38 +52,37 @@ int RamCacheManager::Open(const shash::Any &id) {
 
 int RamCacheManager::DoOpen(const shash::Any &id) {
   bool ok;
+  bool is_volatile;
   MemoryBuffer buf;
-  MemoryKvStore *store;
 
   if (regular_entries_.GetBuffer(id, &buf)) {
-    store = &regular_entries_;
+    is_volatile = false;
   } else if (volatile_entries_.GetBuffer(id, &buf)) {
-    store = &volatile_entries_;
+    is_volatile = true;
   } else {
     LogCvmfs(kLogCache, kLogDebug, "miss for %s",
              id.ToString().c_str());
     perf::Inc(counters_.n_openmiss);
     return -ENOENT;
   }
-
-  int fd = AddFd(ReadOnlyFd(id, 0));
+  int fd = AddFd(ReadOnlyFd(id));
   if (fd < 0) {
     LogCvmfs(kLogCache, kLogDebug, "error while opening %s: %s",
              id.ToString().c_str(), strerror(-fd));
     return fd;
   }
-  open_fds_[fd].store = store;
-  ok = store->IncRef(id);
-  assert(ok);
-  if (store == &regular_entries_) {
-    LogCvmfs(kLogCache, kLogDebug, "hit in regular entries for %s",
-             id.ToString().c_str());
-    perf::Inc(counters_.n_openregular);
-  } else {
+  open_fds_[fd].is_volatile = is_volatile;
+  if (is_volatile) {
     LogCvmfs(kLogCache, kLogDebug, "hit in volatile entries for %s",
              id.ToString().c_str());
     perf::Inc(counters_.n_openvolatile);
+  } else {
+    LogCvmfs(kLogCache, kLogDebug, "hit in regular entries for %s",
+             id.ToString().c_str());
+    perf::Inc(counters_.n_openregular);
   }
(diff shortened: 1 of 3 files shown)


cvmfs/cache_ram.h
```diff
@@ -237,14 +240,11 @@ class RamCacheManager : public CacheManager {
 
  private:
   struct ReadOnlyFd {
-    ReadOnlyFd() : handle(shash::Any()), pos(0), store(NULL) { }
-    ReadOnlyFd(const shash::Any &h, uint64_t pos)
-      : handle(h)
-      , pos(pos)
-      , store(NULL) { }
+    ReadOnlyFd() : handle(kInvalidHandle) { }
```
jblomer: The `is_volatile` and `index` fields should be initialized by the constructors, too

cvmfs/cache_ram.cc
```diff
@@ -96,38 +93,38 @@ int64_t RamCacheManager::GetSize(int fd) {
...
-           open_fds_[last_good_idx].handle == kInvalidHandle)
-      last_good_idx--;
-    open_fds_.resize(last_good_idx + 1);
-    LogCvmfs(kLogCache, kLogDebug, "resized fd vector to %u",
-             last_good_idx + 1);
-  }
 
+  size_t index = open_fds_[fd].index;
+  assert(index < fd_index_.size());
+  assert(fd_pivot_ < fd_index_.size());
```
jblomer: Couldn't `fd_pivot == fd_index_.size()` if all file descriptors are used?

cvmfs/cache_ram.h
```diff
@@ -269,12 +269,22 @@ class RamCacheManager : public CacheManager {
...
+  }
+
   int AddFd(const ReadOnlyFd &fd);
   int64_t CommitToKvStore(Transaction *transaction);
   virtual int DoOpen(const shash::Any &id);
 
   uint64_t max_size_;
+  size_t fd_pivot_;
   std::vector<ReadOnlyFd> open_fds_;
+  std::vector<int> fd_index_;
```
jblomer: Could this be a vector of `unsigned`?  It would remove the need for one or two `assert`s and clarify that contents can never be negative.

jblomer:
Very well!

I'd suggest a unit test that stresses opening and closing files a little.  For instance, it could first open as many file descriptors as possible and than randomly closing one and opening another one.  You can find a random number generator in `prng.h`, which you can use in the unit tests like

    Prng prng;
    prng.InitLocaltime();
    prng.Next(CVMFS_NFILES)

cvmfs/cache_ram.h
trshaffer: Those should always be overwritten, but I can add them anyway

cvmfs/cache_ram.cc
trshaffer: I'm using `fd_pivot_` to refer to the next free spot, so it's initialized to zero, and if it reaches `fd_index_.size()`, we've run off the end of the array. To be sure I'm not making any off-by-one errors, a unit test to hammer the fds sounds like a good idea

trshaffer:
@jblomer, I added a unit test to verify the fd limit and do a lot of random opens and closes, and fixed those other issues. Ready for merge?

test/unittests/t_cache_ram.cc
```diff
@@ -23,7 +24,7 @@ namespace cache {
 class T_RamCacheManager : public ::testing::Test {
  public:
   T_RamCacheManager()
-    : ramcache_(4*alloc_size, 3*cache_size, &statistics_) {
+    : ramcache_(4*alloc_size, cache_size, &statistics_) {
```
jblomer: I only realize now, sorry: is the initialization done in the constructor for a reason?  The rest uses `Setup` and `TearDown`, which will be executed for each individual test.  But perhaps that's the same for the constructor...?

trshaffer: Since LRU doesn't have default or copy constructors, I had to do the initialization just so. I strongly suspect that gtest uses a new instance for each test, since I'm not particularly careful about freeing resources between tests. The `OpenClose` test runs out of fds exactly when expected for a fresh instance, and if I comment out the close loop at the end, the test behavior is the same.

PR #1643, opened 2016-08-02
Add legacy option conversion to libcvmfs public interface

jblomer:


jblomer:
@trshaffer there are new functions `cvmfs_options_init_legacy()` and `cvmfs_options_clone_legacy()` which should do the conversion.

PR #1649, opened 2016-08-08
Feature: arena allocator

trshaffer:
I added an arena allocation mode to the CVMFS RamCache, in addition to the usual glibc `malloc()`. This allocator may help reduce fragmentation, and decrease memory use by directly returning unused pages to the kernel. This PR also includes some fixes to the benchmark scripts


trshaffer:
@jblomer are there any issues to be addressed before this is merged? I'd like to try out the allocator from #1658 with the RAM cache

PR #1650, opened 2016-08-09
Expose statistics counters from libcvmfs

trshaffer:
I'd like to be able to get statistics when using libcvmfs via Parrot. Per discussion with @jblomer, this PR adds a library function to get counters reported by `cvmfs_talk`


jblomer:
Looks good!  The only change I'd make is calling the function

    cvmfs_statistics_format

We might have more `cvmfs_statistics_...` in the future.

trshaffer:
Renamed

PR #1676, opened 2016-09-06
Feature heapcache

trshaffer:
I added another allocator for the RAM cache, and can run the ATLAS benchmark on it. The config key is

    CVMFS_CACHE_RAM_MALLOC=heap

It uses the log-structured allocator @jblomer recently added. Since the new allocator needs to have id headers and know about **all** pointers to allocations, making the transactions call into the `KvStore`'s allocator was proving troublesome. This PR makes a clean separation between the scratch memory for transactions (obtained via libc's `malloc`) and the `KvStore` allocated memory. Since transactions should either succeed or fail quickly, lingering allocations that other allocators address shouldn't be a problem. For a shared cache, this separation also fits, since it might not be possible to directly allocate memory through the `KvStore`.

### DIFF
--- a/cvmfs/kvstore.h
+++ b/cvmfs/kvstore.h
@@ -35,11 +35,6 @@ struct MemoryBuffer {
   shash::Any id;
 };
 
-class CacheCallback : public Callbackable<MallocHeap::BlockPtr> {
- public:
-  void OnBlockMove(const MallocHeap::BlockPtr &ptr);
-};
-
 /**
  * @p MemoryKvStore provides a simple RAM-backed key/value store suited to
  * use with @ref RamCacheManager. To insert entries, the caller must allocate
@@ -51,7 +46,7 @@ class CacheCallback : public Callbackable<MallocHeap::BlockPtr> {
  * can attempt to reduce its size by removing the least recently used
  * entries without any outstanding references.
  */
-class MemoryKvStore :SingleCopy {
+class MemoryKvStore :SingleCopy, public Callbackable<MallocHeap::BlockPtr> {
  public:
   enum MemoryAllocator {
     kMallocArena,
@@ -122,7 +117,6 @@ class MemoryKvStore :SingleCopy {
     , max_entries_(cache_entries)
     , entries_(cache_entries, shash::Any(), lru::hasher_any,
         statistics, name)
-    , cb_()
     , heap_(NULL)
     , counters_(statistics, name + ".lru") {
     int retval = pthread_rwlock_init(&rwlock_, NULL);
@@ -133,7 +127,7 @@ class MemoryKvStore :SingleCopy {
       break;
     case kMallocHeap:
       heap_ = new MallocHeap(alloc_size,
-          cb_.MakeCallback(&CacheCallback::OnBlockMove, &cb_));
+          this->MakeCallback(&MemoryKvStore::OnBlockMove, this));
       break;
     default:
       break;
@@ -260,7 +254,9 @@ class MemoryKvStore :SingleCopy {
   int DoMalloc(MemoryBuffer *buf);
   int DoRealloc(MemoryBuffer *buf, size_t size);
   void DoFree(MemoryBuffer *buf);
+  bool DoCommit(const MemoryBuffer &buf);
   size_t GetBufferSize(MemoryBuffer *buf);
+  void OnBlockMove(const MallocHeap::BlockPtr &ptr);
 
   MemoryAllocator allocator_;
   size_t used_bytes_;
@@ -268,7 +264,6 @@ class MemoryKvStore :SingleCopy {
   unsigned int entry_count_;
   unsigned int max_entries_;
   lru::LruCache<shash::Any, MemoryBuffer> entries_;
-  CacheCallback cb_;
   MallocHeap *heap_;
   pthread_rwlock_t rwlock_;
   Counters counters_;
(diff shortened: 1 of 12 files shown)


jblomer:
Cool, many thanks!  I'm on vacation for a few days, I'll have a look in the next few days.

cvmfs/mountpoint.cc
```diff
@@ -212,8 +212,14 @@ bool FileSystem::CreateCache() {
...
         alloc = MemoryKvStore::kMallocArena;
+      } else if (optarg == "heap") {
+        alloc = MemoryKvStore::kMallocHeap;
       }
     }
+    if (cache_bytes % 8 != 0) {
+      // Ensure 8-byte alignment for heap allocator
+      cache_bytes += 8 - (cache_bytes % 8);
+    }
     cache_bytes = max((uint64_t) 200*1024*1024, cache_bytes);
```
jblomer: Not part of this PR, but it would be better to define a constant like `kDefaultRamCacheSize`

cvmfs/mountpoint.cc
```diff
@@ -212,8 +212,14 @@ bool FileSystem::CreateCache() {
     if (options_mgr_->GetValue("CVMFS_CACHE_RAM_MALLOC", &optarg)) {
       if (optarg == "arena") {
         alloc = MemoryKvStore::kMallocArena;
+      } else if (optarg == "heap") {
+        alloc = MemoryKvStore::kMallocHeap;
       }
     }
+    if (cache_bytes % 8 != 0) {
+      // Ensure 8-byte alignment for heap allocator
+      cache_bytes += 8 - (cache_bytes % 8);
```
jblomer: We have `RoundUp8` functions for 32 bit and 64 bit arguments in malloc_arena.h and malloc_heap.h.  I'd suggest to move both of these function in smalloc.h and use them from there.

jblomer:
If the heap memory manager works well, we could remove support for the arena memory manager and system malloc.  What do you think?

cvmfs/malloc_heap.cc
```diff
@@ -107,12 +107,7 @@ MallocHeap::MallocHeap(uint64_t capacity, CallbackPtr callback_ptr)
   assert((capacity_ % 8) == 0);
   heap_ = reinterpret_cast<unsigned char *>(sxmmap(capacity + 7));
   uintptr_t head = uintptr_t(heap_) % 8;
-  if (head > 0)
```
jblomer: The code changes here I don't quite understand.  The original intent was to ensure 8 byte alignment.  It allocates 7 bytes more and then cuts little pieces at the beginning and the end.  Actually it's probably unnecessary because `mmap()` returns something aligned on a page boundary but I didn't find a garantuee for it...

cvmfs/kvstore.h
```diff
@@ -107,22 +111,41 @@ class MemoryKvStore :SingleCopy {
...
+    case kMallocHeap:
+      heap_ = new MallocHeap(alloc_size,
+          this->MakeCallback(&MemoryKvStore::OnBlockMove, this));
+      break;
+    default:
+      break;
     }
   }
 
   ~MemoryKvStore() {
```
jblomer: I think allocator and deallocator became large enough to put them in the .cc file

cvmfs/kvstore.h
```diff
@@ -107,22 +111,41 @@ class MemoryKvStore :SingleCopy {
...
     for (size_t i = 0; i < malloc_arenas_.size(); ++i) {
       delete malloc_arenas_[i];
     }
+    if (heap_) {
```
jblomer: The shouldn't be necessary as `heap_` is initialized to `NULL`

jblomer:
I think we should put a small header before the hash.  For a start, the header can be a single byte to store a version number.  This will simplify our life if we ever have to change the memory layout.

jblomer:
I think it would make sense to have helper functions for the pointer arithmetic that adds/substracts the first part with the key of a memory block. Especially when we extend it with a header and we'd need to write `ptr + 1 + sizefof(shash)`.

jblomer:
Perhaps I'm overlooking something but it looks to me that we still could use the key-value store for open transactions, too.  We could add a `Write` method to the store and set a flag in the header byte that indicates if a block is committed or in a transaction.  It would cost us more lookup calls during a transaction but we would save the memory copy on commit.

cvmfs/malloc_heap.cc
trshaffer: I was running into errors with the `unmap`, which requires page-aligned pointers. Leaving the extra space mapped does waste a little at the edges.

trshaffer:
I'd be inclined to leave glibc `malloc` in place for testing/debugging purposes, but I think the heap allocator could be the default

cvmfs/malloc_heap.cc
jblomer: I see, I think we can than simply run `sxmmap` and add an assert that the resulting pointer is 8 byte aligned.

trshaffer:
OK, I think that covers the revisions. I added a small struct for the allocation header (containing only the hash and a version number), that way the compiler can track the size and alignment. If you think helper functions would be clearer, I can add those too. `arena` alloc is gone, and `heap` is now the default. Would you prefer to make heap the exclusive allocator, @jblomer?

jblomer:
@trshaffer I think it's all right to keep system malloc and heap.

trshaffer:
Travis is passing, so is this ready to merge @jblomer?

jblomer:
Very good!  I think it's ready.  I'll have one more look tomorrow and then merge.

jblomer:
Superseded by #1685

PR #1685, opened 2016-09-15
In-memory key-value store based on heap cache manager

jblomer:
Superseds #1676


jblomer:
@trshaffer I added a couple of fixes and cleanups.  Please have another look.

jblomer:
@trshaffer: By the way, now that the staging of files moved to the `RamCacheManager`, do we still need the realloc logic in the `MemoryKvStore`?

trshaffer:
The main reason I didn't remove it already is that it feels weird to have `malloc` and `free` without `realloc`. Since no one's really using it and it's not generally better than a `malloc` + `free`, I can put in a separate PR to remove it

PR #1695, opened 2016-10-24
New integration tests for OverlayFS

radupopescu:
Implements integration tests for OverlayFS mentioned in ticket CVM-835.

### DIFF
--- a/test/src/700-overlayfsvalidation/main
+++ b/test/src/700-overlayfsvalidation/main
@@ -108,6 +108,50 @@ delete_dir_in_lower_branch () {
     return 0
 }
 
+# 4. Clear suid or sgid bits during writes on overlayfs
+clear_suid_sgid_during_write () {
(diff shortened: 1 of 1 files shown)


jblomer:
Thanks!

I think we need three more things

1. Explicitly excluding the test case for AUFS-only platforms (only SL6, I think) in the test/cloud_testing/platforms/*_test.sh scripts
2. Adding 7XX tests as server integration tests to the test/cloud_testing/platforms/*_test.sh scripts
3. It proved useful to clean up things in a (test-)central cleanup function that is invoked trough a bash `trap`.  The state that needs cleanup (PIDs, mount points, ...) can be stored in global variables of the from `CVMFS_TEST700_XXX`. There are examples in other integration tests.

radupopescu:
Made the change for Ubuntu 12.04.

Code review on Github is nicer than on Bitbucket. At a certain point I kept deleting PR by mistake on Bitbucket...

PR #1703, opened 2016-11-18
Add support for handling special files when using OverlayFS

Atlantic777:
This commit fixes the problem with publishing a transaction
when there are special files added to the repository
and OverlayFS is used as a UnionFS driver.

Whiteout files in OverlayFS are represented as character devices
with both major and minor numbers 0. OverlayFS driver doesn't allow
users to create such files.

Both block and character devices will be ignored with a warning message
printed out.

A new SyncItemType is introduced: kBlockDevice

SyncItem has 4 new methods:
 - GetMajor
 - GetMinor
 - IsBlockDevice
 - IsSpecialFile

Handling special files may change in future in the following ways:
 - fail by default, enable explicitly in config to ignore with warning
 - add special files to the repo

### DIFF
--- a/cvmfs/sync_item.h
+++ b/cvmfs/sync_item.h
@@ -70,12 +70,12 @@ class SyncItem {
     return IsCharacterDevice() || IsBlockDevice();
   }
 
-  inline unsigned int GetMajor()     const {
+  inline unsigned int GetRdevMajor()     const {
     assert(IsSpecialFile());
     StatUnion(true); return major(union_stat_.stat.st_rdev);
   }
 
-  inline unsigned int GetMinor()     const {
+  inline unsigned int GetRdevMinor()     const {
     assert(IsSpecialFile());
     StatUnion(true); return minor(union_stat_.stat.st_rdev);
   }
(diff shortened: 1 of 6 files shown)


cvmfs/sync_item.h
```diff
@@ -57,13 +58,28 @@ class SyncItem {
...
 
   inline bool IsWhiteout()        const { return whiteout_;                    }
   inline bool IsCatalogMarker()   const { return filename_ == ".cvmfscatalog"; }
   inline bool IsOpaqueDirectory() const { return IsDirectory() && opaque_;     }
 
+  inline bool IsSpecialFile()     const {
+    return IsCharacterDevice() || IsBlockDevice();
+  }
+
+  inline unsigned int GetMajor()     const {
```
jblomer: I think it would be better to call it `GetRdevMajor` and `GetRdevMinor` to avoid confusion which of the devices in the stat structure is meant.

cvmfs/sync_mediator.cc
```diff
@@ -75,6 +75,12 @@ void SyncMediator::Add(const SyncItem &entry) {
     return;  // Ignore markers.
   }
 
+  if (entry.IsSpecialFile() && !entry.IsWhiteout()) {
```
jblomer: Perhaps add a comment here that whiteout entries on OverlayFS are character device files

PR #1709, opened 2016-11-21
Add configuration variable to enable ignoring special files

Atlantic777:
This change preserves old behavior for working with special files (stop publishing by default).
The configuration variable CVMFS_IGNORE_SPECIAL_FILES is introduced.

When this variable is set to true, special files will be ignored with and a warning message will be printed for each of them.
Otherwise, on first encounter of a special file (which is not a whiteout) a warning message will also be printed, but publishing will stop.

### DIFF
--- a/cvmfs/sync_mediator.cc
+++ b/cvmfs/sync_mediator.cc
@@ -79,12 +79,12 @@ void SyncMediator::Add(const SyncItem &entry) {
   // and minor numbers equal to 0. Special files will be ignored except if they
   // are whiteout files.
   if (entry.IsSpecialFile() && !entry.IsWhiteout()) {
-    if (params_->ignore_special_files == true) {
+    if (params_->ignore_special_files) {
       PrintWarning("'" + entry.GetRelativePath() + "' "
                   "is a special file, ignoring.");
       return;
     } else {
-      PrintWarning("'" + entry.GetRelativePath() + "' "
+      PrintError("'" + entry.GetRelativePath() + "' "
                    "is a special file, stopping. "
                    "Enable CVMFS_IGNORE_SPECIAL_FILES in the cvmfs config.");
       abort();
--- a/cvmfs/sync_mediator.cc
+++ b/cvmfs/sync_mediator.cc
@@ -79,9 +79,16 @@ void SyncMediator::Add(const SyncItem &entry) {
   // and minor numbers equal to 0. Special files will be ignored except if they
   // are whiteout files.
   if (entry.IsSpecialFile() && !entry.IsWhiteout()) {
-    PrintWarning("'" + entry.GetRelativePath() + "' "
-                 "is a special file, ignoring.");
-    return;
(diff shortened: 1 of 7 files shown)


cvmfs/sync_mediator.cc
```diff
@@ -79,9 +79,16 @@ void SyncMediator::Add(const SyncItem &entry) {
...
   if (entry.IsSpecialFile() && !entry.IsWhiteout()) {
-    PrintWarning("'" + entry.GetRelativePath() + "' "
-                 "is a special file, ignoring.");
-    return;
+    if (params_->ignore_special_files == true) {
+      PrintWarning("'" + entry.GetRelativePath() + "' "
+                  "is a special file, ignoring.");
+      return;
+    } else {
+      PrintWarning("'" + entry.GetRelativePath() + "' "
```
jblomer: `PrintWarning` --> `PrintError`

cvmfs/sync_mediator.cc
```diff
@@ -79,9 +79,16 @@ void SyncMediator::Add(const SyncItem &entry) {
   // and minor numbers equal to 0. Special files will be ignored except if they
   // are whiteout files.
   if (entry.IsSpecialFile() && !entry.IsWhiteout()) {
-    PrintWarning("'" + entry.GetRelativePath() + "' "
-                 "is a special file, ignoring.");
-    return;
+    if (params_->ignore_special_files == true) {
```
jblomer: This can simply be `if (params_->ignore_special_files) {`

test/src/639-specialfiles/main
```diff
@@ -1,6 +1,7 @@
 cvmfs_test_name="Handling special files"
 
 export CVMFS_TEST_UNIONFS="overlayfs"
+export CVMFS_IGNORE_SPECIAL_FILES="true"
```
jblomer: This would better be in /etc/cvmfs/repositories.d/$CVMFS_TEST_REPO/server.conf.  Like

    echo "PARAM=value" | sudo tee -a /etc/cvmfs/...

We should also add a test that confirms that cvmfs aborts on special files if this parameter is not set

Atlantic777: I will try to refactor this test case a little bit and add more tests into it. I will do it in separate commits which can be added to this PR before it's merged.

test/src/639-specialfiles/main
```diff
@@ -1,12 +1,36 @@
 cvmfs_test_name="Handling special files"
 
-export CVMFS_TEST_UNIONFS="overlayfs"
-# export CVMFS_IGNORE_SPECIAL_FILES="true"
 test_repo_config="/etc/cvmfs/repositories.d/$CVMFS_TEST_REPO/server.conf"
 
+filename=""
```
jblomer: Sorry, one more thing I overlooked: the global variables in the integration test scripts should be named like

    CVMFS_TEST639_XYZ

this avoids that one test case mistakenly sets another test case's parameter

PR #1710, opened 2016-11-23
Refactor cvmfs server

radupopescu:
This branch contains an attempt to simplify the `cvmfs_server` script.

I'm putting this here to do review and keep track of the chages, but it's not ready to be merged.

The idea is to split the script (~6770 lines) into sub-scripts that each contain the functionality related to each command (transaction, abort, publish etc.). Common functionality of these scripts is also factored out into separate scripts. The main script should only do some minimal setup of the environment and source the other scripts. Subscripts are placed in the `cvmfs/server` subdirectory.

Current status:
* [x] Added `cvmfs/server/cvmfs_server_util.sh` and `cvmfs/server/cvmfs_server_common.sh` scripts, containing simple utility functions and, respectively, larger functions used by the different subcommands
* [x] Transaction, publish and abort subcommands have been factored out into the `cvmfs/server/cvmfs_server_{transaction,publish,abort}.sh` files.
* [x] Finish factoring out of remaining subcommands into separate scripts.
* [x] Deployment: the new multi-script program should be re-assembled into a single script for deployment, for convenience. A new `make_cvmfs_server` script can be added to assemble all the components for deployment
* [ ] Testing: must ensure that the new script behaves the same as the old one. There should be no regressions in the integration tests.
* [x] Update Cmake scripts to use the `make_cvmfs_server` script


jblomer:
Looks good to me!

I only wonder if we can have a single list of shell snippets.  Perhaps the make script can look into `cvmfs_server_devel` to collect the snippets, or there can be another file that provides the list for `cvmfs_server_devel` and the make script.

CVM-984 is partially solved by the PR. I think we can drop the part about `cvmfs_config` and simply resolve the ticket once this is merged.

radupopescu:
Re: the single list of includes: I would actually remove the `cvmfs_server_devel` script. I think it's no longer useful. `make_cmvfs_server.sh` is called automatically during the compilation of CVMFS.

I didn't think about the `cvmfs_config` script. It would be really nice if `cvmfs_server` and `cvmfs_config` used some shared code, but I would do this in another PR scheduled for CVMFS 2.4. I'll update CVM-984 and add a new one about `cvmfs_config`.

jblomer:
Sounds good.

PR #1724, opened 2016-12-07
Updated OverlayFS for RedHat 7 (CVM-835)

radupopescu:
This branch:
* [x] Updates the function `check_overlayfs_version` in `cvmfs_server`. The new success condition is either kernel 4.2.0 or newer on all Linux distributions, or, if running RedHat, kernel 3.10.0-493 or newer AND /var/spool/cvmfs should be located on an ext4 partition.
* [x] Updates the integration test scripts to run the server tests on RedHat (CentOS) 7, setting up `/var/spool/cvmfs` on an ext4 partition.

### DIFF
--- a/test/cloud_testing/platforms/slc6_x86_64_test.sh
+++ b/test/cloud_testing/platforms/slc6_x86_64_test.sh
@@ -73,10 +73,66 @@ CVMFS_TEST_CLASS_NAME=ServerIntegrationTests                                  \
                               -x src/518-hardlinkstresstest                   \
                                  src/524-corruptmanifestfailover              \
                                  src/585-xattrs                               \
+                                 src/700-overlayfsvalidation                  \
                                  --                                           \
                                  src/5*                                       \
                                  src/6*                                       \
+                                 src/7*                                       \
                               || retval=1
+
+
+echo -n "starting FakeS3 service... "
+s3_retval=0
+fakes3_pid=$(start_fakes3 $FAKE_S3_LOGFILE) || { s3_retval=1; retval=1; echo "fail"; }
+echo "done ($fakes3_pid)"
+
+if [ $s3_retval -eq 0 ]; then
+  echo "running CernVM-FS server test cases against FakeS3..."
+  CVMFS_TEST_S3_CONFIG=$FAKE_S3_CONFIG                                      \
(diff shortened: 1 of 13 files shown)


test/cloud_testing/platforms/slc6_x86_64_test.sh
```diff
@@ -73,12 +73,68 @@ CVMFS_TEST_CLASS_NAME=ServerIntegrationTests                                  \
...
                                  src/585-xattrs                               \
+                                 src/700-overlayfsvalidation                  \
                                  --                                           \
                                  src/5*                                       \
                                  src/6*                                       \
+                                 src/7*                                       \
                               || retval=1
 
 
+echo -n "starting FakeS3 service... "
```
jblomer: The S3 tests unfortunately need to be disabled for the time being.  They all fail, I think because of FakeS3.  But we didn't look into it yet.

radupopescu: Alright. I will go through all the test scripts and disable the S3 tests, to reduce the noise in the test output.

jblomer:
Let's check if the overlay-fs validation test works on Ubuntu before merging.  Otherwise we'd activate it only on Fedora and CentOS7.

radupopescu:
I think I may have made a mistake in the last commit. I need to double-check

radupopescu:
I reverted the previous commit. The test no longer hangs and I added more detailed status reporting.

It seems that the first test case, trying to delete files in the lower ro branch of the union fs is failing.

I will run the integration test on CentOS 7, too.

Jakob, could you please check the file `src/700-overlayfsvalidation/main`?

radupopescu:
I ran the test on CentOS 7. The same test case is failing as on Ubuntu.

radupopescu:
I've triggered another run of the integration tests (5** and 7**) in my CentOS 7 VM with Ext4 scratch. Will post here the results when they're done.

jblomer:
@radupopescu I guess this is ready, or do you want to add anything for xfs with fstype=1 to this PR?

radupopescu:
I'd like to try the OverlayFS tests on a volume with `ftype=1` and then we can close.

PR #1730, opened 2016-12-19
New `swissknife_lease` command to work with repository leases

radupopescu:
This PR implements the resolution to JIRA ticket [CVM-1130](https://sft.its.cern.ch/jira/browse/CVM-1130).

The new `swissknife_lease` command is used to acquire a new repository lease from a running instance of the CVMFS repository services and also to drop an existing repository lease.

When a new lease is obtained, the session token associated with the lease is written in the file `/var/spool/cvmfs/<REPO_NAME>/session_token`. This session token is used for all future requests associated with this lease.

### DIFF
--- a/cvmfs/swissknife_lease.cc
+++ b/cvmfs/swissknife_lease.cc
@@ -22,6 +22,17 @@ bool CheckParams(const swissknife::CommandLease::Parameters& p) {
 
 namespace swissknife {
 
+enum LeaseError {
+  LEASE_SUCCESS = 0,
+  LEASE_UNUSED = 1,
+  LEASE_PARAM_ERROR = 2,
+  LEASE_CURL_INIT_ERROR = 3,
+  LEASE_FILE_OPEN_ERROR = 4,
+  LEASE_FILE_DELETE_ERROR = 5,
+  LEASE_PARSE_ERROR = 6,
+  LEASE_CURL_REQ_ERROR = 7,
+};
+
 CommandLease::~CommandLease() {
 }
 
@@ -43,15 +54,15 @@ int CommandLease::Main(const ArgumentList& args) {
   params.lease_fqdn = *(args.find('p')->second);
 
   if (!CheckParams(params)) {
-    return 2;
+    return LEASE_PARAM_ERROR;
   }
 
   // Initialize curl
   if (curl_global_init(CURL_GLOBAL_ALL)) {
-    return 3;
+    return LEASE_CURL_INIT_ERROR;
   }
 
-  int ret = 0;
+  LeaseError ret = LEASE_SUCCESS;
   if (params.action == "acquire") {
     CurlBuffer buffer;
     if (MakeAcquireRequest(params.user_name, params.lease_fqdn, params.repo_service_url, buffer)) {
@@ -65,13 +76,13 @@ int CommandLease::Main(const ArgumentList& args) {
           token_file << session_token;
         } else {
           LogCvmfs(kLogCvmfs, kLogStderr, "Error opening file: %s", std::strerror(errno));
-          ret = 4;
+          ret = LEASE_FILE_OPEN_ERROR;
         }
       } else {
-        ret = 5;
+        ret = LEASE_PARSE_ERROR;
       }
     } else {
-      ret = 6;
+      ret = LEASE_CURL_REQ_ERROR;
     }
   } else if (params.action == "drop") {
     // Try to read session token from repository scratch directory
@@ -91,21 +102,21 @@ int CommandLease::Main(const ArgumentList& args) {
           success = true;
         } else {
           LogCvmfs(kLogCvmfs, kLogStderr, "Could not drop active lease");
-          ret = 7;
+          ret = LEASE_PARSE_ERROR;
         }
       } else {
         LogCvmfs(kLogCvmfs, kLogStderr, "Error making DELETE request");
-        ret = 8;
+        ret = LEASE_CURL_REQ_ERROR;
       }
     } else {
       LogCvmfs(kLogCvmfs, kLogStderr, "Error reading session token from file");
-      ret = 9;
+      ret = LEASE_FILE_OPEN_ERROR;
     }
 
     token_file.close();
     if (success && std::remove(token_file_name.c_str())) {
       LogCvmfs(kLogCvmfs, kLogStderr, "Error deleting session token file");
-      ret = 10;
+      ret = LEASE_FILE_DELETE_ERROR;
     }
   }
 
--- a/cvmfs/swissknife_lease.cc
+++ b/cvmfs/swissknife_lease.cc
@@ -23,26 +23,26 @@ bool CheckParams(const swissknife::CommandLease::Parameters& p) {
 namespace swissknife {
 
 enum LeaseError {
-  LEASE_SUCCESS = 0,
-  LEASE_UNUSED = 1,
-  LEASE_PARAM_ERROR = 2,
-  LEASE_CURL_INIT_ERROR = 3,
-  LEASE_FILE_OPEN_ERROR = 4,
+  LEASE_SUCCESS           = 0,
+  LEASE_UNUSED            = 1,
+  LEASE_PARAM_ERROR       = 2,
+  LEASE_CURL_INIT_ERROR   = 3,
+  LEASE_FILE_OPEN_ERROR   = 4,
   LEASE_FILE_DELETE_ERROR = 5,
-  LEASE_PARSE_ERROR = 6,
-  LEASE_CURL_REQ_ERROR = 7,
+  LEASE_PARSE_ERROR       = 6,
+  LEASE_CURL_REQ_ERROR    = 7,
 };
 
 CommandLease::~CommandLease() {
 }
 
 ParameterList CommandLease::GetParams() const {
-    ParameterList r;
-    r.push_back(Parameter::Mandatory('u', "repo service url"));
-    r.push_back(Parameter::Mandatory('a', "action (acquire or drop)"));
-    r.push_back(Parameter::Mandatory('n', "user name"));
-    r.push_back(Parameter::Mandatory('p', "lease path"));
-    return r;
+  ParameterList r;
+  r.push_back(Parameter::Mandatory('u', "repo service url"));
+  r.push_back(Parameter::Mandatory('a', "action (acquire or drop)"));
+  r.push_back(Parameter::Mandatory('n', "user name"));
+  r.push_back(Parameter::Mandatory('p', "lease path"));
+  return r;
 }
 
 int CommandLease::Main(const ArgumentList& args) {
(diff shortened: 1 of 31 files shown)


radupopescu:
This PR is safe to merge and it doesn't affect any existing functionality.

radupopescu:
This PR passed all checks and, since it affect any existing functionality, is completely safe to merge.

cvmfs/swissknife_lease.cc
```diff
@@ -0,0 +1,141 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include "swissknife_lease.h"
+
+#include <fstream>
```
jblomer: Is it a larger effort to get rid of the C++ file and string streams? Everywhere else the C I/O + cvmfs helper functions from util/string.cc and util/posix.cc are used.

cvmfs/swissknife_lease.cc
```diff
@@ -0,0 +1,141 @@
...
+
+  return true;
+}
+
+}  // namespace
+
+namespace swissknife {
+
+enum LeaseError {
+  LEASE_SUCCESS           = 0,
```
jblomer: The naming would rather be `kLeaseSuccess`, `kLeaseUnused`, ...

cvmfs/swissknife_lease.cc
```diff
@@ -0,0 +1,141 @@
...
+    CurlBuffer buffer;
+    if (MakeAcquireRequest(params.user_name,
+                           params.lease_fqdn,
+                           params.repo_service_url,
+                           &buffer)) {
+      std::string session_token;
+      if (ParseAcquireReply(buffer, &session_token)) {
+        // Save session token to /var/spool/cvmfs/<REPO_NAME>
+        // TODO(radu): Is there a special way to access the scratch directory?
+        std::string token_file_name = "/var/spool/cvmfs/" +
```
jblomer: "/var/spool/cvmfs" could be a `char *` / `string` constant

cvmfs/swissknife_lease.cc
```diff
@@ -0,0 +1,141 @@
...
+            params.lease_fqdn + "/session_token";
```
jblomer: I think we should be prepared to have multiple concurrent sessions on the same release manager machine, e.g. by adding a seesion uuid to the file name

cvmfs/swissknife_lease_json.h
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+#ifndef CVMFS_SWISSKNIFE_LEASE_JSON_H_
+#define CVMFS_SWISSKNIFE_LEASE_JSON_H_
+
+#include "swissknife_lease_curl.h"
+
+#include <string>
+
+bool ParseAcquireReply(const CurlBuffer& buffer, std::string* session_token);
```
jblomer: Maybe overkill maybe not: we might want to encapsulate parsing the JSON objects in a class.

cvmfs/swissknife_lease_curl.cc
```diff
@@ -0,0 +1,99 @@
...
+
+  return my_buffer->data.size();
+}
+
+CURL* PrepareCurl(const char* method) {
+  CURL* h_curl = curl_easy_init();
+
+  if (h_curl) {
+    curl_easy_setopt(h_curl, CURLOPT_NOPROGRESS, 1L);
+    curl_easy_setopt(h_curl, CURLOPT_USERAGENT, "curl/7.47.0");
```
jblomer: I think we should use cvmfs and the cvmfs version as user agent

cvmfs/swissknife_lease_curl.cc
```diff
@@ -0,0 +1,99 @@
...
+}
+
+bool MakeAcquireRequest(const std::string& user_name,
+                        const std::string& lease_fqdn,
+                        const std::string& repo_service_url,
+                        CurlBuffer* buffer) {
+  CURLcode ret = static_cast<CURLcode>(0);
+
+  CURL* h_curl = PrepareCurl("POST");
+  if (h_curl) {
```
jblomer: Probably easier to read with an early error return, like

    if (!h_curl)
      return false

    ...

cvmfs/swissknife_lease_curl.cc
```diff
@@ -0,0 +1,99 @@
...
+    // Prepare payload
+    std::string payload = "{\"user\" : \"" + user_name +
+        "\", \"path\" : \"" + lease_fqdn + "\"}";
+
+    // Make request to acquire lease from repo services
+    curl_easy_setopt(h_curl,
+                     CURLOPT_URL,
+                     (repo_service_url + "/api/leases").c_str());
```
jblomer: We should add versioning to the API URLs, e.g. /api/v1/leases

cvmfs/swissknife_lease_curl.h
```diff
@@ -0,0 +1,29 @@
...
+
+struct CurlBuffer {
+  std::string data;
+};
+
+size_t RecvCB(void *buffer, size_t size, size_t nmemb, void *userp);
+
+CURL* PrepareCurl(const char* method);
+
+bool MakeAcquireRequest(const std::string& user_name,
```
jblomer: We might want to have a class around it.  Perhaps it can be a single class with the functions of this compilation unit plus the JSON parsing.

cvmfs/swissknife_lease_json.cc
```diff
@@ -0,0 +1,94 @@
...
+
+#include "json.h"
+#include "json_document.h"
+#include "util/pointer.h"
+
+#include "logging.h"
+
+bool ParseAcquireReply(const CurlBuffer& buffer, std::string* session_token) {
+  UniquePtr<JsonDocument> reply(JsonDocument::Create(buffer.data));
+  if (reply->IsValid()) {
```
jblomer: Probably better to use an early failure return.

cvmfs/swissknife_lease_curl.cc
```diff
@@ -0,0 +1,99 @@
...
+  return false;
+}
+
+bool MakeDeleteRequest(const std::string& session_token,
+                       const std::string& repo_service_url,
+                       CurlBuffer* buffer) {
+  CURLcode ret = static_cast<CURLcode>(0);
+
+  CURL* h_curl = PrepareCurl("DELETE");
+  if (h_curl) {
```
jblomer: Probably better to use an early failure return.

cvmfs/swissknife_lease_json.cc
```diff
@@ -0,0 +1,94 @@
...
+    JSON* result = JsonDocument::SearchInObject(reply->root(),
+                                                "status",
+                                                JSON_STRING);
+    if (result != NULL) {
+      std::string status = result->string_value;
+      if (status == "ok") {
```
jblomer: I think we should use `string` / `char *` constants for the possible return codes.

cvmfs/swissknife_lease_json.cc
```diff
@@ -0,0 +1,94 @@
...
+  return false;
+}
+
+bool ParseDropReply(const CurlBuffer& buffer) {
+  UniquePtr<JsonDocument> reply(JsonDocument::Create(buffer.data));
+  if (reply->IsValid()) {
+    JSON* result = JsonDocument::SearchInObject(reply->root(),
+                                                "status",
+                                                JSON_STRING);
+    if (result != NULL) {
```
jblomer: Probably better to use an early failure return.

radupopescu:
Thanks for the feedback!
Making comments inline and putting changes in a new PR.

cvmfs/swissknife_lease_json.cc
radupopescu: Yes, ok.

cvmfs/swissknife_lease_json.cc
radupopescu: ok.

cvmfs/swissknife_lease_json.cc
radupopescu: ok.

cvmfs/swissknife_lease_curl.cc
radupopescu: ok.

cvmfs/swissknife_lease_curl.cc
radupopescu: Makes sense. I'll put this in a separate PR in sync with one in the services repo.

cvmfs/swissknife_lease_curl.cc
radupopescu: ok.

cvmfs/swissknife_lease_curl.cc
radupopescu: ok.

cvmfs/swissknife_lease.cc
radupopescu: ok.

cvmfs/swissknife_lease.cc
radupopescu: I can check {string,posix}.cc for which functions to use, but what was the problem is C++ std streams?

cvmfs/swissknife_lease.cc
radupopescu: ok.

cvmfs/swissknife_lease_json.h
radupopescu: I don't know about this one. I don't see the point of making classes for everything, like in Java.
These functions already encapsulate the needed functionality and there would be no state to maintain in the instance of the class. The same applies to the `swissknife_lease_curl.{h,cc}` files, too.

I can put these functions in namespaces, if you think it brings more clarity...

Ok, for the JSON unit tests.

jblomer: Ok, makes sense.  I don't think we should add more namespaces, I have the feeling they are rather adding to the confusion...

cvmfs/swissknife_lease.cc
radupopescu: I think it's a good idea to use, from the beginning, a file name which allows multiple concurrent sessions. The problem with a `UUID` or a random string is that it's not possible to recognize which token file belongs to you when it's time to make the `DROP` request.

I suggest appending the `"_<path>"` suffix, instead, where `<path>` is a subpath in the repo. There should never be two locks on the same subpath.

cvmfs/swissknife_lease.cc
jblomer: I can't really pinpoint a particular problem.  The Google style guide has a couple of cons (https://google.github.io/styleguide/cppguide.html#Streams).  Not using streams mostly means one less thing we need to know and care about.

cvmfs/swissknife_lease.cc
jblomer: Sounds good.  There are already Base64 utility functions in cvmfs if you're looking for escaping.

PR #1731, opened 2016-12-21
Add CVMFS_CONFIG_REPO_REQUIRED option

DrDaveD:
See [CVM-1111](https://sft.its.cern.ch/jira/browse/CVM-1111).

This is for the devel branch.  Next I will find out if it merges cleanly into the cvmfs-2.3 branch, and if necessary make a separate pull request; if not necessary I will note it in the comments.

Is there a place in readthedocs for documentation changes before the corresponding code is released?


DrDaveD:
I verified that this also can be merged into the cvmfs-2.3 branch.  I tried it with "git cherry-pick c3efd00 b325a0e" in a new branch based on cvmfs-2.3, built it, and verified that it passed test 050-configrepo.

test/src/050-configrepo/main
```diff
@@ -89,12 +98,17 @@ cvmfs_run_test() {
 
   CVMFS_TEST_050_CONF_EDIT=1
 
+  # because of the different CVMFS_MOUNT_DIR, this cvmfs_mount fails,
+  #  but call it anyway to set up the config
   cvmfs_mount grid.cern.ch
-  ls /cvmfs/grid.cern.ch || return 21
+  mkdir grid.cern.ch
+  mount -t cvmfs grid.cern.ch $PWD/grid.cern.ch || return 21
```
jblomer: This must run as sudo doesn't it?

jblomer: There should be a trap that ensures unmounting if any of the following checks fail

cvmfs/options.cc
```diff
@@ -171,8 +171,17 @@ void BashOptionsManager::ParsePath(const string &config_file,
   }
   if (!fconfig) {
     if (external && !DirectoryExists(config_path)) {
+      string repo_required;
+      if (GetValue("CVMFS_CONFIG_REPO_REQUIRED", &repo_required) &&
+        IsOn(repo_required)) {
+          LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogErr,
+               "required configuration repository directory does not exist: %s",
+               config_path.c_str());
+          exit(1);
```
jblomer: For consistency with other code paths, this should probably rather be `abort()`

test/src/050-configrepo/main
DrDaveD: You're right.  I'm sorry, that must have been at the beginning of my testing when I was forgetting that the tests were supposed to be run as an unprivileged user.

DrDaveD: good point

DrDaveD:
I updated the pr based on Jakob's comments, please take another look.

jblomer:
I was just about to merge it but the integration test fails with error 21.  Does it run through on your machine?

I noticed something else: the `CVMFS_CONFIG_REPO_REQUIRED` string should be added to the list of switches at the beginning of cvmfs_config.

DrDaveD:
Yes it works for me on my el5 test system, even after merging with latest devel branch.  I am building now for el6 to see if it works there.  Meanwhile, can you tell me what the error messages say?
I set CVMFS_TEST_USER and CVMFS_TEST_PROXY.   

I will push your requested change a little later today.

jblomer:
I says "Invalid repository: /tmp/cvmfs-test/workdir/050-configrepo/grid.cern.ch" on CentOS 7.  Somehow it must take the "device name" grid.cern.ch and create a full path from it from the existing sub path.

DrDaveD:
Yes you're right, I was able to reproduce that error on SL7.    I worked around it by using a name that was different than the repository name.

I did an strace on the mount command when using the repository name, and the full path is passed to the mount.cvmfs command, so it must be a bug in mount itself.

PR #1732, opened 2016-12-21
Pr #1731 to merge against cvmfs-2.3

DrDaveD:
This is exactly the same as pr #1731 except based on cvmfs-2.3.  The code changes are the same. I don't know if you prefer a separate pull request, but I thought it would at least be good to submit for the travis check.


radupopescu:
Hi!

The CI failure on MacOS is due to Travis removing old OSX images. We ask for 7.1 (internal Travis version number) and we get routed to 7.3.1, where we don't need to install `cmake`. To fix this issue, the `.travis.yml` file in the `devel` branch needs to be backported into the `2.3` branch.

This is just a quick fix, however. In January I've planned some work to ensure that we are again able to build on the latest version of MacOS - we are now building on OSX Yosemite 10.10, due to the removal of OpenSSL headers in 10.11 and 10.12. One this is done, this should make our Travis jobs and the entire MacOS build and packaging process more reliable.

DrDaveD:
The travis failures are on OSX only and do not appear to be related to this change.

PR #1742, opened 2016-12-30
Add .cvmfs_last_gc timestamp file (CVM-1107)

DrDaveD:


jblomer:
Can we make this information part of the repositories.json file?  I think it would be nice to have only one URL for status information.

DrDaveD:
It's possible, but then we'd be setting a new precedent and would need to move at least .cvmfs_last_snapshot information also into repositories.json.   Writing to repositories.json would have to be made more robust, setting a lock and reading prior information at the end of every successful snapshot or gc, as opposed to generating it from scratch at every repo creation or removal.

jblomer:
I think it would be good to maintain the last_snapshot information also in the repositories.json, both writings protected with file locking.

DrDaveD:
What about the case where the repositories are owned by non-root users?  Then /srv/cvmfs/info would have to be writable by those users.  You're OK with that new requirement?  Typically on a stratum 1 they would all be one user, but on a stratum 0 there could definitely be multiple owners that could run a gc for their repository.  /srv/cvmfs/info would have to be world writable, or it would have to be done through cvmfs_suid_helper.

What if the upstream storage is not a local disk?  How would repositories.json be locked?  I suppose the lock could go in /var/spool/cvmfs, but how would we handle permissions so that all owners on a stratum 0 could lock and write to this file?

This looks to me like a can of worms, probably more than I would be willing to take on personally.  It would be too much for anyone for 2.3.3, I think.

DrDaveD:
On the other hand if we added a per-repo json file to keep all such status information it would be a lot easier.  There's already code to prevent more than one publish or gc from happening on a given repo at a time, and one of my other pending pull requests prevents snapshots and gcs from happening at the same time.  No extra locking would need to be done, and no permissions complications.

jblomer:
I like the idea of a per-repo and stratum 1 JSON file.  We could then easily extend this file with more information in the future.

DrDaveD:
I can probably implement that next week.  This week I should write a CHEP paper.

DrDaveD:
Now it is creating .cvmfs_status.json instead of .cvmfs_last_gc.  The title of the pull request is now incorrect but I don't see a way to change it.

PR #1744, opened 2016-12-31
Fix builds with the Ninja generator

bbockelm:
[Ninja](https://ninja-build.org/) is an optional CMake generator; ninja focuses on speed and parallelism.  For example, calling `ninja-build` on a fully-built directory (meaning there is no actual buildign to do) returns in 50ms.

This commit allows CVMFS to be built using Ninja with the following changes:

- Ninja requires all dependencies to be declared a-priori.  The libtbb target previously declared it output a single library but had several others created as a side-effect.  `cvmfs_swissknife` dependency on those other libraries, but Ninja didn't know how to create them.  Seems to be stricter than GNU make in this respect.
- Ninja creates output directories that are not already present.  This breaks the TBB build as the output directory is created as a symlink during the libtbb target.
- cvmfs_server dependencies aren't exposed to CMakeLists; this explicitly declares the dependency so the cvmfs_server script doesn't have to be regenerated each time.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -107,6 +107,43 @@ set (CVMFS_CLIENT_SOURCES
   xattr.cc xattr.h
 )
 
+set (CVMFS_SERVER_SOURCES
+    server/cvmfs_server_sys.sh
+    server/cvmfs_server_prelude.sh
+    server/cvmfs_server_util.sh
+    server/cvmfs_server_ssl.sh
+    server/cvmfs_server_apache.sh
+    server/cvmfs_server_json.sh
+    server/cvmfs_server_common.sh
+    server/cvmfs_server_health_check.sh
+    server/cvmfs_server_compat.sh
+    server/cvmfs_server_transaction.sh
+    server/cvmfs_server_abort.sh
+    server/cvmfs_server_publish.sh
+    server/cvmfs_server_import.sh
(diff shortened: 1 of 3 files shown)


bbockelm:
@jblomer - now that things are starting to roll in the new year, can you take a look at this one?  Should be fairly simple / harmless.

jblomer:
@bbockelm: I'm still on vacation.  I'll have a look next week.

bbockelm:
Ah - sorry, I thought you were back last week given some other activity.  Enjoy your vacation!

jblomer:
Cool, many thanks!  

I'd like to merge it after having merged the changes from CVM-1040 to the documentation repository.

radupopescu:
Great!

I'm also making PR with the same changes for the `devel` branch. This commit can't be cherry-picked due to some refactoring that has taken place on `devel`.

bbockelm:
@radupopescu - I don't understand your comment.  This PR is against the `devel` repo, no?

radupopescu:
Hello @bbockelm!

There is a mixup somewhere. My comment wasn't meant for the PR. I'm working on some other issue that affects the stable and devel branches and I must have commented on the wrong thread.

I'm sorry for the noise and confusion.

radupopescu:
I think we can close this, as it is superseded by #1780. Compatibility with the Ninja build system is a fortunate side effect of the changes made in that branch.

PR #1745, opened 2017-01-03
Avoid expanding command line repo wildcards too early (CVM-1151)

DrDaveD:


DrDaveD:
Even though this is a 2.4 pr and these OSX travis failures seemed to happen only for 2.3 prs, this looks like the same kind of failure to me.

PR #1751, opened 2017-01-13
Fix linking error of c-ares on CentOS 7.3

radupopescu:
This addresses the linking error encoutered when building the `c-ares` dependency on CentOS 7.3 (CVM-1153). The error appears when linking some `c-ares` internal executables (`ahost`, `acountry` etc.) and causes the entire CVMFS process to fail.

The problem was not present on CentOS 7.2, but it stems from the fact that CVMFS passes some additional C and C++ compiler flags when building its external dependencies. I suspect most of these are needed to ensure consistency of the resulting CVMFS binary. The flag `-fvisibility=hidden` is an exception, since one cannot impose this on an external library. If the library does not have visibility annotations for the symbols of its API, this flag will make it impossible to use the library as a shared object.

This PR is for the CVMFS 2.3 branch, I'm preparing a separate PR for the `devel` branch, since the changes cannot just be cherry-picked, due to some refactoring that happened in the meantime.

### DIFF
--- a/externals/c-ares/src/configureHook.sh
+++ b/externals/c-ares/src/configureHook.sh
@@ -4,4 +4,4 @@
 #              Please install XCode and run `xcode-select --install` afterwards.
 #              Also remember to start XCode at least once and agree to the EULA.
 
-sh configure LDFLAGS="$LDFLAGS -rdynamic" CPPFLAGS="$CPPFLAGS -D_FILE_OFFSET_BITS=64" CFLAGS="$CFLAGS -fno-strict-aliasing -fasynchronous-unwind-tables -fno-omit-frame-pointer -fno-optimize-sibling-calls -fPIC"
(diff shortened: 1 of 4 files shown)


radupopescu:
The PR has been updated s.t. the "-fvisibility=hidden" flag is only removed for c-ares. All other external dependencies are untouched.

radupopescu:
Just as a reminder: the Travis-CI failure on this PR is due to the fact that the OS X image we selected (7.1) no longer exists.

I just checked the .travis.yml configuration on the `cvmfs-2.3` and the `devel` branches. On `cvmfs-2.3` we have selected the OS X image `7.1` which doesn't exist any more, while on the `devel` branch we use `6.4` (!) which will be retired this Friday.

At that point we will no longer have any way to compile for Mac on Travis, until we fix the OpenSSL issue.

jblomer:
We have to do it anyway, so let's start to put effort into it now.

radupopescu:
I can assign higher priority to that issue, but I don't think it's not a blocker for this PR, nor #1752 .

radupopescu:
@jblomer I reverted to using "-fvisibility=hidden" for c-ares, too. The build scripts now disable the shared library (libcares.so).

Building only the c-ares static library, I have seen no problems on CentOS 7.3, Ubuntu.

PR #1752, opened 2017-01-16
Fix compilation of c-ares for the devel branch

radupopescu:
This PR implements the fixes from #1751, this time for the `devel` branch.


jblomer:
Can you try if it's enough to run `make libcares.la` in the `makeHook.sh`?  This should only build the library without the demo applications.  Perhaps this is enough and we don't need to change symbol visibility.  Although c-ares is not a common library, we might still see linking surprises which I'd like to avoid at least in a bugfix release.

radupopescu:
This branch now includes the same changes as in #1751:
"-fvisibility=hidden" is enabled again for c-ares, but only the static library is built and used.

PR #1754, opened 2017-01-18
Disable/adjust ASGC stratum 1

jblomer:


DrDaveD:
Did you get word from Taiwan that they no longer want to provide a public stratum 1 for any of the big 3 domains?

PR #1763, opened 2017-01-23
Use LibreSSL on macOS (CVM-1112)

radupopescu:
This PR addresses ticket CVM-1112, making it possible to build CVMFS on macOS 10.11 and newer, which no longer ship with public OpenSSL library and headers.

Changes:
* On macOS (only), LibreSSL is used. It's compiled and linked at build time, like the other dependencies.
* We are only using the LibreSSL static libraries (`libssl.a`, `libcrypto.a`)
* `libcurl` is also linked with LibreSSL, on macOS

The modifications introduced here are not activated on Linux; there we always use the system version of OpenSSL.

Tested on macOS 10.10, 10.12 and on Linux (Ubuntu 16.04).


externals/libcurl/src/configureHook.sh
```diff
@@ -6,6 +6,15 @@ cd $cares_location
 sh makeHook.sh
 cd $cdir
 
+curl_ssl_config="--with-ssl"
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
jblomer: The `[[ ... ]]` is not a "bashism", is it?  In other places, we used

    case $(uname) in
      Darwin )
      ...
      ;;
    esac

but I'm not sure if it makes a difference

externals/ssl/src/CVMFS-CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+include ("${EXTERNALS_LIB_LOCATION}/CMake-Register_External_Lib.txt")
+
+set (SSL_ARCHIVE "${SSL_BUILTIN_LOCATION}/ssl_install/lib/libssl.a;${SSL_BUILTIN_LOCATION}/ssl_install/lib/libcrypto.a")
```
jblomer: I think in cvmfs/CMakeLists.txt we'd need to add libssl as a dependency on OS X everywhere where we need `OPENSSL_LIBRARIES`.  We'd also need to add `SSL_ARCHIVE` where we need `OPENSSL_LIBRARIES`.

radupopescu: I think I covered this in the root `CMakeLists.txt` (line 414). There is a conditional based on platform. On macOS, I set `OPENSSL_LIBRARIES` to the value of `SSL_ARCHIVE`.

jblomer: Oh, I see.  Neat trick.

The dependency is still needed, I think, for parallel builds, to make sure that the libssl.a is available before it's time to link it.

radupopescu: This is not an issue. On macOS, `libcurl` depends on `ssl` and builds it at the beginning of its own configure step, like it does with `c-ares`. Check `externals/libcurl/src/configureHook.sh`. Basically, if the configuration step has completed successfully, `libssl.a` and `libcrypto.a` are already built. This holds also for parallel builds.

jblomer: Right.

bootstrap.sh
```diff
@@ -107,7 +107,7 @@ replace_in_external() {
 
 mkdir -p $externals_build_dir
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
+if [[ x"$(uname)" == x"Darwin" ]]; then
```
jblomer: I think the `[[ ]]` is a bashism, i.e. might fail under dash.

externals/ssl/src/CVMFS-CMakeLists.txt
radupopescu: I think the dependency handling in our CMake scripts can be simplified and made more standard and I would schedule this work for 2.4.

* I don't think there is a good reason to have some dependencies compiled at configure-time and others at build-time. The `bootstrap.sh` script should do everything - configuring and building all the deps, in a specific order.
* The dependencies could all be installed to a subdirectory in the project (e.g. `externals_install`) that is ignored by Git. There are two main benefits to this:
  1. This prefix could then be passed to CMake at once and the standard `find_package` command can be used for picking up each dependency. No need to define new CMake variables like ****_ARCHIVE. 
  2. The contents of `externals_install` can be archived and reused (on the same platform). This would allow you to make different types of builds for CVMFS, during development, without always recompiling the build depends.

I think this could make things simpler and remove a lot of custom boilerplate.

PR #1789, opened 2017-01-31
Chirp cache plugin

trshaffer:
This is a first draft of a CVMFS cache plugin that uses a [Chirp](http://ccl.cse.nd.edu/software/chirp) server as a storage backend. It's a straightforward implementation that mostly just forwards everything to the Chirp server. There's still plenty of room to improve performance, especially by minimizing the number of remote I/O operations. I added a new configure flag so that this plugin (which depends on CCTools) isn't built by default. Running CMake with `-DBUILD_CCTOOLS=yes` will enable this feature. Once the configuration syntax is finalized, this plugin might make a good lower-tier cache to share data among several workers without stressing the Squids. Also note that this is written in the CCTools way (C, tabbed indentation, etc.) to the style might not be suitable for CVMFS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,6 +193,7 @@ option (BUILD_UNITTESTS_DEBUG   "Build the CernVM-FS unit test set with verbose 
 option (BUILD_UBENCHMARKS       "Build the CernVM-FS micro benchmarks"                             OFF)
 option (BUILD_DOCUMENTATION     "Build the CerVM-FS documentation using Doxygen"                   OFF)
 option (BUILD_COVERAGE          "Compile to collect code coverage reports"                         OFF)
+option (BUILD_CCTOOLS           "Build CCTools integration"                                        OFF)
 option (BUILD_ALL               "Build client, server, lib, preload, octopus, unit tests"          OFF)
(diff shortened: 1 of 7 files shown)


radupopescu:
Thanks @trshaffer for this! It's great to have a new cache plugin implementation.
I wonder if this isn't a good opportunity to set up a new CVMFS contrib repository in the CVMFS GitHub organization.

Ideally, things like cache or authorization plugins could go there and could use CVMFS through the standard CMake `find_package` system. That would also ensure that the artifacts related to the cache plugin system are installed properly.

I could create this new repo tomorrow and help out with the CMake stuff, if needed.

Cheers!

trshaffer:
That sounds like a great idea, since I felt a little odd adding my dependencies into the main build. I'd be happy to move this into a contrib repo

radupopescu:
Great!

To avoid losing any momentum, I went ahead and created the repo: https://github.com/cvmfs/cvmfs_contrib. The repos is quite bare right now, so feel free to create some directory structure (e.g. `plugins/cache`). In any case, we can improve things as we go.

When you have time, you can make the PR there. Don't forget to add yourself to `CONTRIBUTORS.md`.

Cheers,

jblomer:
Closing as this is a contribution for the cvmfs_contrib space.

PR #1791, opened 2017-02-06
Fixes from PR 1790 ported to devel branch

jblomer:


radupopescu:
I fixed the conflicts here, this can also be merged, to avoid that we needlessly diverge between 2.3 and devel.

PR #1796, opened 2017-02-07
Last minute integration test fixes

radupopescu:
* Install `java-1.8.0-openjdk` on CC7 to fix proxy failover integration test
* Disable overlayfs validation test on Ubuntu 14.04: the OS image needs to be update run Linux 3.18+ in order to support overlayfs. Currently running 3.13.

### DIFF
--- a/test/cloud_testing/platforms/ubuntu_x86_64_test.sh
+++ b/test/cloud_testing/platforms/ubuntu_x86_64_test.sh
@@ -33,7 +33,6 @@ CVMFS_TEST_CLASS_NAME=ServerIntegrationTests                                  \
                                  --                                           \
                                  src/5*                                       \
                                  src/6*                                       \
-#                                 src/7*                                       \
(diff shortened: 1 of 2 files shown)


test/cloud_testing/platforms/ubuntu_x86_64_test.sh
```diff
@@ -33,7 +33,7 @@ CVMFS_TEST_CLASS_NAME=ServerIntegrationTests                                  \
                                  --                                           \
                                  src/5*                                       \
                                  src/6*                                       \
-                                 src/7*                                       \
+#                                 src/7*                                       \
```
jblomer: I think this wouldn't work: the final line break is ignored because the line is a comment.

radupopescu: I didn't know this rule, thanks!

PR #1809, opened 2017-02-23
First attempt at a read-only lower level cache.

bbockelm:
Fixes CVM-1183.

(currently untested, just meant to give an example of the approach)


jblomer:
Superseded by #1810

PR #1810, opened 2017-02-23
Allow for read-only lower cache layer

jblomer:
+ integration tests


jblomer:
In the worst case, we might need to add another flag `CAP_ALL_V2` or similar in the future.

bbockelm:
That's a good idea!  Can we rename `CAP_ALL` to `CAP_ALL_V1`?  Seems that it's OK to rename until we actually release this code.

PR #1811, opened 2017-02-27
Stratum 1 agent and release manager client for triggered replication

jblomer:
also adds Mongoose as a new server dependency


jblomer:
I forgot to reply: for the Perl script, I only checked that the modules are available down to SL6, assuming that every other distribution would provide the modules, too.  Before packaging, this should be checked a bit more thoroughly indeed.

For the Mongoose patch: if we change the signal handling in cvmfs, there is a small race directly after the Mongoose main loop is started but before `signal()` is called.  I think at some point we'd anyway want to switch to civetweb

PR #1815, opened 2017-03-07
Fixing failing unittest on macOS

radupopescu:
T_Util WaitForChild was failing on macOS due to the following call: `execvp("/bin/true", argv)`. On macOS, the correct path is `/usr/bin/true`.


jblomer:
Thanks!

PR #1816, opened 2017-03-07
Improvement: smart pointers

radupopescu:
This branch add Boost smart pointers, providing a safe automatic memory management through reference counting.

Summary:

* Adds the Boost `smart_ptr` module as an external dependency. This module is header-only, it introduces no new library dependencies. The `smart_ptr` headers and their requirements are extracted from the Boost sources, to avoid exploding the size of the CVMFS sources (The complete Boost sources are ~80MB compressed, while the `smart_ptr` sources are ~400kB compressed).
* The smart pointer classes are: `scoped_ptr` - single copy owning pointer, like `UniquePtr` from CVMFS, `shared_ptr` - a shared ownership pointer based on reference counting, `weak_ptr` - a pointer for maintaining weak, non-owning references to `shared_ptr` and `intrusive_ptr` - reference-counted pointer which maintains the reference count in the managed object itself.
* The smart pointer classes from Boost are the basis of the design of the smart pointers from the C++11 standard. Using the smart pointers from Boost makes a potential future upgrade to C++11 smart pointers very simple (substitute `boost` namespace for `std`, substitute `scoped_ptr` for `unique_ptr` etc.).
* The Boost smart pointers do not require any special C++ compiler support, they compile on any C++98 compiler. The branch has been tested on all the platforms that we support (see results) and also on CentOS 5 with the default compiler - GCC 4.1 (Note: on CentOS 4 and 5 CVMFS is compiled with a newer version of GCC from our yum repos).
* Includes unit tests exercising the `scoped_ptr`, `shared_ptr` and `weak_ptr` classes and demonstrating usage and semantics.

Merging this branch opens the possibility of migrating CVMFS to a safer memory management strategy. This migration should be incremental: new code can safely be written using smart pointers, existing code can gradually be migrated as the opportunity arises.

Jenkins cross-platform build results are here:
https://phsft-jenkins.cern.ch/view/CVMFS/job/CvmfsIncrementalBuild/653/


jblomer:
Thanks!  I think moving to safer memory handling is a good path!

I'm still trying to understand the risk a bit more.  Could we

1. Build and run the unit tests also on Power8 and AArch64 (which I think are currently disabled)
2. Have a look at stack traces with smart pointers.  I'd assume, nothing changes, I just want to know for sure.
3. We are adding some 24000 lines of C++.  If, in the future, this code should stop compiling or if tests start failing, do you think we will be able to understand and fix the problem?

(just a minor issue I noticed: the tarball contains the .git sub durectory.)

radupopescu:
About the tarball: good catch! I'm trying to be careful with this, but it must have slipped. In the case of this PR, I plan to squash the commits at the end such that the tarball with the `.git` folder isn't added to the history.

radupopescu:
I made a forced push of this branch to remove all traces of the old tarball which contained `.git`.

radupopescu:
To answer your questions:

1. POWER8, ARM64

I submitted a new CvmfsIncrementalBuild job to test building+unit tests for the latest version of the branch. I included the Arm64 configuration this time (but it may also hang like the nightly builds...)

(Update: Arm64 is hanging during the CvmfsSetupBuildEnvironment stage). 

I don't know what difference these platforms will make, though. The implementation of smart pointers is very basic. I expect that in the smart pointer code there is some branching based on compiler, less branching based on platform. 

2. Stack traces

You will not see anything, unless you capture a stack trace at the exact moment the smart pointer is constructed or destroyed. In any case, shared pointers are implemented without "magic" - just a reference count somewhere, either an atomic integer, or an integer and a lock.Intrusive smart pointers (I haven't used their implementation from Boost or the C++11 standard lib) are more... intrusive (to have a type be reference countable with intrusive pointers, the type needs to be derived from the intrusive pointer itself, which ensures that the reference count is with the object, not elsewhere), but even there I wouldn't expect to see anything in the stack trace.

3. Adding a new C++ dependency.

It's not difficult to write a basic shared pointer implementation. On the other hand, writing something that works consistently across platforms and covers any corner case is much more difficult. Another example: we can easily write a simple key-value store. Re-writing LevelDB for our purpose is not feasable. 

The `smart_ptr` module is quite old and have seen a lot of use, which is why it represents the template for the smart pointer implementation and API from C++11. I think that when it comes to risk you can do much worse than with a core module of Boost. One can disagree with their designs sometimes, but the core libraries are really good quality.

On the other hand, it may happen that in the future there is a compilation error because I didn't include all the headers files (I had to extract them manually from the Boost sources). In that case it's just the matter of adding the missing file to fix the problem (I already had to do this twice, on macOS and Fedora).

Bottom line, I'm not worried about depending on Boost smart pointers. However, I am worried about memory errors and leaks with manual memory management.

I was wondering if we are running Valgrind as part of the CI suite? I can't find it in Jenkins. I'm running Valgrind manually on the unit tests and it definitely sees memory leaks. Of course, some of them would be in the server tools (maybe less severe), hopefully none or few are in the FUSE client (very bad, I imagine).

bbockelm:
What's the oldest supported platform?  The RHEL6 `gcc` has support for smart pointers, no?

radupopescu:
@bbockelm - Very old compilers are supported. As I was saying, I checked `smart_ptr` on CentOS 5 with the default compiler, GCC 4.1, and there are no issues. CVMFS is compiled with 4.1 on both SLC4 and SLC5.

PR #1824, opened 2017-03-14
Allow for keys directory in stratum 1 commands -- devel

jblomer:
See #1823


cvmfs/swissknife_check.h
```diff
@@ -40,7 +40,7 @@ class CommandCheck : public Command {
     r.push_back(Parameter::Optional('t', "temp directory (default: /tmp)"));
     r.push_back(Parameter::Optional('l', "log level (0-4, default: 2)"));
     r.push_back(Parameter::Optional('s', "check subtree (nested catalog)"));
-    r.push_back(Parameter::Optional('k', "public key of the repository"));
+    r.push_back(Parameter::Optional('k', "public key of the repository / dir"));
```
DrDaveD: This would make more sense as "public key / dir of the repository"

PR #1832, opened 2017-03-20
Use HMAC for signing HTTP gateway API requests

radupopescu:
Addresses [CVM-1191](https://sft.its.cern.ch/jira/browse/CVM-1191) and [CVM-1193](https://sft.its.cern.ch/jira/browse/CVM-1193).

* All the requests (new lease, drop lease, submit payload) made to the HTTP gateway API are now signed with an HMAC computed with a pre-shared secret key.
* The shared key is configured through the `CVMFS_KEY` configuration option, written in the server onfiguration file (`server.conf`) of the repository. Right now the key has a test format - first line is the public key id, second line is the secret - but this can be updated in the future to use standard key formats.
* The gateway services (`cvmfs_services`) checks the HMAC added to all the requests.

### DIFF
--- a/cvmfs/server/cvmfs_server_abort.sh
+++ b/cvmfs/server/cvmfs_server_abort.sh
@@ -8,6 +8,7 @@
 cvmfs_server_abort() {
   local names
   local user
+  local key_id
   local spool_dir
   local exact=0
   local force=0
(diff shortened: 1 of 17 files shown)


jblomer:
Actually, we might perhaps decide to call `CVMFS_KEY` `CVMFS_GATEWAY_KEY` or similar to distinguish it from the other keys (repo public/private key pairs).

PR #1837, opened 2017-03-22
Install systemd/autofs patch configuration through RPM where necessary -- 2.3

jblomer:


DrDaveD:
Shouldn't you also undo the creation of the corresponding file under /etc/systemd?

jblomer:
No, everything else stayed as is.  If `cvmfs_config setup` discovers that autofs' `KillMode` is not `process`, it will still drop a patch file into /etc/systemd.  So there can be situations in which the cleanup is necessary.

jblomer:
Sorry, that's about removal of the file in /etc/systemd.  About creation: if the `KillMode` already is `process`, nothing needs to be dropped and the file is not created.  So we can leave that in.

DrDaveD:
Alright, cool

jblomer:
Thanks for testing!

PR #1849, opened 2017-04-19
Fix atlas nightlies config

jblomer:


DrDaveD:
This repository is also available on the BNL stratum 1.  I think it would be better to just delete config.d/atlas-nightlies.cern.ch.conf and let clients probe to find it on one of the two stratum ones that have it.  That's what we do with cms-ib.cern.ch which is at CERN & FNAL.

PR #1852, opened 2017-04-27
Add support for branching

jblomer:
Adds the `cvmfs_server checkout` command that allows to branch off a given tag.  Makes most sense with the vitual catalog that provides the /.cvmfs/snapshots directory.  The tag database now has a new column `branch`, and a new table that stores parent-child branch relationship and the initial revision of branches.  A branch is removed if all the tags using it get removed.  Children of removed branches are automatically re-parented to their grand-parents.

Root catalogs store their branch.  Comes with a new extended attribute `branch`.

The recycle bin table in the tag database was unused since the introduction of the reflog.  The table is now flushed and deprecated, since we were anyway dealing with the SQL code around the tag database.

Fixes a bug in the creation of the snapshots directory discovered during testing.  Also, stratum 0 garbage collection does not modify anymore the history (and hence the manifest) anymore.  Adds the missing `update-repoinfo` to the new cvmfs_server build system.


radupopescu:
I'm wondering if including the branch name in a root catalog is the proper design choice here. It seems to be different to Git branches (which may not be an example of good UI, I don't know)?

jblomer:
That's a good question.  One reason is to ensure that root catalogs are unique.  So far, storing the revision was enough, but now the revision became a two dimensional (branch, revision).  To ensure root catalogs are unique, we could instead store a UUID in the root catalog.  As for the user interface, users can lookup the catalog revision already through extended attributes.  So it might make sense for them to be able to query both to distinguish catalog revision X on two different branches.

Then again, I expect that users access branched tags anyway through the .cvmfs/snapshots directory of the latest catalog. And we can also argue that tags and branches are sort of annotations to an existing repository.

I guess storing the branch doesn't hurt much for the time being.  I could remove the access to the branch through extended attributes so that we don't commit to the branch being stored in the root catalog.

PR #1855, opened 2017-05-04
Geosort servers by distance to fallback proxies when there are no regular proxies

DrDaveD:
See [CVM-769](https://sft.its.cern.ch/jira/browse/CVM-769)


jblomer:
I'm not sure I understand the need for the `+PXYSEP+`.  Couldn't the client simply use the fallback proxy when making the Geo-API request?  On the server side, [the host name of the fallback proxy would be picked up](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/webapi/cvmfs_geo.py#L95) and used as `gir_rem`, no?

DrDaveD:
The problem is it doesn't know which fallback proxy is the closest.  This allows the sorting to be done in one request.  Otherwise yes it could first ask to sort the fallback proxies and then request that the servers be sorted relative to closest fallback proxy.

PR #1856, opened 2017-05-08
Fixes for OpenSSL 1.1 interface changes

jblomer:
@bbockelm the changes in authz_curl.cc are untested at the moment, perhaps you'd like to take a second look.


bbockelm:
The `authz_curl.c` changes, at the outset, look reasonable to me.

The `#ifdef` macros are getting quite ugly (`OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)`); might I suggest combining this into a macro like `SSL_API_INTERFACE_V2` (or similarly named)?

PR #1861, opened 2017-05-12
CVM-1195

radupopescu:
This PR contains the implementation of the catalog reconciliation logic in the gateway services [CVM-1195](https://sft.its.cern.ch/jira/browse/CVM-1195).

This closes the operation "pipeline" of the new workflow based on multiple release manager machines and gateway services.


cvmfs/swissknife_lease_curl.h
```diff
@@ -1,6 +1,6 @@
 /**
  * This file is part of the CernVM File System.
- */
+xs */
```
jblomer: Is this indended?

radupopescu: This is from Emacs C-x C-s. Why does it compile? :))

cvmfs/receiver/catalog_merge_tool.cc
```diff
@@ -0,0 +1,234 @@
...
+      path_(other.path_),
+      xattrs_(other.xattrs_),
+      entry1_(new catalog::DirectoryEntry(*(other.entry1_))),
+      entry2_(other.entry2_ ? new catalog::DirectoryEntry(*(other.entry2_))
+                            : NULL) {}
+
+CatalogMergeTool::ChangeItem::~ChangeItem() {
+  if (entry1_) {
+    delete entry1_;
+    entry1_ = NULL;
```
radupopescu: I think this is necessary. Instances of `ChangeItem` are supposed to go into a `std::vector` so I can't use `UniquePtr` inside them. I'm using raw pointers since `entry2_` can be NULL.

cvmfs/receiver/catalog_merge_tool.h
```diff
@@ -0,0 +1,94 @@
...
+  struct ChangeItem {
+    enum ChangeType { kAddition, kRemoval, kModification };
+    ChangeItem(ChangeType type, const PathString& path,
+               const catalog::DirectoryEntry& entry1);
+    ChangeItem(ChangeType type, const PathString& path,
+               const catalog::DirectoryEntry& entry1, const XattrList& xattrs);
+    ChangeItem(ChangeType type, const PathString& path,
+               const catalog::DirectoryEntry& entry1,
+               const catalog::DirectoryEntry& entry2);
+    ChangeItem(const ChangeItem& other);
```
radupopescu: I don't understand this. I actually need the copy constructor and assignment operator, since I'm putting objects of `ChangeItem` in a `std::vector`. I added an explicit definition of `operator=`.

cvmfs/receiver/catalog_merge_tool.cc
jblomer: What I mean is that the `delete` operator can deal with `nullptr` input, and I'd think that we wouldn't need the `entry1_ = NULL` either because the `ChangeItem` is going away.

cvmfs/receiver/catalog_merge_tool.cc
```diff
@@ -0,0 +1,234 @@
...
+}
+
+void CatalogMergeTool::ReportModification(
+    const PathString& path, const catalog::DirectoryEntry& entry1,
+    const catalog::DirectoryEntry& entry2) {
+  changes_.push_back(ChangeItem(ChangeItem::kModification, MakeRelative(path),
+                                entry1, entry2));
+}
+
+bool CatalogMergeTool::InsertChangesIntoOutputCatalog() {
```
radupopescu: I haven't even thought about that. I don't know how `.cvmfscatalog` works. This case should be covered in a unit or integration test.

cvmfs/receiver/catalog_merge_tool.cc
radupopescu: Oh, yes. Obviously!

cvmfs/receiver/catalog_merge_tool.cc
```diff
@@ -0,0 +1,234 @@
...
+          // From file to directory
+          output_catalog_mgr_->RemoveFile(change.path_.c_str());
+          output_catalog_mgr_->AddDirectory(*change.entry2_, parent_path);
+
+        } else if (change.entry1_->IsDirectory() &&
+                   (change.entry2_->IsRegular() || change.entry2_->IsLink())) {
+          // From directory to file
+          const catalog::DirectoryEntryBase* base_entry =
+              static_cast<const catalog::DirectoryEntryBase*>(change.entry2_);
+          output_catalog_mgr_->RemoveDirectory(change.path_.c_str());
```
radupopescu: Yes!

radupopescu:
So, the mocked catalog and catalog manager classes are very incomplete and need some work to make them really usable.

I made a new ticket [CVM-1290](https://sft.its.cern.ch/jira/browse/CVM-1290) about developing a QuickCheck test suite for the `CatalogMergeTool` classes, which involves completing the mocked catalog classes, such that the unit testing can be done without disk based catalogs. I would schedule this for after the integration tests are added and we have a deployment solution for the cvmfs services (tickets [CVM-1270](https://sft.its.cern.ch/jira/browse/CVM-1270) and [CVM-1273](https://sft.its.cern.ch/jira/browse/CVM-1273)).

All the other issues here have been addressed, so I would go ahead with merging this.

jblomer:
Great!  Let's kick off another round of integration tests after the merge.

radupopescu:
Great! I'll wait for Travis to finish and then I'll merge this. I'm fixing up PR #1867, so that can be soon merged, too.

PR #1859, opened 2017-05-10
Add support for Yubikey 4 & NEO to cvmfs-server

DrDaveD:
See [CVM-1259](https://sft.its.cern.ch/jira/browse/CVM-1259)

PR #1863, opened 2017-05-12
Add cvmfs_server resign -n option

DrDaveD:
Important note: Must be merged after pr #1859; all of those changes are included here because edits to cvmfs_server_ssl.sh in that pr were too close to the edits here to be independent.

See [CVM-1265](https://sft.its.cern.ch/jira/browse/CVM-1265)


DrDaveD:
Ok I changed -n to -w <path>. 

When I was almost finished it occurred to me that it is still requiring /etc/cvmfs/keys/<fqrn>.masterkey, but it doesn't require that if there's a masterkeycard.  So it's still probably worth allowing the existing whitelist to be in a flexible location.

PR #1865, opened 2017-05-15
Fixes for linking against protobuf

jblomer:


radupopescu:
I was wondering if this is related to the race condition at build time? Good job!

jblomer:
I hope so.  It's a shot in the dark intended to fix the race.

PR #1877, opened 2017-05-29
Perform a fail-over if manifest or whitelilst are corrupt

jblomer:


cvmfs/manifest_fetch.cc
```diff
@@ -162,4 +165,31 @@ Failures Fetch(const std::string &base_url, const std::string &repository_name,
...
+  {
+    LogCvmfs(kLogCache, kLogDebug | kLogSyslogWarn,
+             "failed to fetch manifest (%d - %s), trying another stratum 1",
+             result, Code2Ascii(result));
+    download_manager->SwitchHost();
+    result =
+      DoFetch(base_url, repository_name, minimum_timestamp, base_catalog,
+              signature_manager, download_manager, ensemble);
+  }
+  return result;
```
DrDaveD: So this only tries one more Stratum 1?  Is it too difficult to try them all like happens with file downloads?

jblomer: I guess we could.  The `DoFetch` routine can do fail-overs internally if hosts are unreachable.  I was afraid of timeouts piling up.  Also, corrupted whitelist/manifest is a pretty rough state for a stratum 1, I'd hope we will never see two at the same time in this state...

PR #1880, opened 2017-05-31
change default graft size from 32 to 24 MB

DrDaveD:
See [CVM-1291](https://sft.its.cern.ch/jira/browse/CVM-1291)

### DIFF
--- a/cvmfs/swissknife_graft.cc
+++ b/cvmfs/swissknife_graft.cc
@@ -19,6 +19,8 @@
 #include "util/posix.h"
 #include "util/string.h"
 
+const std::string swissknife::CommandGraft::kDefaultChunkSizeStr = "24";
+
 bool swissknife::CommandGraft::ChecksumFdWithChunks(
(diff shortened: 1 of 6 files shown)


jblomer:
Many thanks!

I think it would be good to introduce a private constant like `kDefaultChunkSize` for the `CommandGraft` class so that the number only needs to be changed at one place.

DrDaveD:
Ok, I pushed an update, but since I need to use a string it isn't quite as clean as an integer constant.   C++03 does not allow setting a per-class static constant string that is initialized at the same time as it is defined.  Instead I have to define it in the class definition but initialize it in swissknife_graft.cc.   I can initialize an integer when it is defined, but GetDescription() wants a string.  C++11 has a new 'constexpr' way to assign strings the same time as they're defined, but we're not using C++11.

jblomer:
You can use an `unsigned` constant and the `StringifyUint` helper from util/string.h

DrDaveD:
Ok how about now

PR #1882, opened 2017-06-01
change usage of getgrnam_r and getpwnam_r to follow man page API

DrDaveD:
See [CVM-1304](https://sft.its.cern.ch/jira/browse/CVM-1304)

### DIFF
--- a/cvmfs/util/posix.cc
+++ b/cvmfs/util/posix.cc
@@ -898,13 +898,12 @@ vector<string> FindDirectories(const string &parent_dir) {
 bool GetUidOf(const std::string &username, uid_t *uid, gid_t *main_gid) {
   struct passwd pwd;
   struct passwd *result = NULL;
-#ifdef _SC_GETPW_R_SIZE_MAX
-  int bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
-#else
   int bufsize = 16 * 1024;
-#endif
   char *buf = static_cast<char *>(smalloc(bufsize));
-  getpwnam_r(username.c_str(), &pwd, buf, bufsize, &result);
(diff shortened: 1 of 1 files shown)


jblomer:
Many thanks!

Since the suggestion is anyway to increase the buffer on ERANGE, how about dropping the calls to `sysconf`.  Instead both functions can continue doubling the buffer until `errno != ERANGE`.

DrDaveD:
The man page says the API is to use those values.  If you want me to go ahead and stick with 16K anyway and ignore the API, please confirm and I will change it.  The value for getgrnam_r does start out at only 1K (at least on el6) so it takes 4 doublings to get to 16K and is less efficient when it has to do that.  

The API for getpwnam_r is different than getgrnam_r.  According to the el6 man page, the former says to just use the value of the sysconf variable and that tells you the maximum size.   On the other hand, I just checked and the el7 man page's description is just like getgrnam_r, where you are to retry with a larger buffer if you get ERANGE.  Oh, I see that the el6 man page says there is also ERANGE if you give it too small of a value, so it should be alright there too.

The Mac getgrnam_r man page doesn't say anything about sysconf nor does it say that it can return an ERANGE error if the buffer is too small.  It does say that it can return an error value but doesn't say what the errors can be.   On the other hand the Mac getpwnam_r man page does say that it can return an ERANGE, so probably getgrnam_r can too and it is just missing from the man page.

jblomer:
Yep, please leave the current 16kB initial buffer but having buffer doubling on `ERANGE`.  I just tested on OS X, `getpwnam_r` behaves as expected.

DrDaveD:
Ok, done

PR #1888, opened 2017-06-13
SharedPtr  - a reference counted smart pointer class

radupopescu:
Addresses [CVM-1204](https://sft.its.cern.ch/jira/browse/CVM-1204).

SharedPtr is a reference counted smart pointer class which provides shared ownership semantics. The implementation and interface is modelled after the `std::shared_ptr` class.


cvmfs/atomic.h
```diff
@@ -17,92 +17,74 @@ namespace CVMFS_NAMESPACE_GUARD {
...
-{
-  return __sync_bool_compare_and_swap(a, cmp, newval);
+static int32_t inline __attribute__((used))
+atomic_cas32(atomic_int32 *a, int32_t cmp, int32_t newval) {
+  return __sync_val_compare_and_swap(a, cmp, newval);
 }
 
+static int64_t inline __attribute__((used))
+atomic_cas64(atomic_int64 *a, int64_t cmp, int64_t newval) {
+  return __sync_val_compare_and_swap(a, cmp, newval);
```
jblomer: I think this breaks fuse_remount.cc:40

Before it returned `true` (1) if the operation was successful.  Now a successful `cas(a, 0, 1)` returns 0/false.

cvmfs/util/shared_ptr.h
```diff
@@ -0,0 +1,183 @@
...
+
+  SharedPtr() {  // never throws
+    value_ = NULL;
+    count_ = NULL;
+  }
+
+  template <class Y>
+  explicit SharedPtr(Y* p) {
+    if (static_cast<element_type*>(p) == value_) {
+      abort();
```
jblomer: Requires stdlib

cvmfs/util/shared_ptr.h
```diff
@@ -0,0 +1,183 @@
...
+
+  template <class Y>
+  SharedPtr(SharedPtr<Y> const& r)
+      : value_(r.value_), count_(r.count_) {  // never throws
+    if (count_) {
+      atomic_inc64(count_);
+    }
+  }
+
+  SharedPtr& operator=(SharedPtr const& r) {  // never throws
```
jblomer: Wouldn't I want to reset my current state first?

cvmfs/util/shared_ptr.h
```diff
@@ -0,0 +1,183 @@
...
+    if (atomic_read64(count_) == 0) {
+      delete value_;
+      delete count_;
+    }
+    value_ = static_cast<element_type*>(p);
+    count_ = new atomic_int64;
+    atomic_write64(count_, 1);
+  }
+
+  T& operator*() const {  // never throws
```
jblomer: This can throw after 'Reset()', no?

cvmfs/util/shared_ptr.h
```diff
@@ -0,0 +1,183 @@
...
+  return SharedPtr<T>(dynamic_cast<T*>(r.value_));
+}
+
+template <class T, class U>
+SharedPtr<T> ReinterpretPointerCast(SharedPtr<U> const& r) {  // never throws
+  return SharedPtr<T>(reinterpret_cast<T*>(r.value_));
+}
+
+template <class E, class T, class Y>
+std::basic_ostream<E, T>& operator<<(std::basic_ostream<E, T>& os,
```
jblomer: We can probably remove this since we don't use streams anyway

cvmfs/atomic.h
radupopescu: I think it was no clean. It was calling `__sync_bool_compare_and_swap` but implicitly converting to an int32_t, since this was the return value of the `atomic_cas32` function.

I missed the use of this function in fuse_remount.

cvmfs/util/shared_ptr.h
radupopescu: How are exceptions handled in CVMFS?

jblomer: They aren't, cvmfs doesn't use exceptions.

PR #1889, opened 2017-06-13
Fix deprecation warnings for OSSpinLock on macOS 10.12+

radupopescu:
Addresses [CVM-1185](https://sft.its.cern.ch/jira/browse/CVM-1185).

Fixes the warning about the deprecation of `OSSpinLock` on macOS 10.12+. On this version of macOS and newer, `os_unfair_lock` is used instead.

### DIFF
--- a/cvmfs/platform_linux.h
+++ b/cvmfs/platform_linux.h
@@ -7,6 +7,8 @@
 #ifndef CVMFS_PLATFORM_LINUX_H_
 #define CVMFS_PLATFORM_LINUX_H_
 
+#include <sys/types.h>  // contains ssize_t needed inside <attr/xattr.h>
+#include <sys/xattr.h>
 #include <attr/xattr.h>  // NOLINT(build/include_alpha)
 #include <dirent.h>
 #include <errno.h>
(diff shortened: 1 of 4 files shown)


cvmfs/platform_linux.h
```diff
@@ -92,10 +89,8 @@ inline bool platform_umount(const char* mountpoint, const bool lazy) {
       return false;
     }
     FILE *fmntnew = setmntent(mntnew.c_str(), "w+");
-    if (!fmntnew &&
-        (chmod(mntnew.c_str(), mtab_info.st_mode) != 0) &&
-        (chown(mntnew.c_str(), mtab_info.st_uid, mtab_info.st_gid) != 0))
-    {
+    if (!fmntnew && (chmod(mntnew.c_str(), mtab_info.st_mode) != 0) &&
```
jblomer: I wondered for a while, are these reformattings automatically done by the editor?  I wondered because they add a number of "things" to the pull requests besides the actual changes.

radupopescu: I have clang-format running on save. Most of the time it's a time saver, but it's true that it creates some noise the first time it's applied on files.

Sometimes it's bad, like today when it broke compilation in this branch, since it reordered the include files alphabetically in a file where they needed to be in a specific order.

PR #1892, opened 2017-06-17
add cvmfs_server resign -p option

DrDaveD:
See [CVM-1140](https://sft.its.cern.ch/jira/browse/CVM-1140)


cvmfs/server/cvmfs_server_resign.sh
```diff
@@ -35,12 +39,13 @@ cvmfs_server_resign() {
   [ -n "$whitelist_path" ] || check_multiple_repository_existence "$names"
 
   # sanity checks
-  is_root || die "Only root can resign repositories"
+  [ $sign_published -eq 0 ] || [ $require_repo_config -eq 1 ] || die "Cannot use both -n and -p"
```
jblomer: I think this condition refers to an older version of the patch set.

DrDaveD: Good catch, fixed.

cvmfs/server/cvmfs_server_resign.sh
```diff
@@ -35,12 +39,13 @@ cvmfs_server_resign() {
   [ -n "$whitelist_path" ] || check_multiple_repository_existence "$names"
 
   # sanity checks
-  is_root || die "Only root can resign repositories"
+  [ $sign_published -eq 0 ] || [ $whitelist_path -eq 1 ] || die "Cannot use both -w and -p"
```
jblomer: Isn't this still wrong?  The `whitelist_path` parameter carries a string not an integer.

DrDaveD: Yes, sorry, and thank you!  I'm sorry I didn't adequately test it.  Fixed now.

PR #1899, opened 2017-06-23
prevent external server url from inheriting regular server url

DrDaveD:
See [CVM-1322](https://sft.its.cern.ch/jira/browse/CVM-1322)

It was easy to prevent the inheritance, but now if somebody attempts to use an external server it says "malformed URL" instead of "host returned HTTP error".  It's slightly more informative, but not a lot.  I would like to add an error message saying that the external url was not set, but I can't figure out exactly where to put it.  The debug log shows "failed to fetch Part of ..." but it isn't clear to me where the external fetcher is called from nor how easy it is to check inside of Fetcher::Fetch() for this condition.


DrDaveD:
@jblomer thank you for merging but can you also please help me figure out where I can insert an error message about there being no external server url set?

PR #1902, opened 2017-06-27
Clean up and improve logging in cvmfs_receiver

radupopescu:
* Makes the `cvmfs_receiver` tool log to syslog.
* Adds some extra logging for different operations during publication.


cvmfs/receiver/payload_processor.cc
```diff
@@ -67,11 +67,15 @@ void PayloadProcessor::ConsumerEventCallback(
...
   } else {
     // kEmpty - this is an error.
-    LogCvmfs(kLogCvmfs, kLogStderr, "Event received with unknown object.");
+    LogCvmfs(kLogReceiver, kLogDebug | kLogSyslogErr,
+             "Event received with unknown object.");
     num_errors_++;
     return;
   }
 
+  LogCvmfs(kLogReceiver, kLogDebug | kLogSyslog,
```
jblomer: Would it make sense to put this only in the debug log?  I can imagine that under load it quickly fills syslog.  We can also attach a TODO here if we want to do it later.

jblomer: Probably the same for https://github.com/cvmfs/cvmfs/pull/1902/files#diff-1e4aca86201436e103fcd7ff1b7ff156R51

radupopescu: You may have a point here.

At the same time, here was my reasoning: Having this logged to syslog (which is the only logging being done in "production") would give me some way to diagnose things in case of problems. I also imagine that the syslog surely has log rotation enabled such that I don't fill up the root partition with this.

What do you think is the best approach?

jblomer:
I guess what we want is support in `LogCvmfs` to log neither into syslog nor into the debug log but in a pre-defined log file; for things like an access_log.  Let me take a look.

radupopescu:
That would be amazing. On the services' side, I can easily configure the logger to send to syslog and/or file, console etc.

I see that there is a private API for `LogCvmfs` to set a file for logging, but only in debug mode. There is also the option of setting my own logging function, also through a private API in `logging_internal.h`.

jblomer:
See #1908, which adds support for up to three custom log files.  It is used like

    LogSetCustomFile(0, "my-file");
    LogCvmfs(kLogCvmfs, kLogCustom0, ...);

radupopescu:
This is great, thanks!

radupopescu:
Sorry for the late reply.

I'm thinking about where to put the log files. In the `cvmfs_services` application the `lager_syslog` backend is configured with a `cvmfs_services` identity and an rsyslog configuration file for the application is installed into `/etc/rsyslog.d`. The log files are created and owned by the `syslog` user. I think this is a clean way to do it, probably facilitates the integration with system-wide log rotation tools.

Wouldn't it be better to configure the logging here to use the same approach (configure syslog to handle things for us)?

cvmfs/receiver/commit_processor.cc
```diff
@@ -73,7 +78,8 @@ CommitProcessor::Result CommitProcessor::Process(
 
   // Current catalog from the gateway machine
   if (!manifest.IsValid()) {
-    LogCvmfs(kLogCvmfs, kLogStderr, "Could not open repository manifest");
+    LogCvmfs(kLogReceiver, kLogCustom1,
+             "Could not open repository manifest\n");
```
jblomer: Ah, sorry, did I forget to add the newline inside `LogCvmfs`?  Sorry about that!  I'll fix it.

jblomer: Done in #1913

PR #1907, opened 2017-06-28
Add cvmfs_server resign -d option

DrDaveD:
See [CVM-1279](https://sft.its.cern.ch/jira/browse/CVM-1279)


jblomer:
Thanks!  In principal it looks fine to me.  Please don't block whitelists longer than 30 days.  There probably should be a warning and a confirmation instead, with a `-f` flag to suppress the confirmation.

I find it confusing to have different defaults depending on the signing method.  I'd prefer having both at 30 days, especially since it is so easy to overwrite the default.

DrDaveD:
We agreed on the development call to leave the drop to 7 days for the default to masterkeycard.

I have now pushed a change for an undocumented '-f' option to override the 30 days maximum.

DrDaveD:
I changed it to prompt if greater than 30 days, and to have the -f be a documented way to skip the question.

PR #1922, opened 2017-07-07
Add support for repository revision blacklist entries

DrDaveD:
See [CVM-992](https://sft.its.cern.ch/jira/browse/CVM-992)


test/src/047-blacklist/main
```diff
@@ -19,17 +20,42 @@ cvmfs_run_test() {
...
+  cvmfs_mount grid.cern.ch || return 31
+
+  echo "blacklist the grid.cern.ch revision"
+  pub="`curl -f -s http://cvmfs-stratum-one.cern.ch/cvmfs/grid.cern.ch/.cvmfspublished`" || return 32
+
+  echo "$pub" | cat -v | awk '/^S/{r=substr($1,2);print "<grid.cern.ch",r+1;}' \
+    | sudo sh -c "tee /etc/cvmfs/blacklist"
+
+  echo "force remount which should cause blacklisted repo to unmount"
+  sudo cvmfs_talk -i grid.cern.ch remount
```
jblomer: We might be able to avoid the `sleep` by using the new `remount sync`

DrDaveD: I tried it, it didn't help.  It makes sense, because the blacklist check will not return to CheckSynchronously().  The watchdog process only checks once per second to unmount after the main process terminates. so it takes a while.

By the way, I noticed that remount sync is not documented,  I don't know if that is intentional or not.

DrDaveD:
I'm sorry I hadn't noticed the cpplint errors from travis, they're fixed now.

jblomer:
Replaced by #1937

PR #1924, opened 2017-07-09
Optionally omit generation of bulk hashes for chunked files

jblomer:
For the time being, the default is still to create bulk hashes except for the graft utility.  The graft utility does not (yet) have a stable interface.  

As a side effect, large hard links are now chunked, too, which wasn't the case before.

**Note**: When bulk hashes are omitted in gc-enabled repositories, all stratum 1s have to update, too.  Otherwise they will remove chunks from files without a bulk hash during garbage collection.

This PR also reverts to zlib 1.2.8, version 1.2.11 introduced bug fixes that affect cvmfs.  The PR also fixes building of the `cvmfs_swissknife_debug` binary.


cvmfs/file_processing/chunk_detector.h
```diff
@@ -125,18 +118,24 @@ class Xor32Detector : public ChunkDetector {
...
+    return abs(static_cast<int32_t>(xor32_) - kMagicNumber) < threshold_;
   }
 
  private:
+  typedef std::pair<size_t, uint32_t> Threshold;
+  typedef std::vector<Threshold> Thresholds;
+
+  // xor32 only depends on a window of the last 32 bytes in the data stream
+  static const size_t kXor32Window = 32;
+  static const int32_t kMagicNumber;
```
DrDaveD: I don't understand how the kMagicNumber can be defined here without a value.

jblomer: It's only declared here, defined in the corresponding .cc file.

PR #1934, opened 2017-07-13
Fixing some issues reported by Coverity

radupopescu:


cvmfs/gateway_util.cc
```diff
@@ -27,7 +27,9 @@ bool ReadKeys(const std::string& key_file_name, std::string* key_id,
   std::string line;
   if (!GetLineFile(key_file_fd, &line)) {
     return false;
+    fclose(key_file_fd);
```
jblomer: That should be before the `return`, I guess.

PR #1937, opened 2017-07-14
Add support for repository revision blacklist entries -- updated

jblomer:
Updates #1922


jblomer:
@DrDaveD I made several adjustments, please have another look.

cvmfs/fuse_remount.cc
```diff
@@ -34,6 +35,16 @@ FuseRemounter::Status FuseRemounter::Check() {
...
+
+  LogCvmfs(kLogCvmfs, kLogDebug, "checking revision against blacklists");
+  if (mountpoint_->ReloadBlacklists() &&
+      mountpoint_->catalog_mgr()->IsRevisionBlacklisted())
+  {
+    LogCvmfs(kLogCatalog, kLogDebug | kLogSyslogErr,
+            "repository revision blacklisted, aborting");
+    abort();
+  }
+
```
DrDaveD: I think there should probably be a debug message here indicating that the blacklist check is finished and that the remount is about to happen.  Maybe the message at the beginning should just say "catalog TTL expired" (or be incorporated into the checking... message) and a message here that says "remounting".

cvmfs/mountpoint.cc
```diff
@@ -1017,26 +1017,28 @@ bool MountPoint::CheckBlacklists() {
   string blacklist;
   if (!options_mgr_->GetValue("CVMFS_BLACKLIST", &blacklist))
     blacklist = kDefaultBlacklist;
+  blacklist_paths_.push_back(blacklist);
```
DrDaveD: This assumes that blacklist_paths_ is empty and that CheckBlacklists() will only ever be called once (as is the case now, but maybe that will change one day).  Maybe it should not append here, but instead overwrite, just as the "append" boolean does for the contents of the blacklists.

PR #1938, opened 2017-07-14
Change catalog warn threshold to checks for catalog and file size limits

DrDaveD:
See [CVM-1094](https://sft.its.cern.ch/jira/browse/CVM-1094) and [CVM-1123](https://sft.its.cern.ch/jira/browse/CVM-1123)

PR #1940, opened 2017-07-17
Change catalog warn threshold to checks for catalog and file size limits -- update

jblomer:
Replaces #1938


cvmfs/server/cvmfs_server_migrate.sh
```diff
@@ -373,6 +373,14 @@ _migrate_137() {
   fi
 
   echo "--> updating server.conf"
+  local warn_threshold="`sed -n -e 's/^CVMFS_CATALOG_ENTRY_WARN_THRESHOLD=//p' $server_conf`"
+  if [ -n "$warn_threshold" ]; then
+    sed -i -e '/^CVMFS_CATALOG_ENTRY_WARN_THRESHOLD=/d' $server_conf
+    local kcatalog_limit="$(($warn_threshold / 1000))"
+    echo "CVMFS_ROOT_KCATALOG_LIMIT=${kcatalog_limit}" >> $server_conf
```
DrDaveD: This is not good, because it means that every pre-existing repository that had the default setting of CVMFS_CATALOG_ENTRY_WARN_THRESHOLD (that is, 500000) will not check for the new lower limit of 200k on the root catalog.  That is why I only added the new entries if $kcatalog_limit != 500.   I included CVMFS_NESTED_KCATALOG_LIMIT in the check to make it easier to change the defaults in the future.  Please put that check back.

PR #1949, opened 2017-07-19
Cache geoapi DNS lookups in a threaded shared wsgi python interpreter

DrDaveD:
See [CVM-1349](https://sft.its.cern.ch/jira/browse/CVM-1349)

This changes the geoapi to cache all DNS lookup results for 5 minutes.  If a DNS lookup fails and there is an old result available, it will be reused.  The interpreter is multithreaded so the cache is protected with a lock.

This also changes the WSGI apache config to all be in one shared cvmfs.webapi.conf apache config file instead of in every cvmfs.\<repo\>.conf file.  I hadn't been able to figure out how to do that before or I would have done it earlier.  The key was to put the WSGIProcessGroup inside of a <Directory> section for the wsgi script.  Also WSGIApplicationGroup is needed to make them all use the same python interpreter.  Only one WSGIDaemonProcess is started but with 64 threads.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ os:
 
 compiler:
   - gcc
-  - clang
 
 sudo: required
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,11 +15,12 @@ message ("Running CMake version ${CMAKE_VERSION}")
 # The version numbers
 #
 # DON'T DELETE
-## CVMFS_VERSION 2.4.1
+
+## CVMFS_VERSION 2.5.0
 #---------------------
 set (CernVM-FS_VERSION_MAJOR 2)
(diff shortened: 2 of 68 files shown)


DrDaveD:
We've been running this version of the geo api plugin with essentially the same httpd configuration in production on the Fermilab stratum 1 for the last couple of hours.  No problems seen so far.

DrDaveD:
I have now updated this PR so it has a separate migrate step.  The PR is against the devel branch but should still apply equally to the cvmfs-2.4 branch.

DrDaveD:
I updated the migrate function so versions 137 & 138 reuse the same function and fixed the problem noted at the end of [CVM-1357](https://sft.its.cern.ch/jira/browse/CVM-1357).

DrDaveD:
Replaced by #1983

PR #1962, opened 2017-08-03
Fix off-by-one error for chunk size.

bbockelm:
Previously, a new chunk would be emitted after `N+(buffer size)` bytes when the user requested N bytes for chunk size.  So, if the user wanted a 24MB chunk, CVMFS would create 24MB+16KB chunks for `cvmfs_swissknife graft`.

This corrects the issue; 24MB chunks are created when 24MB chunks are requested for grafts.

(This seems trivial, but it is actually really annoying when trying to validate other implementations of grafting!)

### DIFF
--- a/cvmfs/swissknife_graft.cc
+++ b/cvmfs/swissknife_graft.cc
@@ -89,7 +89,7 @@ bool swissknife::CommandGraft::ChecksumFdWithChunks(
     }
 
     // Start a new hash if current one is above threshold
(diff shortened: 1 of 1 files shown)


jblomer:
Many thanks!

Can you resubmit against the 2.4 branch?  We'd make this part of the 2.4.1 patch release.

Also, please submit some documentation on grafting to the docs repository.

bbockelm:
Closing - this PR is superseded by #1982

PR #1963, opened 2017-08-03
Make `true` to be a valid boolean option.

bbockelm:
Currently, `YES`, `ON`, or `1` indicate truth for a boolean option; this also allows `TRUE`.


jblomer:
Looks good to me, there's just a linter error.

bbockelm:
hm.  GitHub got confused by how I did the rebase - the other two commits are already part of  `devel`.

PR #1964, opened 2017-08-04
Minor fixups from debugging external cache plugins.

bbockelm:
For the most part, this just adds a few logging messages to make misconfigurations more obvious.  

Two significant changes:
- Make an invalid locator string a fatal error (instead of indefinitely retrying it).
- Launch the external plugin as the `cvmfs` user.

### DIFF
--- a/cvmfs/cache_extern.cc
+++ b/cvmfs/cache_extern.cc
@@ -185,8 +185,16 @@ int ExternalCacheManager::ConnectLocator(const std::string &locator) {
   } else {
     return -EINVAL;
   }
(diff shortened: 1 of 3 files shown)


jblomer:
Thanks!  Please re-submit against the 2.4 branch, we'd like to put it into a 2.4.1 patch release.

bbockelm:
As noted [here](https://github.com/cvmfs/cvmfs/pull/1968#issuecomment-325363920), this will be merged first to the 2.4 branch, then through to `devel`.  Accordingly, closing for now.

PR #1967, opened 2017-08-04
Tiered cache txn should succeed if upper tier txn does.

bbockelm:
Previously, the tiered cache counted a transaction as successful if *both* caches had a successful transaction.  Thus, the reliability of the tiered cache was the _AND_ of both caches -- i.e., slightly decreased.

This changes the success metric so the transaction commit succeeds if and only if the upper cache transaction works (the upper cache *must* work because we may have handed out a FD to the cache via OpenFromTxn already).

@jblomer - will you merge `cvmfs-2.4` into `devel`?  Or should I do a separate PR?

### DIFF
--- a/cvmfs/cache_tiered.cc
+++ b/cvmfs/cache_tiered.cc
@@ -187,7 +187,13 @@ int TieredCacheManager::CommitTxn(void *txn) {
     lower_result = lower_->CommitTxn(txn2);
   }
 
-  return (upper_result < 0) ? upper_result : lower_result;
+  // Transaction is successful as long as the storage was successful
+  // in the upper cache (which may have already handed out a FD via
+  // OpenFromTxn).
+  //
+  // If there wasn't the OpenFromTxn semantics, we could have this
+  // call succeed if *either* transaction was successful.
+  return upper_result;
 }
 
 
--- a/cvmfs/cache_tiered.cc
+++ b/cvmfs/cache_tiered.cc
@@ -9,6 +9,7 @@
 #include <string>
 #include <vector>
 
+#include "logging.h"
 #include "platform.h"
 #include "quota.h"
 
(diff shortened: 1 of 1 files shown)


jblomer:
Hm, I don't think we should silently ignore problems with the lower cache layer.  I'd prefer to close the file descriptor given by the upper layer if the lower layer runs into an error.

bbockelm:
What about a debug message into syslog?

While I also dislike ignoring errors, it seems preferable to failing the entire operation and causing an IO error for the user...

jblomer:
A syslog error would be good.  Hm, I guess it really depends on what we want from the lower layer.  We already can mark it as "read-only", I think we should add another option that allows users to ignore errors.

bbockelm:
Ok - @jblomer, please take another look.

I decided to only go the logging route - it felt like it's such an obscure corner, it would only confuse the user to provide this as an configuration option.

cvmfs/cache_tiered.cc
```diff
@@ -187,7 +188,17 @@ int TieredCacheManager::CommitTxn(void *txn) {
...
   }
 
-  return (upper_result < 0) ? upper_result : lower_result;
+  // Transaction is successful as long as the storage was successful
+  // in the upper cache (which may have already handed out a FD via
+  // OpenFromTxn).
+  //
+  // If there wasn't the OpenFromTxn semantics, we could have this
+  // call succeed if *either* transaction was successful.
+  if (!upper_result && lower_result) {
```
jblomer: I find the boolean treatment a bit confusing with file descriptors, please change to `... >= 0 / < 0`.

cvmfs/cache_tiered.cc
```diff
@@ -187,7 +188,17 @@ int TieredCacheManager::CommitTxn(void *txn) {
...
+    LogCvmfs(kLogCache, kLogSyslogWarn | kLogDebug, "Commit of transaction "
```
jblomer: Do we need a rate limit?  If there is a systematic issue with the lower layer, we might easily flood the system log, no?

jblomer:
Expired, the lower layer cache is now explicitly marked as read/write or read/only.

PR #1968, opened 2017-08-04
Minor fixups from debugging external cache plugins.

bbockelm:
For the most part, this just adds a few logging messages to make misconfigurations more obvious.

Two significant changes:
- Make an invalid locator string a fatal error (instead of indefinitely retrying it).
- Launch the external plugin as the `cvmfs` user.

(cherry picked from commit 8310517ba9dcfa6304e68faa5dd09e1e86865d10; backport of #1964)

### DIFF
--- a/cvmfs/cache_extern.cc
+++ b/cvmfs/cache_extern.cc
@@ -185,8 +185,16 @@ int ExternalCacheManager::ConnectLocator(const std::string &locator) {
   } else {
     return -EINVAL;
   }
-  if (result < 0)
+  if (result < 0) {
+    if (errno) {
+      LogCvmfs(kLogCache, kLogDebug | kLogStderr,
+               "Failed to connect to socket: %s", strerror(errno));
+    } else {
+      LogCvmfs(kLogCache, kLogDebug | kLogStderr,
+               "Failed to connect to socket (unknown error)");
+    }
     return -EIO;
+  }
   LogCvmfs(kLogCache, kLogDebug | kLogSyslog,
            "connected to cache plugin at %s", locator.c_str());
   return result;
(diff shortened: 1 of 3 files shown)


jblomer:
Thanks!  This should be opened against the 2.4 branch, I think, not against devel.  PR #1964 can then be closed as we'll merge from 2.4 back into devel.

cvmfs/cache_extern.cc
```diff
@@ -677,7 +691,7 @@ bool ExternalCacheManager::SpawnPlugin(const vector<string> &cmd_line) {
   retval = ManagedExec(cmd_line,
                        preserve_filedes,
                        map_fildes,
-                       false,  // drop_credentials
+                       true,   // drop_credentials
```
jblomer: Like the fuse module, at the moment plugins are effectively run by the cvmfs user but they can regain root persona.  Depending on the security settings of the distribution, regaining root is necessary to have the watchdog successfully connect gdb to the main process for a stack trace.  We could instead also let the plugin decide if it wants to give up on its hidden superpowers.  Or do it automatically in the libcvmfs_cache library after the watchdog has been spawned or when the plugins switches to listen mode.

bbockelm: Is the watchdog really going to connect to this process though?  I thought the extern processes (like the authz one) is outside the purview here.

jblomer: The fuse processes watchdog is out of the loop but a plugin can spawn its own watchdog with `cvmcache_spawn_watchdog()`.

jblomer: @bbockelm I merged all but this change into the 2.4 branch, which will be merged into the devel branch.

jblomer:
Closing, please feel free to reopen if further work on the `ManagedExec` invocation is necessary. Other changes are merged and released.

PR #1969, opened 2017-08-04
Fix off-by-one error for chunk size.

bbockelm:
Previously, a new chunk would be emitted after `N+(buffer size)` bytes when the user requested N bytes for chunk size.  So, if the user wanted a 24MB chunk, CVMFS would create `24MB+16KB` chunks.

This corrects the issue; 24MB chunks are created when 24MB chunks are requested for grafts.

(cherry picked from commit 6ea4d9504e2bf313e9740b9d824b01e5d22958ef; backport of #1962)

### DIFF
--- a/cvmfs/swissknife_graft.cc
+++ b/cvmfs/swissknife_graft.cc
@@ -89,7 +89,7 @@ bool swissknife::CommandGraft::ChecksumFdWithChunks(
     }
 
     // Start a new hash if current one is above threshold
-    if (do_chunk && (*file_size - chunk_offsets->back() > chunk_size_)) {
+    if (do_chunk && (*file_size - chunk_offsets->back() >= chunk_size_)) {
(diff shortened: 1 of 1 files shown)


bbockelm:
@jblomer - did this PR get overlooked?

jblomer:
Apologies, I didn't follow up on this.  I think this can be closed, and #1962 should be opened against the 2.4 branch (it's currently against devel).  As of the 2.4 release branch, we start to merge patches in the release branch and then merge the release branch into the devel branch in order to avoid two PRs for every fix.  I'd like to merge the PR after having merged documentation on grafting.

bbockelm:
Doh.  I had intended to open this against the 2.4 release branch.  Will fix that in a moment.

bbockelm:
Closing in favor of #1982.

PR #1970, opened 2017-08-04
do not do health check on resign

DrDaveD:
See [CVM-1358](https://sft.its.cern.ch/jira/browse/CVM-1358)

Please commit to both devel and cvmfs-2.4 branches.


DrDaveD:
Further testing made me realize that when the repository keys change you always have to do a cvmfs_server resign after a cvmfs_server resign -p, because the fingerprint changes.   So the health check on regular resign also gets in the way.

Meanwhile in my testing I have been able to work around the problem by doing two transaction/abort pairs, but it is ugly and I'm not sure if it will work in all cases one might want do this recovery.  I really don't see any good reason to do a health check in order to resign anything, and I think it is better to just not do it.

PR #1975, opened 2017-08-17
CMake will see Manjaro as an Arch based system

dpmungo:


CMakeLists.txt
```diff
@@ -41,9 +40,9 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   if (EXISTS /etc/debian_version)
     set (DEBIAN TRUE)
   endif (EXISTS /etc/debian_version)
-  if (EXISTS /etc/arch-release)
+  if (EXISTS /etc/arch-release OR EXISTS /ect/manjaro-release)
```
jblomer: This should be `/etc` instead of `/ect`, I assume

CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
-#
```
jblomer: To keep the patch minimal, please restore the empty comment line

dpmungo:
Two tipos in three lines should be a World Record.
Thank you for pointing them out

PR #1976, opened 2017-08-17
Fix some issues discovered during large publications with GW backend

radupopescu:
This addresses [CVM-1359](https://sft.its.cern.ch/jira/browse/CVM-1359).

* Fixes memory error in swissknife_lease.cc causing crash during `cvmfs_server abort`.
* Fixes recursion bug in `CatalogDiffTool` causing catalog corruption when directories were deleted
* Adds more tests cases and checks to the repository service integrations tests.

### DIFF
--- a/test/src/801-repository_services_slow/main
+++ b/test/src/801-repository_services_slow/main
@@ -61,7 +61,7 @@ run_transactions() {
     echo "  Copying the payload into repository"
     rm -v /cvmfs/test.repo.org/new_repository
     if [ ! -d /tmp/linux-4.12.8 ]; then
-        curl -o /tmp/linux.tar.xz https://cdn.kernel.org/pub/linux/kernel/v4.x/linux-4.12.8.tar.xz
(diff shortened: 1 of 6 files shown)


test/src/801-repository_services_slow/main
```diff
@@ -58,19 +58,40 @@ run_transactions() {
     echo "  Starting transaction 1"
     cvmfs_server transaction -e test.repo.org
 
-    echo "  Copying the contents of /usr/bin into repository"
+    echo "  Copying the payload into repository"
     rm -v /cvmfs/test.repo.org/new_repository
-    cp -rv /usr/bin/* /cvmfs/test.repo.org/
+    if [ ! -d /tmp/linux-4.12.8 ]; then
+        curl -o /tmp/linux.tar.xz https://cdn.kernel.org/pub/linux/kernel/v4.x/linux-4.12.8.tar.xz
```
jblomer: We should probably store a copy on our web server

radupopescu: You're right.

PR #1977, opened 2017-08-18
Fix catalog processing and uploading with gateway backend

radupopescu:
Addresses [CVM-1364](https://sft.its.cern.ch/jira/browse/CVM-1364).

Spooler::ProcessCatalog should be used to submit catalogs, instead of Spooler::Upload, to ensure that the catalogs are properly handled as content-addressed files by the receiver process on the gateway.

### DIFF
--- a/cvmfs/catalog_mgr_rw.cc
+++ b/cvmfs/catalog_mgr_rw.cc
@@ -1128,15 +1128,16 @@ WritableCatalogManager::SnapshotCatalogsSerialized(
     // Compress and upload catalog
     shash::Any hash_catalog(spooler_->GetHashAlgorithm(),
                             shash::kSuffixCatalog);
+    const std::string temp_compressed_path = (*i)->database_path() + ".compressed";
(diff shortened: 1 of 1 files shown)


cvmfs/catalog_mgr_rw.cc
```diff
@@ -1128,15 +1128,16 @@ WritableCatalogManager::SnapshotCatalogsSerialized(
     // Compress and upload catalog
     shash::Any hash_catalog(spooler_->GetHashAlgorithm(),
                             shash::kSuffixCatalog);
+    const std::string temp_compressed_path = (*i)->database_path() + ".compressed";
     if (!zlib::CompressPath2Path((*i)->database_path(),
```
jblomer: There might be a CompressPath2Null function in compression.cc, which avoids writing the temporary file.

radupopescu:
@jblomer: I will merge this after the integration tests run this evening.

PR #1980, opened 2017-08-25
WIP: Adding initial support for SciTokens plugins

djw8605:
This adds support for SciTokens, which is an OAuth Bearer
token based authentication scheme.  The bearer token is
retrieved from an external authz helper.

This pull request is not complete, and needs to be reviewed and tested.  The purpose of this pull request is to begin gathering feedback on this feature.  

@bbockelm


cvmfs/authz/authz_curl.cc
```diff
@@ -231,18 +270,33 @@ void AuthzAttachment::LogOpenSSLErrors(const char *top_message) {
...
-  delete p;
 
-  // Calls X509_free on each element, then frees the stack itself
-  sk_X509_pop_free(chain, X509_free);
-  EVP_PKEY_free(pkey);
+  assert(info_data);
+  // How can we tell if info_data is OAuth or SSL?
+  AuthzToken* token = reinterpret_cast<AuthzToken*>(info_data);
+  if (token->type == kTokenSciToken) {
+    delete (char*)token->data;
```
bbockelm: Use `static_cast` instead of C-style casts.

cvmfs/authz/authz_fetch.cc
```diff
@@ -448,6 +448,20 @@ bool AuthzExternalFetcher::ParsePermit(
...
+    unsigned size = strlen(json_token->string_value);
+    binary_msg->permit.token.size = size;
+    if (size > 0) {
+      // The token is passed to the AuthzSessionManager, which takes care of
+      // freeing the memory
+      binary_msg->permit.token.data = smalloc(size);
+      memcpy(binary_msg->permit.token.data, json_token->string_value, size);
+    }
+  }
+
```
bbockelm: 1) At the `debug` level, log that you got a JSON token back.
2) Log an error if no auth was returned.

cvmfs/authz/authz_curl.cc
```diff
@@ -98,6 +98,36 @@ CURLcode AuthzAttachment::CallbackSslCtx(
 }
 
 
+bool AuthzAttachment::ConfigureSciTokenCurl(
```
jblomer: I think it would be better to create another class `AuthzOauthAttachement` derived from `download::CredentialsAttachment`.  The current `AuthzAttachement` should then probably renamed to `AuthzX509Attachement`.  Some slight adjustment between base class and derived classes is probably necessary in order to line things up in the `MountPoint` class.

bbockelm:
Oh - and we need to think about the unit (and/or integration) test strategy here too.

jblomer:
Right, I think we should have an integration test similar to the `600-securecvmfs` test.  A unit test requires that we first make the `DownloadManager` more unit-testable.

cvmfs/authz/authz_curl.cc
djw8605: I looked into the `MountPoint` class.  At that time, when it creates the `AuthzAttachement`, it doesn't know what type of authentication it is going to be returned by the helper.  Also, there are circumstances, such as when we are transitioning LIGO from x509 to Bearer tokens, when the authentication plugin may be able to return both.

jblomer: Hm, that's right.  Thanks for checking.  Please ignore the comment.

jblomer:
@djw8605 @bbockelm Please let me know when you think it's ready for merging.

bbockelm:
@djw8605 - what's the latest on this PR?  Did you guys get to discuss this at the meeting I missed today?

djw8605:
Only briefly discussed this change.  I wanted to wait until I had a good server that could understand bearer tokens to make sure everything works.

djw8605:
I still owe a test similar to `600-securecvmfs`

jblomer:
libcurl got native support for bearer tokens: https://github.com/curl/curl/pull/2102/files
But I guess it doesn't change a lot.

@djw8605 Please let us know when we can take another look at the PR.

djw8605:
@jblomer Integration tests are in and working (in my test environment).  Please review!

jblomer:
Fantastic, thanks a lot!  I'm on holidays for one more week, @radupopescu can you meanwhile take a look and test run the integration test?

test/cloud_testing/platforms/centos7_x86_64_setup.sh
```diff
@@ -55,6 +55,9 @@ install_from_repo tree
 # traffic shaping
 install_from_repo trickle
 
+# Install test dependency for 647
+install_from_repo python-flask          || die "fail (installing python-flask)"
```
radupopescu: I think this test is enabled by default on all platforms (all tests `src/6*` are run).
We have two options: either the python-flask should be added to the other platform setup scripts, or the tests should be blacklisted on other platforms, if it's only relevant to SLC6 and CC7

djw8605: This test is similar to the x509 tests.  I believe that test is only enabled on slc6.  I would like the 647 test to run only on slc6 and CC7.

Where do I specify the blacklist?

radupopescu: Each integration platform has a setup and a test script. In the test scripts the `run.sh` script runs the various test suites. For example, in the CC7 test script, the server integration scripts are started as follows:
```
./run.sh $SERVER_TEST_LOGFILE -o ${SERVER_TEST_LOGFILE}${XUNIT_OUTPUT_SUFFIX} \
                               -x src/518-hardlinkstresstest                   \
                                  src/585-xattrs                               \
                                  src/600-securecvmfs                          \
                                  src/602-libcvmfs                             \
                                  src/628-pythonwrappedcvmfsserver             \
                                  src/7*                                       \
                                  src/8*                                       \
                               || retval=1
```
Tests are blacklisted by listing them after `-x` and before `--` in the invocation of `run.sh`

The test should be blacklisted on the following platforms:
```
fedora_test.sh
slc5_i386_test.sh
slc5_x86_64_test.sh
ubuntu_s3_test.sh
ubuntu_test.sh
```

Thanks!

radupopescu:
@djw8605 I just wanted to confirm that I'm running the tests, it's just taking a bit longer than expected, since we've had some failures in the CI infrastructure.

djw8605:
They are failing for different reasons...

I see that in the cc7, httpd service isn't coming up after writing the configuration.  Is it possible that mod_wsgi isn't installed?  Are you able to get logs from the test machine after running the test?  I would be curious as to the logs for the httpd service.

radupopescu:
I checked that the packages are installed. They are, which is expected, since the machines are provisioned from scratch, according to the platform setup scripts, at the beginning of each integration run.

Looking at the HTTPS service status on CentOS 7, I see this:
```
[centos@server-r-q8unjk4i-0 ~]$ sudo systemctl status httpd
● httpd.service - The Apache HTTP Server
   Loaded: loaded (/usr/lib/systemd/system/httpd.service; disabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Wed 2018-08-01 07:56:59 UTC; 7min ago
     Docs: man:httpd(8)
           man:apachectl(8)
  Process: 18498 ExecStop=/bin/kill -WINCH ${MAINPID} (code=exited, status=1/FAILURE)
  Process: 18497 ExecStart=/usr/sbin/httpd $OPTIONS -DFOREGROUND (code=exited, status=1/FAILURE)
 Main PID: 18497 (code=exited, status=1/FAILURE)

Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch systemd[1]: Starting The Apache HTTP Server...
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch httpd[18497]: AH00526: Syntax error on line 5 of /etc/httpd/conf.d/t...nf:
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch httpd[18497]: Invalid command 'SSLEngine', perhaps misspelled or def...ion
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch systemd[1]: httpd.service: main process exited, code=exited, status...LURE
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch kill[18498]: kill: cannot find process ""
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch systemd[1]: httpd.service: control process exited, code=exited status=1
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch systemd[1]: Failed to start The Apache HTTP Server.
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch systemd[1]: Unit httpd.service entered failed state.
Aug 01 07:56:59 server-r-q8unjk4i-0.cern.ch systemd[1]: httpd.service failed.
```
The test-specific HTTPD config file has been deleted during the test cleanup phase, so I can't retrieve it.

Maybe this message helps...

djw8605:
Interesting, SSLEngine is a command defined by mod_ssl.  Can you make sure mod_ssl is installed (it should be, since it's also required by test 600-securecvmfs).  I wonder if the module isn't being loaded?  It has the same command as the 600-securecvmfs.

radupopescu:
@djw8605 @bbockelm 
Indeed `mod_ssl` isn't installed. I think this package is missing, at least on CentOS7, because test `600-securecvmfs` is blacklisted there. Shouldn't it be enabled for both CentOS7 and SLC6?

radupopescu:
I manually ran the test after installing `mod_ssl`. Another error when starting HTTPD due to missing `hostcert-cvmfs.pem`. Do you know where this file is supposed to come from.

djw8605:
That's right, that was an issue I had with that as well.

In CC7, apache (httpd) is configured to have a private temp enabled, so the outside /tmp (where the test files are located, as well as the test certs) cannot be accessed by httpd.

Currently, the test bind mounts `/etc/grid-security` into a directory `/tmp`.  Let me change the test to just put files in that directory `/etc/grid-security`.  It will be less clean, but necessary.

djw8605:
@radupopescu I pushed the change to the test.

bbockelm:
Hm - could we just use a different bind-mount?  Override the systemd private tmp setting?

I've previously just run one-at-a-time integration tests on a local host; really liked the fact this one didn't clobber the host's settings.

djw8605:
To overwrite the systemd private tmp setting, the test would:

1. See if systemd is managing httpd (check if /usr/lib/systemd/system/httpd.service exists?)
2. Write out a file in `/etc/systemd/system/httpd.service.override` to set privatetmp to false.
3. Make sure to clean up the override file.

For the different bind mount, the test directory is in /tmp... I'm not sure where else we could write the cert files to bind mount.  Any suggestions?

radupopescu:
I think it's good to be able to run integration tests without clobbering the host's configs, but I'm not sure we are always taking steps to avoid that. Overriding the systemd service file could be an option, but that may lead to other issues which need to be addressed.

Meanwhile, I submitted another integration run with your latest changes, @djw8605.

bbockelm:
Hi Derek - if all we need to do is those 3 steps, it seems a lot better than clobbering the hostcert.

djw8605:
@bbockelm @radupopescu I have updated the commit with the plan as I laid out in the comments.  I override the systemd configuration of httpd, and clean it up after the test.

radupopescu:
A small update:

We're almost there! I ran the integration tests after the last set of changes you made, @djw8605.
The test passes on CC7, however fails on SLC6. Could it be due to the lack of systemd on SLC6?
I attach the log of the failing test.
[test_server.log.slc6.txt](https://github.com/cvmfs/cvmfs/files/2257146/test_server.log.slc6.txt)

djw8605:
I wanted to let you know that I am looking into this on EL6.

@radupopescu, if you have a minute, could you look and / or send the httpd logs from the sl6 host?

radupopescu:
Great, thanks for looking into it @djw8605 !

I attach the log file from the SLC6 run.
[test_server.log.txt](https://github.com/cvmfs/cvmfs/files/2263665/test_server.log.txt)

djw8605:
Hi @radupopescu.  I built an el6 host and fixed the test.  There was only the 1 issue, b4ffd1c.

But, https://github.com/cvmfs/cvmfs/commit/d1012e5c078b86a81ba04f4a29e085f699e85198 also affected the test, which was fixed just a few weeks ago.  I merged in devel branch into this pull request, so you may have to update.

I also re-ran the test on EL7, and it still works with these changes.

radupopescu:
@djw8605 Sorry for the delay and thanks for the fixes!

I'll run a quick integration run and if all is well we should merge this.

radupopescu:
Integration test is passing on both CentOS 6 and CentOS 7.
Thanks for the good work, @djw8605 !

PR #1981, opened 2017-08-28
Restart authz_helper process after a cool-off period.

bbockelm:
Although I haven't quite figured out why, we have a few crashing `cvmfs_x509_helper` processes on one of the OSG sites.

In order to avoid permanent black-hole nodes, this allows the CVMFS authz helper to restart after `kChildTimeout` seconds (I considered an immediate restart, but I am worried about continuously churning processes).

### DIFF
--- a/cvmfs/authz/authz_fetch.cc
+++ b/cvmfs/authz/authz_fetch.cc
@@ -46,6 +46,7 @@ AuthzExternalFetcher::AuthzExternalFetcher(
   , pid_(-1)
   , fail_state_(false)
   , options_manager_(options_manager)
+  , next_start_(-1)
 {
   InitLock();
 }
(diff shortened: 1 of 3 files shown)


bbockelm:
@jblomer - added an integration test for this functionality as part of `632-authzhelpers`.

In general - any advice on how one might be able to grab a back trace (or core dump) of the authz helper?  I can probably wrap things in a shell script, but it'd be easier to tap into a preexisting infrastructure.

jblomer:
@bbockelm: good question... cvmfs has the `Watchdog` class but we are probably not able to use it between `fork()` and `execve()` because of the unknown state of the cvmfs2 process at this point.  We might be able to cvmfs2 itself as a trampoline process, so that we can do

    fork()
    execve() --> cvmfs2 with special parameters
    spawn off watchdog
    execve into the actual helper

PR #1986, opened 2017-09-01
WIP add POSIX cache plugin

MatteoDome:
Adds a POSIX cache plugin.

Known issue: breaks with warm cache.


jblomer:
Expired, see CVM-1823 for further development

PR #1987, opened 2017-09-12
Trivial patch to display file sizes in swissknife diff command

bcouturi:
Trivial patch to display the size of the entries being added or removed as well as the names.

### DIFF
--- a/cvmfs/swissknife.h
+++ b/cvmfs/swissknife.h
@@ -72,7 +72,7 @@ typedef std::map<char, std::string *> ArgumentList;
 
 class Command : public ServerTool {
  public:
-  // TODO comment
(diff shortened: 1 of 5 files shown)


radupopescu:
Hi @bcouturi,

Thanks for the patch, it looks good.

The checks are failing because the source code is not properly formatted. We have the `ci/run_cpplint.sh` script that tells you what lines are not compliant.

For the record, changes to the `swissknife_diff` output will also break integration test 641, but I can fix it in another PR, since running the integration tests is not straightforward.

bcouturi:
Hi @radupopescu, many thanks!
I have just fixed the cpplint warnings.

PR #1993, opened 2017-09-20
Fix gateway backend submission errors 

radupopescu:
This addresses [CVM-1375](https://sft.its.cern.ch/jira/browse/CVM-1375).

In the process of debugging this JIRA issue, a series of race conditions and swallowed errors were fixed in the part of the code related to the gateway backend, while logging in this parts was improved.

### DIFF
--- a/cvmfs/session_context.cc
+++ b/cvmfs/session_context.cc
@@ -356,7 +356,7 @@ bool SessionContext::DoUpload(const SessionContext::UploadJob* job) {
 
   const bool ok = (reply == "{\"status\":\"ok\"}");
   if (!ok) {
-    LogCvmfs(kLogUploadGateway, kLogStdout,
+    LogCvmfs(kLogUploadGateway, kLogStderr,
              "SessionContext - curl_easy_perform reply: %s", reply.c_str());
   }
 
--- a/cvmfs/session_context.cc
+++ b/cvmfs/session_context.cc
@@ -63,7 +63,7 @@ size_t RecvCB(void* buffer, size_t size, size_t nmemb, void* userp) {
 }
 
 SessionContextBase::SessionContextBase()
-    : upload_results_(1000, 1000),
+    : upload_results_(1000000, 1000000),
       api_url_(),
       session_token_(),
       key_id_(),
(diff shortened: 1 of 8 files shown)


cvmfs/session_context.cc
```diff
@@ -63,7 +63,7 @@ size_t RecvCB(void* buffer, size_t size, size_t nmemb, void* userp) {
 }
 
 SessionContextBase::SessionContextBase()
-    : upload_results_(1000, 1000),
+    : upload_results_(1000000, 1000000),
```
jblomer: These are the increased timeouts, right?  I think we should declare a static constant unsigned in the class for the value.

cvmfs/session_context.cc
```diff
@@ -337,7 +342,6 @@ bool SessionContext::DoUpload(const SessionContext::UploadJob* job) {
   curl_easy_setopt(h_curl, CURLOPT_USERAGENT, "cvmfs/" VERSION);
   curl_easy_setopt(h_curl, CURLOPT_MAXREDIRS, 50L);
   curl_easy_setopt(h_curl, CURLOPT_CUSTOMREQUEST, "POST");
-  curl_easy_setopt(h_curl, CURLOPT_TCP_KEEPALIVE, 1L);
```
jblomer: That's interesting!  I have to remember to ask you what the effect of this change was to the system.

radupopescu: This has no effect visible effect in the end, I just wanted to go back to the default. The problem was coming from the Erlang Cowboy server, who had a timeout by default which was making my requests fail.

Now, about CURLOPT_TCP_KEEPALIVE, this is a bit misleading, in my opinion. This only controls sending TCP keepalive probes not the connection keepalive mechanism. The default is 0 (disabled) which is what I want.

cvmfs/session_context.cc
radupopescu: These aren't timeouts, they are max queue sizes, but in the end this change isn't really needed, since we can use large object packs without problem. The number of upload results is in the 10s in this case.

This change dates from when I was debugging the problem, I should have removed it maybe.

Anyway, I can make it a const somewhere.

PR #1994, opened 2017-09-20
Check XFS ftype parameter in cvmfs_server_util.sh

radupopescu:
Addresses [CVM-1385](https://sft.its.cern.ch/jira/browse/CVM-1385).

Add check for the value of the ftype parameter, when /var/spool/cvmfs is
located on an XFS partition. "ftype=1" is supported. If "ftype=0" error
out and warn user.

### DIFF
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -502,7 +502,7 @@ check_overlayfs_version() {
         return 0
       fi
       if [ "x$scratch_fstype" = "xxfs" ] ; then
-        if [ "x$(xfs_info /var/spool/cvmfs | grep ftype=1)" != "x" ] ; then
+        if [ "x$(xfs_info /var/spool/cvmfs 2>/dev/null | grep ftype=1)" != "x" ] ; then
           return 0
         else
           echo "XFS with ftype=0 is not supported for /var/spool/cvmfs. XFS with ftype=1 is required"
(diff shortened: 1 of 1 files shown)


cvmfs/server/cvmfs_server_util.sh
```diff
@@ -501,7 +501,15 @@ check_overlayfs_version() {
       if [ "x$scratch_fstype" = "xext3" ] || [ "x$scratch_fstype" = "xext4" ] ; then
         return 0
       fi
-      echo "overlayfs scratch /var/spool/cvmfs is type $scratch_fstype, but ext3 or ext4 required"
+      if [ "x$scratch_fstype" = "xxfs" ] ; then
+        if [ "x$(xfs_info /var/spool/cvmfs | grep ftype=1)" != "x" ] ; then
```
jblomer: Can we expect `xfs_info` to be present on the system?  Otherwise we should either use /proc/mounts or add the corresponding package dependency.

radupopescu: I don't know how to get this information from /proc/mounts (I don't see a mention of ftype).

The xfs_info command is in xfsprogs, I checked with a fresh CC7 image on the CERN Openstack and the command is available by default.

jblomer: Ok, thanks for checking.  Looking again at the `if` clause, I think it will fail correctly if `xfs_info` does not exist, anyway.

radupopescu: I checked, too. If the command isn't found, the proper code is returned.

DrDaveD: Sure but it would print an extraneous error message, right?  Should stderr be redirected to /dev/null?

jblomer:
Shall we make this against the cvmfs-2.4 branch for the 2.4.2 patch release?

PR #1998, opened 2017-09-27
Enhanced testsuite for the CatalogMergeTool class

radupopescu:
Addresses [CVM-1290](https://sft.its.cern.ch/jira/browse/CVM-1290).

* Adds unit tests for the CatalogMergeTool class
* Adds RapidCheck (C++ implementation of QuickCheck) as an optional depedency w/ example 
* Adds QuickCheck randomized test for the CatalogMergeTool


cvmfs/util/posix.cc
```diff
@@ -37,13 +37,16 @@
...
+#include <set>
 #include <string>
+#include <vector>
 
 #include "fs_traversal.h"
 #include "logging.h"
 #include "platform.h"
 
-using namespace std;  // NOLINT
+//using namespace std;  // NOLINT
```
jblomer: Interesting.  So the ``std`` namespace caused trouble?

radupopescu: If I remember correctly it broke compilation on my Macbook, for some reason.

Given the short function names with hight risk of collision, maybe we should try to avoid "using namespace std;"

jblomer:
Very cool!

I'd like to have a closer look, for which I'd need a little moment.

test/quickcheck/main.cc
```diff
@@ -0,0 +1,23 @@
...
+
+#include <gtest/gtest.h>
+
+#include <cassert>
+
+#include "env.h"
+#include "monitor.h"
+
+int main(int argc, char **argv) {
+  Watchdog *watchdog = Watchdog::Create("/tmp/stacktrace.cvmfs_unittests");
```
jblomer: We might want to store the stack trace under `/tmp/stacktrace.cvmfs_qc` or similar

test/common/catalog_test_tools.h
```diff
@@ -0,0 +1,122 @@
...
+  bool AddDir(const std::string& name, const std::string& parent);
+  bool RmDir(const std::string& name, const std::string& parent);
+  bool HasDir(const std::string& name) const;
+
+  ItemList items_;
+  std::set<std::string> dirs_;
+};
+
+
+class CatalogTestTool : public ServerTool {
```
jblomer: I think a comment on the classes purpose would be helpful.

test/quickcheck/main.cc
```diff
@@ -0,0 +1,23 @@
...
+  assert(watchdog != NULL);
+  // watchdog->Spawn();
+  CvmfsEnvironment* env = new CvmfsEnvironment(argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  ::testing::FLAGS_gtest_death_test_style = "threadsafe";
+  ::testing::AddGlobalTestEnvironment(env);
+  int result = RUN_ALL_TESTS();
```
jblomer: Shouldn't this call something like `rc::check()`?  Or is this invocation part of RapidCheck's gtest module?

test/quickcheck/qc_catalog_merge_tool.cc
```diff
@@ -0,0 +1,204 @@
...
+          while (idx < item_index) {
+            ++it;
+            ++idx;
+          }
+          out.RemoveItemRec(it->first);
+        }
+      }
+      break;
+      case ChangeType::ModifyFile:
+        //LogCvmfs(kLogCvmfs, kLogStdout, "ModifyFile");
```
jblomer: Perhaps add `TODO implement`

test/quickcheck/qc_catalog_merge_tool.cc
```diff
@@ -0,0 +1,204 @@
...
+  params.min_weight = 0;
+
+  return params;
+}
+
+}  // namespace
+
+class T_CatalogMergeTool : public ::testing::Test {};
+
+RC_GTEST_FIXTURE_PROP(T_CatalogMergeTool, InOut, ()) {
```
jblomer: A comment would we good with a brief high-level explanation about what is being tested.

test/quickcheck/qc_catalog_merge_tool.cc
```diff
@@ -0,0 +1,204 @@
...
+#include <rapidcheck/gtest.h>
+
+#include "catalog_test_tools.h"
+#include "receiver/catalog_merge_tool.h"
+#include "receiver/params.h"
+#include "testutil.h"
+#include "xattr.h"
+
+namespace {
+const char* hashes[] = {"b026324c6904b2a9cb4b88d6d61c81d1000000",
```
jblomer: These are not needed, are they?

test/quickcheck/main.cc
radupopescu: RapidCheck doesn't have any test runner or test harness. `rc::check` is used to run individual test cases.

If it's integrated with GoogleTest, then the test runner part is specific to GTest and we define
our QuickCheck cases using `RC_GTEST_FIXTURE_PROP`.

I don't know, maybe I misunderstood your question.

jblomer: Thanks, that answers my question.

PR #2000, opened 2017-10-02
Enforce explicit ttl setting in root catalog

jblomer:


DrDaveD:
This appears to be only updating the TTL if it was already explicitly set.  So if the default was previously being taken (which is the vast majority of cases), it does nothing. Did I read it right? That's not what I thought you said you were planning to do in [CVM-1388](https://sft.its.cern.ch/jira/browse/CVM-1388).  Did you change your mind?

jblomer:
What it should do: write the provided TTL into the catalog if either it deviates from what's currently in the catalog or if nothing has been written yet into the catalog.  Does it something different?

DrDaveD:
I'm sorry, I missed the "!" on the call to HasExplicitTTL().  It looks fine.

PR #2002, opened 2017-10-04
Add support for cvmcache_get_session()

jblomer:


jblomer:
@bbockelm Please have a look if you can use the `cvmcache_get_session()` API for shared data caches

PR #2009, opened 2017-10-23
Remove a nested catalog's subtree in constant time

radupopescu:
Addresses [CVM-1405](https://sft.its.cern.ch/jira/browse/CVM-1405).

Performance optimization.

Adds a parameter to the `WritableCatalogManager::RemoveNestedCatalog` function to allow removing the nested catalog including its contents. By default, the behaviour of the function is unchanged (i.e. the nested catalog is removed and its contents are merged into its parent).

For the moment, the new functionality is only be used on the repository gateway, in the `CatalogMergeTool` class.

Nested catalog handling in  `CatalogMergeTool` is already tested in the integration test `802-repository_services_nested_catalogs`.

### DIFF
--- a/cvmfs/catalog_rw.cc
+++ b/cvmfs/catalog_rw.cc
@@ -647,6 +647,26 @@ void WritableCatalog::MergeIntoParent() {
 }
 
 
+void WritableCatalog::RemoveFromParent() {
+  assert(!IsRoot() && HasParent());
+  WritableCatalog *parent = GetWritableParent();
+
+  // Remove the nested catalog reference for this nested catalog.
+  // From now on this catalog will be dangling!
+  Catalog* child_catalog;
+  parent->RemoveNestedCatalog(this->mountpoint().ToString(), &child_catalog);
+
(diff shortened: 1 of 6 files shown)


cvmfs/catalog_rw.cc
```diff
@@ -647,6 +647,26 @@ void WritableCatalog::MergeIntoParent() {
...
+  WritableCatalog *parent = GetWritableParent();
+
+  // Remove the nested catalog reference for this nested catalog.
+  // From now on this catalog will be dangling!
+  Catalog* child_catalog;
+  parent->RemoveNestedCatalog(this->mountpoint().ToString(), &child_catalog);
+
+  const Counters& child_counters = child_catalog->GetCounters();
+
+  parent->delta_counters_.subtree.directories = -1;
```
jblomer: The `parent->delta_counters_` could already have some non-zero value from previous operations which would get overwritten here, no?

radupopescu: Most likely. I've been trying to understand how these counters work, by reading the code.

PR #2013, opened 2017-11-07
Turn cvmfs-config into Debian virtual package, default cvmfs-config-default

DrDaveD:
https://sft.its.cern.ch/jira/browse/CVM-1420

I have tested this on my Debian machine with cvmfs-config-default and cvmfs-config-none in a local apt repository, works as expected.

### DIFF
--- a/packaging/debian/config-default/control
+++ b/packaging/debian/config-default/control
@@ -12,6 +12,7 @@ Replaces: cvmfs-keys
 Conflicts: cvmfs-keys
 Breaks: cvmfs-keys, cvmfs (<< 2.1.20), cvmfs-server (<< 2.1.20)
 Provides: cvmfs-config
+Conflicts: cvmfs-config
 Description: CernVM File System Default Configuration
  Defaulf configuration and public keys for mounting common CernVM-FS
  repositories.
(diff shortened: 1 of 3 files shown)


jblomer:
Are you sure about the ``Conflicts:`` clause?  Wouldn't that mean that no two config packages can be installed concurrently?

Please resubmit against the 2.4 branch.

DrDaveD:
I changed the base; I branched it off of cvmfs-2.4 in the first place, it was just a mistake on the PR submission.  

Yes the Conflicts clause prevents two `cvmfs-config` packages from being installed concurrently.  I thought that was what we wanted.  I still think that's what we want; I think that the name `cvmfs-config` should be reserved for a base configuration package which does conflict with other base configuration packages, even if they don't install the same files.  For example, they probably would all (except for `cvmfs-config-none`) define a config repository.  Of course people can create other cvmfs-config-* packages which do not Provide & Conflict with `cvmfs-config`.  

Even though `cvmfs-config-none` does not set a config repository, its intention is to allow people to not install any base config repository, so I think it should also Provide & Conflict with `cvmfs-config`.   In fact when I was testing this PR I found that I had accidentally installed an old version of `cvmfs-config-none` that did not have the Conflict, and I was having trouble figuring out why it wasn't installing `cvmfs-config-default`.   

By having the Conflicts statement there, when somebody does `apt-get install cvmfs-config` it says that `cvmfs-config` is a virtual package and lists all the possible packages to choose from.

PR #2026, opened 2017-11-29
Have geoapi try $REMOTE_ADDR if $HTTP_X_FORWARDED_FOR has no geoinfo

DrDaveD:
See [CVM-1442](﻿﻿https://sft.its.cern.ch/jira/browse/CVM-1442)


jblomer:
@DrDaveD should I wait for the discussion in CVM-1442 to conclude before merging?

DrDaveD:
I don't think it needs to wait.  Even if it isn't exactly the same problem as what Ryan is having, it is a fix for the problem I was able to demonstrate.  Hopefully he can confirm before we want to freeze 2.4.4 though.

PR #2027, opened 2017-11-30
React to a change of DNS server (macOS)

radupopescu:
Addresses [CVM-496](https://sft.its.cern.ch/jira/browse/CVM-496).

On macOS, when roaming across wireless networks, a change of DNS server was not detected, leading to I/O errors.

This is addressed here by watching the "/etc/resolv.conf" file for changes in a background thread. When the file is modified the new nameserver address is read and given to all `DownloadManager` instances.

This solution is limited to macOS, for the moment, as for Linux there are alternative solutions available (see JIRA discussion).

The feature is controlled by the `CVMFS_DNS_ROAMING` repository parameter - by default it is ON.


cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+namespace file_watcher {
+
+FileWatcherKqueue::FileWatcherKqueue() {}
+
+FileWatcherKqueue::~FileWatcherKqueue() {}
+
+bool FileWatcherKqueue::RunEventLoop(const FileWatcher::HandlerMap& handlers,
+                                     int control_pipe) {
+  kq_ = kqueue();
+  if (kq_ == -1) {
```
jblomer: We can probably assert that this must work.

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+
+namespace file_watcher {
+
+FileWatcherKqueue::FileWatcherKqueue() {}
+
+FileWatcherKqueue::~FileWatcherKqueue() {}
+
+bool FileWatcherKqueue::RunEventLoop(const FileWatcher::HandlerMap& handlers,
+                                     int control_pipe) {
+  kq_ = kqueue();
```
jblomer: I think the file descriptor is never closed.

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+  if (kq_ == -1) {
+    LogCvmfs(kLogCvmfs, kLogDebug, "FileWatcherKqueue - Cannot create kqueue.");
+    return false;
+  }
+
+  // Control pipe sending the stop event.
+  struct kevent watch_event;
+  EV_SET(&watch_event, control_pipe, EVFILT_READ, EV_ADD | EV_ENABLE | EV_CLEAR,
```
jblomer: I guess we can assert on a positive result here, too.

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+    RegisterFilter(it->first, it->second);
+  }
+
+  bool stop = false;
+  while (!stop) {
+    std::vector<struct kevent> triggered_events(handlers.size() + 1);
+    int nev = kevent(kq_, NULL, 0, &triggered_events[0],
+                     triggered_events.size(), NULL);
+    if (nev == -1) {
+      LogCvmfs(kLogCvmfs, kLogDebug,
```
jblomer: This could be logged to `kLogSyslogErr` together with the `errno`.

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+      return false;
+    }
+    if (nev == 0) {
+      continue;
+    }
+    for (int i = 0; i < nev; ++i) {
+      struct kevent& current_ev = triggered_events[i];
+      if (current_ev.ident == static_cast<uint64_t>(control_pipe)) {
+        char buffer[1];
+        read(control_pipe, &buffer, 1);
```
jblomer: --> `ReadPipe`

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+        LogCvmfs(kLogCvmfs, kLogDebug, "FileWatcherKqueue - Stopping.\n");
+        stop = true;
+        continue;
+      } else {
```
jblomer: The `else` branch could be dropped, making the control flow a bit easier to follow.

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+      }
+    }
+  }
+  return true;
+}
+
+void FileWatcherKqueue::RemoveFilter(int fd) {
+  struct kevent remove_event;
+  EV_SET(&remove_event, fd, EVFILT_VNODE, EV_DELETE, NULL, 0, 0);
+  if (kevent(kq_, &remove_event, 1, NULL, 0, NULL) == -1) {
```
jblomer: assert?

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+               NOTE_RENAME | NOTE_REVOKE,
+           0, 0);
+
+    if (kevent(kq_, &watch_event, 1, NULL, 0, NULL) == -1) {
+      LogCvmfs(kLogCvmfs, kLogDebug,
+               "FileWatcherKqueue - Could not register event with kqueue.");
+      return;
+    }
+
+    watch_records_[fd] = WatchRecord(file_path, handler);
```
jblomer: I'm not sure, are the established filters and there file descriptors cleaned up on destruction?

cvmfs/file_watcher_kqueue.cc
```diff
@@ -0,0 +1,164 @@
...
+      throttle.Throttle();
+      continue;
+    }
+    struct kevent watch_event;
+    EV_SET(&watch_event, fd, EVFILT_VNODE, EV_ADD | EV_ENABLE | EV_CLEAR,
+           NOTE_DELETE | NOTE_WRITE | NOTE_EXTEND | NOTE_ATTRIB | NOTE_LINK |
+               NOTE_RENAME | NOTE_REVOKE,
+           0, 0);
+
+    if (kevent(kq_, &watch_event, 1, NULL, 0, NULL) == -1) {
```
jblomer: assert?

cvmfs/resolv_conf_event_handler.cc
```diff
@@ -0,0 +1,85 @@
...
+#include "resolv_conf_event_handler.h"
+
+#include <string>
+#include <vector>
+
+#include "backoff.h"
+#include "logging.h"
+#include "util/string.h"
+
+/*
```
jblomer: `/**` for the Doxygen syntax

cvmfs/resolv_conf_event_handler.cc
```diff
@@ -0,0 +1,85 @@
...
+ * Returns any  nameserver IP addresses from file
+ *
+ * Reads a file, line by line. Returns the IP addresses corresponding
+ * to nameservers (i.e. from lines "nameserver <IP_ADDRESS>")
+ */
+bool GetDnsAddresses(const std::string& resolv_file,
```
jblomer: static bool

cvmfs/resolv_conf_event_handler.cc
```diff
@@ -0,0 +1,85 @@
...
+    : download_manager_(download_manager)
+    , external_download_manager_(external_download_manager) {}
+
+ResolvConfEventHandler::~ResolvConfEventHandler() {}
+
+bool ResolvConfEventHandler::Handle(const std::string& file_path,
+                                    file_watcher::Event /*event*/,
+                                    bool* clear_handler) {
+  std::vector<std::string> addresses;
+  GetDnsAddresses(file_path, &addresses, NULL);
```
jblomer: We might need a bit more logic here.  Users can set whether they want to prefer IPv4 or IPv6 through `CVMFS_IPFAMILY_PREFER`.  We should ask the download manager and, depending on the result, if there are IPv4 and IPv6 nameservers select accordingly.

cvmfs/mountpoint.cc
```diff
@@ -1233,6 +1237,27 @@ bool MountPoint::CreateDownloadManagers() {
   return SetupExternalDownloadMgr(false);
 }
 
+bool MountPoint::CreateResolvConfWatcher() {
+  if (!options_mgr_->IsDefined("CVMFS_DNS_ROAMING") ||
```
jblomer: Let's make the default off, no?  We can explicitly enable it then in mount/default.d/10-osx.conf

test/unittests/t_file_watcher_kqueue.cc
```diff
@@ -0,0 +1,69 @@
...
+};
+
+class T_FileWatcherKqueue : public ::testing::Test {
+ protected:
+  void SetUp() { counters_.clear(); }
+
+  Counters counters_;
+};
+
+TEST_F(T_FileWatcherKqueue, ModifyThenDelete) {
```
jblomer: It would be good if we can avoid the sleeping.  The first can perhaps be removed entirely?  The second and third one can become an endless loops on `counters.find()`?

cvmfs/file_watcher.h
```diff
@@ -0,0 +1,75 @@
...
+
+#ifndef CVMFS_FILE_WATCHER_H_
+#define CVMFS_FILE_WATCHER_H_
+
+#include <pthread.h>
+
+#include <map>
+#include <string>
+
+namespace file_watcher {
```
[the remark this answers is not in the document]
radupopescu: I don't know. I find namespaces useful for organizing things, I would probably use them more, not less. It avoids having very long class and function names and you can even have "using namespace ***" clauses in scoped blocks.

Having a top level `cvmfs` namespace doesn't stop you from having nested namespaces per system.

cvmfs/file_watcher.h
```diff
@@ -0,0 +1,75 @@
...
+   *                        handler remove from the loop after this call
+   *
+   * Setting the clear_handler parameter to false means that the FileWatcher
+   * object will attempt to re-register the handler for the same file name
+   * in the case a file was delete - useful for continuously watching a file
+   * which may be deleted and recreated
+   */
+  virtual bool Handle(const std::string& file_path,
+                      Event event,
+                      bool* clear_handler) = 0;
```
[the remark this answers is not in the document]
radupopescu: Since this is part of the public API of the feature, I wanted to have it like this:
* return value signals a failure in the handler
* clear_handler output variable is for deciding if the file watcher should clear or keep the handler.

cvmfs/file_watcher_kqueue.cc
radupopescu: I think EV_SET returns void, not sure what to assert.

cvmfs/file_watcher.cc
```diff
@@ -0,0 +1,86 @@
...
+  // Use the control pipe to signal readiness to the main thread,
+  // before continuing with the event loop.
+  write(watcher->control_pipe_[1], "s", 1);
+
+  if (!watcher->RunEventLoop(watcher->handler_map_,
+                             watcher->control_pipe_[0])) {
+    LogCvmfs(kLogCvmfs, kLogDebug, "Error running event loop.");
+  }
+
+  pthread_exit(NULL);
```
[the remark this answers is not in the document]
radupopescu: I'm used to terminating pthreads with this call. I think this only makes a difference if one registers a cleanup handler with `pthread_cleanup_push`. Apparently, the handlers are only called with `pthread_exit`.

cvmfs/file_watcher_kqueue.cc
radupopescu: They are now at file_watcher_kqueue.cc:115

cvmfs/file_watcher.h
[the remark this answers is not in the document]
jblomer: I see.  Let's add a comment then on the meaning of the return value.

radupopescu:
Once the Travis-CI report is green, this PR can be merged.

All the comments have been addressed, only 2 issues are left for later:

1. Linux version based on inotify
2. Integration test (make it when either the Linux version becomes available or when the integration tests start running on macOS).

PR #2028, opened 2017-12-05
Add support for AWSv4 S3 authorization

jblomer:
If `CVMFS_S3_REGION` is specified in the S3 configuration, use AWSv4 style authorization requests.

This PR also disables multi-bucket support from the S3 backend.  Distributing contents over multiple buckets and accounts was a workaround for one specific storage product which is not sold anymore.


radupopescu:
Great that multi-bucket support is removed!

DrDaveD:
Apparently this was closed and the branch deleted before the travis check was finished?

jblomer:
The last two commits where documentation/add-on updates.  The check before succeeded.

PR #2043, opened 2017-12-18
FileWatcher background thread should be spawned after process fork

radupopescu:
Related to [CVM-496](https://sft.its.cern.ch/jira/browse/CVM-496).

This addresses the issue where the background thread of the `FileWatcher` object responsible for reacting to changes in the `/etc/resolv.conf` file was spawned too early, before the main client process forks to background.

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1864,6 +1864,7 @@ static void Spawn() {
 
   cvmfs::mount_point_->download_mgr()->Spawn();
   cvmfs::mount_point_->external_download_mgr()->Spawn();
+  cvmfs::mount_point_->resolv_conf_watcher()->Spawn();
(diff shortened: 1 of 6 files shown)


radupopescu:
I checked this with `CVMFS_HTTP_PROXY="http://ca-proxy.cern.ch:3128;DIRECT"`. It works as expected - the download manager will try a list of different IPs for the proxy, before switching to the DIRECT one. It will also try different stratum 1 servers, but eventually the request will succeed using the new DNS.

jblomer:
That's interesting, I wonder why it tries different S1 servers. I would have expected that the first S1 server with a DIRECT proxy works. Did the DNS settings change before the S1 switches started?

radupopescu:
I don't know, I tried it again and this time it just loops over proxy IPs, then switches to DIRECT and the request succeeds.

Do you think the previous loop over S1 is related to the DNS roaming?

PR #2051, opened 2018-02-02
Add doc about dependencies necessary to build the CVMFS

siscia:


jblomer:
Thanks!  It's good to mention the dependencies in the INSTALL file.  We have a few duplicates of listing dependency already.  In order to avoid creating another list that we need to update, I'd suggest to point from the INSTALL file to the deb and rpm package sources (under packaging).

PR #2053, opened 2018-02-06
Splitting sync union

siscia:
I splitted the the sync_union implementation into different files.

Then I made it compile and change the dependences of swissknife_sync.

Moreover, I remove useless (I believe) includes (shouldn't they big catch at compile time?)

Finally I apply the standard google formatter, overall I like the formatter but somewhere, made questionable choices. (Eg line: 682-685 of cvmfs/swissknife_sync.cc)

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -244,6 +244,8 @@ set (CVMFS_SWISSKNIFE_SOURCES
   sync_item.cc
   sync_mediator.cc
   sync_union.cc
+  sync_union_overlayfs.cc
+  sync_union_aufs.cc
   upload.cc
   upload_facility.cc
   upload_gateway.cc
--- a/cvmfs/swissknife_sync.cc
+++ b/cvmfs/swissknife_sync.cc
@@ -45,6 +45,8 @@
 #include "sanitizer.h"
 #include "sync_mediator.h"
 #include "sync_union.h"
+#include "sync_union_aufs.h"
+#include "sync_union_overlayfs.h"
 #include "util/string.h"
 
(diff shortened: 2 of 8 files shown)


cvmfs/swissknife_sync.cc
```diff
@@ -678,16 +679,10 @@ int swissknife::CommandSync::Main(const swissknife::ArgumentList &args) {
...
-      hash_algorithm,
-      params.compression_alg,
-      params.generate_legacy_bulk_chunks,
-      params.use_file_chunking,
-      params.min_file_chunk_size,
-      params.avg_file_chunk_size,
-      params.max_file_chunk_size,
-      params.session_token_file,
-      params.key_file);
+      params.spooler_definition, hash_algorithm, params.compression_alg,
```
siscia: Here, maybe the original was better, the formatter just automatically change them...

cvmfs/CMakeLists.txt
```diff
@@ -244,6 +244,8 @@ set (CVMFS_SWISSKNIFE_SOURCES
   sync_item.cc
   sync_mediator.cc
   sync_union.cc
+  sync_union_overlayfs.cc
+  sync_union_aufs.cc
```
jblomer: These changes should be applied to the unit test CMakeLists.txt, too

siscia: Actually I am checking it right now, but there is not even a sync_union.cc in that CMakeList, it makes sense since it is not tested at all.

I think would be more logical to add them as soon as they are needed.

jblomer: Ah, ok.  Yes, that makes sense.

PR #2054, opened 2018-02-08
Fix error message when attempting to mount an already mounted repo

DrDaveD:
See [CVM-1477](https://sft.its.cern.ch/jira/browse/CVM-1477)

### DIFF
--- a/mount/mount.cvmfs.cc
+++ b/mount/mount.cvmfs.cc
@@ -139,7 +139,9 @@ static bool CheckProxy() {
 }
 
 
-static bool CheckConcurrentMount(const string &fqrn, const string &workspace, string *mountpointp) {
+static bool CheckConcurrentMount(const string &fqrn,
+                                 const string &workspace,
+                                 string *mountpointp) {
   // Try connecting to cvmfs_io socket
   int socket_fd = ConnectSocket(workspace + "/cvmfs_io." + fqrn);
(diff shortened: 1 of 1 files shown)


jblomer:
Many thanks!  Please have a look at the Travis issues.

I think for this issue we can also add an integration test to ensure it doesn't come up again in the future.

DrDaveD:
@jblomer Ok I fixed the cpplint error.  What exactly do you want in an integration test?  To cause the error and check the contents of the error message?  That seem strange to me. Is there precedent for that?

olifre:
> To cause the error and check the contents of the error message? That seem strange to me. 

That's exactly what I would expect at least, and what is commonly done in other projects (e.g. charliecloud's test suite in bats, runc's test suite in bats).

jblomer:
@DrDaveD yep, that's what I had in mind.  I'm not sure from the top of my head if there is a precedent.  I'd classify this as a "presentation error", which got introduced when we rewrote the mount logic.  As it is a regression, it made sense to me adding a test for the next time we modify the mount code.

DrDaveD:
Ok.  I am on vacation next week though so I won't get to it anytime soon.  Maybe you could add it to the PR or make another one.  Maybe that's best anyway for setting a new precedent.

jblomer:
Replaced by #2078

PR #2055, opened 2018-02-09
[WIP] Upgrade gtest 1.8

siscia:
Work in progress for updating google test to 1.8

It just build on my machine and I would like to test on the CI


externals/googletest/src/configureHook.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
 
-CXXFLAGS="$CVMFS_BASE_CXX_FLAGS -DGTEST_HAS_TR1_TUPLE=0 -fPIC" cmake \
-    -DCMAKE_INSTALL_PREFIX=$EXTERNALS_INSTALL_LOCATION .
+mkdir build
+cd build
+
+CXXFLAGS="$CVMFS_BASE_CXX_FLAGS -fPIC" cmake ..
```
jblomer: If I remember correctly, the `-DGTEST_HAS_TR1_TUPLE=0` was necessary for certain older platforms with broken partial C++11 implementation.

externals/googletest/src/makeHook.sh
```diff
@@ -1,7 +1,24 @@
 #!/bin/sh
 
-make clean
-make
+cd build 
+make
```
jblomer: Actually I forgot why we have the `make clean`.  Can you check how it is done in the other externals?  Or was there a particular reason for removal?

externals/googletest/src/configureHook.sh
siscia: I was guessing something like that!

siscia: Actually I wasn't able to keep it with `GTEST_HAS_TR1_TUPLE=0` I am trying to use the flag `GTEST_USE_OWN_TR1_TUPLE=1` that  should use the google version of tuple... We should check if it is enough...

Ref googletest: https://github.com/google/googletest/tree/master/googletest#choosing-a-tr1-tuple-library
Ref googlemock: https://github.com/google/googletest/tree/master/googlemock#choosing-a-tr1-tuple-library

siscia:
Before to merge we should test if everything works fine in the old platform.

That problem with the c++ tuples.

I used the GTest version of tuple so I am not expecting problems, but still...

jblomer:
Yep, indeed.  You should be able to test it with `centos6` and `centos5` docker containers.

siscia:
👍

jblomer:
I think that's ok, it's most likely due to the local docker environment (perhaps not a privileged container).  The test runs fine on our test servers.

jblomer:
Almost there.  We have a build problem on SLES11:

  https://epsft-jenkins.cern.ch/job/CvmfsIncrementalBuild/911/CVMFS_BUILD_ARCH=docker-x86_64,CVMFS_BUILD_PLATFORM=sles11/console

Let's look into it tomorrow

PR #2057, opened 2018-02-14
add patch to the makefile in order to execute it on sles11

siscia:


bootstrap.sh
```diff
@@ -177,6 +177,7 @@ build_lib() {
       ;;
     googletest)
         do_extract "googletest"   "googletest-release-${GOOGLETEST_VERSION}.tar.gz"
+	patch_external "googletest"	"cmake_compatibility.patch"
```
jblomer: Just a nit pick: a space seems missing.

siscia:
Even if travis give us the green light I would suggest anyway to re-run the matrix integration

PR #2060, opened 2018-02-20
Swallow libarchive

siscia:
In this PR we swallow libarchive.

We add the "compressed" source as an external dependency. (We also include a small README.txt that explain how to get from the official sources to the compressed one).

We configure, build and install the software (in configureHook.sh and makeHook.sh) and get around some issues created by the compression.

We modify the bootstrap file to actually work on libarchive.

Finally we add the dependency in CMakeList.

I wasn't able to cleanly remove the commit (810526c, checkpoint) the one that started the whole PR, it should automagically merge though.


externals/libarchive/src/configureHook.sh
```diff
@@ -0,0 +1,24 @@
...
+    --without-bz2lib \
+    --without-lz4 \
+    --without-lzma \
+    --without-lzo2 \
+    --without-cng \
+    --without-nettle \
+    --without-openssl \
+    --without-xml2 \
+    --without-expat \
+    --disable-maintainer-mode \
```
jblomer: I forgot, what is this for?

siscia: From the help: 
 --disable-maintainer-mode
                          disable make rules and dependencies not useful (and
                          sometimes confusing) to the casual installer

I was just removing as much as possible in order to make it as simple as possible to compile, however, digging deeper it does not seems like a good idea to keep it around.

It basically don't regenerate the Makefiles, configure scripts, etc when something changes in Makefile.in . It is pretty neutral for us, but I don't think is a good idea...

Removing it.

CMakeLists.txt
```diff
@@ -306,6 +306,8 @@ set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${GTEST_INCLUDE_DIRS})
 find_package (VJSON REQUIRED)
 set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${VJSON_INCLUDE_DIRS})
 
+find_package (LibArchive REQUIRED)
```
jblomer: One more thing: shouldn't that come together with a FindLibArchive cmake module?

siscia: Added! :)

siscia: Moreover, I confirm that on SLE11 and on my machine it actually pick the correct libarchive.a from external install.

PR #2064, opened 2018-03-01
Fix the creation of libcvmfs_cache.a on macOS

radupopescu:
On macOS (maybe also on Linux?), the libtool command can be used
instead of the "combine_libs" script to create a stand-alone
libcvmfs_cache.a library.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -152,6 +152,7 @@ set (CVMFS_CACHE_RAM_SOURCES
   malloc_heap.cc
   statistics.cc
   util_concurrency.cc
+  util/posix.cc
   util/string.cc
 )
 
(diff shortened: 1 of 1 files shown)


cvmfs/CMakeLists.txt
```diff
@@ -541,23 +541,31 @@ if (BUILD_LIBCVMFS_CACHE)
...
+    add_custom_command(
+        OUTPUT libcvmfs_cache.a
+        COMMAND libtool -static -o libcvmfs_cache.a ${LIBCVMFS_CACHE_LIBS}
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        DEPENDS ${LIBCVMFS_CACHE_LIBS}
+      )
+  else (MACOSX)
+    #set (LIBCVMFS_CACHE_PUBLIC_SYMS "-public" "${CMAKE_CURRENT_SOURCE_DIR}/cache_plugin/libcvmfs_cache_public_syms.txt")
+    add_custom_command(
+        OUTPUT libcvmfs_cache.a
```
jblomer: Perhaps we can add a comment/todo here?  Along the lines that libtool might work for both but we didn't try yet.

radupopescu:
Merging this. Travis tests are actually green, but this isn't shown here due to DDoS?

PR #2065, opened 2018-03-02
Enable CORS in default apache configuration

saurvs:
For web browser JavaScript to download cvmfs data.


jblomer:
Many thanks!  We'll keep it open until the change has been discussed with our main infrastructure providers.  If approved, we'd also need to add the migration routine to inject the configuration into existing Apache configurations.

jblomer:
For the moment the infrastructure team decided it would be better to provide a dedicated URL for JS clients with this header set. That can avoid accidental overload of the regular backend server.

jblomer:
The repositories hosted by hepvm.cern.ch can be used for testing and development for the time being.

PR #2067, opened 2018-03-06
Bump minimum supported macOS version to 10.12

radupopescu:


radupopescu:
Apple usually provide security updated for two previous OS version, so we will bump the minimum supported version after macOS 10.14 is released.

PR #2091, opened 2018-04-03
Export checksum to workspace

jblomer:


jblomer:
@btovar Based on #2070 this PR writes the checksum to the cvmfs "workspace". The workspace can be set through the `CVMFS_WORKSPACE` parameter, which defaults to the cache directory.  In the legacy options scheme, it is the `lock_directory` parameter that sets the workspace explicitly. Please let us know if this works for you.

btovar:
Thanks @jblomer, I'll give it a try.

PR #2097, opened 2018-04-06
Tarball Ingestion

siscia:
This PR add the new functionality of tarball ingestion.

It is mostly ready and at least big errors are found the main architecture will stay this one.

Being this a quite big PR any suggestion to split it in more manageable chunk is very welcome.

Moreover, if any helps is needed in the code review phase in the form of documentation or addition of comments or any other write up I am happy to oblige.

Still I would like to check some details especially on the SyncItem class where I changed the visibility of some fields. Nevertheless I would like to have travis running on this.
Also the last additions makes quite difficult to have a solid unittest suite, that for the moment is commented out.

I do need some time to think about the best way to provide unittest for this PR.


siscia:
CC @jblomer @radupopescu @DrDaveD

jblomer:
I assume that #2098 requires some adjustments for this PR, too, so you might want to resubmit after rebasing.

siscia:
:+1: It could definitely.

siscia:
We rebase on #2101

PR #2098, opened 2018-04-06
Improve memory utilization in ingestion pipeline

jblomer:
The I/O payload pushed through the pipeline is allocated separately from the system malloc. Also, the number of files in flight is limited. Both measures combined should keep the memory envelope of `cvmfs_swissknife sync` within a 4G envelope independent of the change set.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -168,6 +168,7 @@ set (CVMFS_FSCK_SOURCES
 
 
 set (CVMFS_SWISSKNIFE_SOURCES
+  backoff.cc
   catalog.cc
   catalog_counters.cc
   catalog_mgr_ro.cc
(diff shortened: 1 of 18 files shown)


siscia:
I got 651 and 652 integration test fails on my machine.

This is what #2099 is fixing?

jblomer:
Actually these two test cases have not much to do with any of the pull requests at hand. Are other integration tests working for you? Let's see what the automatic test runs say.

siscia:
I was rebasing on top of this PR and so I was re-running the integration test I wrote for the ingestion of tarballs (from 651 to 657) these two were 651 and 652 and ran as well.

The tarballs one where just fine, however these two failed...

jblomer:
Ok, let's keep that in mind for the time being. Please also disantangle the test case numbers. As we were working on branches in parallel, we ended up with 651, 652 issued multiple times.

siscia:
:tada: 

I believe this solved the excessive memory consunption when ingesting big tarballs!

PR #2099, opened 2018-04-06
Generate systemd mount unit on the fly if necessary

jblomer:
While working on #2098 I figured that the change introduced in #2089 would make `cvmfs_suid_helper rdonly_mount` not generate the systemd unit under certain circumstances.  This patch recreates the unit on the fly if necessary.


jblomer:
On second though, this is not what we want. The original code runs as root, the new code does not.  It might be necessary though to run `systemd-fstab-generator` during the package update or during migrate for the following case:

  1. A repository is created with the old software, so /etc/fstab is updated but no systemd mount unit is created
  2. The cvmfs_server package is updated and the machine has not been rebooted

PR #2100, opened 2018-04-06
Adding new FileWatcher implementation for Linux based on Inotify

radupopescu:


cvmfs/file_watcher_inotify.cc
```diff
@@ -0,0 +1,167 @@
...
+  poll_set[0].events = POLLHUP | POLLIN;
+  poll_set[0].revents = 0;
+  poll_set[1].fd = inotify_fd_;
+  poll_set[1].events = POLLIN;
+  poll_set[1].revents = 0;
+
+  bool stop = false;
+  while (!stop) {
+    int ready = poll(poll_set, 2, -1);
+    if (ready == -1) {
```
jblomer: This might be ok when `errno == EINTR`

cvmfs/file_watcher_inotify.cc
```diff
@@ -0,0 +1,167 @@
...
+            event = file_watcher::kRenamed;
+          } else if (inotify_event->mask & IN_ATTRIB) {
+            // Attributes
+            event = file_watcher::kAttributes;
+          } else if (inotify_event->mask & IN_IGNORED) {
+            // Attributes
+            event = file_watcher::kIgnored;
+          }
+          bool clear_handler = true;
+          if (event != file_watcher::kInvalid ||
```
jblomer: Shouldn't this be "event not invalid _and_ event not ignored"?

cvmfs/file_watcher_inotify.cc
```diff
@@ -0,0 +1,167 @@
...
+    if (poll_set[0].revents & POLLIN) {
+      char buffer[1];
+      ReadPipe(read_pipe, &buffer, 1);
+      LogCvmfs(kLogCvmfs, kLogDebug, "FileWatcherInotify - Stopping.\n");
+      stop = true;
+      continue;
+    }
+
+    const size_t event_size = sizeof(struct inotify_event);
+    const size_t buffer_size = event_size + 100;
```
jblomer: The 100 might be better a constant explaining its meaning.

cvmfs/file_watcher_inotify.cc
```diff
@@ -0,0 +1,167 @@
...
+
+  close(inotify_fd_);
+
+  return true;
+}
+
+void FileWatcherInotify::RegisterFilter(const std::string& file_path,
+                                        EventHandler* handler) {
+  bool done = false;
+  BackoffThrottle throttle(1000, 10000, 50000);
```
jblomer: Could move as constants to the header.

cvmfs/file_watcher_inotify.cc
radupopescu: Good catch.

PR #2102, opened 2018-04-09
Tarball ingestion after rebase

siscia:
This PR is exactly like #2097 but it is been rebased on top of #2098

For convenience the commit message of 2097

This PR add the new functionality of tarball ingestion.

It is mostly ready and at least big errors are found the main architecture will stay this one.

Being this a quite big PR any suggestion to split it in more manageable chunk is very welcome.

Moreover, if any helps is needed in the code review phase in the form of documentation or addition of comments or any other write up I am happy to oblige.

Still I would like to check some details especially on the SyncItem class where I changed the visibility of some fields. Nevertheless I would like to have travis running on this.
Also the last additions makes quite difficult to have a solid unittest suite, that for the moment is commented out.

I do need some time to think about the best way to provide unittest for this PR.


CMakeLists.txt
```diff
@@ -311,6 +310,11 @@ set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${VJSON_INCLUDE_DIRS})
 find_package (LibArchive REQUIRED)
 set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${LibArchive_INCLUDE_DIRS})
 
+if (MACOSX)
```
siscia: Is this necessary?

radupopescu: I don't recognize these lines. Does it compile without?

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,130 @@
...
+ */
+
+#ifndef CVMFS_INGESTION_INGESTION_SOURCE_H_
+#define CVMFS_INGESTION_INGESTION_SOURCE_H_
+
+#include <pthread.h>
+#include <stdio.h>
+#include <string>
+
+#include "archive.h"
```
siscia: Use duplex

CMakeLists.txt
siscia: The mac compiles happily without!

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,129 @@
...
+  virtual std::string GetPath() const = 0;
+  virtual bool Open() = 0;
+  virtual ssize_t Read(void* buffer, size_t nbyte) = 0;
+  virtual bool Close() = 0;
+  virtual bool GetSize(uint64_t* size) = 0;
+};
+
+class FileIngestionSource : public IngestionSource {
+ public:
+  explicit FileIngestionSource(const std::string& path) : path_(path) {}
```
siscia: fd_ should be set to -1

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,129 @@
...
+
+#include <pthread.h>
+#include <stdio.h>
+#include <string>
+
+#include "duplex_libarchive.h"
+#include "platform.h"
+#include "sync_item.h"
+#include "util/posix.h"
+#include "util/single_copy.h"
```
siscia: errno header

cvmfs/server/cvmfs_server_ingest_tarball.sh
```diff
@@ -313,10 +323,4 @@ cvmfs_server_ingest_tarball() {
   publish_after_hook $name
   publish_succeeded  $name
 
-  # force and health check with repeairs.
-  # it unmount and remount the fs making the new extracted file "appear"
-  # then it checks that everything is ok that doesn't hurt
-  health_check -r $name
```
siscia: I don't understand why this command does not unmount and remount automatically the repo making the new folders appears

siscia: healt_check was making also this job

test/unittests/t_sync_union_tarball.cc
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  std::string CreateTarFile(const std::string& tar_filename,
+                            const std::string& base64_data) {
+    int tar;
+    std::string data_binary;
+    Debase64(base64_data, &data_binary);
+
+    std::string tmp_dir = CreateTempDir("test_sync_union");
+    assert(!tmp_dir.empty());
```
siscia: We discuss about using gtest assert. Unfortunately it is not possible since ASSERT_TRUE returns from the function and it does return void, this particular function need to return a std::string

siscia:
@jblomer I believe I addressed all your comments in the share code review session.

I could have definitely have miss something but I believe that this PR could definitely benefit from your code review.

siscia:
@jblomer I was refering to this PR before

cvmfs/sync_item_tar.h
```diff
@@ -0,0 +1,176 @@
...
+        readlink((this->GetUnionPath()).c_str(), slnk, PATH_MAX);
+    assert(length >= 0);
+    dirent.symlink_.Assign(slnk, length);
+  }
+
+  if (this->IsCharacterDevice() || this->IsBlockDevice()) {
+    dirent.size_ = makedev(GetRdevMajor(), GetRdevMinor());
+  }
+
+  if (dirent.IsDirectory() && dirent.size_ == 0) dirent.size_ = 4096;
```
jblomer: What is the reasoning behind the directory size?

cvmfs/sync_union.h
```diff
@@ -34,6 +34,8 @@
 #include <string>
 
 #include "sync_item.h"
+#include "sync_mediator.h"
```
jblomer: If possible, a forward declaration of `AbstractSyncMediator` instead of the include would be preferable

cvmfs/sync_union_tarball.h
```diff
@@ -11,40 +11,63 @@
...
+  std::string UnwindWhiteoutFilename(SharedPtr<SyncItem> entry) const;
+  bool IsOpaqueDirectory(SharedPtr<SyncItem> directory) const;
+  bool IsWhiteoutEntry(SharedPtr<SyncItem> entry) const;
+
  private:
+  struct archive *src;
   const std::string tarball_path_;
   const std::string base_directory_;
-  std::set<std::string> to_recur_;
+  const std::string to_delete_; /* entity to delete */
```
jblomer: Use doxyen style comments

cvmfs/sync_union_tarball.h
```diff
@@ -11,40 +11,63 @@
...
+  std::set<std::string>
+      know_directories_; /* directory that we know already exist */
```
jblomer: Doxygen style comments.  The comment should explain a bit more, namely the problem of random order of contents in the tarball which makes it necessary to fix-up directories later on.

cvmfs/sync_union_tarball.h
```diff
@@ -11,40 +11,63 @@
...
 #include "sync_union.h"
 
+#include <pthread.h>
+
+#include <map>
 #include <set>
 #include <string>
 
+#include "duplex_libarchive.h"
+#include "sync_mediator.h"
```
jblomer: A forward declaration of `AbstractSyncMediator` might be enough

cvmfs/sync_union_tarball.h
```diff
@@ -11,40 +11,63 @@
...
+  struct archive *src;
   const std::string tarball_path_;
   const std::string base_directory_;
-  std::set<std::string> to_recur_;
+  const std::string to_delete_; /* entity to delete */
+  std::set<std::string>
+      know_directories_; /* directory that we know already exist */
+  std::set<std::string> to_create_catalog_dirs_;
+  std::map<std::string, SharedPtr<SyncItem> > dirs_;
+  Signal *read_archive_signal_;
```
jblomer: `Signal *` --> `Signal`

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+SyncUnionTarball::~SyncUnionTarball() { delete read_archive_signal_; }
+
+bool SyncUnionTarball::Initialize() {
+  bool result;
+
+  src = archive_read_new();
+  assert(ARCHIVE_OK == archive_read_support_format_tar(src));
+  assert(ARCHIVE_OK == archive_read_support_format_empty(src));
+
+  if (tarball_path_ == "--") {
```
jblomer: `--` --> `-`

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+ * This whole process is not necessary for directories since we don't
+ * actually need to read data from them.
  */
-SyncUnionTarball::SlyncUnionTarball(SyncMediator *mediator,
-                                    const std::string &tarball_path,
-                                    const std::string &base_directory)
-    : tarball_path_(tarball_path), base_directory_(base_directory) {}
+void SyncUnionTarball::Traverse() {
+  assert(this->IsInitialized());
+  struct archive_entry *entry = archive_entry_new();
```
jblomer: Should be after the `to_delete_` handling

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
 
-/*
- * We traferse the tar and then we keep track (the path inside a set) of each
- * .tar we find, then we go back and repeat the procedure. Should we be worry by
- * tar-bombs? Maybe, we will think about it later.
+  /*
+   * As first step we eliminate the directories we are request.
```
jblomer: directories we are request --> requested directories

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+
+    for (vector<string>::iterator s = to_eliminate_vec.begin();
+         s != to_eliminate_vec.end(); ++s) {
+      std::string parent_path;
+      std::string filename;
+      SplitPath(*s, &parent_path, &filename);
+      if (parent_path == ".") parent_path = "";
+      SharedPtr<SyncItem> sync_entry =
+          CreateSyncItem(parent_path, filename, kItemDir);
+      mediator_->Remove(sync_entry);
```
jblomer: We should have a test on what happens if one of the paths is not a directory but a file. Or if it doesn't exist.

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+    }
+  }
+
+  while (true) {
+    /* Get the lock, wait if lock is not available yet */
```
jblomer: `/*` --> `//`

This facilitates commenting larger blocks with `/* */`

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+    read_archive_signal_->Wait();
+
+    int result = archive_read_next_header2(src, entry);
+
+    switch (result) {
```
jblomer: We should have a `default` case that aborts to make sure we didn't forget about a return value.

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+      case ARCHIVE_FATAL: {
+        LogCvmfs(kLogUnionFs, kLogStderr,
+                 "Fatal error in reading the archive.");
+        return;
+        break; /* Only exit point with error */
```
jblomer: `/*` --> `//`

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+    /* Get the lock, wait if lock is not available yet */
+    read_archive_signal_->Wait();
+
+    int result = archive_read_next_header2(src, entry);
+
+    switch (result) {
+      case ARCHIVE_FATAL: {
+        LogCvmfs(kLogUnionFs, kLogStderr,
+                 "Fatal error in reading the archive.");
+        return;
```
jblomer: This needs to be an `abort()`, I think. Otherwise we might publish half an archive.

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+        }
+        return; /* Only successful exit point */
+        break;
+      }
+
+      case ARCHIVE_WARN: {
+        LogCvmfs(kLogUnionFs, kLogStderr,
+                 "Warning in uncompression reading, going on. \n %s",
+                 archive_error_string(src));
+        /* We actually want this to enter the ARCHIVE_OK case */
```
jblomer: `/*` --> `//`

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+        for (set<string>::iterator dir = to_create_catalog_dirs_.begin();
+             dir != to_create_catalog_dirs_.end(); ++dir) {
+          assert(dirs_.find(*dir) != dirs_.end());
+          SharedPtr<SyncItem> to_mark = dirs_[*dir];
+          assert(to_mark->IsDirectory());
+          to_mark->SetCatalogMarker();
+          to_mark->AlreadyCreatedDir();
+          ProcessDirectory(to_mark);
+        }
+        return; /* Only successful exit point */
```
jblomer: `/*` --> `//`

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+  while (true) {
+    /* Get the lock, wait if lock is not available yet */
+    read_archive_signal_->Wait();
+
+    int result = archive_read_next_header2(src, entry);
+
+    switch (result) {
+      case ARCHIVE_FATAL: {
+        LogCvmfs(kLogUnionFs, kLogStderr,
+                 "Fatal error in reading the archive.");
```
jblomer: I guess we want `archive_error_string(src)` in this case, too

cvmfs/sync_union_tarball.cc
```diff
@@ -4,30 +4,223 @@
...
+                 "Warning in uncompression reading, going on. \n %s",
+                 archive_error_string(src));
+        /* We actually want this to enter the ARCHIVE_OK case */
+      }
+
+      case ARCHIVE_OK: {
+        std::string archive_file_path(archive_entry_pathname(entry));
+        std::string complete_path(base_directory_ + "/" + archive_file_path);
+
+        if (*complete_path.rbegin() == '/') {
```
jblomer: Should use `MakeCanonicalPath()` from util/posix.h

cvmfs/sync_item.h
```diff
@@ -153,6 +162,10 @@ class SyncItem {
     return relative_parent_path_;
   }
 
+  virtual IngestionSource *CreateIngestionSource() const;
+  virtual void AlreadyCreatedDir() const { assert(false); }
```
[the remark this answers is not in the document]
jblomer: Ehh, that's of course wrong, too.  How about: `IsPlaceholderDirectory`

cvmfs/util/shared_ptr.h
```diff
@@ -120,6 +120,8 @@ class SharedPtr {
     return static_cast<void*>(value_);
   }
 
+  element_type* get() const { return Get(); }
```
jblomer: Add a comment why this method is needed (Google Mock, right?)

test/src/655-tarball_ingest_special_files/main
```diff
@@ -0,0 +1,119 @@
+
```
jblomer: Here I'd avoid re-testing was was tested before. Restrict it only to the more complex files.

test/src/657-tarball_ingest_from_stdin/main
```diff
@@ -0,0 +1,90 @@
+
```
jblomer: This can be a very simple tarball, as it should only test the stdin input way

test/src/658-tarball_remove_directory/main
```diff
@@ -0,0 +1,144 @@
+
```
jblomer: This can again be a test simpler tarball.  I'd suggest one file and one nested sub directory to test the recursive delete.

test/src/659-tarball_remove_base_directory/main
```diff
@@ -0,0 +1,143 @@
+
```
jblomer: I think the extra case of this test can be combined with the previous one.

test/unittests/t_ingestion_stress.cc
```diff
@@ -467,18 +469,33 @@ TEST_F(T_IngestionStress, ProcessMultipleFilesInSeparateWavesSlow) {
...
   uploader_->ClearResults();
 
   // second wave...
   // some small and medium sized files with file chunking enabled
   // one big file without file chunking
-  pipeline.Process(GetEmptyFile(), true);
-  pipeline.Process(GetSmallFile(), true);
-  pipeline.Process(GetBigFile(), true);
-  pipeline.Process(GetHugeFile(), false, shash::kSuffixCatalog);
+  std::string empty_path_2(GetEmptyFile());
```
jblomer: This could be more concise as `pipeline.Process(new FileIngestionSource(GetEmptyFile()), ...)` to avoid the extra definition of a `string` and a `IngestionSource` variable.

test/unittests/t_sync_union_tarball.cc
```diff
@@ -0,0 +1,188 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include <sync_union_tarball.h>
```
jblomer: Should be further down in the include list as `#include "sync_union_tarball.h"`

test/unittests/t_sync_union_tarball.cc
```diff
@@ -0,0 +1,188 @@
...
+    tar = open(tmp_tar_filename_.c_str(), O_WRONLY);
+
+    assert(tar > 0);
+    assert(SafeWrite(tar, data_binary.c_str(), data_binary.size()));
+
+    return tmp_tar_filename_;
+  }
+
+  virtual void TearDown() {
+    remove(tmp_tar_filename_.c_str());
```
jblomer: I'd use the more precise `unlink`, which requires `#include <unistd.h>` (this one would require `#include <cstdio>`).

test/unittests/t_sync_union_tarball.cc
```diff
@@ -0,0 +1,188 @@
...
+TEST_F(T_SyncUnionTarball, Init) {
+  std::string tar_filename = CreateTarFile("tar.tar", simple_tar);
+  publish::SyncUnionTarball sync_union(m_sync_mediator_, "", tar_filename,
+                                       "/tmp/lala", "");
+
+  EXPECT_CALL(*m_sync_mediator_, RegisterUnionEngine(_)).Times(1);
+  EXPECT_TRUE(sync_union.Initialize());
+}
+
+///*
```
jblomer: The following lines should either be activated or removed.

cvmfs/server/cvmfs_server_ingest_tarball.sh
```diff
@@ -0,0 +1,326 @@
...
+
+# This file depends on fuctions implemented in the following files:
+# - cvmfs_server_util.sh
+# - cvmfs_server_common.sh
+
+
+# TODO Most of this code is replicated and shared between different scrips,
+# it would be a good idea to refactor common patterns into coherent functions.
+
+cvmfs_server_ingest_tarball() {
```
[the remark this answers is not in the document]
siscia: Do you mean like what we do with the normal publish. `cvmfs_server transaction ...` then the `cvmfs_server ingest-tarball ...` and finally `cvmfs_server publish ...` or simply open a transaction inside the `cvmfs_server ingest-tarball` command?

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,117 @@
...
+class FileIngestionSource : public IngestionSource {
+ public:
+  explicit FileIngestionSource(const std::string& path)
+      : path_(path), fd_(-1) {}
+  ~FileIngestionSource() {}
+
+  std::string GetPath() const { return path_; }
+
+  bool Open() {
+    fd_ = open(path_.c_str(), O_RDONLY);
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,117 @@
...
+    if (fd_ < 0) {
+      LogCvmfs(kLogCvmfs, kLogStderr,
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,117 @@
...
+               "Err: Impossible to open the file: %s (%d)\n %s", path_.c_str(),
+               errno, strerror(errno));
+      return false;
+    }
+    return true;
+  }
+
+  ssize_t Read(void* buffer, size_t nbyte) {
+    assert(fd_ >= 0);
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,117 @@
...
+    if (read < 0) {
+      LogCvmfs(kLogCvmfs, kLogStderr, "failed to read the file: %s (%d)\n %s",
+               fd_, path_.c_str(), errno, strerror(errno));
+    }
+    return read;
+  }
+
+  bool Close() {
+    if (fd_ == -1) return true;
+    int ret = close(fd_);
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,117 @@
...
+    fd_ = -1;
+    return (ret == 0);
+  }
+
+  bool GetSize(uint64_t* size) {
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/ingestion/item.h
```diff
@@ -71,6 +74,13 @@ class FileItem : SingleCopy {
   uint64_t GetNumChunks() { return chunks_.size(); }
   FileChunkList *GetChunksPtr() { return &chunks_; }
 
+  bool Open() { return source_->Open(); }
```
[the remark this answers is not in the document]
siscia: Done! :+1:  I get rid of the path_ member altogether

cvmfs/server/cvmfs_server_common.sh
```diff
@@ -695,6 +695,21 @@ close_transaction() {
 
   is_in_transaction $name || return 0
 
+  remount_repo $name $use_fd_fallback
+}
```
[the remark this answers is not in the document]
siscia: @jblomer 
I am not sure I follow, the old interface is exactly the same, the only thing that changes is that I factored out the function that remount the repo. The old users of the functions should be left untouched, shouldn't them?

Am I missing something obvious?

cvmfs/swissknife_tarball.cc
```diff
@@ -0,0 +1,199 @@
...
+
+  if (!InitVerifyingSignatureManager(params.public_keys,
+                                     params.trusted_certs)) {
+    return 3;
+  }
+
+  bool with_gateway =
+      spooler_definition.driver_type == upload::SpoolerDefinition::Gateway;
+
+  UniquePtr<manifest::Manifest> manifest;
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/swissknife_tarball.cc
```diff
@@ -0,0 +1,199 @@
...
+
+    sync->Traverse();
+  }
+
+  if (!params.authz_file.empty()) {
+    LogCvmfs(kLogCvmfs, kLogDebug,
+             "Adding contents of authz file %s to"
+             " root catalog.",
+             params.authz_file.c_str());
+    int fd = open(params.authz_file.c_str(), O_RDONLY);
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/swissknife_tarball.cc
```diff
@@ -0,0 +1,199 @@
...
+    if (fd == -1) {
+      LogCvmfs(kLogCvmfs, kLogStderr,
+               "Unable to open authz file (%s)"
+               "from the publication process: %s",
+               params.authz_file.c_str(), strerror(errno));
+      return 7;
+    }
+
+    std::string new_authz;
+    const bool read_successful = SafeReadToString(fd, &new_authz);
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/swissknife_tarball.cc
```diff
@@ -0,0 +1,199 @@
...
+    close(fd);
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/sync_item.h
```diff
@@ -29,11 +37,11 @@ enum SyncItemType {
   kItemSocket,
   kItemNew,
   kItemMarker,
-  kItemUnknown
+  kItemUnknown,
+  kItemTarfile
```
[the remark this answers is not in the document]
siscia: @jblomer I though about that and I really like your idea. The only concern I have is that it is quite a big change, am I wondering if it is better to make it in a different PR, so we avoid make this even bigger and it is simpler to catch issues.

cvmfs/sync_item.h
```diff
@@ -197,7 +210,8 @@ class SyncItem {
     return rdonly_type_ == expected_type;
   }
 
- private:
```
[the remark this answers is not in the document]
siscia: @jblomer I believe it is protected, I do agree that the diff was a little complex to read, but reading the whole file it seems to me that `rdonly_type_` is already protected. Am I wrong?

cvmfs/sync_item_tar.h
```diff
@@ -0,0 +1,176 @@
+/**
```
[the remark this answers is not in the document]
siscia: Done! :+1:

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,117 @@
...
+               errno, strerror(errno));
+      return false;
+    }
+    return true;
+  }
+
+  ssize_t Read(void* buffer, size_t nbyte) {
+    assert(fd_ >= 0);
+    ssize_t read = SafeRead(fd_, buffer, nbyte);
+    if (read < 0) {
```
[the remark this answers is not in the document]
siscia: @jblomer I have mixed opinions here. From one side I do agree that one check could be enough, on the other hand, I do prefer to check also here since it is more specific and can provide better information about something that went wrong. In either case, something should be done, either remove this log or adding an extra one in the relative tar method.

cvmfs/ingestion/ingestion_source.h
```diff
@@ -0,0 +1,144 @@
...
+  std::string GetPath() const { return path_; }
+
+  bool Open() {
+    assert(size_ >= 0);
+    return true;
+  }
+
+  ssize_t Read(void* external_buffer, size_t nbytes) {
+    return archive_read_data(archive_, external_buffer, nbytes);
+  }
```
siscia: If something goes wrong we should set the errno that is logged by the caller (`TaskRead`)

cvmfs/sync_union_tarball.cc
siscia: @jblomer  I am enhancing the delete path in another PR, if you don't mind I believe we can carry this point on the next PR.

siscia:
BUILD PR

siscia:
BUILD PR

cvmfs/server/cvmfs_server_common.sh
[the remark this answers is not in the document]
jblomer: Ah, sorry, my mistake.  You're right.

cvmfs/server/cvmfs_server_ingest_tarball.sh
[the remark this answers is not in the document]
jblomer: I was thinking of the latter: a call to `cvmfs_server ingest` would call `cvmfs_server_transaction` at the beginning and `cvmfs_server_publish` or `cvmfs_server_abort` at the end.  Between these calls, it uses `cvmfs_swissknife ingest` in order to process a tarball instead of the normal `sync`.

cvmfs/sync_item.h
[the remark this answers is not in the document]
jblomer: Sounds good.

siscia:
@jblomer I believe I addressed all the issues you raised.

However I am getting some error on jenkis for an old version of Fedora: https://epsft-jenkins.cern.ch/job/CvmfsIncrementalBuildOnPR/4/

It is very weird because in the previous CI cycle it worked flawlessly. I will re-run the build again.

siscia:
@jblomer I believe I addressed all your concerns.
Finally also the CI is green.

siscia:
Next PR will be:
1. https://github.com/siscia/cvmfs/pull/8
2. https://github.com/siscia/cvmfs/pull/9
3. https://github.com/siscia/cvmfs/pull/11

jblomer:
I pushed a copy here: https://github.com/cvmfs/cvmfs/tree/siscia-pulling_tarball

Making a full build and integration test run now.

cvmfs/util_concurrency.cc
```diff
@@ -36,8 +36,9 @@ Signal::Signal() : fired_(false) {
 
 
 Signal::~Signal() {
-  pthread_cond_destroy(&signal_);
-  pthread_mutex_destroy(&lock_);
+  assert(IsSleeping());
+  assert(0 == pthread_cond_destroy(&signal_));
```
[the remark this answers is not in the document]
siscia: Correct, thanks for catching it! Will fix it!

jblomer:
Excellent! Let's see what the integration tests give back.

PR #2105, opened 2018-04-11
Make Signal multi shot

siscia:
Signal is now SingleShot.

Once Wakeup is called, every successive call to Wait will simply not wait.

Wakeup set fired_ to true and the next invocation of Wait will acquire the lock but it will not enter the main loop.

This PR reset the Wait to false so that we are forced to have a sequence of Wait -> Wakeup -> Wait -> Wakeup

 I believe we could discuss if we want another class that has the MultiShot behaviour and leaving signal SingleShot, however, this one, is a quite tricky API to manage and I figure out this SingleShot behaviour only by chance...

### DIFF
--- a/cvmfs/util_concurrency.cc
+++ b/cvmfs/util_concurrency.cc
@@ -47,6 +47,7 @@ void Signal::Wait() {
     int retval = pthread_cond_wait(&signal_, &lock_);
     assert(retval == 0);
   }
+  fired_ = false;
 }
 
 


jblomer:
Very well!  Are there any other users currently of the Signal class?  If so we need to check them one by one and perhaps leave the default behaviour single-shot but allow for a constructor parameter that turns it into a multi-shot Signal.

siscia:
I believe we are save.

There is only one other place where Signal is used: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/cache_extern.h#L252-L258

However the Signal is stack initialized, make it `Wait` and immediately destroyed here: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/cache_extern.cc#L111-L120

Then there is a single `Wakeup`: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/cache_extern.cc#L542 

So after this last Wakeup Signal just get destroyed.

The change is such that it modify the behaviour only with multiple call of Wait, in our case we are safe.

Anyway we should discuss if we want to maintain the Single Shot semantic for Signal and create a new class for Multi Shot.

jblomer:
Hm, might be a problem here: if `Wakeup` is called _before_ `Wait`, then `Wait` never returns. Let's discuss.

PR #2109, opened 2018-04-19
Fix crashing of cvmfs_receiver for chunked files

radupopescu:
Addresses [CVM-1511](https://sft.its.cern.ch/jira/browse/CVM-1511).

### DIFF
--- a/cvmfs/catalog_diff_tool.h
+++ b/cvmfs/catalog_diff_tool.h
@@ -8,6 +8,7 @@
 #include <string>
 
 #include "directory_entry.h"
+#include "file_chunk.h"
 #include "shortstring.h"
 #include "statistics.h"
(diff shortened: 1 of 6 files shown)


cvmfs/catalog_diff_tool.h
```diff
@@ -53,7 +54,8 @@ class CatalogDiffTool {
  protected:
   virtual void ReportAddition(const PathString& path,
                               const catalog::DirectoryEntry& entry,
-                              const XattrList& xattrs) = 0;
+                              const XattrList& xattrs,
+                              const FileChunkList& chunks) = 0;
```
jblomer: Don't we need that change for `ReportModification`, too?

radupopescu: Good point

PR #2114, opened 2018-04-27
Improve error message in cvmfs_server_mkfs

saurvs:
If `cvmfs_server mkfs` isn't run as root, `modprobe` fails quietly and the script prints `neither AUFS nor OverlayFS detected on the system!` even if those modules exist.


jblomer:
Thanks!

PR #2117, opened 2018-05-08
Optionally print GC stats with CVMFS_SERVER_FLAGS=-+stats

jblomer:


DrDaveD:
What exactly would one set CVMFS_SERVER_FLAGS to?  It's not obvious from the changed code.

jblomer:
@DrDaveD It's in the title, `CVMFS_SERVER_FLAGS=-+stats` enables the statistics output.

DrDaveD:
That literal string "-+stats"?  It gets passed directly to swissknife gc, why isn't that in the swissknife help?  Why doesn't it follow the pattern of all other swissknife options?

jblomer:
Because we're not there yet. For the time being, it is a hot fix to help LHCb to get a grasp on their GC performance. We have a summer student project for producing cvmfs_server command stats properly.

DrDaveD:
Ok, good to know. 

I was still mystified by how it might be working, but I see now that it is done in swissknife_main.cc where matching kGenericParam (+) and "stats" results in a call to command->statistics().

PR #2124, opened 2018-05-17
Handle linkcount == 0 for catalog entries

radupopescu:
Partially addresses [CVM-1549](https://sft.its.cern.ch/jira/browse/CVM-1549).

* `CatalogDiffTool` - abort if any entry with `linkcount == 0` is encountered
*  `swissknife_check` - check that entries have `linkcount > 0`


jblomer:
Apologies, I merged too fast. This would make sense to open against the 2.5 branch, I think.  I'll revert.

jblomer:
Hm, I guess in this case we can also simply cherry-pick the two commits

PR #2126, opened 2018-05-21
Lower log verbosity when files are skipped in gateway

jblomer:


jblomer:
Opened against wrong branch.

PR #2133, opened 2018-05-24
Use OSG aliases for U.S. stratum 1s in cern.ch.conf

DrDaveD:
See [CVM-1556](https://sft.its.cern.ch/jira/browse/CVM-1556)

### DIFF
--- a/mount/domain.d/cern.ch.conf
+++ b/mount/domain.d/cern.ch.conf
@@ -15,6 +15,6 @@
 # Use cvmfs_config showconfig to get the effective parameters.
 #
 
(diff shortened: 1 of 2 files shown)


DrDaveD:
@jblomer Shall I update this PR to also increase the rpm version number?

jblomer:
That would be great, thanks!

DrDaveD:
Ok, done.

Note that this is submitted against the cvmfs-2.5 branch.  I did that because I think that's what you prefer, but it could easily be rebased against development if that's better.

PR #2136, opened 2018-05-25
Fix gcc pragma syntax for older compilers

jblomer:


siscia:
@jblomer this is actually showing several warning in my compiler, gcc 5.4.

It is expected?

jblomer:
@siscia Thanks, I noticed as well; the fix in this PR makes the code compile at all on older compilers.  Getting rid of the warning requires an additional step.

PR #2142, opened 2018-06-01
Add hardlinks support to tarball ingestion

siscia:
In this PR we add support for hardlinks.

It was required to add an aadditional function `Link` to the catalog manager

PR #2147, opened 2018-06-13
Handling hardlink inside tarballs.

siscia:
Was necessary to remembering every single hardlink we found in an
appropriate data structure and, after all the normal files have been
spooled and register in the catalog, add the hardlinks.

It was implemented a new functionality in the WritableCatalogManager,
Clone, that clone a file changing just its name.

Supersedes #2142

### DIFF
--- a/cvmfs/catalog_mgr_rw.cc
+++ b/cvmfs/catalog_mgr_rw.cc
@@ -287,6 +287,31 @@ void WritableCatalogManager::RemoveDirectory(const std::string &path) {
   SyncUnlock();
 }
 
+/**
+ * Clone the file called `source` changing its name into `destination`, the
+ * source file is keep intact
+ * @params destination, the name of the new file, complete path
+ * @params source, the name of the file to clone, which must be already in the
+ * repository
+ * @return void
+ */
+void WritableCatalogManager::Clone(const std::string destination,
+                                   const std::string source) {
+  const std::string parent = MakeRelativePath(source);
+
+  std::string destination_dirname;
+  std::string destination_filename;
+  SplitPath(destination, &destination_dirname, &destination_filename);
+
+  DirectoryEntry to_dirent;
+  LookupPath(parent, kLookupSole, &to_dirent);
+
+  DirectoryEntry destination_dirent(to_dirent);
+  destination_dirent.name_.Assign(
+      NameString(destination_filename.c_str(), destination_filename.length()));
+
+  this->AddFile(destination_dirent, empty_xattrs, destination_dirname);
+}
 
 /**
  * Add a new directory to the catalogs.
--- a/cvmfs/sync_union_tarball.cc
+++ b/cvmfs/sync_union_tarball.cc
@@ -10,6 +10,7 @@
 
 #include <cassert>
 #include <cstdio>
+#include <list>
 #include <set>
 #include <string>
 #include <vector>
@@ -146,43 +147,7 @@ void SyncUnionTarball::Traverse() {
       }
 
       case ARCHIVE_OK: {
-        std::string archive_file_path(archive_entry_pathname(entry));
-        std::string complete_path =
-            MakeCanonicalPath(base_directory_ + "/" + archive_file_path);
-
-        std::string parent_path;
-        std::string filename;
-        SplitPath(complete_path, &parent_path, &filename);
-
-        CreateDirectories(parent_path);
-
-        SharedPtr<SyncItem> sync_entry = SharedPtr<SyncItem>(new SyncItemTar(
-            parent_path, filename, src, entry, read_archive_signal_, this));
-
-        if (sync_entry->IsDirectory()) {
-          if (know_directories_.find(complete_path) !=
-              know_directories_.end()) {
-            sync_entry->IsPlaceholderDirectory();
-          }
-          ProcessUnmaterializedDirectory(sync_entry);
-          dirs_[complete_path] = sync_entry;
-          know_directories_.insert(complete_path);
-
-          read_archive_signal_->Wakeup();  // We don't need to read data and we
-                                           // can read the next header
-
-        } else if (sync_entry->IsRegularFile()) {
-          ProcessFile(sync_entry);
-          if (filename == ".cvmfscatalog") {
-            to_create_catalog_dirs_.insert(parent_path);
-          }
-        } else if (sync_entry->IsSymlink() || sync_entry->IsFifo() ||
-                   sync_entry->IsSocket()) {
-          ProcessFile(sync_entry);
-          read_archive_signal_->Wakeup();
-        } else {
-          read_archive_signal_->Wakeup();
-        }
+        ProcessArchiveEntry(entry);
         break;
       }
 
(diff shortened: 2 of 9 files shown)


cvmfs/sync_union_tarball.cc
```diff
@@ -199,6 +164,74 @@ void SyncUnionTarball::Traverse() {
...
+  CreateDirectories(parent_path);
+
+  SharedPtr<SyncItem> sync_entry = SharedPtr<SyncItem>(new SyncItemTar(
+      parent_path, filename, src, entry, read_archive_signal_, this));
+
+  if (NULL != archive_entry_hardlink(entry)) {
+    const std::string hardlink =
+        base_directory_ + "/" + std::string(archive_entry_hardlink(entry));
+
+    printf("Hardlink %s -> %s\n", hardlink.c_str(), complete_path.c_str());
```
jblomer: This should become a `LogCvmfs` statement, probably with increased verbosity level

cvmfs/sync_union_tarball.cc
```diff
@@ -199,6 +164,74 @@ void SyncUnionTarball::Traverse() {
...
+      sync_entry->IsPlaceholderDirectory();
+    }
+    ProcessUnmaterializedDirectory(sync_entry);
+    dirs_[complete_path] = sync_entry;
+    know_directories_.insert(complete_path);
+
+    read_archive_signal_->Wakeup();  // We don't need to read data and we
+                                     // can read the next header
+
+  } else if (sync_entry->IsRegularFile()) {
```
jblomer: I think we should combine this case and the following case on, if the file is `.cvmfscatalog`, validate that we deal with a regular file.  Otherwise one could add a symlink .cvmfscatalog without magic behavior, which would be confusing.

cvmfs/sync_union_tarball.cc
```diff
@@ -199,6 +164,74 @@ void SyncUnionTarball::Traverse() {
...
+    ProcessFile(sync_entry);
+    if (filename == ".cvmfscatalog") {
+      to_create_catalog_dirs_.insert(parent_path);
+    }
+  } else if (sync_entry->IsSymlink() || sync_entry->IsFifo() ||
+             sync_entry->IsSocket()) {
+    ProcessFile(sync_entry);
+    read_archive_signal_->Wakeup();
+  } else {
+    read_archive_signal_->Wakeup();
```
jblomer: Shouldn't we have an `abort()` or an error message here?  I guess in this branch we deal with an unexpected file.

cvmfs/sync_union_tarball.h
```diff
@@ -62,12 +65,18 @@ class SyncUnionTarball : public SyncUnion {
   const std::string to_delete_;  ///< entity to delete before to extract the tar
   std::set<std::string>
       know_directories_;  ///< directory that we know already exist
-  std::set<std::string> to_create_catalog_dirs_;
+
   ///< directories where we found catalog marker, after the main traverse we
```
jblomer: This might be broken doxygen syntax.  Multiline comments need to be in the format

    /**
     * Comment
     */

cvmfs/sync_union_tarball.h
```diff
@@ -62,12 +65,18 @@ class SyncUnionTarball : public SyncUnion {
...
   ///< iterate through them and we add the catalog
-  std::map<std::string, SharedPtr<SyncItem> > dirs_;
+  std::set<std::string> to_create_catalog_dirs_;
+
   ///< map of all directories found, we need them since we don't know, at
```
jblomer: Same: Doxygen syntax

cvmfs/sync_union_tarball.h
```diff
@@ -62,12 +65,18 @@ class SyncUnionTarball : public SyncUnion {
...
   ///< priori, where the catalog files appears
+  std::map<std::string, SharedPtr<SyncItem> > dirs_;
+
+  ///< hardlink -> path of file that points to the same hardlink
```
jblomer: Doxygen syntax

test/src/654-tarball_ingest_special_files/main
```diff
@@ -34,6 +36,9 @@ produce_tarball() {
 
   mkfifo tarball_foo/fifo
 
+  ln tarball_foo/empty_file.txt tarball_foo/to_empty_hardlink.txt
+  ln tarball_foo/d/1.txt tarball_foo/d_1_hardlink.txt
```
jblomer: I think it would be good to have a hardlinke with more than 2 files, too.

cvmfs/sync_union_tarball.cc
siscia: I am quite keen to just remove this line... Do you think is necessary?

jblomer: Removing is fine, I think.

cvmfs/sync_union_tarball.cc
siscia: I am not quite sure I follow you on this one.

At the moment:
1. We check that it is a regular file
2. We process it
3. If is a catalog marker we keep track of it

A symlink to a catalog marker should not enter this branch at all, am I wrong?

Are you suggesting to don't process the symlink if it points to a catalog marker?

siscia:
@jblomer is there anything missing from this PR?

cvmfs/catalog_mgr_rw.cc
```diff
@@ -287,6 +287,42 @@ void WritableCatalogManager::RemoveDirectory(const std::string &path) {
...
+  SplitPath(destination, &destination_dirname, &destination_filename);
+
+  DirectoryEntry to_dirent;
+  if (!LookupPath(relative_source, kLookupSole, &to_dirent)) {
+    LogCvmfs(kLogCatalog, kLogStderr,
+             "catalog for file '%s' cannot be found aborting",
+             source.c_str());
+    assert(false);
+  }
+  if (!to_dirent.IsRegular()) {
```
jblomer: I think it should be `to_dirent.IsDirectory()` because we can have hardlinks/clones of symlinks and special files.

cvmfs/sync_union_tarball.cc
```diff
@@ -206,14 +206,28 @@ void SyncUnionTarball::ProcessArchiveEntry(struct archive_entry *entry) {
...
              sync_entry->IsSocket()) {
-    ProcessFile(sync_entry);
+    // we avoid to add an entity called as a catalog marker if it is not a
+    // regular file
+    if (filename != ".cvmfscatalog") {
+      ProcessFile(sync_entry);
+    } else {
+      LogCvmfs(kLogUnionFs, kLogStderr,
+               "Found entity called as a catalog marker '%s' that however is "
+               "not a regular file, skipping it", complete_path.c_str());
```
jblomer: I'd `abort()` here, if we see such files, chances are that there is a deeper problem that the user should not ignore.

cvmfs/sync_union_tarball.h
```diff
@@ -66,19 +66,28 @@ class SyncUnionTarball : public SyncUnion {
   std::set<std::string>
       know_directories_;  ///< directory that we know already exist
 
-  ///< directories where we found catalog marker, after the main traverse we
-  ///< iterate through them and we add the catalog
+  /** 
+   * directories where we found catalog marker, after the main traverse we
+   * iterate through them and we add the catalog
+   **/
```
jblomer: The Doxygen syntax uses a single star at the end, I think.  Like

    /**
     * Comment
     */

jblomer:
Ok, I think we are almost there!

PR #2151, opened 2018-06-25
Configuration templating variables

samysweb:
Within the configuration file values it is now possible to use variables by inserting `@foo@` where `foo` is the name of a variable.
The first variable available across all configuration files (where applicable[1]) is `@fqrn@` which is resolved to the repository name.
This will enable repository dependent configuration values and will be used for the tracer (to allow one trace log per repository).

To make this change in a way that new configuration template variables can easily be added, an `OptionsTemplatingManager` was introduced which manages the available templating variables and is passed to the `ParsePath` methods of the `OptionsManager`s.

The unit tests in `T_Options` were extended to check the behaviour described above.

[1] It seemed to me that there is no fqrn available in cvmfs/cache_plugin/libcvmfs_cache_options.cc and cvmfs/libcvmfs_options.cc. Therefore the fqrn is not resolved when the configuration is parsed through invocation of one of these 2 functions.

### DIFF
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1388,9 +1388,11 @@ void MountPoint::CreateTables() {
 
 /**
  * Will create a tracer for the current mount point
- * Tracefile path, Trace buffer size and trace buffer flush threshold can be configured by the options:
- * CVMFS_TRACEFILE, CVMFS_TRACEBUFFER, CVMFS_TRACEBUFFER_THRESHOLD (respectively)
- * VMFS_TRACEBUFFER and CVMFS_TRACEBUFFER_THRESHOLD will silently fallback to default values if configuration values don't exist or are invalid
+ * Tracefile path, Trace buffer size and trace buffer flush threshold
+ * can be configured by the options: CVMFS_TRACEFILE,
+ * CVMFS_TRACEBUFFER, CVMFS_TRACEBUFFER_THRESHOLD(respectively)
+ * VMFS_TRACEBUFFER and CVMFS_TRACEBUFFER_THRESHOLD will silently fallback
+ * to default values if configuration values don't exist or are invalid
  */
 bool MountPoint::CreateTracer() {
   string optarg;
@@ -1406,11 +1408,11 @@ bool MountPoint::CreateTracer() {
     uint64_t tracebuffer_threshold = kTracerFlushThreshold;
 
     if (options_mgr_->GetValue("CVMFS_TRACEBUFFER", &optarg)) {
-      tracebuffer_size = String2Uint64(optarg.c_str());
+      tracebuffer_size = String2Uint64(optarg);
     }
     if (options_mgr_->GetValue("CVMFS_TRACEBUFFER_THRESHOLD",
       &optarg)) {
-      tracebuffer_threshold = String2Uint64(optarg.c_str());
+      tracebuffer_threshold = String2Uint64(optarg);
     }
     assert(tracebuffer_size <= INT_MAX
       && tracebuffer_threshold <= INT_MAX);
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -11,8 +11,8 @@
 #endif
 #include <inttypes.h>
 #include <stdint.h>
-#include <unistd.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #include <algorithm>
 #include <cassert>
@@ -1126,7 +1126,7 @@ MountPoint *MountPoint::Create(
     return mountpoint.Release();
   if (!mountpoint->CreateTracer())
     return mountpoint.Release();
-  
+
   mountpoint->ReEvaluateAuthz();
   mountpoint->CreateTables();
   mountpoint->SetupBehavior();
@@ -1406,13 +1406,15 @@ bool MountPoint::CreateTracer() {
 
     string tracebufferSizeOpt;
     string tracebufferThresholdOpt;
-    if(options_mgr_->GetValue("CVMFS_TRACEBUFFER", &tracebufferSizeOpt)){
+    if (options_mgr_->GetValue("CVMFS_TRACEBUFFER", &tracebufferSizeOpt)) {
       tracebufferSize = atoi(tracebufferSizeOpt.c_str());
     }
-    if(options_mgr_->GetValue("CVMFS_TRACEBUFFER_THRESHOLD", &tracebufferThresholdOpt)){
+    if (options_mgr_->GetValue("CVMFS_TRACEBUFFER_THRESHOLD",
+      &tracebufferThresholdOpt)) {
       tracebufferThreshold = atoi(tracebufferThresholdOpt.c_str());
     }
-    LogCvmfs(kLogCvmfs, kLogDebug, "Initialising tracer with buffer size %i and threshold %i",
+    LogCvmfs(kLogCvmfs, kLogDebug,
+      "Initialising tracer with buffer size %i and threshold %i",
       tracebufferSize, tracebufferThreshold);
     tracer_->Activate(tracebufferSize, tracebufferThreshold, optarg);
   }
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -12,6 +12,7 @@
 #include <inttypes.h>
 #include <stdint.h>
 #include <unistd.h>
+#include <stdlib.h>
 
 #include <algorithm>
 #include <cassert>
@@ -1125,7 +1126,7 @@ MountPoint *MountPoint::Create(
     return mountpoint.Release();
   if (!mountpoint->CreateTracer())
     return mountpoint.Release();
-
+  
   mountpoint->ReEvaluateAuthz();
   mountpoint->CreateTables();
   mountpoint->SetupBehavior();
@@ -1385,7 +1386,12 @@ void MountPoint::CreateTables() {
   inode_tracker_ = new glue::InodeTracker();
 }
 
-
+/**
+ * Will create a tracer for the current mount point
+ * Tracefile path, Trace buffer size and trace buffer flush threshold can be configured by the options:
+ * CVMFS_TRACEFILE, CVMFS_TRACEBUFFER, CVMFS_TRACEBUFFER_THRESHOLD (respectively)
+ * VMFS_TRACEBUFFER and CVMFS_TRACEBUFFER_THRESHOLD will silently fallback to default values if configuration values don't exist or are invalid
+ */
 bool MountPoint::CreateTracer() {
   string optarg;
   tracer_ = new Tracer();
@@ -1395,7 +1401,20 @@ bool MountPoint::CreateTracer() {
       boot_status_ = loader::kFailOptions;
       return false;
     }
-    tracer_->Activate(kTracerBufferSize, kTracerFlushThreshold, optarg);
+    int tracebufferSize = kTracerBufferSize;
+    int tracebufferThreshold = kTracerFlushThreshold;
+
+    string tracebufferSizeOpt;
+    string tracebufferThresholdOpt;
+    if(options_mgr_->GetValue("CVMFS_TRACEBUFFER", &tracebufferSizeOpt)){
+      tracebufferSize = atoi(tracebufferSizeOpt.c_str());
+    }
+    if(options_mgr_->GetValue("CVMFS_TRACEBUFFER_THRESHOLD", &tracebufferThresholdOpt)){
+      tracebufferThreshold = atoi(tracebufferThresholdOpt.c_str());
+    }
+    LogCvmfs(kLogCvmfs, kLogDebug, "Initialising tracer with buffer size %i and threshold %i",
+      tracebufferSize, tracebufferThreshold);
+    tracer_->Activate(tracebufferSize, tracebufferThreshold, optarg);
   }
   return true;
 }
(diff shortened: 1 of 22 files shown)


cvmfs/options.h
```diff
@@ -15,6 +15,25 @@
 namespace CVMFS_NAMESPACE_GUARD {
 #endif
 
+/**
+ * Templating manager used for variable replacement in the config file
+ */
+class OptionsTemplatingManager{
```
jblomer: A space before `{`

cvmfs/options.h
```diff
@@ -15,6 +15,25 @@
...
+ public:
+  void SetVal(std::string name, std::string val);
+  std::string GetVal(std::string name);
+  bool HasVal(std::string name);
+ private:
+    std::map<std::string, std::string> vars;
```
jblomer: Naming style: `vars_` (perhaps better: `templates_`)

cvmfs/options.cc
```diff
@@ -311,14 +320,34 @@ void OptionsManager::PopulateParameter(
...
+  if (opt_templ_mgr == NULL) {
+    // No template manager -> nothing to do here
+    return;
+  }
+  vector<string> tokens = SplitString(val->value, '@');
+  if (tokens.size() > 2) {
+    string resStr = "";
+    for (unsigned int i = 0; i < tokens.size(); i++) {
+      if (i%2 == 0) {
+        resStr+=tokens[i];
```
jblomer: Naming convention: `res_str` (or perhaps: `result`), white spaces around operators

cvmfs/options.cc
```diff
@@ -311,14 +320,34 @@ void OptionsManager::PopulateParameter(
...
+      } else {
+        resStr+=opt_templ_mgr->GetVal(tokens[i]);
```
jblomer: white spaces around operator

jblomer: I think we should through a `LogCvmfs` statement, at least at the debug level, if there is an unrecognized template identifier.

cvmfs/options.cc
```diff
@@ -311,14 +320,34 @@ void OptionsManager::PopulateParameter(
...
   }
 }
 
+void OptionsManager::ParseValue(ConfigValue *val,
+  OptionsTemplatingManager *opt_templ_mgr) {
+  if (opt_templ_mgr == NULL) {
+    // No template manager -> nothing to do here
+    return;
+  }
+  vector<string> tokens = SplitString(val->value, '@');
```
jblomer: Apologies: I forgot that we already have a `ReplaceAll` utility function in util/string.cc. So you can simply iterate through all available template identifier and replace them.  In fact, this could be done in the `OptionsTemplateManager` that can iterate over all the templates.

cvmfs/options.cc
```diff
@@ -434,6 +463,26 @@ void OptionsManager::UnsetValue(const string &key) {
     unsetenv(key.c_str());
 }
 
+const std::string DefaultOptionsTemplatingManager
+  ::fqrnTemplateIdentifier = "fqrn";
+DefaultOptionsTemplatingManager::DefaultOptionsTemplatingManager(
```
jblomer: Empty line before

cvmfs/options.h
```diff
@@ -162,7 +183,9 @@ class OptionsManager {
  *  "KEY=VALUE".
  *
  *  No comments (#) are allowed.
+ *  No templating variables (@foo@) are allowed.
```
jblomer: This comment is not (anymore) accurate, is it?

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,22 @@ bool MountPoint::CreateTracer() {
       boot_status_ = loader::kFailOptions;
       return false;
     }
-    tracer_->Activate(kTracerBufferSize, kTracerFlushThreshold, optarg);
+    int tracebufferSize = kTracerBufferSize;
```
jblomer: Naming style: `tracebuffer_size`

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,22 @@ bool MountPoint::CreateTracer() {
...
+    int tracebufferThreshold = kTracerFlushThreshold;
```
jblomer: `tracebuffer_threshold`

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,22 @@ bool MountPoint::CreateTracer() {
...
+
+    string tracebufferSizeOpt;
```
jblomer: For consistency, should be changed to reuse `optarg` for the options values.  The previously stored trace file name should then be saved in a separate variable.

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,22 @@ bool MountPoint::CreateTracer() {
...
+    string tracebufferThresholdOpt;
+    if (options_mgr_->GetValue("CVMFS_TRACEBUFFER", &tracebufferSizeOpt)) {
+      tracebufferSize = atoi(tracebufferSizeOpt.c_str());
```
jblomer: `atoi` --> `String2Uint64`

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,22 @@ bool MountPoint::CreateTracer() {
...
+    }
+    if (options_mgr_->GetValue("CVMFS_TRACEBUFFER_THRESHOLD",
+      &tracebufferThresholdOpt)) {
+      tracebufferThreshold = atoi(tracebufferThresholdOpt.c_str());
```
jblomer: `atoi` --> `String2Uint64`

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,22 @@ bool MountPoint::CreateTracer() {
...
+    }
+    LogCvmfs(kLogCvmfs, kLogDebug,
+      "Initialising tracer with buffer size %i and threshold %i",
```
jblomer: `int` --> `unsigned`

samysweb:
OK, thank you for the feedback! I'm working on including the requested changes now...

cvmfs/options.h
```diff
@@ -15,6 +15,25 @@
...
+class OptionsTemplatingManager{
+ public:
+  void SetVal(std::string name, std::string val);
+  std::string GetVal(std::string name);
+  bool HasVal(std::string name);
+ private:
+    std::map<std::string, std::string> vars;
+};
+
+class DefaultOptionsTemplatingManager : public OptionsTemplatingManager {
```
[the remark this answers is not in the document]
samysweb: So should the `OptionsTemplatingManager` then take the fqrn as a constructor argument or should I just use the "normal" `OptionsTemplatingManager` and then manually set the fqrn template variable?
I'm asking because I would have added the @org@ extraction logic to the `DefaultOptionsTemplatingManager` constructor otherwise.

cvmfs/options.cc
samysweb: In this case I would not be able to detect undefined template variables though (previous note). Depends on what is more important...
Alternatively one could use the simple finite state machine system that You had already mentioned (probably best implemented in the `OptionsTemplateManager` then).

cvmfs/options.h
samysweb: Yes, this was meant to be removed already...

samysweb:
Concerning the `SwitchTemplateManager` method: Should this reevaluate everything or only symbols which have not previously been resolved? If we were to go for a full reevaluation I would propose to evaluate the templating variables upon retrieval in `GetValue` (not during charging). This would avoid having to store all configuration options twice (once in its template variable resolved version, once in the original version).
This is only feasible though if the calls to `GetValue` mostly happen during charging (not during the file system system calls)

samysweb:
@jblomer  This should include all requested changes now...

samysweb:
Please do not merge yet, since I found out that the current implementation of this branch is breaking the Tracer functionality in the following pull request (and crashes the program on an assertion if the Tracer is being activated even in this version).
I'm underway of fixing this issue...

samysweb:
Bug is fixed.

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,24 @@ bool MountPoint::CreateTracer() {
       boot_status_ = loader::kFailOptions;
       return false;
     }
-    tracer_->Activate(kTracerBufferSize, kTracerFlushThreshold, optarg);
+    string tracebuffer_file = optarg;
+    uint64_t tracebuffer_size = kTracerBufferSize;
+    uint64_t tracebuffer_threshold = kTracerFlushThreshold;
+
+    if (options_mgr_->GetValue("CVMFS_TRACEBUFFER", &optarg)) {
+      tracebuffer_size = String2Uint64(optarg.c_str());
```
jblomer: `String2Uint64` takes a `std::string`.

cvmfs/mountpoint.cc
```diff
@@ -1395,7 +1401,24 @@ bool MountPoint::CreateTracer() {
...
+    }
+    if (options_mgr_->GetValue("CVMFS_TRACEBUFFER_THRESHOLD",
+      &optarg)) {
+      tracebuffer_threshold = String2Uint64(optarg.c_str());
```
jblomer: - `c_str()`

cvmfs/mountpoint.cc
```diff
@@ -1385,7 +1386,12 @@ void MountPoint::CreateTables() {
   inode_tracker_ = new glue::InodeTracker();
 }
 
-
+/**
+ * Will create a tracer for the current mount point
+ * Tracefile path, Trace buffer size and trace buffer flush threshold can be configured by the options:
```
jblomer: Line wrap for maximum 80 chars / line

cvmfs/mountpoint.h
```diff
@@ -469,6 +469,10 @@ class MountPoint : SingleCopy, public BootFactory {
   void CreateFetchers();
   bool CreateCatalogManager();
   void CreateTables();
+  /**
```
jblomer: We should have only one comment for the method, in the .cc file.  This keeps the header quickly digestible and the comments all in one place.

cvmfs/options.cc
```diff
@@ -71,6 +71,23 @@ string OptionsManager::TrimParameter(const string &parameter) {
...
+    opt_templ_mgr_ = opt_templ_mgr_param;
+  } else {
+    opt_templ_mgr_ = new OptionsTemplateManager();
+  }
+  for (std::map<std::string, std::string>::iterator it
+    = templatable_values_.begin();
+    it != templatable_values_.end();
+    it++) {
+    config_[it->first].value = it->second;
+    opt_templ_mgr_->ParseString(&(config_[it->first].value));
```
jblomer: Shouldn't this be a call to `PopulateParameter` to change the environment as well?

cvmfs/options.cc
```diff
@@ -434,6 +457,72 @@ void OptionsManager::UnsetValue(const string &key) {
     unsetenv(key.c_str());
 }
 
+const std::string DefaultOptionsTemplateManager
```
jblomer: Ah, I forgot about this the first time: for the `const static` members, let's use only plain old data types, that is in this case `char *`. The reason is that they get constructed at some, unclear point at program start, and with more complex C++ objects being constructed at this point, funny things can happen.

cvmfs/options.cc
```diff
@@ -434,6 +457,72 @@ void OptionsManager::UnsetValue(const string &key) {
...
+  bool has_vars = false;
+  int mode = 0;
+  std::string stock;
+  for (std::string::size_type i = 0; i < in.size(); i++) {
+    switch (mode) {
+      case 0:
+        if (in[i] == '@') {
+          mode = 1;
+        } else {
+          result+=in[i];
```
jblomer: space around operators

cvmfs/options.cc
```diff
@@ -434,6 +457,72 @@ void OptionsManager::UnsetValue(const string &key) {
...
+          result += GetTemplate(stock);
+          stock = "";
+          has_vars = true;
+        } else {
+          stock += in[i];
+        }
+      break;
+    }
+  }
+  if (mode == 1) {
```
jblomer: Good thinking!

cvmfs/options.h
```diff
@@ -24,8 +46,31 @@ namespace CVMFS_NAMESPACE_GUARD {
...
+  explicit OptionsManager(OptionsTemplateManager *opt_templ_mgr_param)
+      : taint_environment_(true) {
+    if (opt_templ_mgr_param != NULL) {
+      opt_templ_mgr_ = opt_templ_mgr_param;
+    } else {
+      opt_templ_mgr_ = new OptionsTemplateManager();
+    }
+  }
+
+  OptionsManager(const OptionsManager& opt_mgr) {
```
jblomer: `taint_environment` is missing.

cvmfs/options.cc
samysweb: I have just submitted a commit including the requested changes. Concerning this change request I believe it could be better to have an extra method which changes the parameter based environment variables. The disadvantage of using `PopulateParameter` would be that a protected parameters can no longer be updated based on template variables.

For now I have implemented the method I proposed above, but I can also change this in the way described you described...

cvmfs/options.h
```diff
@@ -55,7 +55,8 @@ class OptionsManager {
     }
   }
 
-  OptionsManager(const OptionsManager& opt_mgr) {
+  OptionsManager(const OptionsManager& opt_mgr)
+    : taint_environment_(true) {
```
jblomer: What I meant is: wouldn't it need to be

    taint_environment_ = opt_mgr.taint_environment_

?

samysweb: Yes, absolutely. Sorry about that - just made a push that fixes this.

PR #2155, opened 2018-06-26
Add libcvmfs calls that list and stat nested catalogs

nhazekam:
Added calls for getting information about nested catalogs using libcvmfs. Also added were several tests for the new calls.

PR #2157, opened 2018-06-26
Add extended cvmfs stat call to libcvmfs

nhazekam:
This addition to libcvmfs provides a functions for retrieving not only the stat information, but also information specific to CVMFS like checksum and the stored names. This also adds tests for stat and ext_stat in t_libcvmfs.

Note: this coincides with #2155 so one will need to be rebased off the other prior to merging.

### DIFF
--- a/cvmfs/directory_entry.h
+++ b/cvmfs/directory_entry.h
@@ -205,16 +205,22 @@ class DirectoryEntryBase {
   inline struct cvmfs_stat GetCVMFSStatStructure() const {
     struct cvmfs_stat s;
     memset(&s, 0, sizeof(s));
-    s.version = 1;
-    s.struct_size = sizeof(s);
-    s.inode = inode_;
-    s.mode = mode_;
-    s.linkcount = linkcount();
-    s.uid = uid();
-    s.gid = gid();
-    s.size = size();
+    s.version  = 1;
+    s.size     = sizeof(s);
+    s.st_ino   = inode_;
+    s.st_mode  = mode_;
+    s.st_nlink = linkcount();
+    s.st_uid   = uid();
+    s.st_gid   = gid();
+    s.st_rdev  = rdev();
+    s.st_size  = size();
+    s.st_blksize = 4096;  // will be ignored by Fuse
+    s.st_blocks = 1 + size() / 512;
     s.mtime = mtime_;
-    s.checksum = (const void *)&checksum_;
+    s.cvm_checksum = (const void *)&checksum_;
+    s.cvm_symlink  = symlink_.ToString().c_str();
+    s.cvm_name     = name_.ToString().c_str();
+    s.cvm_has_xattr = has_xattrs_;
     return s;
   }
 
--- a/cvmfs/directory_entry.h
+++ b/cvmfs/directory_entry.h
@@ -220,7 +220,7 @@ class DirectoryEntryBase {
     s.cvm_checksum = (const void *)&checksum_;
     s.cvm_symlink  = symlink_.ToString().c_str();
     s.cvm_name     = name_.ToString().c_str();
-    s.cvm_has_xattr = has_xattrs_;
+    s.cvm_xattrs   = NULL;
     return s;
   }
 
--- a/cvmfs/directory_entry.h
+++ b/cvmfs/directory_entry.h
@@ -202,26 +202,23 @@ class DirectoryEntryBase {
    * information such as hash.
    * @return the struct cvmfs_stat for this DirectoryEntry
    */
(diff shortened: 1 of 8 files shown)


siscia:
I find curious that you opted for `char*` instead of `std::string`, is there any technical reason?

nhazekam:
If you are referring inside of the `cvmfs_stat` structure, I am using chars so that this is accessible in c, not just c++.  The exceptions are for checksum and xattr so that the CVMFS structured data remains. If a c application were interested in this data, there would likely have to be a conversion or helper function to access the data.

nhazekam:
@jblomer Can you take a look when you get the chance.

cvmfs/directory_entry.h
```diff
@@ -195,6 +196,31 @@ class DirectoryEntryBase {
     return s;
   }
 
+  /**
+   * Converts to a cvmfs_stat struct. This provides the
+   * information provided by stat, but also CVMFS specific
+   * information such as hash.
+   * @return the struct cvmfs_stat for this DirectoryEntry
+   */
+  inline void GetCVMFSStatStructure(struct cvmfs_attr *attr) const {
```
jblomer: I think we should move this to libcvmfs, and convert from the other side, like `cvmfs_attr::FromDirent`

cvmfs/directory_entry.h
```diff
@@ -18,6 +18,7 @@
 #include "bigvector.h"
 #include "compression.h"
 #include "hash.h"
+#include "libcvmfs.h"
```
jblomer: It might work fine but I'd prefer not to include this header in the internal code.

cvmfs/directory_entry.h
```diff
@@ -195,6 +196,31 @@ class DirectoryEntryBase {
...
+    attr->size     = sizeof(*attr);
+    attr->st_ino   = inode_;
+    attr->st_mode  = mode_;
+    attr->st_nlink = linkcount();
+    attr->st_uid   = uid();
+    attr->st_gid   = gid();
+    attr->st_rdev  = rdev();
+    attr->st_size  = size();
+    attr->mtime    = mtime_;
+    attr->st_blksize = 4096;  // will be ignored by Fuse
```
jblomer: Perhaps we can skip this and the next entry, as they are anyway meaningless

test/common/catalog_test_tools.h
```diff
@@ -127,6 +126,7 @@ class CatalogTestTool : public ServerTool {
   bool Apply(const std::string& id, const DirSpec& spec);
   bool ApplyAtRootHash(const shash::Any& root_hash, const DirSpec& spec);
   bool AddNestedCatalog(const shash::Any& root_hash, const std::string& path);
+  bool FindEntry(const shash::Any& root_hash, const std::string& path, catalog::DirectoryEntry *entry);
```
jblomer: Line break to stay within the 80 chars/line limit.

test/unittests/t_libcvmfs.cc
```diff
@@ -63,10 +69,16 @@ class T_Libcvmfs : public ::testing::Test {
...
 
+  SimpleOptionsParser options_mgr_;
+
   string tmp_path_;
+  string repo_path_;
   string alien_path_;
   string opt_cache_;
   FILE *fdevnull_;
+
+  int fd_cwd_;
```
jblomer: This and the following member are unused, I think.

test/unittests/t_libcvmfs.cc
```diff
@@ -264,3 +278,164 @@ TEST_F(T_Libcvmfs, Templating) {
...
+  EXPECT_FALSE(attr->cvm_xattrs);
+  cvmfs_attr_free(attr);
+
+  /* Lookup non-existent file */
+  attr = cvmfs_attr_init();
+  retval = cvmfs_stat_attr(ctx, "/dir/file40", attr);
+  EXPECT_EQ(-1, retval);
+  cvmfs_attr_free(attr);
+
+  /* Finalize and close repo and options */
```
jblomer: Missing `cvmfs_detach_repo()`

nhazekam:
@jblomer I think this is ready for another look. This was also rebased on the recent changes and addresses a similar issue in the test. I also moved the version and struct size to the constructor function for `cvmfs_attr`. If this should still be in the setting function please let me know.

nhazekam:
@jblomer I updated the calls you mentioned. Also, I agree and updated `cvmfs_stat_attr` so that it looks at the link information. I checked that the test would fail if the call resolved the link, but if there are additional checks you would like me to add please let me know.

PR #2159, opened 2018-07-03
S3: Improve logging and make network timeouts configurable

jblomer:
New configuration parameter: `CVMFS_S3_MAX_RETRIES` and `CVMFS_S3_MAX_TIMEOUT`


jblomer:
Should be opened against 2.5 branch

PR #2152, opened 2018-06-25
Tracer implementation

samysweb:
**Note: This pull request should only be considered once #2151 is merged**

This pull request contains the implementation of a tracer which allows the tracing of the following file system operations:
- `opendir` and `open`
- `lookup`
- `statfs` and `getattr`
- `listxattr` and `getxattr`
- `readlink`

The tracer was implemented in a way that should minimize the overhead when the tracer is deactivated and used the base implementation already available in `tracer.cc`.

An integration test was added to check the functionality described above.

The tracer can be activated by defining `CVMFS_TRACEFILE` in the configuration file.
Additionally `CVMFS_TRACEBUFFER` (size of the tracebuffer) and `CVMFS_TRACEBUFFER_THRESHOLD` (threshold for a tracebuffer flush) can be defined to further customize the behaviour of the tracer.

PR #2161, opened 2018-07-03
Feature tracer ready

samysweb:
Newer version of pull request #2152 (based on results of #2151)

**Quote:**
This pull request contains the implementation of a tracer which allows the tracing of the following file system operations:
- `opendir` and `open`
- `lookup`
- `statfs` and `getattr`
- `listxattr` and `getxattr`
- `readlink`

The tracer was implemented in a way that should minimize the overhead when the tracer is deactivated and used the base implementation already available in `tracer.cc`.

An integration test was added to check the functionality described above.

The tracer can be activated by defining `CVMFS_TRACEFILE` in the configuration file.
Additionally `CVMFS_TRACEBUFFER` (size of the tracebuffer) and `CVMFS_TRACEBUFFER_THRESHOLD` (threshold for a tracebuffer flush) can be defined to further customize the behaviour of the tracer.


samysweb:
@jblomer This is ready for review

cvmfs/cvmfs.cc
```diff
@@ -487,6 +506,7 @@ static void cvmfs_getattr(fuse_req_t req, fuse_ino_t ino,
 
   fuse_remounter_->fence()->Enter();
   ino = mount_point_->catalog_mgr()->MangleInode(ino);
+  TraceInode(Tracer::kEventGetAttr, ino, "getattr()");
```
jblomer: I think this should move down, until after the checks and just before leaving the fence.

cvmfs/cvmfs.cc
```diff
@@ -338,6 +337,26 @@ static bool GetPathForInode(const fuse_ino_t ino, PathString *path) {
...
+  bool found = GetPathForInode(ino, &path);
+  if (!found) {
+    LogCvmfs(kLogCvmfs, kLogDebug,
+      "Tracing: Could not find path for inode %" PRIu64, uint64_t(ino));
+  } else {
+    mount_point_->tracer()->Trace(event, path, msg);
+  }
+}
+
+static void inline TraceInode(const int event,
```
jblomer: Because this is a common pattern, I'd add a `bool found_inode` to the parameter list and only trace if the inode is actually valid. This avoids the `if (found) ...` in all the places where TraceInode is called.

cvmfs/cvmfs.cc
```diff
@@ -581,7 +601,7 @@ static void cvmfs_opendir(fuse_req_t req, fuse_ino_t ino,
   ino = catalog_mgr->MangleInode(ino);
   LogCvmfs(kLogCvmfs, kLogDebug, "cvmfs_opendir on inode: %" PRIu64,
            uint64_t(ino));
-
+  TraceInode(Tracer::kEventOpenDir, ino, "opendir()");
```
jblomer: Move down below the `LogCvmfs` statement

cvmfs/cvmfs.cc
```diff
@@ -788,6 +808,7 @@ static void cvmfs_open(fuse_req_t req, fuse_ino_t ino,
     fuse_reply_err(req, ENOENT);
     return;
   }
+  mount_point_->tracer()->Trace(Tracer::kEventOpen, path, "open()");
```
jblomer: Move down under the `CheckVoms` block

cvmfs/cvmfs.cc
```diff
@@ -1178,6 +1199,8 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   struct statvfs info;
   memset(&info, 0, sizeof(info));
 
+  TraceInode(Tracer::kEventStat, ino, "statfs()");
```
jblomer: We should change the constant to `kEventStatFs` to avoid confusion with the stating of files.

cvmfs/cvmfs.cc
```diff
@@ -1232,7 +1255,7 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   LogCvmfs(kLogCvmfs, kLogDebug,
            "cvmfs_getxattr on inode: %" PRIu64 " for xattr: %s",
            uint64_t(ino), name);
-
+  TraceInode(Tracer::kEventGetAttr, ino, "getxattr()");
```
jblomer: Move down after `CheckVoms`

python/spec_builder.py
```diff
@@ -0,0 +1,152 @@
+import argparse
```
jblomer: Add our standard comment "This file is part of the CernVM File System"

python/spec_test.py
```diff
@@ -0,0 +1,76 @@
...
+    help="Calls which should be filtered")
+  return argparser.parse_args()
+
+def main():
+  args = parse_args()
+  traceParser = TraceParser(args)
+  traceParser.read_log()
+
+if __name__ == "__main__":
+    main()
```
jblomer: Missing newline at the end of the file

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
+cvmfs_test_name="tracing"
+
+cvmfs_run_test() {
+    logfile=$1
+    sudo sh -c "echo \"export GLITE_VERSION='util'\" > /etc/cvmfs/config.d/grid.cern.ch.local"
+    repo="grid.cern.ch"
+    echo "Setup..."
+    rm -f /tmp/cvmfs-$repo.trace.log 2> /dev/null
+    echo "Mounting test repository..."
+    cvmfs_mount $repo "CVMFS_TRACEFILE=/tmp/cvmfs-@fqrn@.trace.log" "CVMFS_TRACEBUFFER=4096" "CVMFS_TRACEBUFFER_THRESHOLD=2048" > /dev/null
```
jblomer: It should be possible to add `GLITE_VERSION=util` as parameter to `cvmfs_mount`

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
+cvmfs_test_name="tracing"
+
+cvmfs_run_test() {
+    logfile=$1
+    sudo sh -c "echo \"export GLITE_VERSION='util'\" > /etc/cvmfs/config.d/grid.cern.ch.local"
+    repo="grid.cern.ch"
```
jblomer: `local repo=..`

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
...
+    echo "Setup..."
```
jblomer: Prefixing the explicit echo statements with `*** ` makes them easier to find in the output log

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
...
+    rm -f /tmp/cvmfs-$repo.trace.log 2> /dev/null
```
jblomer: `sudo rm -f ...` ?

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
...
+    echo "Mounting test repository..."
+    cvmfs_mount $repo "CVMFS_TRACEFILE=/tmp/cvmfs-@fqrn@.trace.log" "CVMFS_TRACEBUFFER=4096" "CVMFS_TRACEBUFFER_THRESHOLD=2048" > /dev/null
+
+    echo "Checking trace log creation..."
+    sudo cvmfs_talk tracebuffer flush > /dev/null
```
jblomer: `cvmfs_talk -i $repo ...`

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
...
+
+    if [ ! -f /tmp/cvmfs-$repo.trace.log ]; then
+        echo "ERROR: Log file was not created"
+        return 2
+    fi
+
+    cp /dev/null /tmp/cvmfs-$repo.trace.log > /dev/null
```
jblomer: `sudo truncate -s0 /tmp/...`

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
...
+
+    echo "Executing a few operations to check tracing..."
+
+    ls /cvmfs/$repo/ > /dev/null
+    ls /cvmfs/$repo/vc/bin > /dev/null
+    cat /cvmfs/$repo/README > /dev/null
+    df -P /cvmfs/$repo/README > /dev/null
+    getfattr -d -m ".*" /cvmfs/$repo/README > /dev/null
```
jblomer: There is a helper function for getting extended attributes because it's different on Linux and Mac

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
...
+    if [ ! -f /tmp/cvmfs-$repo.trace.log ]; then
+        echo "ERROR: Log file was not created"
+        return 2
+    fi
+
+    cp /dev/null /tmp/cvmfs-$repo.trace.log > /dev/null
+
+    echo "Executing a few operations to check tracing..."
+
+    ls /cvmfs/$repo/ > /dev/null
```
jblomer: You can leave the output of the commands in, it can only help

test/src/080-tracing/main
```diff
@@ -0,0 +1,61 @@
...
+
+        ); then 
+            return 3
+    fi
+    
+    return 0
+}
+
+check_trace() {
+    if ! grep -q "\"$1\",\"$3\",\"$2()\"" /tmp/cvmfs-$repo.trace.log; then
```
jblomer: `sudo grep`?

test/src/080-tracing/main
jblomer: In addition, I think it would be good to set up a signal handler later on that removes the file when the function exists.  This can be done with `trap`.  You can check test case 011 for a blue print.

samysweb:
@jblomer These commits should implement all requested changes.

PR #2163, opened 2018-07-04
Fix Libcvmfs test for Fedora and Ubuntu

nhazekam:
The issue appears to be that the tester instance was being destructed, leaving the d0_hash pointing to nothing. This could be resolved in a number of ways (strdup the c_str, keep a reference of d0_hash as a shash::Any or string). To make it more uniform with the other two hash lookup, strdup was used. This also added some code to remove warnings about comparing signed and unsigned values for length.

I think that this only occurred on some of the systems depending on how aggressive the garbage collector is.

@jblomer take a look when you get a chance. This was tested on Fedora 26, Fedora 27, and Ubuntu 16.04 using docker.

### DIFF
--- a/test/unittests/t_libcvmfs.cc
+++ b/test/unittests/t_libcvmfs.cc
@@ -360,8 +360,9 @@ TEST_F(T_Libcvmfs, Listdir) {
   char **buf = NULL;
   size_t buflen = 0;
   cvmfs_listdir(ctx, "/dir/dir", &buf, &buflen);
-  // The buf length should be at least 3
-  EXPECT_LE(3, buflen);
+  // The buf length should be at least 4
+  size_t expected = 4;
+  EXPECT_LE(expected, buflen);
   // Check that listed info matches specified dir
   EXPECT_FALSE(strcmp(".", buf[0]));
   EXPECT_FALSE(strcmp("..", buf[1]));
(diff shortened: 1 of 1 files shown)


test/unittests/t_libcvmfs.cc
```diff
@@ -360,8 +360,9 @@ TEST_F(T_Libcvmfs, Listdir) {
   char **buf = NULL;
   size_t buflen = 0;
   cvmfs_listdir(ctx, "/dir/dir", &buf, &buflen);
-  // The buf length should be at least 3
-  EXPECT_LE(3, buflen);
+  // The buf length should be at least 4
+  size_t expected = 4;
+  EXPECT_LE(expected, buflen);
```
jblomer: This can be simplified to `EXPECT_LE(4U, buflen)`

test/unittests/t_libcvmfs.cc
```diff
@@ -423,34 +425,33 @@ TEST_F(T_Libcvmfs, StatNestedCatalog) {
...
   nc_attr = cvmfs_nc_attr_init();
   EXPECT_FALSE(cvmfs_stat_nc(ctx, "/dir", nc_attr));
-  orig_nc = tester.manifest()->catalog_hash().ToString().c_str();
-  EXPECT_FALSE(strcmp(orig_nc, nc_attr->hash));
+  EXPECT_FALSE(strcmp(d0_hash, nc_attr->hash));
   EXPECT_FALSE(strcmp("", nc_attr->mountpoint));
   // Size of root catalog is 0 as a Nested Catalog
-  EXPECT_EQ(nc_attr->size, 0);
+  size_t expected = 0;
+  EXPECT_EQ(nc_attr->size, expected);
```
jblomer: `EXPECT_EQ(0U, nc_attr->size)` -- GTest wants the expected value first for proper reporting when it fails.

PR #2176, opened 2018-07-25
Only assume workspace is cwd if in FUSE

nhazekam:
When setting up a cache workspace we set the cache_path to the cwd as it is assumed we had moved to the directory earlier in setup. This only occurs if the type is Fuse, but if the type is Library we don't `chdir`. This checks if the type is Fuse before assuming we are in the cwd for the cache.


nhazekam:
@jblomer @gganis This is a small fix in reference to issue I opened in JIRA CVM-1603. Take a look when you get the chance. If you would like me to address this in another way please let me know.

radupopescu:
Thanks for the fix, @nhazekam !

PR #2178, opened 2018-07-25
Fix snapshot -t to override stale update lock

DrDaveD:
See [CVM-1598](https://sft.its.cern.ch/jira/browse/CVM-1598)


radupopescu:
Thanks!

PR #2180, opened 2018-07-26
Add GC metrics to release manager statistics

dosarudaniel:
Addresses [CVM-1569](https://sft.its.cern.ch/jira/browse/CVM-1569).

- Added one counter for gc: `sz_condemned_bytes` which is used only if `CVMFS_EXTENDED_GC_STATS` variable is set to on/true/yes/1 value in `server.conf` file.
- Write garbage collector counters values in database statistics file into `gc_statistics` table
- Added two new integration tests: `662-merge_stats_tool` and `661-garbage_collector_statistics`
- Check if `repo_name` was found. (`swissknife_main.cc`)

### DIFF
--- a/cvmfs/garbage_collection/garbage_collector_impl.h
+++ b/cvmfs/garbage_collection/garbage_collector_impl.h
@@ -157,6 +157,8 @@ void GarbageCollector<CatalogTraversalT, HashFilterT>::Sweep(
   if (last_char != 'C') {    // if is not a catalog file
     int64_t condemned_bytes = configuration_.uploader->GetObjectSize(hash);
     if (condemned_bytes > 0) {
+      printf("condemned ------ %s\n", hash.MakePath().c_str());
+      printf("condemned size - %ld\n", condemned_bytes);
       condemned_bytes_ += condemned_bytes;
     }
   }
--- a/cvmfs/garbage_collection/garbage_collector_impl.h
+++ b/cvmfs/garbage_collection/garbage_collector_impl.h
@@ -153,7 +153,7 @@ void GarbageCollector<CatalogTraversalT, HashFilterT>::Sweep(
                                                        const shash::Any &hash) {
   ++condemned_objects_;
 
-  if (!hash.HasSuffix()) {  // Count only data, not metadata
+  if (!hash.HasSuffix() || hash.suffix == shash::kSuffixPartial) {
     int64_t condemned_bytes = configuration_.uploader->GetObjectSize(hash);
     if (condemned_bytes > 0) {
       condemned_bytes_ += condemned_bytes;
--- a/cvmfs/garbage_collection/garbage_collector_impl.h
+++ b/cvmfs/garbage_collection/garbage_collector_impl.h
@@ -157,8 +157,6 @@ void GarbageCollector<CatalogTraversalT, HashFilterT>::Sweep(
   if (last_char != 'C') {    // if is not a catalog file
     int64_t condemned_bytes = configuration_.uploader->GetObjectSize(hash);
     if (condemned_bytes > 0) {
-      printf("condemned ------ %s\n", hash.MakePath().c_str());
-      printf("condemned size - %ld\n", condemned_bytes);
       condemned_bytes_ += condemned_bytes;
     }
   }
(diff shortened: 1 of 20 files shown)


jblomer:
I think we need to discuss how to count the condemned bytes in the garbage collection.  This implementation assumes that the data is in a local directory but it can also be in S3.  Perhaps we have to get the size of objects from the catalogs.

dosarudaniel:
@jblomer , @radupopescu 
I think this is ready for code review.
Also, should I make gathering condemned_bytes_ optional?

dosarudaniel:
Integration test 659-662 passed on my machine.

cvmfs/garbage_collection/garbage_collector_impl.h
```diff
@@ -152,6 +153,13 @@ void GarbageCollector<CatalogTraversalT, HashFilterT>::Sweep(
                                                        const shash::Any &hash) {
   ++condemned_objects_;
 
+  if (!hash.HasSuffix()) {  // Count only data, not metadata
```
dosarudaniel: Maybe I should add `|| hash.suffix == shash::kSuffixPartial` here.

cvmfs/statistics_database.cc
```diff
@@ -155,11 +218,28 @@ StatisticsDatabase::~StatisticsDatabase() {
...
+                                        const std::string &finished_time,
+                                        const std::string &command_name,
+                                        const swissknife::ArgumentList &args) {
+  std::string insert_statement;
+  if (command_name == "ingest" || command_name == "sync") {
+    insert_statement = PrepareStatementIntoPublish(statistics, start_time,
+                                                               finished_time);
+  } else if (command_name == "gc") {
+    std::string dry_run_ = "0";
+    const bool dry_run = (args.count('d') > 0);
```
dosarudaniel: I think I don't need dry_run variable here.

radupopescu: I'm thinking the same thing. Dry runs should be ignored - no need to save any stats in the SQLite db in this case.

radupopescu: Indeed, dry runs should not be recorded in the database.

I would remove the dry_run field from the schema and only save actual GC runs, not the dry runs.

cvmfs/statistics_database.h
```diff
@@ -11,6 +11,7 @@
 #include "options.h"
 #include "sql.h"
 #include "statistics.h"
+#include "swissknife.h"
```
radupopescu: If you are not storing dry runs in the database, there is no need to include `swissknife.h` here. The `args` parameter of the `StoreStatistics` function can also go.

A design point, for future reference: if we weren't removing the `args` parameter, we would still improve the function by observing that a single argument from the `args` list is used (dry_run in this case). It's better to read the value of the argument in the caller function and have the `StoreStatistics` function receive the value directly. This way, we don't need to be aware of any type from  `swissknife.h` in this file and we don't need to include it.
You really want to avoid complicated include relationships between your files: swissknife uses the `StatisticsDatabase` class, this class shouldn't need to know anything about it's users.

radupopescu:
I think the branch is fine, now.

@jblomer what do you think?

cvmfs/upload_gateway.cc
```diff
@@ -231,6 +231,10 @@ bool GatewayUploader::ReadKey(const std::string& key_file, std::string* key_id,
   return gateway::ReadKeys(key_file, key_id, secret);
 }
 
+int64_t GatewayUploader::DoGetObjectSize(const std::string &file_name) {
+  return 0;
```
jblomer: I think we should return `-EOPNOTSUPP` (or another suitable error code from errno.h) in order to distinguish it from a zero-byte object.

cvmfs/upload_s3.cc
```diff
@@ -398,4 +398,7 @@ bool S3Uploader::PlaceBootstrappingShortcut(const shash::Any &object) const {
   return false;  // TODO(rmeusel): implement
 }
 
+
+int64_t S3Uploader::DoGetObjectSize(const std::string &file_name) { return 0; }
```
jblomer: Same as above.  We can add a TODO here: a `HEAD` request gives us the byte count.

test/unittests/t_garbage_collector.cc
```diff
@@ -54,6 +54,7 @@ class GC_MockUploader : public AbstractMockUploader<GC_MockUploader> {
   }
 
   virtual unsigned GetNumberOfErrors() const { return 0; }
+  virtual int64_t DoGetObjectSize(const std::string &file_name) { return 0;}
```
jblomer: Same as above

PR #2187, opened 2018-08-03
Add start/finished time in statistics database.

dosarudaniel:
Changed `timestamp` column to `start_time`.
Added `finished_time` column.

### DIFF
--- a/cvmfs/statistics_database.cc
+++ b/cvmfs/statistics_database.cc
@@ -57,11 +57,14 @@ struct Stats {
   * @param stats a struct with all values stored in strings
   * @return the insert statement
   */
-std::string PrepareStatement(const perf::Statistics *statistics) {
+std::string PrepareStatement(const perf::Statistics *statistics,
+                            const std::string start_time,
+                            const std::string finished_time) {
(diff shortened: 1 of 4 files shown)


cvmfs/statistics_database.cc
```diff
@@ -105,7 +110,7 @@ bool StatisticsDatabase::CreateEmptyDatabase() {
     "sz_bytes_added INTEGER,"
     "sz_bytes_removed INTEGER,"
     "sz_bytes_uploaded INTEGER,"
-    "CONSTRAINT pk_publish_statistics PRIMARY KEY (timestamp));").Execute();
+    "CONSTRAINT pk_publish_statistics PRIMARY KEY (start_time));").Execute();
```
jblomer: That could become a problem if two consecutive server commands finish in the same second, no?

radupopescu: @jblomer Good point. What would be good solution: going to a finer precision or using a unique counter as primary key?

jblomer: I think I'd go for a sequence number.

PR #2189, opened 2018-08-03
Allow to ingest tarballs with hardlink multiple times

siscia:
This fix is necessary when we are ingesting a tarball with hardlinks, if
there are already file with the same name and content of the hardlink
(basically only when we are ingesting twice the same tarball) we would
try to add twice the same unique hash, which produce a runtime error
violating the UNIQUE constraint in SQLite.

With this PR we simply check if the file is already there and if it is,
we eliminate it.


jblomer:
Please change the PR title to a description of the fix.

PR #2191, opened 2018-08-06
Accept also Character and Block Device

siscia:
Some images include devices like /dev/full which are quite complex to manage, as an example normal tar utility do not extract the /dev/full device.

The other alternative is to don't accept devices, however this will mean to avoid the `abort()` if every check fail.


cvmfs/sync_union_tarball.cc
```diff
@@ -232,8 +233,9 @@ void SyncUnionTarball::ProcessArchiveEntry(struct archive_entry *entry) {
 
   } else {
     LogCvmfs(kLogUnionFs, kLogStderr,
-             "Warning found unexpected file: \n%s\n", filename.c_str());
+             "Fatal error found unexpected file: \n%s\n", filename.c_str());
     read_archive_signal_->Wakeup();
+    abort();
```
siscia: In the PR I am refering to this `abort()`

cvmfs/sync_union_tarball.cc
```diff
@@ -233,6 +234,7 @@ void SyncUnionTarball::ProcessArchiveEntry(struct archive_entry *entry) {
   } else {
     LogCvmfs(kLogUnionFs, kLogStderr,
              "Fatal error found unexpected file: \n%s\n", filename.c_str());
+    read_archive_signal_->Wakeup();
```
jblomer: We crash anyway one line later, so it seems the `Wakeup` is not necessary

jblomer:
I think we should actually explicitly test for special files in the tarballs; it's one of the nice features of cvmfs that we can support such files.

siscia:
What you mean by "explicitly test", would you prefer to an integration test just for the char and block devices or just ingest them is enough?

cvmfs/sync_union_tarball.cc
```diff
@@ -233,6 +234,10 @@ void SyncUnionTarball::ProcessArchiveEntry(struct archive_entry *entry) {
   } else {
     LogCvmfs(kLogUnionFs, kLogStderr,
              "Fatal error found unexpected file: \n%s\n", filename.c_str());
+    // if for any reason this code path change and we don't abort anymore,
+    // remember to wakeup the signal, otherwise we will be stuck in a deadlock
+    //
+    // read_archive_signal_->Wakeup();
```
siscia: @jblomer  do you think this is better?

I left the Wakeup like there from a previous iteration where I spend too much time figuring out why it was deadlocking.

Maybe I should just leave the comment and remove the actual code line?

jblomer: Oh, I see.  Yep, that's fine.

jblomer:
I mean just that a /dev/null device should be somewhere in the integration tests, no need to issue a new test number for that.

siscia:
I am not quite sure, but just move /dev/null into a directory and then compress it using tar I don't believe is a good idea.
I naively tried and it just move an empty file into the directory, files that didn't get well compressed into the tar file.

At the moment I just create a char and block device, but they requires sudo privileges...

test/src/654-tarball_ingest_special_files/main
```diff
@@ -41,6 +41,10 @@ produce_tarball() {
   ln tarball_foo/d/1.txt tarball_foo/d_1_hardlink2.txt
   ln tarball_foo/d/1.txt tarball_foo/d_1_hardlink3.txt
 
+  # use 0 (zero) as major number to automatically allocate a free major number
+  mknod tarball_foo/char_device c 0 42
+  mknod tarball_foo/block_device b 0 24
+
```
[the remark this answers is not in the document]
jblomer: Sorry for the late reply, this got somehow lost.  It is fine, you can use `sudo` in the integration tests, so the two lines would need to change to `sudo mknod ...`

PR #2193, opened 2018-08-06
check for files that start with `./` and remove the prefix

siscia:
Some docker layers have this peculiar structure of having the whole content inside the `.` folder.

Basically the `.` folder is the root of the tarball, and inside it, the filesystem tree develops.

This shows up as two characters `./` prefix when we are working with the file path that breaks some ingestion.

This PR fix the issue by removing the two char if present.

Given the semantic of the two particular characters, I believe that there would no issues. But I could be wrong.

### DIFF
--- a/cvmfs/sync_union_tarball.cc
+++ b/cvmfs/sync_union_tarball.cc
@@ -177,6 +177,12 @@ void SyncUnionTarball::Traverse() {
 
 void SyncUnionTarball::ProcessArchiveEntry(struct archive_entry *entry) {
   std::string archive_file_path(archive_entry_pathname(entry));
+  if (archive_file_path.length() >= 2) {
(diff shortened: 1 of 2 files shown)


jblomer:
We might see more special cases of similar type.  I'd suggest to channel the path names through a `SanitizePath` method which we can extend as needed.

siscia:
The process is a little more involved that I though.

Ideally I could just pass a non-const reference and change the string in place, but the linter seems not so fine with that.

I would like to avoid copy the whole string everytime.

I settled for this solution right now that copies the string only if it need to be modified.

Maybe we can do better, but it seems a very minor performance concern.

Thoughts?

@jblomer

jblomer:
I think no need to worry about the string copy.

PR #2200, opened 2018-08-09
Fix uploaded bytes count

dosarudaniel:
I noticed that I was counting uploaded_bytes for duplicated files too, so I fixed it:

In `StreamedUpload()` I was using the `bytes_written` variable, but I couldn't tell if those bytes were actually duplicated or not.
In `FinalizeStreamedUpload()` function I can count only the real uploaded bytes, but I have to use a system call - `stat`.

I need to make this change for [#2180](https://github.com/cvmfs/cvmfs/pull/2180)'s integration tests.

### DIFF
--- a/cvmfs/upload_local.cc
+++ b/cvmfs/upload_local.cc
@@ -167,7 +167,7 @@ void LocalUploader::FinalizeStreamedUpload(UploadStreamHandle *handle,
               UploaderResults(UploaderResults::kChunkCommit, cpy_errno));
       return;
     }
-    if (!content_hash.HasSuffix()) { // count only data, not metadata
+    if (!content_hash.HasSuffix()) {  // count only data, not metadata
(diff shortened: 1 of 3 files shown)


cvmfs/upload_local.cc
```diff
@@ -168,6 +167,9 @@ void LocalUploader::FinalizeStreamedUpload(UploadStreamHandle *handle,
               UploaderResults(UploaderResults::kChunkCommit, cpy_errno));
       return;
     }
+    if (!content_hash.HasSuffix()) {  // count only data, not metadata
+      CountUploadedBytes(GetFileSize(upstream_path_ + "/" + final_path));
```
jblomer: Sorry for looking only now: this might still not be enough.  Large files can be split in smaller pieces, in which case the hash has the 'P' suffix.

PR #2202, opened 2018-08-13
Added a case for large files ('P' suffix)

dosarudaniel:
Count uploaded_bytes for large files splitted into smaller pieces with 'P' suffix.

### DIFF
--- a/cvmfs/upload_local.cc
+++ b/cvmfs/upload_local.cc
@@ -167,7 +167,7 @@ void LocalUploader::FinalizeStreamedUpload(UploadStreamHandle *handle,
               UploaderResults(UploaderResults::kChunkCommit, cpy_errno));
       return;
     }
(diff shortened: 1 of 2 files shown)


cvmfs/upload_local.cc
```diff
@@ -167,7 +167,7 @@ void LocalUploader::FinalizeStreamedUpload(UploadStreamHandle *handle,
               UploaderResults(UploaderResults::kChunkCommit, cpy_errno));
       return;
     }
-    if (!content_hash.HasSuffix()) {  // count only data, not metadata
+    if (!content_hash.HasSuffix() || content_hash.GetSuffix() == 'P') {
```
jblomer: Thanks!  You can use the constant `shash::kSuffixPartial` instead of 'P'.

PR #2203, opened 2018-08-13
cvmfs_shrinkwrap: Modified files

samysweb:
This pull request contains existing cvmfs code files that had to be modified for the development of cvmfs_shrinkwrap.

This should be merged before the "big" cvmfs_shrinkwrap pull request in #2206


samysweb:
@jblomer I think this is ready for review now.

nhazekam:
@samysweb @jblomer I think we need to add `platform_linux.h` and `platform_osx.h` to this pull request. These were things I changed recently to make the xattr compile for mac. They are existing files that we added to, which I think is the goal of this pull request.

jblomer:
Replaced by #2225

PR #2209, opened 2018-08-14
Disable GEOAPI in incremental build script on Red Hat older than 6

radupopescu:
The GEOAPI component seems to fail to build on RedHat 4 and 5 due to the very old version of Python shipped with the system.
This PR disables building the GEOAPI component in the incremental build script on these platforms.

However, I haven't yet understood why the full build is passing.


radupopescu:
Result of incremental build:
https://epsft-jenkins.cern.ch/view/CVMFS/job/CvmfsIncrementalBuild/1030/

DrDaveD:
The inclusion of geoapi is explicitly disabled on el4 & el5 in the full build because some there was some missing dependencies.

PR #2212, opened 2018-08-16
increase block size

siscia:
Did we already try this?

Visually it seems faster, after lunch I will make some more perf test.


siscia:
Nope, just the way I was running the benchmarks was wrong. This change does not increase the performance at all.

PR #2214, opened 2018-08-23
WIP: Add manual pages and help command

s4n-cz:
### Add manual pages:
  * `man cvmfs_server`
  * `man cvmfs_server <subcommand>` (e.g., `man cvmfs_server mkfs`)

### Add help subcommand:
  * `cvmfs_server help <subcommand>` (e.g., `cvmfs_server help mkfs`)
  * `cvmfs_server <subcommand> --help` does the same


CMakeLists.txt
```diff
@@ -466,6 +466,30 @@ if (BUILD_QC_TESTS)
...
+# Manual pages
+#
+find_program (ASCIIDOCTOR_EXE asciidoctor)
+if (NOT ASCIIDOCTOR_EXE)
+  message (FATAL_ERROR "Could not find asciidoctor. Try yum/gem install asciidoctor")
+endif (NOT ASCIIDOCTOR_EXE)
+add_custom_target (
+  man
+  ALL
+  COMMAND ${CMAKE_CURRENT_BINARY_DIR}/cvmfs/cvmfs_server generate-man ${CMAKE_CURRENT_BINARY_DIR}
```
tstdin: This is the problematic line not passing in Travis. Currently I am not sure how to fix that.

Second thing, on OS X it would also need to first install asciidoctor.

tstdin: Now I found the cause of Travis failing on Ubuntu (I was trying it on Fedora 27 and it worked fine).

It is because of the shebang `#!/bin/sh` in cvmfs_server that indicates usage of a strictly bourne-compatible shell; in the case of Ubuntu, dash is used. And there is no support for `declare` command.

I am using it for declaring associative arrays with `declare -A` that is supported in Bash since v4. Even on RHEL6 it's OK - ["Red Hat Enterprise Linux 6 includes version 4.1 of Bash as its default shell."](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/migration_planning_guide/sect-migration_guide-package_changes-bash)
 It is a really nice key-value store ideal for setting options.
```
_CVMFS_SERVER_MKFS_OPTIONS=(
  [G]="set timespan for auto tags"
  [R]="require masterkeycard key"
  [V]="VOMS authorization"
   ...
)
```
The fix would be change the shebang to `#!/bin/bash` if there is no problem with it.

cvmfs/server/cvmfs_server_abort.sh
```diff
@@ -5,6 +5,14 @@
 #
 # Implementation of the "cvmfs_server abort command"
 
+_CVMFS_SERVER_ABORT_SHORT="Abort transaction and return to the state before"
+_CVMFS_SERVER_ABORT_DESCRIPTION="TODO"
+_CVMFS_SERVER_ABORT_SYNOPSIS="_cvmfs_server abort_ [options] <fqrn>"
+_CVMFS_SERVER_ABORT_OPTIONS=(
```
tstdin: Dash does not support even normal arrays, didn't expect that.

jblomer:
This looks very useful and well-written, many thanks!

One thing that came to my mind is that it would probably be better to name the bash variable `_CVMFS_DOC_...`.  I know, I suggested `_CVMFS_SERVER` in the first place, apologies!

Being under a pile of things, we'd need a little bit of time to digest it all, please be patient.

PR #2215, opened 2018-08-27
Add option to create catalog during ingestion.

siscia:
This pull request add a flag (-c, short for --catalog) that allow to
automagically add a (sub-)catalog when ingesting a tar file.
The catalog is added in the base dir, where the tar is "uncompressed".

I am not so sure about this implementation.
It is not a very clean design, we borrow a method from a nice interface
(AddCatalogMarker from CatalogBalancer), similarly when we are actually
adding the catalog (in SyncMediator::Commit) we are basically repeating
what is happening inside the CatalogBalancer.

However I believe that the implementation is quite reasonable and strike
a balance between functionality added and extremelly clean design.

### DIFF
--- a/cvmfs/catalog_balancer.h
+++ b/cvmfs/catalog_balancer.h
@@ -41,6 +41,17 @@ namespace catalog {
  *
  * c) The number of entries of the catalog is lesser than min_weight_: the
  * catalog gets merged with its father (except the root catalog, obviously).
+ *
+ * Notes: Originally the CatalogBalancer class had only a single function,
+ * Balance, then it is been necessary to expose also the function
+ * AddCatalogMarker.
+ * One of the biggest risk, when working with any kind of interface is to,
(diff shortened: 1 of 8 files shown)


DrDaveD:
FYI a PR owner can relaunch travis by closing & reopening the PR.

siscia:
@jblomer This features would be quite useful for unpacked.cern.ch

Do you see any problem in this PR?

jblomer:
I'm wondering, since the tarball union engine anyway works with virtual files, wouldn't it be simpler to pretend there was a .cvmfscatalog file and let the other parts of cvmfs do the rest?

siscia:
It seems a little bit like a hack.

I don't dislike the idea, though.

I will try to sketch an implementation and see how far we can go, it may actually work.

jblomer:
Replaced by #2227

PR #2219, opened 2018-08-31
X509 Authentication Fixes

bbockelm:
This PR fixes a segfault introduced by @djw8605 in the SciTokens patch - the wrong `void*` was being cast to an `sslctx_info` pointer.

Additionally, I took the opportunity to free up X509 data structures as soon as we hand them off to OpenSSL.


bbockelm:
Whoops - reverted + squashed freeing up the X509 data structures.  We actually cache and reuse these between requests if necessary.

PR #2220, opened 2018-08-31
Improve syslog messages for HTTP failures.

bbockelm:
This splits the source of "host connection error" into three:

1.  Actual connection errors to the host.
2.  Timeouts from the download.
3.  Host responds with insufficient data.

This should hopefully improve our ability to debug failures; the tweaked log messages were immensely useful for figuring out some issues with the LIGO repository.

### DIFF
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -1245,12 +1245,18 @@ bool DownloadManager::VerifyAndFinalize(const int curl_error, JobInfo *info) {
     case CURLE_COULDNT_RESOLVE_HOST:
       info->error_code = kFailHostResolve;
       break;
-    case CURLE_COULDNT_CONNECT:
     case CURLE_OPERATION_TIMEDOUT:
+      info->error_code = (info->proxy == "DIRECT") ?
+                         kFailHostTooSlow : kFailProxyTooSlow;
+      break;
     case CURLE_PARTIAL_FILE:
     case CURLE_GOT_NOTHING:
     case CURLE_RECV_ERROR:
+      info->error_code = (info->proxy == "DIRECT") ?
+                         kFailHostShortTransfer : kFailProxyShortTransfer;
+      break;
     case CURLE_FILE_COULDNT_READ_FILE:
+    case CURLE_COULDNT_CONNECT:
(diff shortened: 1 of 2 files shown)


jblomer:
Many thanks!  

We have several places where we check for `kFailProxyConnection` and `kFailHostConnection` in download.cc, object_fetcher.h and swissknife_pull.cc.  These places now need to check for the additional error codes.  I'd suggest adding static functions like `bool download::IsTransferError(download::Failures error_code)` to have this check in a central place.

I think we should merge this into the 2.5 branch and make the patch part of the 2.5.2 release (no ETA yet).

DrDaveD:
Unfortunately this didn't get finished for 2.5.2.  There should be a jira ticket to go along with it so it doesn't get forgotten again.

Brian asked me to take over finishing this, so I'll create the ticket.

bbockelm:
@DrDaveD - another thing worth pointing out is the log message doesn't say _which_ host is having the connection problem.  Example:

```
Nov 27 11:19:38 treehorn00 cvmfs2: (ligo.osgstorage.org) failed to fetch Part of /frames/O2/hoft_C01/H1/H-H1_HOFT_C01-11646/H-H1_HOFT_C01-1164619776-4096.gwf (hash: 5a8e21dc032a0e418d073a8908e03b6e999b2f0f, error 7 [host connection problem])
```

There's 3-4 different hosts that this could be referring to.  If you can get the sysadmin to pull the right lines, you may be able to deduce this information -- but it'd probably be better to make it self-contained.

DrDaveD:
It turns out that calculating the host name is quite a complicated thing to do, probably requiring reading the proxy & host urls back from the curl handle and parsing them, not to mention following redirects.  The syslog does show when proxies and servers are switched so it should be possible for the sysadmin to get that information, although that also doesn't show redirects.  I wonder if it would be possible & relatively easy to get the socket that was used from curl, and use that to get the IP address.

jblomer:
To get the host and proxy, I think we can store the values in the JobInfo object at the time they are set on the curl handle.

jblomer:
Replaced by #2279

PR #2222, opened 2018-09-07
Script for submitting publisher statistics to a Carbon server

dosarudaniel:
Addresses [CVM-1571](https://sft.its.cern.ch/jira/browse/CVM-1571).

This script sends stats to a carbon server ([graphite](https://graphiteapp.org/)) using the [pickle protocol](https://graphite.readthedocs.io/en/latest/feeding-carbon.html#the-pickle-protocol) .

Usage: `send_stats_to_graphite.py [-h] <db_file> <carbon_server_IP> <carbon_server_PORT>`

This script uses  `last_timestamp_sent` file to send only the new stats data form the `db_file`.
Format of the `last_timestamp_sent` file:
- First line: last start_time sent for publish_statistics table
- Second line: last start_time sent for gc_statistics table

If the `last_timestamp_sent` file does not exist in the same directory as `db_file`, this script will create one and it will write the current UTC timestamp in the it.

### DIFF
--- a/python/send_stats_to_graphite.py
+++ b/python/send_stats_to_graphite.py
@@ -10,27 +10,31 @@ import sys
 import time
 from calendar import timegm
 
+
 def get_data_publish_stats(last_timestamp, db_file):
     conn = sqlite3.connect(db_file)
     c = conn.cursor()
 
     # order DESC by id ---> first element of the list has the last start_time
-    c.execute('SELECT start_time, files_added, files_removed, files_changed, duplicated_files, directories_added, \
-                      directories_removed, directories_changed, sz_bytes_added, sz_bytes_removed, sz_bytes_uploaded \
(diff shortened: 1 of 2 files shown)


radupopescu:
Good job, this is a great start!

Here's what I would improve, though:

1. The Carbon server IP, port etc. should not be hardcoded, but given as commandline parameters. You can use the `argparse` module to implement all the commandline parameter handling.

2. By default, the script should be "one-shot": it starts, it looks for any new rows, pickles them, uploads them and stops. The continuous mode can be optional.

3. You could create an auxiliary file alongside `stats.db` in `/var/spool/cvmfs/<REPO>` to store some information about which rows have already been uploaded. I think storing the number of rows and last uploaded timestamps would be useful. You can then implement the logic for only uploading new rows.

PR #2227, opened 2018-09-28
Add option to create catalog during ingestion using a fake file

siscia:
Following the discussion on https://github.com/cvmfs/cvmfs/pull/2215 here is another implementation.

It use a fake file `$tarball_root/.cvmfscatalog` that is piped into the regular process.

### DIFF
--- a/cvmfs/sync_union_tarball.cc
+++ b/cvmfs/sync_union_tarball.cc
@@ -175,6 +175,8 @@ void SyncUnionTarball::Traverse() {
           archive_entry_set_pathname(catalog, root.c_str());
           archive_entry_set_filetype(catalog, AE_IFREG);
           ProcessArchiveEntry(catalog);
+          // The ProcessArchiveEntry does call Wakeup on the signal, in this
+          // particular corner case we need to re-wait for it.
           read_archive_signal_->Wait();
         }
         first_iteration = false;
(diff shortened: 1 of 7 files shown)


cvmfs/sync_union_tarball.cc
```diff
@@ -158,6 +167,21 @@ void SyncUnionTarball::Traverse() {
       }
 
       case ARCHIVE_OK: {
+        if (first_iteration && create_catalog_on_root_) {
+          std::string root =
+              GetRoot(std::string(archive_entry_pathname(entry)));
+          root = root + "/.cvmfscatalog";
+          struct archive_entry *catalog = archive_entry_clone(entry);
```
jblomer: Can we use `archive_entry_new` instead? I think we cannot tell for sure which entry we get first from the tar file, so it is not necessarily the root directory of the archive. This could lead to unexpected results if we clone a random entry.  We would then also not need `GetRoot()`.

siscia: We can for sure un `archive_entry_new`.
The reason why I went with cloning is to already have most fields set in the entry itself. Stuff like UID and GID, possibly timestamp, etc
I believe that those fields are not so necessary for the `.cvmfscatalog` file, am I correct?

I need to think how to get the `root` where to position the catalog.
Given this command:
```
cvmfs_server ingest --tarball baz.tar --base-dir foo/ foo.cern.ch
```
It will expand into `/cvms/foo.cern.ch/foo/baz/....content_of_baz...

In this case we want the catalog to be in `/cvms/foo.cern.ch/foo/baz/`, not in  `/cvms/foo.cern.ch/foo/baz/`

siscia: Nope, I am wrong.

The ingest works in a different way.

```
cvmfs_server ingest --tarball baz.tar --base-dir foo/ foo.cern.ch
```

It will expand into `It will expand into /cvms/foo.cern.ch/foo/....content_of_baz...`

So yes, I believe we can do without the `GetRoot`

PR #2229, opened 2018-10-02
Fix build for QG

siscia:
This should fix the build of just the quickgen test.

As reported in CVM-1637


siscia:
I am wondering if we should add the QG build to the automated test infrastructure?

radupopescu:
The reason why QC is not ON by default is because it needs "-std=c++11". I didn't have time to think about making a C++11 build for the purposes of CI.

If that can be accommodated somewhere in the CI infrastructure, then yes, it should be enabled. We should also write more tests with it, in that case.

As it sits right now, it's of limited use and will always break.

PR #2236, opened 2018-10-11
Fix cache control headers for S3 backend

jblomer:
Actively set the cache-control header for immutable content-addressed objects under data to 3 days (previously: no setting) and for the .cvmfs... files to 61 seconds to avoid running into the Squid bug where "no cache" objects are cached forever.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,5 @@
 2.5.2:
+  * Fix cache control headers for objects pushed to S3 (CVM-1606)
   * Fix building on macOS 10.14
   * Fix potential memory corruption in catalog traversal
   * Reimplement cvmfs_talk in C++ to improve efficiency
--- a/cvmfs/s3fanout.cc
+++ b/cvmfs/s3fanout.cc
@@ -20,6 +20,10 @@ using namespace std;  // NOLINT
 
 namespace s3fanout {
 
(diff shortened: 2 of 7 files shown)


DrDaveD:
So Jakob, these are on the client side, correct?  Normally cache-control: max-age is set on the server side, but the client can also set it to reduce a time set from the server.  The 61 second squid limitation (with default squid configuration) is I believe only the server side.

DrDaveD:
Also, what squid "no cache" bug are you referring to?

jblomer:
It's during upload, so it's supposed to tell the S3 service the `cache-control` header it is supposed to send when clients ask for it. Regarding Squid, I thought we set the cache-control to 61 seconds because everything below that value is mistakenly cached without a time limitation.

DrDaveD:
Oh, that's what you meant by the "no cache" bug.  That's right, if you're setting server cache headers, it should be a minimum of 61 seconds.  If you go less than that from the server side the max-age is ignored and the squid uses its default, which is typically 3 days I think.  The reason for the 61 seconds is the squid option minimum_expiry_time which defaults to 60 seconds.  So it's a feature, not really a bug.  frontier-squid-3.5.19 and later sets the default minimum to 0, but you can't depend on people running that.

PR #2237, opened 2018-10-11
allow user xattr only on regular files and directories

siscia:
Following the standard I believe we should not allow xattr to anything different from regular files and directories.

https://linux.die.net/man/5/attr

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1547,6 +1547,16 @@ static void cvmfs_listxattr(fuse_req_t req, fuse_ino_t ino, size_t size) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


jblomer:
Hm, I'm not getting from the man page why it should be disallowed for symlinks.  In any case, we do use extended magic attributes on symlinks already to get the raw content of variant symlinks (`$(VAR_NAME)`).

siscia:
It was just breaking the containerd snapshotter. I dig deeper and the man says that we should not have the user extended attributed set for symlinks, so I fix it.
(For which reason I don't understand neither honestly.)

However if we need them let's leave it there.

PR #2241, opened 2018-10-11
Support arbitrary subpaths inside S3 bucket

radupopescu:
Addresses [CVM-1641](https://sft.its.cern.ch/jira/browse/CVM-1641?filter=13573).

Using the recommended DNS-style bucket URL, it's possible to provide an arbitrary subpath for the repository URL. For example, using "/cvmfs", to maintain the familiar URL naming scheme for CVMFS:
```
http://<bucket_name>.s3.cern.ch/cvmfs/<repo_name>
```

This PR provides support for this arbitrary subpath, by encoding it into the "repository_alias" token of the upstream configuration:
```
CVMFS_UPSTREAM_STORAGE=S3,/var/spool/cvmfs/s3test.cern.ch/tmp,cvmfs/s3test.cern.ch@/etc/cvmfs//s3/cern.conf
```
This solution is backward compatible, providing a subpath is optional.


jblomer:
I guess we need to set `CVMFS_S3_DNS_BUCKETS=false` for the integration tests, no?

radupopescu:
Good catch!

PR #2245, opened 2018-10-12
Implement retry support for "cvmfs_server transaction"

radupopescu:
Addresses [CVM-1611](https://sft.its.cern.ch/jira/browse/CVM-1611).

Allow setting the initial and max retry delays, and the maximum
number of retries.

The retry loop starts with the initial delay and doubles it at each
retry, until the max delay is reached.

### DIFF
--- a/cvmfs/server/cvmfs_server_transaction.sh
+++ b/cvmfs/server/cvmfs_server_transaction.sh
@@ -19,13 +19,30 @@ cvmfs_server_transaction() {
   local force=0
   local retcode=0
 
+  local retry=0
+  local init_retry_delay=5 # seconds
+  local max_retry_delay=300 # seconds
+  local num_retries=50
+
   # optional parameter handling
   OPTIND=1
-  while getopts "fe" option
+  while getopts "fri:m:n:" option
   do
     case $option in
       f)
         force=1
+      ;;
+      r)
+        retry=1
+      ;;
+      i)
+        init_retry_delay=$OPTARG
+      ;;
+      m)
+        max_retry_delay=$OPTARG
+      ;;
+      n)
+        num_retries=$OPTARG
       ;;
       ?)
         shift $(($OPTIND-2))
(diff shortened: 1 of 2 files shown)


jblomer:
I think we should make this available to ingestion as well, perhaps in a first approach through good default values when `cvmfs_server ingest` calls `transaction`

radupopescu:
Makes sense, but adding gateway support to the tarball ingestion workflow is to be done in [CVM-1610](https://sft.its.cern.ch/jira/browse/CVM-1610).

I quickly checked `cvmfs_server_ingest.sh`. Right now, the command line flags of that command don't conflict with `cvmfs_server transaction` so we can even expose the same flags as here.

Do the current default retry values make sense?

cvmfs/server/cvmfs_server_transaction.sh
```diff
@@ -78,7 +95,34 @@ cvmfs_server_transaction() {
...
+        while true ; do
+          __swissknife lease -a acquire -u $repo_services_url -k $gw_key_file -p $name"/"$subpath
+          res=$?
+          if [ $res -eq 0 ] || [ $retry -eq 0 ] || [ $r -gt $num_retries ]; then
+            break
+          fi
+          echo "Retry $r/$num_retries, waiting: $delay"
+          r=$((r + 1))
+          sleep $delay
+          delay=$((delay * 2))
```
jblomer: I think that needs to be `delay=$(($delay * 2))`

radupopescu: I think "$" isn't needed inside double parantheses. I mean, I also checked that it works.

jblomer:
Let's take care then of the ingest in a follow-up PR. I'd go slightly higher in the number of retries, perhaps to 100 (~8 hours).

cvmfs/server/cvmfs_server_transaction.sh
jblomer: Cool, I didn't know this is valid shell.

radupopescu: I always have to Google parantheses, brackets, operators etc. in shell!

PR #2248, opened 2018-10-12
Simplify the use of cvmfs_server {abort,publish} with repository gateway

radupopescu:
Addresses [CVM-1601](https://sft.its.cern.ACch/jira/browse/CVM-1601).

In the presence of a repository gateway, the lease subpath
(i.e. repo.cern.ch/some/path/to/lock) only needs to be given to the
cvmfs_server transaction command. The abort and publish commands
can find and use the session token of the current lease.

This PR allows the following workflow:
```
$ cvmfs_server transaction repo.cern.ch/some/path
$ cvmfs_server publish repo.cern.ch
or
$ cvmfs_server abort repo.cern.ch
```
As per the existing convention, if there is a single repository configured on the machine, even the repository name can be omitted.

### DIFF
--- a/cvmfs/server/cvmfs_server_abort.sh
+++ b/cvmfs/server/cvmfs_server_abort.sh
@@ -38,8 +38,8 @@ cvmfs_server_abort() {
 
   for name in $names; do
 
-    # Check if the repo name contains a subpath for locking, e.g. repo.cern.ch/sub/path/for/locking
-    local subpath=$(echo $name | cut -d'/' -f2- -s)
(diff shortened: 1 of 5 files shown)


jblomer:
I guess we can adjust the 8** integration tests accordingly.

radupopescu:
I updated `cvmfs_server {publish,abort}` use in the 8** series integration tests, but there were only a couple of uses to change. I would not change the calls to omit the repo name, since there could be other repos around...

PR #2258, opened 2018-10-24
Add option to disable FUSE notify-based inode invalidation

radupopescu:
Adding client configuration option `CVMFS_FUSE_NOTIFY_INVALIDATION` which can be set to `false` to disable the notification-based inode invalidation process for FUSE.
This feature is causing instability on macOS and will be disabled there by default.


cvmfs/cvmfs.cc
```diff
@@ -1814,8 +1814,18 @@ static int Init(const loader::LoaderExports *loader_exports) {
...
   if (loader_exports->version >= 4)
     channel = loader_exports->fuse_channel;
-  cvmfs::fuse_remounter_ = new FuseRemounter(
-    cvmfs::mount_point_, &cvmfs::inode_generation_info_, channel);
+
+  bool use_fuse_notify_invalidation = true;
+  std::string buf;
+  if (cvmfs::options_mgr_->GetValue("CVMFS_FUSE_NOTIFY_INVALIDATION",
+                                    &buf)) {
+    if (buf == "false") {
```
jblomer: We could check for  `!cvmfs::options_mgr_::IsOn(...)`, which understands all of `true`, `1`, `on`, etc.

cvmfs/fuse_evict.h
```diff
@@ -94,6 +95,8 @@ class FuseInvalidator : SingleCopy {
    */
   atomic_int32 terminated_;
   BigVector<uint64_t> evict_list_;
+
+  static bool use_fuse_notify_invalidation_;
```
jblomer: Let's call it `g_use_fuse_notify_invalidation_` to make clear it is a global entity.

PR #2263, opened 2018-11-09
Update VOMS info cache when it is out of date instead of denying request

DrDaveD:
See [CVM-1653](https://sft.its.cern.ch/jira/browse/CVM-1653)

### DIFF
--- a/cvmfs/authz/authz_session.cc
+++ b/cvmfs/authz/authz_session.cc
@@ -218,11 +218,12 @@ bool AuthzSessionManager::LookupAuthzData(
              session_key.sid, authz_data->membership.c_str(),
              authz_data->status);
     const bool granted = authz_data->IsGranted(membership);
-    if (granted)
+    if (granted) {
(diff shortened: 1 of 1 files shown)


jblomer:
It seems the PR needs to come with adjusted unit tests: https://travis-ci.org/cvmfs/cvmfs/jobs/453071916

DrDaveD:
Please help debug it.  I can't figure out what's supposed to happen in that code, and github says you were the one who checked it in so you hopefully understand it better than I do.  I can't even figure out where the "A" and "B" are coming from in the test.  

For some reason, in the running client the "membership" string passed to LookupAuthzData is the entire list of all DNs, not just the running user's DN.  I suspect that was not how you envisioned it would be used.  Any time that list changes, I believe as read from xattr from the client, in the original code "authz_data->IsGranted(membership);" always returns false because the membership list is different from the cached list.

DrDaveD:
Replaced by #2271

PR #2264, opened 2018-11-13
Fine-grained syncfs control through CVMFS_SYNCFS_LEVEL=[none,default,cautious]

jblomer:


cvmfs/server/cvmfs_server_util.sh
```diff
@@ -1025,7 +1025,9 @@ is_subcommand() {
 
 # Flushes data to disk; we might at some point want to do more than just sync
 syncfs() {
-  sync
+  if [ "x$CVMFS_DISABLE_SYNCFS" != "xtrue" ]; then
+    sync
+  fi
```
DrDaveD: So does this mean the default is false?  I strongly think the default should be the previous behavior of no sync.  It really caused havoc.

rptaylor: Usually the safe option that ensures data integrity is selected as a default, leaving it up to the user or administrator to choose an unsafe option for increased performance (for example write barriers).

jblomer: There must be something else on these nodes, perhaps the choice of the file system? We didn't see any problems at CERN, afaik.  Also, even before 2.5.1 `cvmfs_server` [occasionally called `sync`](https://github.com/cvmfs/cvmfs/pull/2173/files), e.g. after every publish.

jblomer: Ok, that's news to me. @bbockelm can you elaborate a bit on the issues with `sync`?

rptaylor: Perhaps a further optimization is possible that would benefit both cases: when doing a snapshot, only call sync if an update is actually detected and new files are written. I see that the script is calling /bin/sync which flushes all filesystem and kernel buffers on the system, which could be disruptive if it is being called frequently by multiple concurrent updates.

Actually my thought on the original issue https://sft.its.cern.ch/jira/browse/CVM-1552 was that fsync() would be called on specific file handles, rather than a system wide flushing of buffers. I'm trying to recall the details of that discussion ...

rptaylor: It was email subject "crash resiliency of CVMFS server snapshot" in cvmfs-talk. 
The server in question was CO7 with a XFS filesystem with standard settings.

rptaylor: During normal operation on our stratum 1 servers I see a very short blip of about 3% CPU use for sync every 5 minutes.

DrDaveD:
Thank you

PR #2265, opened 2018-11-13
Update apache config to ignore If-Modified-Since

DrDaveD:
See [CVM-1655](https://sft.its.cern.ch/jira/browse/CVM-1655)

### DIFF
--- a/cvmfs/server/cvmfs_server_apache.sh
+++ b/cvmfs/server/cvmfs_server_apache.sh
@@ -270,6 +270,7 @@ Alias /cvmfs/${name} ${storage_dir}
         ForceType application/octet-stream
     </FilesMatch>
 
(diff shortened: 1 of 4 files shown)


cvmfs/server/cvmfs_server_apache.sh
```diff
@@ -271,6 +271,7 @@ Alias /cvmfs/${name} ${storage_dir}
     </FilesMatch>
 
     Header unset Last-Modified
+    RequestHeader unset If-Modified-Since
```
jblomer: I think it would be good to add a comment here, explaining that this is a workaround.

DrDaveD: Ok I added one for both Last-Modified and If-Modified-Since

PR #2266, opened 2018-11-14
 Add support for CVMFS_MAX_[EXTERNAL_]SERVERS

jblomer:
Optional limit for the number of stratum 1 servers. The list of servers is cut off after geo sorting.
The PR also includes a unit test fix for openssl >= 1.1.1, that keeps file descriptors to /dev/random and /dev/urandom open.


test/unittests/t_mountpoint.cc
```diff
@@ -843,3 +844,57 @@ TEST_F(T_MountPoint, History) {
...
+    mp->external_download_mgr()->GetHostInfo(&host_chain, NULL, NULL);
+    EXPECT_EQ(1U, host_chain.size());
+  }
+
+  options_mgr_.SetValue("CVMFS_EXTERNAL_MAX_SERVERS", "10");
+  {
+    UniquePtr<MountPoint> mp(MountPoint::Create("keys.cern.ch", fs.weak_ref()));
+    ASSERT_EQ(loader::kFailOk, mp->boot_status());
+    mp->external_download_mgr()->GetHostInfo(&host_chain, NULL, NULL);
+    EXPECT_EQ(1U, host_chain.size());
```
DrDaveD: Why is this value 1 when it is 2 below where CVMFS_EXTERNAL_MAX_SERVERS=0?

jblomer: I added a comment, as for the settings below: we explicitly set `CVMFS_EXTERNAL_URL` only further down (`options_mgr_.SetValue("CVMFS_EXTERNAL_URL", server_url);`). The `CVMFS_EXTERNAL_MAX_SERVERS=0` setting means "no limit".

PR #2269, opened 2018-11-14
Fix cache cleanup logic for "big files"

jblomer:


cvmfs/cache_posix.cc
```diff
@@ -482,10 +482,16 @@ int PosixCacheManager::StartTxn(
...
     if (size > kBigFile) {
-      assert(quota_mgr_->GetCapacity() >= size);
-      quota_mgr_->Cleanup(quota_mgr_->GetCapacity() - size);
+      uint64_t cache_size = quota_mgr_->GetSize();
+      uint64_t cache_capacity = quota_mgr_->GetCapacity();
+      assert(cache_capacity >= size);
+      if ((cache_size + size) > cache_capacity) {
+        uint64_t leave_size =
+          std::min(cache_capacity / 2, cache_capacity - size);
+        quota_mgr_->Cleanup(leave_size);
```
DrDaveD: Ok so this immediately cleans up down to the size that an async cleanup would do.  I thought there was perhaps a concern of that taking too long and thought you might instead somehow send an indication to the async cleanup to also do its cleaning.  But if you think this is fine, OK.

jblomer: That is a good point!  I didn't consider it but now that I think about it I believe it should be ok.  The reason is that the actual `unlink()` calls are anyway asynchronous.  The synchronous part traverses the SQlite database and collects all the chunks that should be removed from the cache. With this modification, the list of items to be removed is larger.  But given all the the other operations that are part of the `Cleanup()` call, compiling a larger list should only add a little bit to the overall duration.

PR #2271, opened 2018-11-16
Clear authz session cache when upstream membership requirement changes

jblomer:
Replaces #2263 
@DrDaveD Can you test whether this change behaves as expected?


DrDaveD:
Will do.  I will need help from @tpdownes to change the authorization data.  Tom: although my fix worked for the test case it was the wrong way to do it.  Jacob wants us to try this better way.  I will build it and start a test run.

DrDaveD:
Yes the test was successful.

PR #2273, opened 2018-11-19
Log to syslog when session authorization disappears

DrDaveD:
See [CVM-1658](https://sft.its.cern.ch/jira/browse/CVM-1658)

### DIFF
--- a/cvmfs/authz/authz_session.cc
+++ b/cvmfs/authz/authz_session.cc
@@ -122,7 +122,8 @@ bool AuthzSessionManager::GetPidInfo(pid_t pid, PidKey *pid_key) {
 
   FILE *fp_stat = fopen(pid_path, "r");
   if (fp_stat == NULL) {
-    LogCvmfs(kLogAuthz, kLogDebug, "Failed to open status file.");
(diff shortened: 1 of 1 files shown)


cvmfs/authz/authz_session.cc
```diff
@@ -122,7 +123,11 @@ bool AuthzSessionManager::GetPidInfo(pid_t pid, PidKey *pid_key) {
 
   FILE *fp_stat = fopen(pid_path, "r");
   if (fp_stat == NULL) {
-    LogCvmfs(kLogAuthz, kLogDebug, "Failed to open status file.");
+    LogCvmfs(kLogAuthz, kLogDebug, 
+             "Failed to open status file /proc/%d/stat: (errno=%d) %s",
+             pid, errno, strerror(errno));
+    LogCvmfs(kLogAuthz, kLogSyslogWarn | kLogDebug,
```
jblomer: I'm not quite sure about this change. Isn't there a chance that this will flood the syslog with entries?

DrDaveD: In my testing where I was running 6 parallel processes under one session over ssh, and the ssh session died so the parent shell died, there was only one message in the debug log.  I think this is a relatively rare occurrence and so don't think it will flood the syslog.  Under normal circumstances jobs should run to completion before their parent dies.

PR #2278, opened 2018-11-26
Add support for CVMFS_DNS_{MIN,MAX}_TTL

jblomer:


radupopescu:
This look correct to me.

The only question I have is whether in `Resolver::set_{min,max}_ttl` it would be useful to enforce some lower/upper bounds for the values, to avoid some abusive cases like `max ttl` of 1s. It is also the responsibility of the user to ensure that `max ttl > min ttl`.

traylenator:
@jblomer this is in seconds?

PR #2284, opened 2018-12-05
Enable creation of nested catalogs during ingestion

siscia:
The first important fix is about the test.
The final `$` is necessary to tell grep to match a new line. Without it,
grep match the few catalogs that have been normally added in the test.
(The catalog under /a and /a/b)

Then we simply fix the logic, adding a new empty file called
.cvmfscatalog into the repository. Earlier it was a file called as the
base_directory where the tar is ingested.... No idea why, but the
failure of the test I believe was shadowing the error.

Finally to make everything work I had to implement a method on the
DummyDirectory. Without the implementation it would call a method in a
base class that simply assert(false), it is not bad to implement that
method for the Dummy, otherwise we would have too re-structure a lot of
code for benefits that I fail to see.

Also address: #1163

### DIFF
--- a/cvmfs/sync_item_dummy.h
+++ b/cvmfs/sync_item_dummy.h
@@ -25,7 +25,7 @@ class SyncItemDummyDir : public SyncItemNative {
  public:
   catalog::DirectoryEntryBase CreateBasicCatalogDirent() const;
   SyncItemType GetScratchFiletype() const;
-  virtual void IsPlaceholderDirectory() const { rdonly_type_ = kItemDir; }
+  virtual void MakePlaceholderDirectory() const { rdonly_type_ = kItemDir; }
 
  protected:
   SyncItemDummyDir(const std::string &relative_parent_path,
--- a/cvmfs/sync_item_dummy.h
+++ b/cvmfs/sync_item_dummy.h
@@ -25,6 +25,7 @@ class SyncItemDummyDir : public SyncItemNative {
  public:
   catalog::DirectoryEntryBase CreateBasicCatalogDirent() const;
   SyncItemType GetScratchFiletype() const;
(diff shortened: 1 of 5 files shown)


cvmfs/sync_union_tarball.cc
```diff
@@ -336,6 +338,7 @@ void SyncUnionTarball::CreateDirectories(const std::string &target) {
       new SyncItemDummyDir(dirname, filename, this, kItemDir));
 
   ProcessUnmaterializedDirectory(dummy);
+  dirs_[target] = dummy;
```
jblomer: Is this change related to the creation of `.cvmfscatalog`?

siscia: Yes, it is the one access on L153 (https://github.com/cvmfs/cvmfs/pull/2284/files#diff-a076033a7bf9d98a9175b0dce22c8c8cR153).

siscia: We are allowing the code to create catalogs also in directory that are not in the tar file, like the root directory where we extract the tar itself.

@jblomer

cvmfs/sync_item_dummy.h
```diff
@@ -25,6 +25,7 @@ class SyncItemDummyDir : public SyncItemNative {
  public:
   catalog::DirectoryEntryBase CreateBasicCatalogDirent() const;
   SyncItemType GetScratchFiletype() const;
+  virtual void IsPlaceholderDirectory() const { rdonly_type_ = kItemDir; }
```
jblomer: One more thing, can we rename this to `MakePlaceholderDirectory` or similar?  I got for a moment confused and thought this is a boolean function.

siscia: It seems a quite good idea, I get confused too actually. Fixing it.

PR #2285, opened 2018-12-07
Updating curl to 7.62.0

djw8605:
Have not run the integration tests yet, I have only build CVMFS with the new curl.

I will try to run the integration tests (not the easiest thing to do).


jblomer:
Thanks!  I'll replace it by another PR to bump to 7.63.0.
Meanwhile, I think we will be affected by this change: https://curl.haxx.se/docs/CVE-2018-1000007.html

jblomer:
@djw8605 @bbockelm Can you check whether we need to set the `CURLOPT_UNRESTRICTED_AUTH` option given the new behavior described in https://curl.haxx.se/docs/CVE-2018-1000007.html

PR #2294, opened 2019-01-15
Added scan mode and updated stats

nhazekam:
Added scan feature (used to track expected package size).

Updated stats for added clarity of data files vs file tree. This included changes to count and size tracking for accuracy and clarity.

Moved print loop into main traversal loop and wait loop(loop after traversal is done, but copy is not). The print loop was originally in copy thread, which leads to no or limited printing when updating large images with only a few changes.

Let me know if there are any changes and I will watch to see that formatting and unit tests still pass.


nhazekam:
This is an expired pull, which contained a scan mode and the updated stats. The stat update was moved to #2306. I will reopen the scan mode in a new pull request once the code is brought up to date and rebased on master.

PR #2295, opened 2019-01-17
remove cache invalidation

siscia:
This PR remove cache invalidation after ingesting chunk of files.


jblomer:
Looking again into René's presentation: https://indico.cern.ch/event/469775/contributions/2148267/attachments/1285151/1911122/cvmfs_talk.pdf

Can we try invalidating the file that has just been processed instead of the newly created file?  That should avoid the flushing problem but bring the same or more savings in page cache.

PR #2297, opened 2019-01-18
Move kernel cache on the correct side of the pipeline

siscia:
Moving the removal of the cache invalidation on the correct side of the pipeline.

Performances tests are coming.

### DIFF
--- a/cvmfs/ingestion/ingestion_source.h
+++ b/cvmfs/ingestion/ingestion_source.h
@@ -75,6 +75,11 @@ class FileIngestionSource : public IngestionSource {
 
   bool Close() {
     if (fd_ == -1) return true;
+
+    // tell to the OS that we are not going to access the file again in the
+    // foreaseable future.
+    (void)platform_invalidate_kcache(fd_, 0, 0);
+
     int ret = close(fd_);
     fd_ = -1;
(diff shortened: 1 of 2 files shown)


cvmfs/ingestion/ingestion_source.h
```diff
@@ -75,6 +75,11 @@ class FileIngestionSource : public IngestionSource {
 
   bool Close() {
     if (fd_ == -1) return true;
+
+    // tell to the OS that we are not going to access the file again in the
+    // foreaseable future.
+    (void)platform_invalidate_kcache(fd_, 0, 0);
```
jblomer: Don't we need to set the last parameter to the file size?

siscia:
First test show no difference in time when invalidating the cache after finishing reading the file or not invalidating the cache at all.

My suggestion would be to just drop the invalidation completely, so merging #2295 and closing this one (#2297).

But both work fine!

@jblomer, your call!

siscia:
(No idea what is wrong with travis)

jblomer:
Thanks!  I'm merging this into the devel branch.  Can you make another PR for the 2.5 branch (which doesn't have the ingestion logic yet)?

PR #2302, opened 2019-01-25
add ducc and cmake to compile it

siscia:
Add ducc into the main repository


jblomer:
The version is indeed tricky.  Let's check at least for a `go` compiler though.

siscia:
@jblomer this version should be better indeed.

radupopescu:
While I am working on fixing the Mac build, I needed to take the two nodes offline for testing.
I'll write back here once this is solved and Jenkins jobs can resume.

radupopescu:
I brought back the old Mac node (running 10.13) so you can go ahead with the merges, if needed. The new node running 10.14 still needs work.

siscia:
@jblomer would you mind have a look at the latest commit https://github.com/cvmfs/cvmfs/pull/2302/commits/68b1cb541ec25fb3fda2fbe1a122c10123063b4b and being sure that it is how we want to build ducc

PR #2306, opened 2019-02-05
Improve shrinkwrap statistics output

nhazekam:
Add/clarify several stat fields for Shrinkwrap. 

The stat print cycle was move to the main loop to avoid unresponsive periods when retraversing existing repositories.

Also removed unused retries option/variable.

### DIFF
--- a/cvmfs/shrinkwrap/fs_traversal.cc
+++ b/cvmfs/shrinkwrap/fs_traversal.cc
@@ -5,7 +5,6 @@
 #include <errno.h>
 #include <pthread.h>
 #include <stdio.h>
-#include <time.h>
 
 #include <fstream>
 #include <map>
@@ -15,6 +14,7 @@
 #include "cvmfs_config.h"
 #include "libcvmfs.h"
 #include "logging.h"
+#include "platform.h"
 #include "shrinkwrap/fs_traversal.h"
 #include "shrinkwrap/fs_traversal_interface.h"
 #include "shrinkwrap/fs_traversal_libcvmfs.h"
@@ -92,7 +92,7 @@ class RecDir {
 
 unsigned             num_parallel_ = 0;
 bool                 recursive = true;
-int                  stat_update_period_ = 10;
+uint64_t             stat_update_period_ = 0;  // Off for testing
 int                  pipe_chunks[2];
 // required for concurrent reading
 pthread_mutex_t      lock_pipe = PTHREAD_MUTEX_INITIALIZER;
@@ -564,7 +564,7 @@ bool SyncFull(
   struct fs_traversal *src,
   struct fs_traversal *dest,
   perf::Statistics *pstats,
-  time_t last_print_time) {
+  uint64_t last_print_time) {
   if (dirs_.empty()) {
     dirs_.push_back(new RecDir("", true));
   }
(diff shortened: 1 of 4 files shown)


nhazekam:
@jblomer This is the bug fix related to the stat print cycle and more descriptive stat names. This sets the default print cycle to 10 seconds, which I can update if needed.

cvmfs/shrinkwrap/fs_traversal.cc
```diff
@@ -694,10 +714,18 @@ int SyncInit(
     spec_tree_ = SpecTree::Create(spec);
   }
 
+  time_t last_print_time = 0;
   add_dir_for_sync(base, recursive);
-  int result = !SyncFull(src, dest, pstats);
+  int result = !SyncFull(src, dest, pstats, last_print_time);
 
   while (atomic_read64(&copy_queue) != 0) {
+    if (time(NULL)-last_print_time > stat_update_period_) {
```
jblomer: See above

cvmfs/shrinkwrap/fs_traversal.cc
```diff
@@ -564,6 +578,14 @@ bool SyncFull(
       return false;
     }
 
+    if (stat_update_period_ > 0 &&
+        time(NULL)-last_print_time > stat_update_period_) {
```
jblomer: You'll find a `platform_monotonic_time()` in platform_linux.h and platform_osx.h, which is more appropriate here.

cvmfs/shrinkwrap/shrinkwrap.cc
```diff
@@ -94,7 +94,7 @@ void Usage() {
         " -z --dest-config Dest config\n"
         " -t --spec-file   Specification file [default=$REPO.spec]\n"
         " -j --threads     Number of concurrent copy threads [default:2*CPUs]\n"
-        " -r --retries     Number of retries on copying file [default:0]\n"
+        " -p --stat-period Frequency of stat prints, 0 disables [default:10]\n"
```
jblomer: Why are we removing the `retries` option? I guess retry on download is covered by cvmfs options?

nhazekam: At the time of removal this option did not actually impact retries. I can add this back in and have it passed through to the CVMFS options. It was intended for setting retries in shrinkwrap, but as the CVMFS client handled this its redundant and could cause excessive retries if CVMFS does a full round of retries for each shrinkwrap retry.

jblomer: I see, thanks!  Let's remove the `-r` option then.

cvmfs/shrinkwrap/shrinkwrap.cc
```diff
@@ -162,14 +164,44 @@ int main(int argc, char **argv) {
         params.spec_trace_path = optarg;
         break;
       case 'j':
-        params.num_parallel = atoi(optarg);
-        break;
-      case 'n':
-        params.retries = atoi(optarg);
+        errno = 0;
+        endptr = NULL;
+        tmp_value = strtoll(optarg, &endptr, 10);
```
nhazekam: This now is nearly to `String2Uint64Parse` and checks that the option was fully and correctly parsed. I can refactor to use the existing function if you like, but I did not directly as I would have needed to convert optarg to String. This would be a small change so please let me know. @jblomer

jblomer: @nhazekam yes, please go ahead.

nhazekam:
@jblomer Let me know if there are any other changes you would like for this pull.

PR #2307, opened 2019-02-06
Lease on ingest

siscia:
@radupopescu I am not sure I understand.

This PR seems to work but I never invoke 
```bash
__swissknife lease -a acquire -u $repo_services_url -k $gw_key_file -p $name"/"$base_dir
```

In a similar PR I invoke it, but it was like it was invoked twice... and so never completed...

### DIFF
--- a/cvmfs/server/cvmfs_server_ingest.sh
+++ b/cvmfs/server/cvmfs_server_ingest.sh
@@ -23,11 +23,6 @@ cvmfs_server_ingest() {
   local force_native=0
   local force_external=0
 
-  local retry=0
-  local init_retry_delay=5 # seconds
-  local max_retry_delay=300 # seconds
-  local num_retries=100
-
   # if we use the gateway we cannot easily accept multiple deletion
   local multiple_delete=0
 
--- a/cvmfs/server/cvmfs_server_ingest.sh
+++ b/cvmfs/server/cvmfs_server_ingest.sh
@@ -220,7 +220,7 @@ cvmfs_server_ingest() {
 
   if [ ! x"$to_delete" = "x" ]; then
     ingest_command="$ingest_command -D $to_delete"
-  fiGG
+  fi
 
   if [ "$create_catalog" = true ]; then
     ingest_command="$ingest_command -C true"
@@ -231,7 +231,7 @@ cvmfs_server_ingest() {
   fi
 
         ############# WIP ##########
-   
+
   local upstream_storage=$CVMFS_UPSTREAM_STORAGE
   local upstream_type=$(get_upstream_type $upstream_storage)
   gw_key_file=/etc/cvmfs/keys/${name}.gw
@@ -250,15 +250,15 @@ cvmfs_server_ingest() {
     if [ ! x"$tar_file" = "x" ]; then
       subpath=$tar_file
     fi
-    
+
     if [ ! x"$to_delete" = "x" ]; then
       subpath=$to_delete
     fi
 
     local repo_services_url=$(echo $upstream_storage | cut -d',' -f3)
- 
+
     set +e
- 
+
     local res=1
     local r=1
     local delay=$init_retry_delay
@@ -276,9 +276,9 @@ cvmfs_server_ingest() {
         delay=$max_retry_delay
       fi
     done
- 
+
     set -e
- 
+
     if [ $res -ne 0 ]; then
       die "Could not acquire a new lease for repository $name"
     fi
--- a/cvmfs/server/cvmfs_server_ingest.sh
+++ b/cvmfs/server/cvmfs_server_ingest.sh
@@ -29,8 +29,8 @@ cvmfs_server_ingest() {
   local num_retries=100
 
   # if we use the gateway we cannot easily accept multiple deletion
-  local multiple_delete=0 
-  
+  local multiple_delete=0
+
   while [ "$2" != "" ]; do
     case $1 in
       -b | --base_dir )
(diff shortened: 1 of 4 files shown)


radupopescu:
If you are able to publish without acquiring a lease, I suspect that you are using the session token that was already acquired through an earlier call to `cvmfs_server transaction`.

I think the `cvmfs_swissknife ingest` command should take the following steps:
1. Try to get a lease for the relevant path - if not possible, abort and inform the user (they can call `cvmfs_server abort` if there is an existing lease that conflicts).
2. Perform the tarball ingestion
3. Make sure that `FinalizeSession(true, ....)` is called on the `Spooler` object to commit the transaction on the gateway (see how it's done in swissknife_sync.cc):

https://github.com/cvmfs/cvmfs/blob/ce33b7b16724b2c16a597deee6586e873e0b2f39/cvmfs/swissknife_sync.cc#L915

4. If everything is ok, we are done. If there was an error, drop the active lease (`cvmfs_swissknife lease -a drop ...`).

Makes sense?

siscia:
@radupopescu How do I know if I am using an earlier lease?

Similarly, is there a way to know what paths are leased at the moment?

radupopescu:
@siscia 

There can only be one active lease on each publisher machine and the session token is saved in `/var/spool/cvmfs/<REPO>/session_token_`. `cvmfs_swissknife lease -a acquire` should give an error if there is an active lease.

Right now, I don't think it's possible to know which paths are leased.

I guess that with the ingestion, it becomes possible to have two concurrent "transactions" per repo on the same machine. Some thought needs to be given to how to handle multiple session token files.

siscia:
@radupopescu , we should look into this together, that the session token is create and destructed even without the `cvmfs_swissknife lease` command invoked inside the `cvmfs_server_ingest`.

It must be invoked somewhere else that we should track down and understand if correct.

Are there any way that the session token get created?

siscia:
Todo:
[ ] re-run test, the commit ab2ec

siscia:
I believe this PR is now ready for revision!

@jblomer @radupopescu

cvmfs/swissknife_ingest.h
```diff
@@ -38,6 +38,9 @@ class Ingest : public Command {
     r.push_back(Parameter::Optional(
         'C', "create a new catalog where the tar file is extracted"));
 
+    r.push_back(Parameter::Optional('P', "session_token_file"));
```
radupopescu: How are these two parameters actually used?

cvmfs/server/cvmfs_server_ingest.sh
```diff
@@ -210,6 +228,15 @@ cvmfs_server_ingest() {
     ingest_command="$ingest_command -+stats"
   fi
 
+  local upstream_storage=$CVMFS_UPSTREAM_STORAGE
+  local upstream_type=$(get_upstream_type $upstream_storage)
+  gw_key_file=/etc/cvmfs/keys/${name}.gw
+
+  if [ x"$upstream_type" = xgw ]; then
+    ingest_command="$ingest_command -H $gw_key_file -P ${spool_dir}/session_token"
```
radupopescu: You are passing the gw key and the session token. Below, I don't see how these are used.

cvmfs/server/cvmfs_server_ingest.sh
```diff
@@ -62,8 +65,23 @@ cvmfs_server_ingest() {
     die "Please set the base directory where to extract the tarball, use -b \$BASE_DIR or --base_dir \$BASE_DIR or don't provide the base directory to simply delete entities from the repository"
   fi
 
+  if [ x"$upstream_type" = xgw ]; then
+
+    if [ $multiple_delete -eq 1 ]; then
```
radupopescu: What is the cause of these limitations?

cvmfs/swissknife_ingest.h
siscia: Here: https://github.com/cvmfs/cvmfs/blob/74efe86614d6e145eb75e93d145adfd6865a4035/cvmfs/swissknife_ingest.cc#L76-L90

You don't think is necessary?

cvmfs/server/cvmfs_server_ingest.sh
siscia: I could delete files in two completely different directories, and I would need to take a lease on the common parent of those directory. Correct?

It seems wiser to avoid this complexity and just drive the user to make smaller deletes.

cvmfs/swissknife_ingest.h
radupopescu: I understand now, thanks.

cvmfs/server/cvmfs_server_ingest.sh
radupopescu: Never mind this, you explained below.

cvmfs/server/cvmfs_server_ingest.sh
radupopescu: Thanks for the clarifications. Makes sense, I agree that this can be avoided now and implemented later if there is an actual need.

siscia:
Radu,

if I understand correctly you are looking for this call here: https://github.com/cvmfs/cvmfs/blob/74efe86614d6e145eb75e93d145adfd6865a4035/cvmfs/server/cvmfs_server_ingest.sh#L358 correct? Is that enough?

For the 80x tests I don't have much experience with those, I will need some time...

radupopescu:
@siscia  Yes, that's the call I was looking for. Thanks!

About the test, no rush, but I think it's really useful to have it with the PR. I would use test 800 as a starting point:

https://github.com/cvmfs/cvmfs/blob/devel/test/src/800-repository_gateway/main

siscia:
@radupopescu I took heavy inspiration from the 800 test.
Let me know what you think!

siscia:
Do NOT merge yet.

The test added in the latest commit fails. Investigating now.

radupopescu:
Great for the extra tests!

Let me know when ready...

siscia:
@radupopescu we should be good now!

radupopescu:
Great!

PR #2308, opened 2019-02-07
[externals] Update c-ares to version 1.15.0

jblomer:
@radupopescu Perhaps that helps with the new macOS machine?  The [comment](https://github.com/cvmfs/cvmfs/blob/devel/externals/c-ares/src/configureHook.sh#L3) in the c-ares build script could be relevant, too.


radupopescu:
@jblomer The Mac problem is fixed now, was related to some header files which were removed from the base macOS 10.14, but were made available through a separate package.

Updating the version of c-ares should not cause any problems, though.

PR #2324, opened 2019-02-21
Remove raw mutex and use MutexLockGuards

siscia:
Trying to address https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-694?filter=allopenissues&orderby=cf%5B11250%5D+ASC%2C+priority+DESC%2C+updated+DESC

All almost done, the raw locks that are in the code base are better to leave in. IMHO.

The only file that I was not able to clean up is: cvmfs/logging.cc

It gives some problem when importing "util_concurrency.h" and it necessitate more investigation, nevertheless I would suggest to merge it.

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -694,14 +694,15 @@ static void cvmfs_opendir(fuse_req_t req, fuse_ino_t ino,
     stream_listing.capacity = 0;
 
   // Save the directory listing and return a handle to the listing
-  pthread_mutex_lock(&lock_directory_handles_);
-  LogCvmfs(kLogCvmfs, kLogDebug,
-           "linking directory handle %d to dir inode: %" PRIu64,
-           next_directory_handle_, uint64_t(ino));
-  (*directory_handles_)[next_directory_handle_] = stream_listing;
-  fi->fh = next_directory_handle_;
-  ++next_directory_handle_;
-  pthread_mutex_unlock(&lock_directory_handles_);
+  {
+    MutexLockGuard m(&lock_directory_handles_);
+    LogCvmfs(kLogCvmfs, kLogDebug,
+             "linking directory handle %d to dir inode: %" PRIu64,
+             next_directory_handle_, uint64_t(ino));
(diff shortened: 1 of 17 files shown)


cvmfs/cvmfs.cc
```diff
@@ -721,20 +722,19 @@ static void cvmfs_releasedir(fuse_req_t req, fuse_ino_t ino,
...
-  DirectoryHandles::iterator iter_handle =
-    directory_handles_->find(fi->fh);
-  if (iter_handle != directory_handles_->end()) {
-    if (iter_handle->second.capacity == 0)
-      smunmap(iter_handle->second.buffer);
-    else
-      free(iter_handle->second.buffer);
-    directory_handles_->erase(iter_handle);
-    pthread_mutex_unlock(&lock_directory_handles_);
-    perf::Dec(file_system_->no_open_dirs());
```
siscia: Here I am not sure, since the call to perf::Dec was not in the protected section... Move it outside the protected section was quite complex.

cvmfs/fetch.cc
```diff
@@ -94,31 +98,32 @@ int Fetcher::Fetch(
...
+    // Synchronization point: either act as a master thread for this object or
+    // enqueue to the list of waiting threads.
+    MutexLockGuard m(lock_queues_download_);
+    ThreadQueues::iterator iDownloadQueue = queues_download_.find(id);
+    if (iDownloadQueue != queues_download_.end()) {
+      LogCvmfs(kLogCache, kLogDebug, "waiting for download of %s",
+               name.c_str());
+
+      iDownloadQueue->second->push_back(tls->pipe_wait[1]);
+      ReadPipe(tls->pipe_wait[0], &fd_return, sizeof(int));
```
siscia: Also here, the call to ReadPipe is outside the lock in the original code while I put it inside.

I don't believe is a problem but is worth a second look...

cvmfs/cvmfs.cc
jblomer: That should be fine.

PR #2331, opened 2019-03-01
Fix issues with char device and overlayfs

siscia:
Overlayfs use char devices with number 0/0 to mark whiteouts.

In the previous implementation, every char device was assigned 0/0 as number.

I fix the issue assigning the correct dev numbers.

Still running the tests.


siscia:
@jblomer I also add the test to the "quick" suites.

PR #2341, opened 2019-03-29
fix transaction lock, use base_dir instead of tar_file

siscia:
As discussed in the mail list the lock should be taken on the directory and not on the tarfile ingested.


siscia:
@DrDaveD

jblomer:
Thanks!  Please open against the 2.6 branch.

siscia:
Sup by #2342

PR #2343, opened 2019-04-05
Libfuse3 support

jblomer:
This provides basic support for libfuse3 without exploiting the new features yet.

The cvmfs client decides dynamically whether to use libfuse2 or libfuse3.  To do so, the cvmfs2 binary now loads the libcvmfs_fuse_stub or libcvmfs_fuse3_stub library, depending on availability and on the new `libfuse=[2,3]` mount option. The libcvmfs_fuse_stub library thus contains the code the previously was in the cvmfs2 binary.  The libcvmfs_fuse3_stub library loads the corresponding libcvmfs_fuse3 library.

This allows to add libfuse3 support simply by deploying the additional libcvmfs_fuse3 and libcvmfs_fuse3_stub libraries. For FC29, the sub package cvmfs-fuse3 does just that. It also ensures that updating and hotpatching existing libfuse2 based clients continues to work.


DrDaveD:
I am working on a fuse3 installation for epel7, and with my test version installed, the current version of this builds the libcvmfs_fuse3 packages but then rpmbuild dies because they are not packaged.  I suggest removing them during the install step `%if ! 0%{?build_fuse3}`.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -370,6 +370,11 @@ rm -f $RPM_BUILD_ROOT/etc/cvmfs/domain.d/*.conf
 rm -f $RPM_BUILD_ROOT/etc/cvmfs/default.d/*.conf
 rm -f $RPM_BUILD_ROOT/etc/cvmfs/serverorder.sh
 
+# Don't install coincidentially built libfuse3 libraries
+%if ! 0%{?build_fuse3}
+rm -f $RPM_BUILD_ROOT%{_libdir}/libcvmfs_fuse3
```
DrDaveD: Shouldn't there be an asterisk on the end of that?

DrDaveD:
I noticed the libfuse3 files are in the cvmfs rpm, not the cvmfs-fuse3 rpm.  The latter has only /usr/share/doc files.  Was that your intention?

PR #2348, opened 2019-04-10
Whitelist data/txn/receiver in swissknife scrub

radupopescu:
Addresses [CVM-1704](https://sft.its.cern.ch/jira/browse/CVM-1704).

### DIFF
--- a/test/src/800-repository_gateway/main
+++ b/test/src/800-repository_gateway/main
@@ -21,7 +21,7 @@ compare_file_checksum() {
     echo "Checksum of $file_name is $checksum"
     if [ "$checksum" != "$target_checksum" ]; then
         echo "Checksum mismatch for $file_name. Expected $target_checksum. Found $checksum"
-        exit 1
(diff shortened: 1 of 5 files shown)


jblomer:
Thanks!  I think we should add the `cvmfs_server check -i $reponame` command somewhere in our quick integration tests, too.

test/src/800-repository_gateway/main
```diff
@@ -180,15 +180,21 @@ run_transactions() {
...
+    local integrity_check_res=$?
 
     echo "  Check that the named tag has been created"
     specialtag_line=$(cvmfs_server tag -l test.repo.org | grep specialtag)
 
     ## Check results
 
+    if [ $? -ne 0 ]; then
+        echo "Repository integrity check failed: $integrity_check_res"
+        exit 1
```
jblomer: `exit` might end the test cases altogether, `return 1` should do the trick.

PR #2351, opened 2019-04-30
Fix potential hang during reload if a config repository is used

jblomer:


DrDaveD:
Doesn't this apply even if there is no blacklist in the config repository?  Just the fact that the other repos look for a blacklist in the config repo could be a problem.

How about also having a test that does many reloads on many repositories many times, to try to catch race conditions like this one?

jblomer:
I guess that is correct.  Having a config repository should be enough.

Regarding reload stress tests, there are tests [023](https://github.com/cvmfs/cvmfs/blob/cvmfs-2.6/test/src/023-reload_safe_path_traversal/main) and [024](https://github.com/cvmfs/cvmfs/blob/cvmfs-2.6/test/src/024-reload-during-asetup/main). Both of them should use cvmfs-config.cern.ch as a config repository, but only one "real" repository.  Apparently that and our comparably small test node fleet has not been enough to ever trigger the problem.

Based on the FNAL numbers, can you estimate how many repositories and trials we'd need to get into a range where we could expect the problem to be triggered every now and then?

DrDaveD:
I counted the number of IP addresses that have contacted the squid today and yesterday, and it is 1880.  There were about 20 repositories mounted on the machine we investigated, so that's about 38k trials to repeat the Fermilab tests.  There were about 50 hung nodes, so you should get a failure about every 750 trials with 20 repositories.

DrDaveD:
The blacklist is checked on every reload so shortening the ttl would probably make it happen faster

PR #2352, opened 2019-05-13
Include reflog checksum in the repository manifest

radupopescu:
Addresses [CVM-1732](https://sft.its.cern.ch/jira/browse/CVM-1732).

* This PR makes the hash of the repository reflog available as an optional member of the manifest. This is needed for cases where `/var/spool/cvmfs/<REPO>/reflog.chksum` is not available, such as on publisher machines connected to a repository gateway.
* The manifest signing tool (`signing_tool.cc`) now obtains the reflog hash from the manifest, if the name of a file on disk containing the hash is not given.
* The `cvmfs_server check` command now falls back to the reflog hash contained in the manifest, if the "reflog.chksum" file is not present, for validating the reflog.

### DIFF
--- a/cvmfs/manifest.cc
+++ b/cvmfs/manifest.cc
@@ -68,6 +68,7 @@ Manifest *Manifest::Load(const map<char, string> &content) {
   bool garbage_collectable = false;
   bool has_alt_catalog_path = false;
   shash::Any meta_info;
+  shash::Any reflog_checksum;
 
   if ((iter = content.find('B')) != content.end())
     catalog_size = String2Uint64(iter->second);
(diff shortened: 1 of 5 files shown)


cvmfs/manifest.cc
```diff
@@ -68,6 +68,7 @@ Manifest *Manifest::Load(const map<char, string> &content) {
   bool garbage_collectable = false;
   bool has_alt_catalog_path = false;
   shash::Any meta_info;
+  shash::Any reflog_checksum;
```
jblomer: I'd rather call it hash than checksum, consistent with the other fields.

PR #2355, opened 2019-05-23
Set CVMFS_CATALOG_WATERMARK on publisher nodes

jblomer:


DrDaveD:
Wouldn't it be better to make the watermark default be calculated by some algorithm based on the setting of CVMFS_NFILES?  Certainly it should be lower than CVMFS_NFILES if somebody sets a lower value.

jblomer:
That would be better, yes. E.g. 1/4 of the effective fd limit, unless overwritten by `CVMFS_CATALOG_WATERMARK`.  I'll have a look.

jblomer:
Closing, should be covered by the latest changes in #2354

PR #2356, opened 2019-05-24
build fuse3 on rhel >= 6

DrDaveD:
[CVM-1744](https://sft.its.cern.ch/jira/browse/CVM-1744)

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -30,7 +30,8 @@
 %define systemd_autofs_patch 1
 %endif
 
-%if 0%{?fedora} >= 29
+# fuse3 is in epel starting with epel6
+%if 0%{?fedora} >= 29 || 0%{?rhel} >= 6
 %define build_fuse3 1
 %endif
 


packaging/rpm/cvmfs-universal.spec
```diff
@@ -30,7 +30,8 @@
 %define systemd_autofs_patch 1
 %endif
 
-%if 0%{?fedora} >= 29
+# fuse3 is in epel starting with epel6
+%if 0%{?fedora} >= 29 || 0%{?rhel} >= 6
```
jblomer: Shouldn't that be `el` instead of `rhel`?

DrDaveD: [Nope](https://fedoraproject.org/wiki/Packaging:DistTag#Conditionals)

jblomer: Ok, so I understand both work in the same way.  As everything else uses `el` in the spec file, for consistency reasons I'd prefer to stick to it.

DrDaveD: There is no %{el}, only individual %{el#} numbers.  When they went to the general number they used %{rhel}.  Inconsistent, but that's the way it is.
```
$ rpm --eval '%{el}'
%{el}
$ rpm --eval '%{rhel}'
6
```

jblomer: Oh, I see. Thanks!

PR #2361, opened 2019-06-06
[WIP] Support fuse premount in mount.cvmfs helper

DrDaveD:
With these changes I am able to run mount.cvmfs directly from the new fuse premounting function in [singularity PR 3679](https://github.com/sylabs/singularity/pull/3679). 

There are still a few issues I have noticed that I may want to add fixes for in this PR, so for now consider this a WIP, but these changes are needed minimally.  The issues I have noticed are:
1. fuse_session_loop_mt reports that it cannot support file descriptor cloning and disables it.  I don't know if this happens normally or not.
2. With singularity -p (which uses a process namespace) there are many defunct sh processes and a couple of defunct cvmfs processes.
3. Without singularity -p, cvmfs2 processes remain running after singularity exits.


DrDaveD:
Closing and reopening to re-try the macos travis falure.

DrDaveD:
I'm closing this since Jakob & I decided that we could bypass mount.cvmfs and directly use cvmfs2

PR #2362, opened 2019-06-06
Turn failure to set requested CVMFS_NFILES into a warning

DrDaveD:
Something like this is needed along with pr #2361 in order for an unprivileged user to not have to select a specific CVMFS_NFILES matching their ulimit -Hn.

### DIFF
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -59,6 +59,7 @@ struct CvmfsOptions {
   char *config;
   int uid;
   int gid;
+  int system_mount;
   int grab_mountpoint;
   int cvmfs_suid;
   int disable_watchdog;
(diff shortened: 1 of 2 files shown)


DrDaveD:
@jblomer how about this way?

cvmfs/loader.cc
```diff
@@ -814,10 +819,20 @@ int FuseMain(int argc, char *argv[]) {
...
-      return kFailPermission;
+      if (system_mount_) {
+        LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogErr,
+                 "Failed to set maximum number of open files, "
+                 "insufficient permissions");
+        return kFailPermission;
+      }
+      unsigned soft_limit, hard_limit;
+      GetLimitNoFile(&soft_limit, &hard_limit);
+      LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogWarn,
```
jblomer: I think we should use stdout if it is just a warning.

jblomer:
Thanks, looks good to me except for the minor comment.

PR #2364, opened 2019-06-10
Reload geodb after a new one is downloaded

DrDaveD:
See [CVM-1739](https://sft.its.cern.ch/jira/browse/CVM-1739).

Also move stratum1 used in the service test back from unl to fnal.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -983,7 +983,7 @@ if (BUILD_SERVER)
     )
 
     install(
-      FILES      webapi/cvmfs_api.py webapi/cvmfs_geo.py webapi/cvmfs_globals.py
+      FILES      webapi/cvmfs_api.py webapi/cvmfs_geo.py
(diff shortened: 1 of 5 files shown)


jblomer:
@DrDaveD Thanks, Dave!  Please let me know whether the unit test failure is understood.

DrDaveD:
Don't merge the current version even though it will probably pass tests; it doesn't work with the second reopen DB

DrDaveD:
Ok this version works. 

Should I get rid of the cvmfs_globals.CVMFS_UNITTESTS and cvmfs_globals.py now?  I don't know if anything else is expecting it.

DrDaveD:
I did a build after removing the python cvmfs_globals and nothing complained about them being missing, and it passed the 614-geoservice test, so I added the commit to this PR.

DrDaveD:
This one should be ready to go.

jblomer:
Ah, I forgot to double check, this went into the devel branch.  Could you please reopen against the 2.6 branch?

PR #2365, opened 2019-06-11
Explicitly specify python2 in webapi/cvmfs-api.py shebang

janpriessnitz:
As of Fedora 30, shebangs with unspecified python versions are
forbidden. Let's explicitly set the shebang to python2

### DIFF
--- a/cvmfs/webapi/cvmfs-api.wsgi
+++ b/cvmfs/webapi/cvmfs-api.wsgi
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#!/usr/bin/env python2
 
 import os, sys, re
 import cvmfs_api


packaging/rpm/cvmfs-universal.spec
```diff
@@ -83,6 +83,13 @@ BuildRequires: pkgconfig
 BuildRequires: python-devel
 BuildRequires: unzip
 
+%if 0%{?el4} || 0%{?el5} || 0%{?el6}
+BuildRequires: python2-devel
```
jblomer: Is `python2-devel` and `python3-devel` different from `python-devel` (required a few lines above)?

packaging/rpm/cvmfs-universal.spec
```diff
@@ -83,6 +83,13 @@ BuildRequires: pkgconfig
 BuildRequires: python-devel
 BuildRequires: unzip
 
+%if 0%{?el4} || 0%{?el5} || 0%{?el6}
```
jblomer: We should have this if statement only once: somewhere at the beginning of the file we should decide whether this is a python 2 or a python 3 platform and set a custom variable.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -83,6 +83,13 @@ BuildRequires: pkgconfig
...
+BuildRequires: python2-devel
+%else
+BuildRequires: python3-devel
+%endif
+BuildRequires: /usr/bin/pathfix.py
```
jblomer: Is this part of... the default installation on all platforms?

packaging/rpm/cvmfs-universal.spec
jblomer: I guess SLESS11/12 and possibly openSuSE 13 are also python2 platforms, at least to be checked

jblomer:
One more thing, please also add an entry to the rpm changelog

janpriessnitz:
@DrDaveD Hey, I'm having some issues with python shebang in https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/webapi/cvmfs-api.wsgi and I have a question: On which version of python (2/3) is the script meant to run (or is it universal)? Thanks!

DrDaveD:
It was developed with python2.  I haven't tested it with python3 so I don't know if there are any issues with that or not.

PR #2367, opened 2019-06-12
WIP: add test for looking at file before kcache timeout

DrDaveD:
This test addition demonstrates the failure reported in [CVM-1759](https://sft.its.cern.ch/jira/browse/CVM-1759)

### DIFF
--- a/test/src/611-idleclient/main
+++ b/test/src/611-idleclient/main
@@ -91,6 +91,23 @@ cvmfs_run_test() {
     return 21
   fi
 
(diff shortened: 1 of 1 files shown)


DrDaveD:
@jblomer Any progress on investigating this?

The test is ready, it is only marked WIP because it currently fails, because the current behavior is wrong.

jblomer:
@DrDaveD I unfortunately did not yet manage to look into this, hope to be able to do it next week.

jblomer:
Superseded by #2376

PR #2368, opened 2019-06-21
Add --force flag and $PATH to singularity build on DUC

siscia:
Singularity interface changed on the latest release.

These two fixes were explored with @rptaylor and it seems like they are stable enough.

### DIFF
--- a/ducc/lib/image.go
+++ b/ducc/lib/image.go
@@ -250,7 +250,7 @@ func (img Image) DownloadSingularityDirectory(rootPath string) (sing Singularity
 	}
 	defer os.RemoveAll(singularityTempCache)
 	err = ExecCommand("singularity", "build", "--force", "--sandbox", dir, img.GetSingularityLocation()).Env(
-		"SINGULARITY_CACHEDIR", singularityTempCache).Start()
+		"SINGULARITY_CACHEDIR", singularityTempCache).Env("PATH", os.Getenv("PATH")).Start()
 	if err != nil {
 		LogE(err).Error("Error in downloading the singularity image")
 		return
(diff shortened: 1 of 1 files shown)


jblomer:
Thanks!  Does it work with old and new Singularity alike?  We should probably fix it then for the 2.6.1 release.

siscia:
I tested with the 3.0 release (oldish) and the 3.2 (the latest stable release) and it was fine for both.

Going down to the 2.x version require to compile the software from source, I can try and see, but since I cannot install them from repos I believe is quite unlikely that somebody will have them...

DrDaveD:
It's true that 2.6.x isn't current anywhere anymore, and the old rpms are gone from EPEL, but they are still in the [OSG repo](https://repo.opensciencegrid.org/osg/3.4/el7/release/x86_64/) and in the [EPEL archive](https://archive.fedoraproject.org/pub/archive/epel/7/x86_64/Packages/s/).

siscia:
Thanks Dave,

indeed I was able to successfully install singularity 2.6 (actually the real problem was some handling with not so well specified dependencies).

I don't report any error on that as well.

I am going to apply the fix to unpacked.cern.ch and hopefully it will run well there as well. If I don't see regression I suggest we merge.

Cheers,
Simone

jblomer:
Thanks for the testing!  @siscia Please re-open the PR against the cvmfs-2.6 branch so that we can make it part of the cvmfs 2.6.1 release.  Is there a JIRA ticket connected to it?

siscia:
Continues on #2370

siscia:
Aw no, I don't believe we have a JIRA for this.

PR #2370, opened 2019-06-24
Add --force flag and $PATH to singularity build on DUCC

siscia:
Reopening #2368 on the correct branch.


siscia:
The test on unpacked went just fine.

I will suggest merging.

PR #2379, opened 2019-07-03
Fix cloudflare geo api with proxy name

DrDaveD:
See [CVM-1774](https://sft.its.cern.ch/jira/browse/CVM-1774)


jblomer:
This should be opened against the cvmfs-2.6 branch, I think

DrDaveD:
Yes, I keep forgetting to change the base.  Changed now.

PR #2383, opened 2019-07-05
Notification refactor

radupopescu:
This introduces some needed changes to the notification system client code, accompanying the migration of the notification server into the repository gateway.

* New notification subscriber is based on Curl, uses server-sent events
* Removed `libwebsockets` (no new third-party library)
* Notification server URL handling is simplified - both publishing and subscription tools now use the gateway API root URL
* Refactoring, cleanup, simplification, improved validation of manifests
* The notification background thread is properly joined on mountpoint reload or unmount


cvmfs/notify/cmd_sub.cc
```diff
@@ -20,37 +23,76 @@ namespace {
...
+    const std::string config_file =
+        "/etc/cvmfs/repositories.d/" + repository_ + "/client.conf";
+    SimpleOptionsParser options;
+    if (!options.TryParsePath(config_file)) {
+      LogCvmfs(kLogCvmfs, kLogError,
+               "SwissknifeSubscriber - could not parse configuration file");
+      return false;
+    }
+
+    dl_mgr_->Init(kMaxPoolHandles, kUseSystemProxy,
```
jblomer: Shouldn't there be a corresponding `Fini()` call?

cvmfs/notify/subscriber_sse.h
```diff
@@ -0,0 +1,67 @@
+
```
jblomer: First empty line can be removed

cvmfs/notify/subscriber_sse.cc
```diff
@@ -0,0 +1,158 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include "subscriber_sse.h"
+
+#include <vector>
+
+#include "cvmfs_config.h"
```
jblomer: This should rather go to the beginning of the includes

cvmfs/notify/subscriber_sse.cc
```diff
@@ -0,0 +1,158 @@
...
+    return false;
+  }
+
+  this->topic_ = topic;
+
+  std::string request = "{\"version\":1,\"repository\":\"" + topic + "\"}";
+
+  const char* user_agent_string = "cvmfs/" VERSION;
+
+  CURL* h_curl = curl_easy_init();
```
jblomer: I guess we can simply assert on `h_curl != NULL`

test/cloud_testing/platforms/common_setup.sh
```diff
@@ -68,9 +66,6 @@ while getopts "s:c:d:k:t:g:l:w:n:p:" option; do
     w)
       GATEWAY_BUILD_URL=$OPTARG
       ;;
-    n)
```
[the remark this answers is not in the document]
radupopescu: I removed uses of this flag from `cernvm/ci-scripts`

PR #2385, opened 2019-07-09
S3fanout: a few improvements + refactor

janpriessnitz:
Refactor:
- move S3 server info (hostname, port, bucket, ...) from JobInfo into S3Fanout
- construct DNS-style bucket hostname in constructor of S3Fanout 

Improvements:
- use pipes instead of jobs_todo_ and thread_upload_run_ variables to communicate jobs and termination between S3Uploader and S3Fanout
- in poll function, use timeout suggested by curl instead of default 100ms

### DIFF
--- a/cvmfs/s3fanout.cc
+++ b/cvmfs/s3fanout.cc
@@ -472,17 +472,17 @@ bool S3FanoutManager::MkV2Authz(const JobInfo &info, vector<string> *headers)
                    content_type + "\n" +
                    timestamp + "\n" +
                    "x-amz-acl:public-read" + "\n" +  // default ACL
-                   "/" + info.bucket + "/" + info.object_key;
+                   "/" + bucket_ + "/" + info.object_key;
   LogCvmfs(kLogS3Fanout, kLogDebug, "%s string to sign for: %s",
            request.c_str(), info.object_key.c_str());
 
   shash::Any hmac;
   hmac.algorithm = shash::kSha1;
-  shash::Hmac(info.secret_key,
+  shash::Hmac(secret_key_,
               reinterpret_cast<const unsigned char *>(to_sign.data()),
               to_sign.length(), &hmac);
 
-  headers->push_back("Authorization: AWS " + info.access_key + ":" +
+  headers->push_back("Authorization: AWS " + access_key_ + ":" +
                      Base64(string(reinterpret_cast<char *>(hmac.digest),
                                    hmac.GetDigestSize())));
   headers->push_back("Date: " + timestamp);
@@ -525,17 +525,15 @@ string S3FanoutManager::GetUriEncode(const string &val, bool encode_slash)
 }
 
 
-string S3FanoutManager::GetAwsV4SigningKey(
-  const JobInfo &info,
-  const string &date) const
+string S3FanoutManager::GetAwsV4SigningKey(const string &date) const
 {
-  string id = info.secret_key + info.region + date;
+  string id = secret_key_ + region_ + date;
   map<string, string>::const_iterator iter = signing_keys_.find(id);
   if (iter != signing_keys_.end())
     return iter->second;
 
-  string date_key = shash::Hmac256("AWS4" + info.secret_key, date, true);
-  string date_region_key = shash::Hmac256(date_key, info.region, true);
+  string date_key = shash::Hmac256("AWS4" + secret_key_, date, true);
+  string date_region_key = shash::Hmac256(date_key, region_, true);
   string date_region_service_key = shash::Hmac256(date_region_key, "s3", true);
   string signing_key =
     shash::Hmac256(date_region_service_key, "aws4_request", true);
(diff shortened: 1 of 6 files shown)


jblomer:
Thanks!  Please have a look at the linter errors.

jblomer:
Please also expand a bit in the PR title and description on the nature of the refactoring.

cvmfs/s3fanout.cc
```diff
@@ -244,29 +245,53 @@ void *S3FanoutManager::MainUpload(void *data) {
...
-    JobInfo *info = NULL;
-    {
-      MutexLockGuard m(s3fanout_mgr->jobs_todo_lock_);
-      if (!s3fanout_mgr->jobs_todo_.empty() &&
-          (jobs_in_flight < s3fanout_mgr->pool_max_handles_)) {
-        info = s3fanout_mgr->jobs_todo_.back();
-        s3fanout_mgr->jobs_todo_.pop_back();
+  while (true) {
+    // Check events with 100ms default timeout or timeout suggested by curl
+    int timeout = 100;
```
jblomer: Better rename it to `timeout_ms`

cvmfs/s3fanout.cc
```diff
@@ -244,29 +245,53 @@ void *S3FanoutManager::MainUpload(void *data) {
...
+    // Cannot use int64_t because curl_multi_timeout header is platform-specific
+    long curl_timeout;  // NOLINT
+    curl_multi_timeout(s3fanout_mgr->curl_multi_, &curl_timeout);
```
jblomer: I think we have to go another route.  As per the documentation, we should not use this function together with the multi socket API (which we use).

cvmfs/s3fanout.cc
```diff
@@ -601,15 +616,15 @@ bool S3FanoutManager::MkV4Authz(const JobInfo &info, vector<string> *headers)
     scope + "\n" +
     hash_request;
 
-  string signing_key = GetAwsV4SigningKey(info, date);
+  string signing_key = GetAwsV4SigningKey(date);
```
jblomer: We should cache the signing key, I guess.  Just something to keep in mind.

janpriessnitz: There is some sort of caching inside the function (the signing_keys_ map), but I think we should go in the opposite direction. Since time moves only forward, it only makes sense to cache these signing keys within one day. I suggest caching just one signing key, which gets overwritten at midnight every day.

cvmfs/s3fanout.h
```diff
@@ -236,24 +206,33 @@ class S3FanoutManager : SingleCopy {
...
+  struct S3Config {
+    std::string access_key;
+    std::string secret_key;
+    std::string hostname_port;
+    AuthzMethods authz_method;
+    std::string region;
+    std::string bucket;
+    bool dns_buckets;
+    uint32_t pool_max_handles;
+    unsigned opt_timeout_sec = 20;
```
jblomer: I think this syntax does not work with older C++ compilers, but there has to be an explicit default constructor.

cvmfs/s3fanout.h
```diff
@@ -236,24 +206,33 @@ class S3FanoutManager : SingleCopy {
...
 
+  struct S3Config {
+    std::string access_key;
+    std::string secret_key;
+    std::string hostname_port;
+    AuthzMethods authz_method;
+    std::string region;
+    std::string bucket;
+    bool dns_buckets;
+    uint32_t pool_max_handles;
```
jblomer: Initialize this member

cvmfs/s3fanout.h
```diff
@@ -236,24 +206,33 @@ class S3FanoutManager : SingleCopy {
...
   static const unsigned kDefaultHTTPPort;
 
+  struct S3Config {
+    std::string access_key;
+    std::string secret_key;
+    std::string hostname_port;
+    AuthzMethods authz_method;
+    std::string region;
+    std::string bucket;
+    bool dns_buckets;
```
jblomer: Initialize the member

cvmfs/s3fanout.h
```diff
@@ -236,24 +206,33 @@ class S3FanoutManager : SingleCopy {
   static const unsigned kThrottleReportIntervalSec;
   static const unsigned kDefaultHTTPPort;
 
+  struct S3Config {
+    std::string access_key;
+    std::string secret_key;
+    std::string hostname_port;
+    AuthzMethods authz_method;
```
jblomer: Initialize the member

cvmfs/s3fanout.cc
```diff
@@ -244,29 +245,48 @@ void *S3FanoutManager::MainUpload(void *data) {
...
-    {
-      MutexLockGuard m(s3fanout_mgr->jobs_todo_lock_);
-      if (!s3fanout_mgr->jobs_todo_.empty() &&
-          (jobs_in_flight < s3fanout_mgr->pool_max_handles_)) {
-        info = s3fanout_mgr->jobs_todo_.back();
-        s3fanout_mgr->jobs_todo_.pop_back();
+  while (true) {
+    // Check events with 100ms timeout
+    int timeout_ms = 100;
+    int retval = poll(s3fanout_mgr->watch_fds_, s3fanout_mgr->watch_fds_inuse_,
```
jblomer: In the client, we saw a performance drop between 100ms timeout and 1ms timeout.  Can you check whether this makes a difference for uploading?

janpriessnitz: I am not sure which performance drop do you mean. For every sent request, a response should arrive sooner than 100 ms after. The `poll` function should therefore block the thread very rarely. On the other hand, a 1ms timeout is very small and borders with ordinary busy waiting. Anyway, the s3benchmark did not show any noticeable difference.

jblomer: Ok, thanks for cross checking.

PR #2389, opened 2019-07-09
Add -g snapshot group option to cvmfs_server add-replica and snapshots -a

DrDaveD:
See [CVM-1779](https://sft.its.cern.ch/jira/browse/CVM-1779).

Note that the original of the test script I modified here [fails](https://sft.its.cern.ch/jira/browse/CVM-1780) in the base cvmfs-2.6 branch.  I tested this by using the modified cvmfs_server with a regular 2.6.0 install, without the branch's cvmfs_swissknife.

### DIFF
--- a/cvmfs/server/cvmfs_server_add_replica.sh
+++ b/cvmfs/server/cvmfs_server_add_replica.sh
@@ -21,10 +21,11 @@ cvmfs_server_add_replica() {
   local configure_apache=1
   local enable_auto_gc=0
   local s3_config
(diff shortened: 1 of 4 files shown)


DrDaveD:
@jblomer I don't yet see this cherry-picked into devel.

jblomer:
Should be there, now.  To get patches from the release branches to the devel branch, I occasionally merge the release branch into devel (which is easier than manually cherry-picking the commits)

PR #2404, opened 2019-08-08
Swissknife filestats feature

janpriessnitz:
New swissknife subcommand 'filestats'. Traverses the catalog tree and exports useful information into sqlite db:

- catalogs table (number of entries, file size)
- objects table (hash, size)
- symlinks table (string length)
- N-to-N files-objects relation (useful for deduplication stats, N-to-N relation useful for chunked files)

### DIFF
--- a/cvmfs/swissknife_filestats.cc
+++ b/cvmfs/swissknife_filestats.cc
@@ -1,0 +1,325 @@
+/**
+ * This file is part of the CernVM File System
+ */
+
+#include "swissknife_filestats.h"
+
+#include <cassert>
+
+#include "hash.h"
+#include "logging.h"
+#include "util/posix.h"
+#include "util/string.h"
+
+using namespace std;  // NOLINT
+
+namespace swissknife {
+
+ParameterList CommandFileStats::GetParams() const {
+  ParameterList r;
+  r.push_back(Parameter::Mandatory(
+              'r', "repository URL (absolute local path or remote URL)"));
+  r.push_back(Parameter::Mandatory('o', "output database file"));
+  r.push_back(Parameter::Optional('n', "fully qualified repository name"));
+  r.push_back(Parameter::Optional('k', "repository master key(s) / dir"));
+  r.push_back(Parameter::Optional('l', "temporary directory"));
+  r.push_back(Parameter::Optional('h', "root hash (other than trunk)"));
+  return r;
+}
+
+int CommandFileStats::Main(const ArgumentList &args) {
+  shash::Any manual_root_hash;
+  const std::string &repo_url  = *args.find('r')->second;
+  db_path_ = *args.find('o')->second;
+  const std::string &repo_name =
+    (args.count('n') > 0) ? *args.find('n')->second : "";
+  std::string repo_keys =
+    (args.count('k') > 0) ? *args.find('k')->second : "";
+  if (DirectoryExists(repo_keys))
+    repo_keys = JoinStrings(FindFilesBySuffix(repo_keys, ".pub"), ":");
+  const std::string &tmp_dir   =
+    (args.count('l') > 0) ? *args.find('l')->second : "/tmp";
+  if (args.count('h') > 0) {
+    manual_root_hash = shash::MkFromHexPtr(shash::HexPtr(
+      *args.find('h')->second), shash::kSuffixCatalog);
+  }
+
+  tmp_db_path_ = tmp_dir + "/cvmfs_filestats/";
+  atomic_init32(&num_downloaded_);
+
+  bool success = false;
+  if (IsHttpUrl(repo_url)) {
+    const bool follow_redirects = false;
+    if (!this->InitDownloadManager(follow_redirects) ||
+        !this->InitVerifyingSignatureManager(repo_keys)) {
+      LogCvmfs(kLogCatalog, kLogStderr, "Failed to init remote connection");
+      return 1;
+    }
+
+    HttpObjectFetcher<catalog::Catalog,
+                      history::SqliteHistory> fetcher(repo_name,
+                                                      repo_url,
+                                                      tmp_dir,
+                                                      download_manager(),
+                                                      signature_manager());
+    success = Run(&fetcher);
+  } else {
+    LocalObjectFetcher<> fetcher(repo_url, tmp_dir);
+    success = Run(&fetcher);
+  }
+
+  return (success) ? 0 : 1;
+}
+
+template <class ObjectFetcherT>
+bool CommandFileStats::Run(ObjectFetcherT *object_fetcher)
+{
+  atomic_init32(&finished_);
+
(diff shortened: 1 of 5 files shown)


cvmfs/swissknife_filestats.h
```diff
@@ -0,0 +1,86 @@
...
+  int Main(const ArgumentList &args);
+
+ protected:
+  string db_path_;
+  string tmp_db_path_;
+  FileStatsDatabase* db_;
+
+  pthread_t thread_processing_;
+  volatile int num_downloaded_;
+  volatile bool running_;
```
jblomer: For multi-threading safety, it would be better to use atomic integers (see atomic.h)

cvmfs/swissknife_filestats.h
```diff
@@ -0,0 +1,86 @@
...
+// Database class for storing repo statistics, the class design
+// follows StatisticsDatabase's design
+class FileStatsDatabase : public sqlite::Database<FileStatsDatabase> {
+ public:
+  bool CreateEmptyDatabase();
+
+  // Following Store* functions return the primary key value of the
+  // inserted entry
+  int StoreCatalog(int num_entries, int file_size);
+  int StoreFile(int catalog_id, int object_id);
```
jblomer: We might have more than 2B objects, so it's safer to use a 64bit integer.

cvmfs/swissknife_filestats.cc
```diff
@@ -0,0 +1,317 @@
...
+  if (DirectoryExists(repo_keys))
+    repo_keys = JoinStrings(FindFilesBySuffix(repo_keys, ".pub"), ":");
+  const std::string &tmp_dir   =
+    (args.count('l') > 0) ? *args.find('l')->second : "/tmp";
+  if (args.count('h') > 0) {
+    manual_root_hash = shash::MkFromHexPtr(shash::HexPtr(
+      *args.find('h')->second), shash::kSuffixCatalog);
+  }
+
+  tmp_db_path_ = "/tmp/cvmfs_filestats/";
```
jblomer: Shouldn't this be the user's selected `tmp_dir`?

cvmfs/swissknife_filestats.cc
```diff
@@ -0,0 +1,317 @@
...
+bool CommandFileStats::Run(ObjectFetcherT *object_fetcher)
+{
+  running_ = true;
+
+  string abs_path = GetAbsolutePath(db_path_);
+  unlink(abs_path.c_str());
+  db_ = FileStatsDatabase::Create(db_path_);
+  db_->InitStatements();
+
+  RemoveTree(tmp_db_path_);
```
jblomer: That's possibly dangerous.  Unless the program crashes, it should not leave temporary files behind anyway.  Can the recursive removal be dropped?

cvmfs/swissknife_filestats.cc
```diff
@@ -0,0 +1,317 @@
...
+void *CommandFileStats::MainProcessing(void *data) {
+  CommandFileStats *repo_stats = static_cast<CommandFileStats *>(data);
+  int processed = 0;
+  while (repo_stats->running_ || processed < repo_stats->num_downloaded_) {
+    if (processed < repo_stats->num_downloaded_) {
+      LogCvmfs(kLogCatalog, kLogStdout, "Processing catalog %d", processed);
+      string db_path = repo_stats->tmp_db_path_ + "/" +
+                       StringifyInt(processed + 1) + ".db";
+      repo_stats->ProcessCatalog(db_path);
+      unlink(db_path.c_str());
```
jblomer: The cvmfs sqlite database base class has the concept of file ownership.  Can this be used to avoid the manual `unlink()`?

cvmfs/swissknife_filestats.cc
```diff
@@ -0,0 +1,317 @@
...
+
+
+
+void CommandFileStats::ProcessCatalog(string db_path) {
+  sqlite::Database<catalog::CatalogDatabase> *cat_db;
+  cat_db = sqlite::Database<catalog::CatalogDatabase>::Open(
+           db_path,
+           sqlite::Database<catalog::CatalogDatabase>::kOpenReadOnly);
+  int64_t file_size = GetFileSize(db_path);
+  sqlite::Sql *catalog_count = new sqlite::Sql(cat_db->sqlite_db(),
```
jblomer: This can be stack allocated

cvmfs/swissknife_filestats.cc
```diff
@@ -0,0 +1,317 @@
...
+                    "ORDER BY md5path_1 ASC, md5path_2 ASC;");
+
+  db_->BeginTransaction();
+
+  while (catalog_list->FetchRow()) {
+    const void *hash = catalog_list->RetrieveBlob(0);
+    int num_bytes = catalog_list->RetrieveBytes(0);
+    int64_t size = catalog_list->RetrieveInt64(1);
+    int flags = catalog_list->RetrieveInt(2);
+    if (flags == 4) {
```
jblomer: Use symbolic flag values

cvmfs/swissknife_filestats.cc
```diff
@@ -0,0 +1,317 @@
...
+    "catalog_id INTEGER PRIMARY KEY,"
+    "num_entries INTEGER,"
+    "file_size INTEGER,"
+    "UNIQUE(catalog_id));").Execute();
+  ret &= sqlite::Sql(sqlite_db(),
+    "CREATE TABLE objects ("
+    "object_id INTEGER PRIMARY KEY,"
+    "hash BLOB,"
+    "size INTEGER,"
+    "UNIQUE(object_id));").Execute();
```
jblomer: a primary key should be automatically unique

cvmfs/swissknife_filestats.cc
jblomer: Flags are combinable, so most likely this should query `(flags & VALUE) == VALUE`

cvmfs/swissknife_filestats.cc
janpriessnitz: The sql statement needs to be destroyed early (before closing the database). I think this is the easiest way.

cvmfs/swissknife_filestats.cc
janpriessnitz: Thanks for the advice. Manually taking file ownership after opening the database deletes the db file upon database object destruction.

cvmfs/swissknife_filestats.cc
janpriessnitz: True, the removal was unnecessary.

cvmfs/swissknife_filestats.cc
janpriessnitz: Yes, that makes more sense. I've changed it so that the `cvmfs_filestats` subfolder gets created in the user-defined temporary folder. The folder itself is used by HTTPFetcher as well.

PR #2408, opened 2019-08-28
Bugfix in dns.cc HostfileResolver

janpriessnitz:
HostfileResolver crashes when parsing lines consisting only of
whitespaces.

Fix this and add a unittest for this case.

### DIFF
--- a/cvmfs/dns.cc
+++ b/cvmfs/dns.cc
@@ -1163,8 +1163,8 @@ void HostfileResolver::ParseHostFile() {
       continue;
     if (ip_end_pos - ip_start_pos > kIpMaxLength) {
       LogCvmfs(kLogDns, kLogSyslogWarn,
-        string("Skipping line in hosts file due to invalid IP address format: "
-        + line).c_str());
+        "Skipping line in hosts file due to invalid IP address format: %s",
+        line.c_str());
       continue;
     }
 
@@ -1189,8 +1189,8 @@ void HostfileResolver::ParseHostFile() {
 
       if (hostname_end_pos - hostname_start_pos > kHostnameMaxLength) {
         LogCvmfs(kLogDns, kLogSyslogWarn,
-          string("Skipping invalid (too long) hostname in hosts file on line: "
-          + line).c_str());
+          "Skipping invalid (too long) hostname in hosts file on line: %s",
+          line.c_str());
         str_offset += hostname_end_pos;
         continue;
       }
--- a/cvmfs/dns.cc
+++ b/cvmfs/dns.cc
@@ -1145,7 +1145,7 @@ void HostfileResolver::ParseHostFile() {
   while (GetLineFile(fhosts_, &line)) {
     char address[kIpMaxLength + 1];
     char hostname[kHostnameMaxLength + 1];
-    int bytes_read = 0;
+    int bytes_read;
     size_t str_offset = 0;
 
     // strip comments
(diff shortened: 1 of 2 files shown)


cvmfs/dns.cc
```diff
@@ -1153,9 +1153,9 @@ void HostfileResolver::ParseHostFile() {
     if (hash_pos != string::npos) line = line.substr(0, hash_pos);
 
     // First token is an IP address
-    int ip_start_pos, ip_end_pos;
-    sscanf(line.c_str(), " %n%*s%n", &ip_start_pos, &ip_end_pos);
-    if (ip_start_pos == ip_end_pos)
+    int ip_start_pos, ip_end_pos, scan_result;
```
jblomer: These three variables should be initialized. The `sscanf` function can fail after parsing only a subset of the variables.

cvmfs/dns.cc
```diff
@@ -1153,9 +1153,9 @@ void HostfileResolver::ParseHostFile() {
...
+    scan_result = sscanf(line.c_str(), " %n%*s%n", &ip_start_pos, &ip_end_pos);
+    if (scan_result == EOF || ip_start_pos == ip_end_pos)
```
jblomer: I think this should test for `scan_result == 2` (instead of only EOF)

cvmfs/dns.cc
```diff
@@ -1153,9 +1153,9 @@ void HostfileResolver::ParseHostFile() {
...
       continue;
     if (ip_end_pos - ip_start_pos > kIpMaxLength) {
       LogCvmfs(kLogDns, kLogSyslogWarn,
```
jblomer: Repeat the faulty input line for easier debugging

cvmfs/dns.cc
```diff
@@ -1168,10 +1168,11 @@ void HostfileResolver::ParseHostFile() {
...
     // Next tokens are hostnames and aliases (we treat them as equal)
     while (str_offset < line.length()) {
+      // check hostname length
       int hostname_start_pos, hostname_end_pos;
-      sscanf(line.c_str() + str_offset, " %n%*s%n",  // check hostname length
-             &hostname_start_pos, &hostname_end_pos);
-      if (hostname_start_pos == hostname_end_pos) {
+      scan_result = sscanf(line.c_str() + str_offset, " %n%*s%n",
+                           &hostname_start_pos, &hostname_end_pos);
+      if (scan_result == EOF || hostname_start_pos == hostname_end_pos) {
```
jblomer: see above

cvmfs/dns.cc
```diff
@@ -1168,10 +1168,11 @@ void HostfileResolver::ParseHostFile() {
 
     // Next tokens are hostnames and aliases (we treat them as equal)
     while (str_offset < line.length()) {
+      // check hostname length
       int hostname_start_pos, hostname_end_pos;
-      sscanf(line.c_str() + str_offset, " %n%*s%n",  // check hostname length
-             &hostname_start_pos, &hostname_end_pos);
-      if (hostname_start_pos == hostname_end_pos) {
+      scan_result = sscanf(line.c_str() + str_offset, " %n%*s%n",
```
jblomer: Check the result of the previous `sscanf` call, too

cvmfs/dns.cc
```diff
@@ -1168,10 +1168,11 @@ void HostfileResolver::ParseHostFile() {
 
     // Next tokens are hostnames and aliases (we treat them as equal)
     while (str_offset < line.length()) {
+      // check hostname length
       int hostname_start_pos, hostname_end_pos;
```
jblomer: initialize

test/unittests/t_dns.cc
```diff
@@ -953,6 +953,15 @@ TEST_F(T_Dns, HostfileResolverMultipleAddresses) {
   EXPECT_EQ(host.ipv6_addresses(), expected_ipv6);
 }
 
+TEST_F(T_Dns, HostfileResolverBlankLines) {
+  CreateHostfile("   \n  #comment\n127.0.0.1 localhost\n");
```
jblomer: Let's add more: an ip address without a host name, a valid line with a comment at the end, an empty line, a too long ip address / host name

cvmfs/dns.cc
janpriessnitz: I can initialize and do further checking to be sure, but the flags in sscanf work in a way that either all or none of the variables are going to be filled.

cvmfs/dns.cc
janpriessnitz: I tested sscanf on my computer and it returns 0 even when all of the variables are filled. It seems that both `%n` and `%*s` do not count as filled variables.

janpriessnitz:
Alright, I added a bunch of asserts and the test cases you requested.

cvmfs/dns.cc
```diff
@@ -1145,44 +1145,61 @@ void HostfileResolver::ParseHostFile() {
...
+    if (scan_result == EOF)
+      continue;
+    assert(ip_start_pos != -1);
+    assert(ip_end_pos != -1);
     if (ip_start_pos == ip_end_pos)
       continue;
     if (ip_end_pos - ip_start_pos > kIpMaxLength) {
       LogCvmfs(kLogDns, kLogSyslogWarn,
-               "Skipping line in hosts file due to invalid IP address format");
+        string("Skipping line in hosts file due to invalid IP address format: "
```
jblomer: It would be more natural to use the form

    "Skipping line in hosts file due to invalid IP address format: %s", line.c_str()

cvmfs/dns.cc
```diff
@@ -1145,44 +1145,61 @@ void HostfileResolver::ParseHostFile() {
...
+        break;
+      assert(hostname_start_pos != -1);
+      assert(hostname_end_pos != -1);
+      if (hostname_start_pos == hostname_end_pos)
+        break;
+
       if (hostname_end_pos - hostname_start_pos > kHostnameMaxLength) {
         LogCvmfs(kLogDns, kLogSyslogWarn,
-                 "Skipping invalid (too long) hostname in hosts file");
+          string("Skipping invalid (too long) hostname in hosts file on line: "
```
jblomer: See above

cvmfs/dns.cc
```diff
@@ -1145,44 +1145,61 @@ void HostfileResolver::ParseHostFile() {
...
+          + line).c_str());
         str_offset += hostname_end_pos;
         continue;
       }
 
-      sscanf(line.c_str() + str_offset, " %s%n", hostname, &bytes_read);
+      bytes_read = -1;
+      scan_result = sscanf(line.c_str() + str_offset, " %s%n", hostname,
+                           &bytes_read);
+      assert(scan_result == 1);
```
jblomer: Can we construct a host file that lets cvmfs crash at this point?

cvmfs/dns.cc
janpriessnitz: Ah, sorry, I totally forgot about this possibility.

cvmfs/dns.cc
janpriessnitz: I think that is currently (intentionally) impossible. The asserts are there to ensure that if I made a mistake and it crashes, we're going to know why. Currently, it should parse every possible sequence of characters without crashing.

janpriessnitz:
@jblomer I added one more test and some logging testing: there is a new static function in T_Dns class compatible with `SetAltLogFunc` which records all logs into a static vector.

test/unittests/t_dns.cc
```diff
@@ -953,6 +978,63 @@ TEST_F(T_Dns, HostfileResolverMultipleAddresses) {
...
+  CreateHostfile("   \n  #comment\n127.0.0.1 localhost\n\n");
+  Host host = hostfile_resolver->Resolve("localhost");
+  EXPECT_EQ(kFailOk, host.status());
+  set<string> expected_ipv4;
+  expected_ipv4.insert("127.0.0.1");
+  EXPECT_EQ(host.ipv4_addresses(), expected_ipv4);
+}
+
+TEST_F(T_Dns, HostfileResolverTooLong) {
+  SetAltLogFunc(TDnsAltLogFunc);
```
jblomer: We just need to set it back to the default at the end of the test in order to not change the environment for following tests.  I'll take care of that.

jblomer:
Replaced by #2410

PR #2409, opened 2019-08-30
Few cvmfs statistics improvements

janpriessnitz:
- `revision` column in `publish_statistics` table of the statistics database: revision number of the publish
- count hardlinks as added files in publish statistics
- change column names in `publish_statistics` table: `duplicated_files` -> `files_duplicated` and `finished_time` -> `finish_time`


cvmfs/statistics_database.cc
```diff
@@ -16,10 +16,11 @@ bool           StatisticsDatabase::compacting_fails        = false;
 namespace {
 
 struct PublishStats {
+  std::string revision;
   std::string files_added;
   std::string files_removed;
   std::string files_changed;
-  std::string duplicated_files;
+  std::string files_duplicated;
   std::string dir_added;
```
jblomer: Just for consistency, let's change these variables to `dirs_added` etc.

janpriessnitz:
Alright, thanks for the advice! I'll also fix the symlinks (CVM-1716) and duplicated chunks (CVM-1720) in this PR, so I do not have to change the layout so frequently.

jblomer:
Let's rename test/src/671-stats_db_upgrade/stats.db to statsdb-rev01.db so that we know what to expect.

jblomer:
Excellent, thanks a lot!

PR #2415, opened 2019-09-16
Refactor cvmfs server to build on a library (step 1)

jblomer:
Add libcvmfs_server library as a basis for server related binaries. This should provide a common building block for the command line utilities as well as for the gateway receiver and possibly other applications.


DrDaveD:
Are you going to wait to merge this until 2.7 was branched off?

jblomer:
We will most likely merge this before but not install libcvmfs_server nor cvmfs_publish.  It is easier to work on the new code base when merged, it already deviated a bit too much.

PR #2416, opened 2019-09-17
Remove raw aborts

siscia:
Draft for switching to raw abort() to the possible use of throw.

The switch is necessary since we want to create a library `libcvmfs_server.so` where a raw aborting is not desiderable. However we would like to keep the original behaviour in the binaries.
The first idea was to have a much simpler function that using pre-compilation condition would either simply `abort()` (for the binaries) or throw an exception string (for the library).
This approach would imply to find good descriptive string for all the abort in the code. Some abort do not have any message associate with (implying that we don't expect them to ever been called), other are preceded by a log (implying that we are facing a known error condition).
Overall it seems complex to find reasonable string messages for all the panics and to format those messages correctly.

The proposed approach will allow to automatically replace either the call to plain `abort(void)` with a call to `Panic(void)` or the call to `LogCvmfs(source, mask, format, ...)` to a call to `Panic(source, mask, format, ...)` leaving untouched the original semantic of the calls.

Downside of this approach is the problem in compilation phase annotate in the comment.
`Panic/3+` should be defined taking as first argument `const LogSource source` while now it is just `const int source` (it is correctly defined in the .cc file). Indeed including `logging.h` break the compilation process, this should be investigate and understood correctly

### DIFF
--- a/cvmfs/util/posix.cc
+++ b/cvmfs/util/posix.cc
@@ -1573,6 +1573,48 @@ bool SafeWriteToFile(const std::string &content,
   return retval;
 }
 
+void Panic() {
+#ifdef LIB_CVMFS_SERVER
+  throw "panic";
+#else
+  abort();
+#endif
+}
+
+// Most likely useless at this stage
+//
+// void Panic(const std::string msg) {
+//   LogCvmfs(kLogCvmfs, kLogDebug, "Panic: %s", msg);
+//
+//   #ifdef LIB_CVMFS_SERVER
+//     throw msg;
+//   #else
+//     abort();
+//   #endif
+// }
+//
+
+void Panic(const LogSource source, const int mask, const char *format, ...) {
+  // Format the message string
+  char *msg_buffer = NULL;
+  va_list variadic_list;
+
+  va_start(variadic_list, format);
+  int retval = vasprintf(&msg_buffer, format, variadic_list);
+  assert(retval != -1);  // else: out of memory
+  va_end(variadic_list);
+
+  const std::string msg(msg_buffer);
+  free(msg_buffer);
+
+  LogCvmfs(source, mask, msg.c_str());
+#ifdef LIB_CVMFS_SERVER
+  throw msg;
+#else
+  abort();
+#endif
+}
+
 #ifdef CVMFS_NAMESPACE_GUARD
 }  // namespace CVMFS_NAMESPACE_GUARD
 #endif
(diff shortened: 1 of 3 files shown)


siscia:
@jblomer This is more a request for comment that a real PR.

cvmfs/util/posix.cc
```diff
@@ -1573,6 +1573,48 @@ bool SafeWriteToFile(const std::string &content,
   return retval;
 }
 
+void Panic() {
```
jblomer: We should rather put the Panic function in a separate file, e.g. util/except

cvmfs/util/posix.cc
```diff
@@ -1573,6 +1573,48 @@ bool SafeWriteToFile(const std::string &content,
...
+#ifdef LIB_CVMFS_SERVER
+  throw "panic";
```
jblomer: We should have an exception base class that derives from `std::runtime_error`.  Perhaps `ECvmfsException`

cvmfs/util/posix.cc
```diff
@@ -1573,6 +1573,48 @@ bool SafeWriteToFile(const std::string &content,
...
+//
+//   #ifdef LIB_CVMFS_SERVER
+//     throw msg;
+//   #else
+//     abort();
+//   #endif
+// }
+//
+
+void Panic(const LogSource source, const int mask, const char *format, ...) {
```
jblomer: It would be useful if we could add the source code file and line where the panic was raised.

cvmfs/util/posix.cc
```diff
@@ -1573,6 +1573,48 @@ bool SafeWriteToFile(const std::string &content,
...
+  va_start(variadic_list, format);
+  int retval = vasprintf(&msg_buffer, format, variadic_list);
+  assert(retval != -1);  // else: out of memory
+  va_end(variadic_list);
+
+  const std::string msg(msg_buffer);
+  free(msg_buffer);
+
+  LogCvmfs(source, mask, msg.c_str());
+#ifdef LIB_CVMFS_SERVER
```
jblomer: I guess we can `#ifdef` the entire function.

cvmfs/util/posix.cc
jblomer: I think for aborting, we don't need the LogSource and mask, and can thus drop the dependency on the logging module.

siscia:
Closed in favor of #2424

PR #2418, opened 2019-09-25
EL8 fixes to build and package system

jblomer:


packaging/rpm/cvmfs-universal.spec
```diff
@@ -160,6 +165,7 @@ Group: Applications/System
 Requires: cvmfs = %{version}
 Requires: fuse3
 Requires: fuse3-libs
+Requires: fuse3
```
DrDaveD: Isn't this a duplicate of the line 2 lines above?

PR #2420, opened 2019-10-01
Ducc ChainID preparatory work. 

siscia:
Preparing work for removing singularity dependency and working with ChainID images.

The main contributions of this PR are two.

First of all, is now possible to compute the ChainID of a docker image.
The `Image` struct gains another method `GetDiffIDs` which returns a list of hashes (the docker DiffIDs). These hashes can be plugged into the new ChainID code to produce the list of ChainIDs on which a docker image is based upon.

Then, we create a new function ApplyDirectory(bottom, top) that simulate an union file system, applying the content of the top directory to the bottom one. The bottom directory is modified.

Whatever feedback is greatly appreciated @jblomer @janpriessnitz


DrDaveD:
I don't have specific comments about the code, but don't we need something like [CVM-1801](https://sft.its.cern.ch/jira/browse/CVM-1801) before storing all the ChainID layers unpacked in order to be reasonably efficient?  I couldn't tell from the code, but is this implementation making symbolic links for all the unchanged files?

siscia:
Works continues on #2515

PR #2423, opened 2019-10-07
Enable cern config repository for debian stretch and newer

janpriessnitz:


mount/default.d/50-cern-debian.conf
```diff
@@ -2,11 +2,18 @@
...
-# Debian 9 (stretch) and newer.  For Ubuntu 16.04 (xenial) and Debian 8
-# (jessie), the CernVM apt repositories contain backported versions of autofs
-# 5.1.2 which are installed by default together with cvmfs.
+# On older platforms, CVMFS_CONFIG_REPOSITORY is turned off because otherwise
+# cvmfs can hang on mount. The config repository can be safely turned on from
+# Debian 9 (stretch) and Ubuntu 18.04 (bionic) onwards. For
+# Ubuntu 16.04 (xenial) and Debian 8 (jessie), the CernVM apt repositories
+# contain backported versions of autofs 5.1.2 which are installed by default
+# together with cvmfs. This default configuration enables the config repository
+# on Debian 10 (buster) / Ubuntu 18.04 (bionic) and newer.
```
DrDaveD: Why enable it only on Debian 10, since the previous comment said it worked on 8 & 9? 

Even though the code says it enables on Ubuntu 18.04 and newer, I do not see any check for that.

/etc/debian_version contains "10.1" on my buster system, and I checked on a debian:8 docker container where it says "8.11".  So grepping for a version by name isn't going to work.

DrDaveD: Ok I see the names in /etc/debian_version on Ubuntu so grepping for names covers Ubuntu but it doesn't actually cover Debian.

janpriessnitz: Hi Dave, thanks for the comment. You're right, the description is misleading, I'll fix it. We do not really test CVMFS for Debian 7 and older and since 8 onwards, the autofs should be fine and the config repo is going to be enabled.
I disabled it for Ubuntu 16.04, because the newer autofs is still installed via the cvmrepo.web.cern.ch on that platform, but that should be fine as well, so I'll enable it (and keep it disabled for 14.04 and older). 
The content of /etc/debian_version for each Ubuntu version can be found here: https://askubuntu.com/questions/445487/what-debian-version-are-the-different-ubuntu-versions-based-on

Do you have any other proposals?

DrDaveD: The updated code still does not work on Debian distributions, which store numbers in /etc/debian_version, not names like on Ubuntu.

DrDaveD: oh, ok, that's fine then.

janpriessnitz:
Cloud-testing job with 050 and 602 tests here: https://epsft-jenkins.cern.ch/view/CVMFS/job/CvmfsCloudTesting/448

PR #2424, opened 2019-10-07
add function to panic and either log + abort or thrown an exception

siscia:
Sup #2416 while keeping the benefits of being easy to integrate in the old code.

Just replace calls to Log + Abort to a call to the PANIC macro.

### DIFF
--- a/cvmfs/swissknife_pull.cc
+++ b/cvmfs/swissknife_pull.cc
@@ -36,6 +36,7 @@
 #include "signature.h"
 #include "smalloc.h"
 #include "upload.h"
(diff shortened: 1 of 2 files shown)


siscia:
I believe it would be wise to understand how we should move this forward. This is signed as a blocker for [CVM-1696](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1696) and I would like to get this in good shape before to work on the blocked issue.

siscia:
I will prepare another PR to apply this changes to the whole codebase.

PR #2425, opened 2019-10-07
cvmfs_config: use `uuid -v4` instead of `uuidgen`

janpriessnitz:
`uuid` tool is a part of 'uuid' package, which is already listed
as a CVMFS dependency. `uuidgen` would require 'uuid-runtime'
package as additional dependency.


cvmfs/cvmfs_config
```diff
@@ -707,7 +707,12 @@ cvmfs_chksetup() {
       done
 
       # Check for network
-      local uuid=$(uuidgen)
+      if which uuidgen 1>/dev/null 2>&1; then
```
jblomer: The `which` utility comes in another package. The check can be done without the dependency on `which` simply by calling `uuidgen -V`.

janpriessnitz: Thanks. Should be fine now.

PR #2427, opened 2019-10-08
'user.repo_metainfo' and 'user.repo_counters' xattrs

janpriessnitz:


cvmfs/catalog_mgr_client.h
```diff
@@ -88,6 +89,8 @@ class ClientCatalogManager : public AbstractCatalogManager<Catalog> {
   std::map<PathString, shash::Any> loaded_catalogs_;
   std::map<PathString, shash::Any> mounted_catalogs_;
 
+  manifest::Manifest *manifest_;
```
jblomer: Let's make it a unique pointer

cvmfs/cvmfs.cc
```diff
@@ -1331,6 +1331,25 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+                                    GetCounters().GetCsvMap();
+  } else if (attr == "user.repo_metainfo") {
+    if (!mount_point_->catalog_mgr()->manifest()) {
+      attribute_value = "Manifest not available";
+    } else {
+      shash::Any hash = mount_point_->catalog_mgr()->manifest()->meta_info();
+      int fd = mount_point_->fetcher()->
+                Fetch(hash, CacheManager::kSizeUnknown,
+                      "metainfo (" + metainfoHash.ToString() + ")",
+                      zlib::kZlibDefault, CacheManager::kTypeVolatile, "");
```
jblomer: I think this should not be a volatile cache entry.  Volatile entries are reserved for file from a volatile repository (e.g., conditions data)

cvmfs/cvmfs.cc
```diff
@@ -1331,6 +1331,25 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+      if (fd < 0) {
+        attribute_value = "Failed to open metadata file";
+      } else {
+        SafeReadToString(fd, &attribute_value);
```
jblomer: The file descriptor is not necessarily a system file descriptor. It is valid only in the context of the cache manager.  That means, we'd need to use cache manager methods to read from it and close it.

cvmfs/cvmfs.cc
```diff
@@ -1331,6 +1331,25 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     attribute_value = string(VERSION) + "." + string(CVMFS_PATCH_LEVEL);
   } else if (attr == "user.pubkeys") {
     attribute_value = mount_point_->signature_mgr()->GetActivePubkeys();
+  } else if (attr == "user.repo_counters") {
+    attribute_value = mount_point_->catalog_mgr()->GetRootCatalog()->
+                                    GetCounters().GetCsvMap();
+  } else if (attr == "user.repo_metainfo") {
```
jblomer: I think it makes sense to move the code of this `if` branch into a helper function

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,31 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
+  }
+  uint64_t actual_size = file_system_->cache_mgr()->GetSize(fd);
+  if (actual_size > max_size) {
+    return "Failed to open: metadata file is too big";
+  }
+  char buffer[actual_size];
+  file_system_->cache_mgr()->Pread(fd, buffer, actual_size, 0);
+  close(fd);
+  return string(buffer, buffer + actual_size);
+}
```
janpriessnitz: @jblomer Is there a better place for the helper function?

test/src/087-xattrs/main
```diff
@@ -0,0 +1,21 @@
+cvmfs_test_name="Test magic extended attributes"
+cvmfs_test_suites="quick"
+
+cvmfs_run_test() {
+  logfile=$1
+
+  cvmfs_mount grid.cern.ch || return 1
+  local attributes_err = $(list_xattrs /cvmfs/grid.cern.ch 2>&1 1>/dev/null)
```
jblomer: Not sure what is being tested here.  Isn't the output always empty by definition?

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,33 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
 }
 
+const uint64_t kMaxMetainfoLength = 64*1024;
+
+string GetRepoMetainfo(uint64_t size) {
+  if (!mount_point_->catalog_mgr()->manifest()) {
+    return "Manifest not available";
+  }
+
+  shash::Any hash = mount_point_->catalog_mgr()->manifest()->meta_info();
```
jblomer: This hash can be the Null hash (`hash.IsNull() == true`).  In this case, the result should be "Metainfo not available"

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,33 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   fuse_reply_statfs(req, &info);
 }
 
+const uint64_t kMaxMetainfoLength = 64*1024;
+
+string GetRepoMetainfo(uint64_t size) {
+  if (!mount_point_->catalog_mgr()->manifest()) {
+    return "Manifest not available";
```
jblomer: String might need to be truncated if size is too small

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,33 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
+
+  shash::Any hash = mount_point_->catalog_mgr()->manifest()->meta_info();
+  // Follow max size suggested by fuse, otherwise 64KiB
+  uint64_t max_size = (size) ? size : kMaxMetainfoLength;
+  int fd = mount_point_->fetcher()->
+            Fetch(hash, CacheManager::kSizeUnknown,
+                  "metainfo (" + hash.ToString() + ")",
+                  zlib::kZlibDefault, CacheManager::kTypeRegular, "");
+  if (fd < 0) {
+    return "Failed to open metadata file";
```
jblomer: String might need to be truncated.

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,33 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
+  }
+  uint64_t actual_size = file_system_->cache_mgr()->GetSize(fd);
+  if (actual_size > max_size) {
+    return "Failed to open: metadata file is too big";
```
jblomer: Might need truncation

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,33 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
+  }
+  char buffer[actual_size];
+  file_system_->cache_mgr()->Pread(fd, buffer, actual_size, 0);
```
jblomer: Check return value to make sure the bytes where actually read

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,33 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   fuse_reply_statfs(req, &info);
 }
 
+const uint64_t kMaxMetainfoLength = 64*1024;
+
+string GetRepoMetainfo(uint64_t size) {
```
jblomer: Size should not be bigger than `kMaxMetainfoLength`

cvmfs/catalog_mgr_client.cc
```diff
@@ -187,6 +187,9 @@ LoadError ClientCatalogManager::LoadCatalog(
     return success_code;
   }
 
+  if (manifest_.IsValid()) delete manifest_;
```
jblomer: That should not be necessary, it should be inherently done by the unique pointer

cvmfs/catalog_mgr_client.h
```diff
@@ -64,6 +64,7 @@ class ClientCatalogManager : public AbstractCatalogManager<Catalog> {
   uint64_t all_inodes() const { return all_inodes_; }
   uint64_t loaded_inodes() const { return loaded_inodes_; }
   std::string repo_name() const { return repo_name_; }
+  manifest::Manifest *manifest() const { return manifest_.weak_ref(); }
```
jblomer: This could simply return a pointer, I guess.

cvmfs/cvmfs.cc
```diff
@@ -1331,6 +1358,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     attribute_value = string(VERSION) + "." + string(CVMFS_PATCH_LEVEL);
   } else if (attr == "user.pubkeys") {
     attribute_value = mount_point_->signature_mgr()->GetActivePubkeys();
+  } else if (attr == "user.repo_counters") {
+    attribute_value = mount_point_->catalog_mgr()->GetRootCatalog()->
```
jblomer: The CSV maps needs to be extracted before `fuse_remounter_->fence()->Leave()`, otherwise there is a race.

cvmfs/cvmfs.cc
```diff
@@ -1331,6 +1358,11 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+                                    GetCounters().GetCsvMap();
+  } else if (attr == "user.repo_metainfo") {
+    attribute_value = GetRepoMetainfo(size);
```
jblomer: It might be more complicated, we should get the hash of the meta info before `fuse_remounter_->fence()->Leave()`, but actually fetch the object only after leaving the fence.

test/src/087-xattrs/main
janpriessnitz: Fuse somehow tests the actual values of the attributes even when just listing them. When I broke the repo_counters attribute, list_xattrs actually outputted an error.

cvmfs/cvmfs.cc
jblomer: Actually, there is not a critical race.

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,41 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
+    return "Failed to open metadata file";
+  }
+  uint64_t actual_size = file_system_->cache_mgr()->GetSize(fd);
+  if (actual_size > max_size) {
+    return "Failed to open: metadata file is too big";
+  }
+  char buffer[actual_size];
+  int bytes_read =
+    file_system_->cache_mgr()->Pread(fd, buffer, actual_size, 0);
+  close(fd);
```
jblomer: This needs to be the cache manager's close function

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,41 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
+                      ? size : kMaxMetainfoLength;
+  int fd = mount_point_->fetcher()->
+            Fetch(hash, CacheManager::kSizeUnknown,
+                  "metainfo (" + hash.ToString() + ")",
+                  zlib::kZlibDefault, CacheManager::kTypeRegular, "");
+  if (fd < 0) {
+    return "Failed to open metadata file";
+  }
+  uint64_t actual_size = file_system_->cache_mgr()->GetSize(fd);
+  if (actual_size > max_size) {
```
jblomer: Needs to close in the if branch

cvmfs/cvmfs.cc
```diff
@@ -1270,6 +1270,41 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   fuse_reply_statfs(req, &info);
 }
 
+const uint64_t kMaxMetainfoLength = 64*1024;
+
+string GetRepoMetainfo(uint64_t size) {
+  if (!mount_point_->catalog_mgr()->manifest()) {
+    return "Manifest not available";
```
jblomer: This string and the other error strings need to be truncated if `size` is smaller than their length

PR #2435, opened 2019-10-18
possible approach for CVM-1735 (wait and retry)

siscia:
This is the approach of trying to make a tag, and if it fails, wait one second and then re-try.


siscia:
Approach of adding milliseconds on #2436

jblomer:
Replaced by #2436

PR #2436, opened 2019-10-18
CVM-1735: add milliseconds accurracy to tag from gateways

siscia:
This PR adds milliseconds accuracy to the generic tag generated by gateways.

It actually sidestep the possible problem in [CVM-1735](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1735)

### DIFF
--- a/cvmfs/receiver/commit_processor.cc
+++ b/cvmfs/receiver/commit_processor.cc
@@ -4,6 +4,8 @@
 
 #include "commit_processor.h"
 
+#include <time.h>
+
 #include <vector>
 
 #include "catalog_diff_tool.h"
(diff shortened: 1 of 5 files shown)


cvmfs/receiver/commit_processor.cc
```diff
@@ -100,13 +102,24 @@ CommitProcessor::Result CommitProcessor::Process(
...
-    char buf[32];
-    time_t now = time(NULL);
+    // format time following the ISO 8601 YYYY-MM-DDThh:mm:ss.sssZ
+    // note the millisecond accurracy
+    struct timespec now;
     struct tm timestamp;
-    gmtime_r(&now, &timestamp);
-    strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%SZ", &timestamp);
-    final_tag.name_ = std::string("generic-") + buf;
+    clock_gettime(CLOCK_REALTIME, &now);
```
jblomer: We should make this a posix.cc utility function; we also need a `platform_...` function because `get_clock` is Linux only

jblomer:
I like this one, let's go for it!

jblomer:
Can we close this in favor of #2437 ?

siscia:
Sure, I open both to decide if we like the idea of refactoring the linux time.

PR #2437, opened 2019-10-21
Use millisecond precision for generic tags in gateway

siscia:
Not sure if this refactor is worth to do.

Feel free to merge or discard

### DIFF
--- a/cvmfs/platform_linux.h
+++ b/cvmfs/platform_linux.h
@@ -336,6 +336,13 @@ inline uint64_t platform_monotonic_time_ns() {
   return static_cast<uint64_t>(tp.tv_sec*1e9 + tp.tv_nsec);
 }
 
+inline uint64_t platform_time() {
+  struct timespec tp;
+  int retval = clock_gettime(CLOCK_REALTIME, &tp);
+  assert(retval == 0);
+  return static_cast<uint64_t>(tp.tv_sec*1e9 + tp.tv_nsec);
+}
+
 inline uint64_t platform_memsize() {
   return static_cast<uint64_t>(sysconf(_SC_PHYS_PAGES)) *
          static_cast<uint64_t>(sysconf(_SC_PAGE_SIZE));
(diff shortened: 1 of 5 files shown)


siscia:
To easy review, only the last commit is different from #2436

cvmfs/platform_linux.h
```diff
@@ -318,22 +318,30 @@ inline std::string platform_getexepath() {
...
 inline uint64_t platform_monotonic_time_ns() {
-  struct timespec tp;
-  int retval = clock_gettime(CLOCK_MONOTONIC, &tp);
-  assert(retval == 0);
-  return static_cast<uint64_t>(tp.tv_sec*1e9 + tp.tv_nsec);
+  struct timespec tp = _time_with_clock(CLOCK_MONOTONIC);
+  return static_cast<uint64_t>(tp.tv_sec * 1e9 + tp.tv_nsec);
+}
+
+inline uint64_t platform_time() {
```
jblomer: We should add the unit and the time source to the function name.  I'd suggest `platform_realtime_ns`

cvmfs/platform_osx.h
```diff
@@ -265,6 +265,8 @@ inline double platform_monotonic_time_ns() {
   return val_ns;
 }
 
+inline uint64_t platform_time() { return platform_monotonic_time_ns(); }
```
jblomer: That would be ticks since boot time, not realtime.  We can simply use C's `gettimeofday()`.

cvmfs/util/posix.cc
```diff
@@ -1630,6 +1630,14 @@ bool SafeWriteToFile(const std::string &content,
   return retval;
 }
 
+/**
+ * Returns current time in nanoseconds
+ * It is not guarantee to be monotomic increasing
+ */
+uint64_t GetTime() {
```
jblomer: --> `GetRealTimeNs` -- or, the user code could use directly the `platform_...` functions.

cvmfs/platform_osx.h
siscia: `gettimeofday()` is obsolete in Posix 2008. So I was guessing that we should not use it...

What do you consult for OSX documentation?

FaizSyed430:
clints

siscia:
???

PR #2441, opened 2019-11-04
Use relative symbolic links in DUCC

siscia:
Address [CVM-1817](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1817)

### DIFF
--- a/ducc/lib/cvmfs.go
+++ b/ducc/lib/cvmfs.go
@@ -119,6 +119,16 @@ func CreateSymlinkIntoCVMFS(CVMFSRepo, newLinkName, toLinkPath string) (err erro
 		return err
 	}
 
+	relativePath, err := filepath.Rel(newLinkName, toLinkPath)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


jblomer:
Thanks!  We should be prepared to migrate the existing repository, let's think about whether there is an easy way to do it with the next software upgrade.

siscia:
Sure! I believe the simplest thing would be to use a simple shell script against the main repo.

siscia:
Yep, something like the following should do the trick:

```
#!/usr/bin/python3

import pathlib, os

p = pathlib.Path('/cvmfs/foo.ch')

for x in p.glob('*/*/*'):
    if x.is_symlink():
        link_to = os.readlink(str(x))
        link = os.path.relpath(link_to, x)
        link = os.sep.join(link.split(os.sep)[1:])
        os.remove(x)
        os.symlink(link, x, target_is_directory=True)
```

PR #2442, opened 2019-11-04
Delete empty "auto" from proxy list

DrDaveD:
See [CVM-1818](https://sft.its.cern.ch/jira/browse/CVM-1818)

### DIFF
--- a/cvmfs/wpad.cc
+++ b/cvmfs/wpad.cc
@@ -211,23 +211,23 @@ string ResolveProxyDescription(
   if ((cvmfs_proxies == "") || (cvmfs_proxies.find("auto") == string::npos))
     return cvmfs_proxies;
 
-  bool use_cache = false;
+  int empty_auto = -1;
   vector<string> lb_groups = SplitString(cvmfs_proxies, ';');
   for (unsigned i = 0; i < lb_groups.size(); ++i) {
     if (lb_groups[i] != "auto")
       continue;
 
     lb_groups[i] = AutoProxy(download_manager);
-    if (lb_groups[i].empty()) {
-      use_cache = true;
-      lb_groups.erase(lb_groups.begin()+i);
-    }
+    if (lb_groups[i].empty())
+      empty_auto = static_cast<int>(i);
   }
 
+  if (empty_auto != -1)
+    lb_groups.erase(lb_groups.begin()+static_cast<unsigned>(empty_auto));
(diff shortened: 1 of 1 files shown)


jblomer:
Many thanks for the fix!  We have a rather stable 2.7 release now, which we'd like to push out as soon as possible.  As this is marked low priority, I suggest to defer it to the 2.7.1 release.

DrDaveD:
I suppose.

cvmfs/wpad.cc
```diff
@@ -218,8 +218,10 @@ string ResolveProxyDescription(
       continue;
 
     lb_groups[i] = AutoProxy(download_manager);
-    if (lb_groups[i].empty())
+    if (lb_groups[i].empty()) {
       use_cache = true;
+      lb_groups.erase(lb_groups.begin()+i);
```
jblomer: Wouldn't that skip one entry?  I.e., if the string is 'auto;auto', will both auto keywords be processed?

jblomer: (We anyway have to merge #2449, so we can merge this one, too)

DrDaveD: I didn't know, so I tried "auto;auto", and according to the debug log, for the second time it tries to lookup "auto" as a hostname.  That is a different behavior from cvmfs-2.6.3, which processes the auto twice.  But does it really make a difference, because what would be the purpose of using "auto" twice?

I also tried just "auto" with no DIRECT for this case with both 2.6.3 and they both behaved the same way when "auto" returned nothing: they exited after the auto failed.

jblomer: I agree, `auto;auto` doesn't make sense, but such slight misconfiguration might still be present.  Cvmfs claims that it replaces `auto` by the list of retrieved proxies (or removes the entry altogether with your fix applied).  I think we should stick to that algorithm.

DrDaveD: Oh, and I see now that the erase statement is inside of a loop where the lb_groups are accessing by index, and erasing one of them messes up the index.  So that's not good.  I'm trying to change it to keep track of an auto that returns empty (any one) and erase it after the loop.

DrDaveD: Done.  It now also processes the "auto" in "auto;auto" twice.  If the misconfiguration is "auto;auto;DIRECT" it still gets the extra message in CVM-1818, but I'm not worried about that case.

PR #2444, opened 2019-11-08
MockHTTPServer & new tests for DownloadManager

janpriessnitz:
Implement a simple HTTP server class. Then use this class to implement simple mock HTTP servers (proxy, redirect, file serving) in t_download.cc. These will be further used for writing additional tests for DownloadManager.

WIP: rewrite t_download.cc using FileSandbox and write more tests.

### DIFF
--- a/test/unittests/c_http_server.cc
+++ b/test/unittests/c_http_server.cc
@@ -1,0 +1,248 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include "c_http_server.h"
+
+HTTPRequestParser::HTTPRequestParser() {
+  state_ = kBegin;
+  body_bytes_read_ = 0;
+}
+
+bool HTTPRequestParser::Parse(char c) {
+  switch (state_) {
+    case kBegin:
+      buffer_ += c;
+      state_ = kMethod;
+      break;
+    case kMethod:
+      if (c == ' ') {
+        request_.method = buffer_;
+        buffer_.clear();
+        state_ = kPath;
+      } else {
+        buffer_ += c;
+      }
+      break;
+    case kPath:
+      if (c == ' ') {
+        request_.path = buffer_;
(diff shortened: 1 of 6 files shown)


test/unittests/c_http_server.h
```diff
@@ -0,0 +1,212 @@
...
+
+#include <cassert>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "atomic.h"
+#include "util/string.h"
+
+typedef std::vector< std::pair<std::string, std::string> > HeaderList;
```
jblomer: As this is global, let's call it `HTTPHeaderList`

test/unittests/c_http_server.h
```diff
@@ -0,0 +1,212 @@
...
+  HeaderList headers;
+  std::string body;
+
+  // If true, the whole request is stored in body and other variables
+  // will not be used in the reply
+  bool raw;
+};
+
+// Class for parsing HTTP request. Create one instance per one request.
+// Works in a state-machine-like manner. Function Parse() parses characters
```
jblomer: Very nice!

test/unittests/c_http_server.cc
```diff
@@ -0,0 +1,401 @@
...
+  MockHTTPServer *server = static_cast<MockHTTPServer *>(data);
+  const int kReadBufferSize = 1000;
+  int listen_sockfd = -1, accept_sockfd = -1;
+  socklen_t clilen;
+  struct sockaddr_in serv_addr, cli_addr;
+  char buffer[kReadBufferSize];
+  int retval = 0;
+
+  // Listen incoming connections
+  listen_sockfd = socket(AF_INET, SOCK_STREAM, 0);
```
jblomer: We should use `MakeTcpEndpoint` from util/posix.cc

PR #2448, opened 2019-11-19
DUCC: add convert-singularity-image command

janpriessnitz:
Also separate ConvertWish into ConvertWishSingularity and ConvertWishDocker

### DIFF
--- a/test/src/401-ducc-convert_singularity/main
+++ b/test/src/401-ducc-convert_singularity/main
@@ -1,0 +1,20 @@
+cvmfs_test_name="DUCC singularity image conversion"
+cvmfs_test_autofs_on_startup=true
+cvmfs_test_suites="ducc"
+
+DOCKER_TEST_IMAGE="https://registry.hub.docker.com/library/centos:centos7"
+
+
+cvmfs_run_test() {
+  echo "*** creating empty repo ***"
+  create_empty_repo $CVMFS_TEST_REPO $CVMFS_TEST_USER || return 1
+
+  echo "*** converting image to singularity ***"
+  cvmfs_ducc convert-singularity-image -r $CVMFS_TEST_REPO $DOCKER_TEST_IMAGE || return 2
+
+  cvmfs_mount $CVMFS_TEST_REPO || return 3
+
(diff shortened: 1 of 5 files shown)


test/src/401-ducc-convert_singularity/main
```diff
@@ -0,0 +1,16 @@
...
+
+cvmfs_run_test() {
+  echo "*** creating empty repo ***"
+  create_empty_repo $CVMFS_TEST_REPO $CVMFS_TEST_USER || return 1
+
+  echo "*** converting image to singularity ***"
+  cvmfs_ducc convert-singularity-image --repo $TEST_REPO $DOCKER_TEST_IMAGE || return 2
+
+  return 0
+}
```
jblomer: A trailing newline is missing

test/src/401-ducc-convert_singularity/main
```diff
@@ -0,0 +1,16 @@
...
+DOCKER_TEST_IMAGE="https://registry.hub.docker.com/library/centos:centos7"
+
+
+cvmfs_run_test() {
+  echo "*** creating empty repo ***"
+  create_empty_repo $CVMFS_TEST_REPO $CVMFS_TEST_USER || return 1
+
+  echo "*** converting image to singularity ***"
+  cvmfs_ducc convert-singularity-image --repo $TEST_REPO $DOCKER_TEST_IMAGE || return 2
+
```
jblomer: There should be some test code here to check that the image contents actually appeared in the repository

DrDaveD:
It's very unusual for a command line client flag to accept `true` and `false`.  It's more common to have separate flags to turn off, for example `--no-convert-docker` or `--dont-convert-docker`.  The `cvmfs_ducc convert` help message also doesn't at all make it clear that one is supposed to pass `--convert-docker=false` to turn it off.  For that matter, it also doesn't say anywhere how to pass a wish file.

siscia:
This is actually a very good point, from an usability point of view I agree that the use of negative flags should be prefered. @janpriessnitz I would take care of it!

PR #2449, opened 2019-11-21
add firmlink to support mounting /cvmfs in Catalina osx 10.15

siscia:
I also removed the `sudo` that should not be necessary since we are running as sudo anyway.


cvmfs/cvmfs_config
```diff
@@ -1833,15 +1833,23 @@ EOF
...
+     # firmlinks from `/cvmfs` to `/Users/shared/cvmfs`
      if sw_vers | grep 10.15; then
-        sudo mkdir -p /Users/shared/cvmfs
-        sudo chown cvmfs:cvmfs /Users/shared/cvmfs
-        sudo chmod 777 /Users/shared/cvmfs
+        mkdir -p /Users/shared/cvmfs
+        chown cvmfs:cvmfs /Users/shared/cvmfs
+        chmod 777 /Users/shared/cvmfs
+        if ! grep -q "cvmfs" /etc/synthetic.conf 2> /dev/null ; then
+          echo -e "# Autogenerated by CernVM-FS, do not delete nor edit" >> /etc/synthetic.conf
```
jblomer: I think we cannot do that.  There might be other parties interested in editing the _file_. We can only set the "auto-generated" marker for our _line_, if that is possible (i.e., if lines can have trailing comments).  Or, we have to wrap the `cvmfs` line between begin and end comment lines.

cvmfs/cvmfs_config
```diff
@@ -1833,15 +1833,23 @@ EOF
...
+     # In order to keep the standard CVMFS mountpoint, we decide to implement
+     # firmlinks from `/cvmfs` to `/Users/shared/cvmfs`
      if sw_vers | grep 10.15; then
-        sudo mkdir -p /Users/shared/cvmfs
-        sudo chown cvmfs:cvmfs /Users/shared/cvmfs
-        sudo chmod 777 /Users/shared/cvmfs
+        mkdir -p /Users/shared/cvmfs
+        chown cvmfs:cvmfs /Users/shared/cvmfs
+        chmod 777 /Users/shared/cvmfs
+        if ! grep -q "cvmfs" /etc/synthetic.conf 2> /dev/null ; then
```
jblomer: Perhaps better `grep -q "^cvmfs\t"` to be more precise

cvmfs/cvmfs_config
siscia: Aw I meant that the line below was autogenerated, not the whole file.

Something like this:

```
# The line below was autogenerated by CernVM-FS, do not delete nor edit it
cvmfs /Users/Shared/cvmfs
```

cvmfs/cvmfs_config
siscia: I intentionally leave it a little unprecise to let user play with it.

With the assumption that they may want to use something different, but that such something will contains `cvmfs` in the name.

I guess is not a big deal to have one more firmlink anyway!

jblomer: It's fine to let the user add more symlinks.  But in this way, if there is some "cvmfs stuff" present in the file, we won't add our version of the symlink.  And I think we should.

cvmfs/cvmfs_config
siscia: The manual of `syntetich.conf` indicate that only the line that starts with `#` are comment, so we should not use trailing comments.

cvmfs/cvmfs_config
siscia: Fixed!

cvmfs/cvmfs_config
```diff
@@ -1833,15 +1833,23 @@ EOF
      # SIP) create directory in root `/`.
      # We decide to move the standard installation directory on
      # `/Users/shared/cvmfs` instead of `/cvmfs`
+     # In order to keep the standard CVMFS mountpoint, we decide to implement
+     # firmlinks from `/cvmfs` to `/Users/shared/cvmfs`
      if sw_vers | grep 10.15; then
```
jblomer: Ah, sorry, I forgot: this should check for >= 10.15, otherwise it will break again with 10.16...

siscia: Addressed in #2450

PR #2451, opened 2019-11-22
Several fixes for integration tests

jblomer:


test/src/054-geoapi/main
```diff
@@ -35,11 +36,11 @@ cvmfs_run_test() {
     *.fnal.gov|*.unl.edu) country=us;;
   esac
 
-  if [ "$country" = "ch" ]; then
+  if [ "x$country" = "xch" ]; then
```
DrDaveD: Why this change?  It has no effect.  The only reason for putting an "x" in such places is for those scripts that do not quote variables, because there's a syntax error if one side of a comparison is completely empty.  A quoted variable is not completely empty, it has two quotes.

jblomer: Ok, I wasn't sure.  It doesn't hurt and just matches the pattern of many of our other scripts.

DrDaveD: Ok.  Unfortunately the pattern is a bad one though, in my opinion.  One should stick to either using the "x" prefix or using quotes, but don't use both.  I prefer quotes.

PR #2453, opened 2019-11-25
 External POSIX cache plugin

janpriessnitz:
Add executable 'cvmfs_cache_posix', which acts as an external
CVMFS cache plugin and uses CVMFS PosixCacheManager.

Currently, it does NOT use QuotaPosixManager, so it does not
support object listing and cache shrinking. User can still clean
cache manually.

### DIFF
--- a/ci/run_unittests.sh
+++ b/ci/run_unittests.sh
@@ -84,6 +84,7 @@ if [ "x$CVMFS_CACHE_PLUGIN" != "x" ]; then
   CVMFS_CACHE_CONFIG=$(mktemp /tmp/cvmfs-unittests-XXXXX)
   echo "CVMFS_CACHE_PLUGIN_SIZE=1000" > $CVMFS_CACHE_CONFIG
   echo "CVMFS_CACHE_PLUGIN_TEST=yes" >> $CVMFS_CACHE_CONFIG
+  echo "CVMFS_CACHE_PLUGIN_PATH=/tmp/cvmfs_cache_test_dir/" >> $CVMFS_CACHE_CONFIG
   i=0
   for plugin in $(echo $CVMFS_CACHE_PLUGIN | tr : " "); do
     if [ -x $plugin ]; then
@@ -99,6 +100,7 @@ if [ "x$CVMFS_CACHE_PLUGIN" != "x" ]; then
       $CVMFS_CACHE_UNITTESTS $CVMFS_CACHE_LOCATOR \
         --gtest_output=xml:${CVMFS_UNITTESTS_RESULT_LOCATION}.$(basename $plugin)
       /bin/kill $plugin_pid
+      rm -rf "/tmp/cvmfs_cache_test_dir/"
     else
       echo "Warning: plugin $plugin not found, skipping"
     fi
--- a/cvmfs/cache_plugin/cvmfs_cache_posix.cc
+++ b/cvmfs/cache_plugin/cvmfs_cache_posix.cc
@@ -1,0 +1,498 @@
+/**
+ * This file is part of the CernVM File System.
+ *
+**/
+
+#include <signal.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#include <climits>
+#include <cstring>
+#include <string>
+
+#include "atomic.h"
+#include "cache_plugin/libcvmfs_cache.h"
+#include "cache_posix.h"
+#include "logging.h"
+#include "smallhash.h"
+#include "util/posix.h"
+#include "util/string.h"
+
+bool operator==(const cvmcache_hash &a, const cvmcache_hash &b) {
+  return memcmp(a.digest, b.digest, 20) == 0;
+}
+
+bool operator!=(const cvmcache_hash &a, const cvmcache_hash &b) {
+  return memcmp(a.digest, b.digest, 20) != 0;
+}
+
+namespace {
+
+struct CacheObject {
+  uint32_t refcnt;
+  int fd;
+  uint64_t size;
+};
+
+struct Txn {
+  struct cvmcache_hash hash;
+  uint64_t size;
+  cvmcache_object_type type;
+  void *txn;
+  int fd;
+};
+
+struct Listing {
+  cvmcache_object_type type;
+  std::vector<std::string> list;
+  std::vector<std::string>::iterator it;
+};
+
+struct Settings {
+  Settings() :
+    is_alien(false), cache_base_defined(false),
+    cache_dir_defined(false), quota_limit(0)
+    { }
+
+  bool IsValid() {
+    if (is_alien && quota_limit > 0) {
+      error_reason = "Alien cache cannot be managed (no quota manager allowed)";
+      return false;
+    }
+    if (is_alien && workspace == "") {
+      error_reason = "Workspace option needs to be set for alien cache";
+      return false;
+    }
+    if ((is_alien ? 1 : 0) +
+        (cache_base_defined ? 1 : 0) +
+        (cache_dir_defined ? 1 : 0) != 1) {
+      error_reason = "CVMFS_CACHE_DIR, CVMFS_CACHE_BASE and CVMFS_CACHE_ALIEN "
+                     "are mutually exclusive. Exactly one needs to be defined.";
+      return false;
+    }
+    return true;
+  }
+
+  bool is_alien;
+  bool cache_base_defined;
+  bool cache_dir_defined;
+  /**
+   * Soft limit in bytes for the cache.  The quota manager removes half the
+   * cache when the limit is exceeded.
+   */
+  int64_t quota_limit;
+  std::string cache_path;
+  /**
+   * Different from cache_path only if CVMFS_WORKSPACE or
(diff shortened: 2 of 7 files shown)


ci/run_unittests.sh
```diff
@@ -84,6 +84,7 @@ if [ "x$CVMFS_CACHE_PLUGIN" != "x" ]; then
   CVMFS_CACHE_CONFIG=$(mktemp /tmp/cvmfs-unittests-XXXXX)
   echo "CVMFS_CACHE_PLUGIN_SIZE=1000" > $CVMFS_CACHE_CONFIG
   echo "CVMFS_CACHE_PLUGIN_TEST=yes" >> $CVMFS_CACHE_CONFIG
+  echo "CVMFS_CACHE_PLUGIN_PATH=/tmp/cvmfs_cache_test_dir/" >> $CVMFS_CACHE_CONFIG
```
jblomer: Since the POSIX cache manager has more options than just the base part, I'd suggest to reuse the mechanism from mountpoint.cc to setup the cache manager and quota manager from a set of options represented by an options manager.

cvmfs/manifest.cc
```diff
@@ -242,6 +246,11 @@ Breadcrumb Manifest::ReadBreadcrumb(
   Breadcrumb breadcrumb;
   const string breadcrumb_path = directory + "/cvmfschecksum." + repo_name;
   FILE *fbreadcrumb = fopen(breadcrumb_path.c_str(), "r");
+  if (!fbreadcrumb) {
+    // Return invalid breadcrumb if not found
+    breadcrumb.timestamp = 0;
```
jblomer: This is taken care of by the default constructor of breadcrumb

cvmfs/manifest.cc
```diff
@@ -242,6 +246,11 @@ Breadcrumb Manifest::ReadBreadcrumb(
...
+    return breadcrumb;
+  }
   char tmp[128];
   int read_bytes;
   if (fbreadcrumb && (read_bytes = fread(tmp, 1, 128, fbreadcrumb)) > 0) {
```
jblomer: Now we know that fbreadcrumb is not NULL (same for line 250)

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+
+  ctx = cvmcache_init(&callbacks);
+  int retval = cvmcache_listen(ctx, locator);
+  if (!retval) {
+    LogCvmfs(kLogCache, kLogStderr | kLogSyslogErr,
+             "failed to listen on %s", locator);
+    return 1;
+  }
+
+  if (test_mode) {
```
jblomer: What exactly would be the test mode?  Foreground?

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+    LogCvmfs(kLogCache, kLogStdout,
+             "Running unsupervised. Quit by SIGINT (CTRL+C)");
+    terminated = false;
+    signal(SIGINT, handle_sigint);
+    while (!terminated) sleep(1);
+  }
+
+  cvmcache_wait_for(ctx);
+
+  delete opened_objects;
```
jblomer: Should be set to NULL, too.

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+
+static struct cvmcache_hash Cpphash2Chash(const shash::Any &hash) {
+  struct cvmcache_hash h;
+  memset(h.digest, 0, 20); // ensure deterministic digest
+  memcpy(h.digest, hash.digest, sizeof(h.digest));
+  h.algorithm = hash.algorithm;
+  return h;
+}
+
+SmallHashDynamic<struct cvmcache_hash, CacheObject> *opened_objects;
```
jblomer: Global object names should start with `g_` to make them instantly recognizable

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+SmallHashDynamic<uint64_t, Txn> *transactions;
+SmallHashDynamic<uint64_t, Listing> *listings;
+PosixCacheManager *cache_mgr;
+cvmcache_context *ctx;
+volatile bool terminated;
```
jblomer: This should probably better be an atomic int

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+    object.size = cache_mgr->GetSize(fd);
+    object.refcnt = 0;
+    pinned_size += object.size;
+  } else if (object.refcnt + change_by < 0) {
+    return CVMCACHE_STATUS_BADCOUNT;
+  }
+
+  object.refcnt += change_by;
+  if (object.refcnt == 0) {
+    if (cache_mgr->Close(object.fd)) {
```
jblomer: Better: `if (cache_mgr->Close(object.fd) != 0)`

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+    return CVMCACHE_STATUS_NOENTRY;
+  }
+  if (offset > object.size) {
+    return CVMCACHE_STATUS_OUTOFBOUNDS;
+  }
+  int64_t bytes_read = cache_mgr->Pread(object.fd, buffer, *size, offset);
+  if (bytes_read < 0) {
+    return CVMCACHE_STATUS_IOERR;
+  }
+  *size = (uint32_t) bytes_read;
```
jblomer: --> static_cast

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+}
+
+static int posix_write_txn(uint64_t txn_id, unsigned char *buffer,
+                           uint32_t size) {
+  Txn transaction;
+  if (!transactions->Lookup(txn_id, &transaction)) {
+    return CVMCACHE_STATUS_NOENTRY;
+  }
+  int64_t bytes_written = cache_mgr->Write(buffer, size, transaction.txn);
+  if (bytes_written < 0) {
```
jblomer: Shouldn't it be `(bytes_written >= 0) && (static_cast<uint32_t>(bytes_written) == size)`?

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+  }
+  CacheObject object;
+  if (!opened_objects->Lookup(transaction.hash, &object)) {
+    int result = cache_mgr->CommitTxn(transaction.txn);
+    if (result) {
+      return CVMCACHE_STATUS_IOERR;
+    }
+    CacheManager::BlessedObject
+      blessed_object(Chash2Cpphash(&transaction.hash));
+    object.fd = cache_mgr->Open(blessed_object);
```
jblomer: Here you would use `OpenFromTxn` before committing in order to avoid a race

cvmfs/cache_plugin/cvmfs_cache_posix.cc
```diff
@@ -0,0 +1,415 @@
...
+    if (object.fd < 0) {
+      transactions->Erase(txn_id);
+      return CVMCACHE_STATUS_IOERR;
+    }
+    object.refcnt = 0;
+    object.size = cache_mgr->GetSize(object.fd);
+    pinned_size += object.size;
+    used_size += object.size;
+  } else {
+    if (cache_mgr->AbortTxn(transaction.txn)) {
```
jblomer: ... test for != 0

cvmfs/cache_plugin/cvmfs_cache_posix.cc
janpriessnitz: I checked functions PosixCacheManager::Write and PosixCacheManager::Flush and the Write function should either return negative value or `size`. Am I missing something? Should I do it as the comment says anyway?

cvmfs/cache_plugin/cvmfs_cache_posix.cc
janpriessnitz: It ensures that the plugin cooperates with https://github.com/cvmfs/cvmfs/blob/devel/ci/run_unittests.sh . I copied this part of code from https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/cache_plugin/cvmfs_cache_ram.cc .

jblomer: Oh, right.

cvmfs/cache_plugin/cvmfs_cache_posix.cc
jblomer: Ok.  Let's still be defensive and explicitly check for size.

PR #2455, opened 2019-11-25
WIP: [CVM-1696] Addressing trailing slash problem

siscia:
The first part of this PR address the trailing slash problem when using the gateway.

The first step was to actually make the test fails if the ingestion of the gateway didn't work out.
The we added a test for checking that the ingestion work also with a trailing slash: [setup](https://github.com/cvmfs/cvmfs/compare/CVM-1696?expand=1#diff-a4ea05080469eebcea0ecbeaf73c7d95R187-R200) and [check](https://github.com/cvmfs/cvmfs/compare/CVM-1696?expand=1#diff-a4ea05080469eebcea0ecbeaf73c7d95R223-R224)

Finally we correct how we check [if a lease is in path](https://github.com/cvmfs/cvmfs/commit/ba4bd30155dead1fa9040e00a162b1666f5df184).

This is still a work in progress, but I am quite unsure on commit https://github.com/cvmfs/cvmfs/commit/d5adde80eb99dd58779123b54a7ca73c172d7b6b which does nothing but heavily annotate the function DiffRec, very useful during debugging, but maybe too heavy on the annotation.

The work is also being documented in this [gist](https://gist.github.com/siscia/df86c59b00de6905ea4606af83e6d1b5)

### DIFF
--- a/cvmfs/catalog_diff_tool_impl.h
+++ b/cvmfs/catalog_diff_tool_impl.h
@@ -94,25 +94,25 @@ RoCatalogMgr* CatalogDiffTool<RoCatalogMgr>::OpenCatalogManager(
 
 template <typename RoCatalogMgr>
 void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
-  LogCvmfs(kLogCvmfs, kLogSyslog,
-                "CatalogDiffTool - DiffRec on %s",
-                path.ToString().c_str());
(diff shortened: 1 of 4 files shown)


jblomer:
In principle looks fine to me!  As discussed, let's keep the comments more concise and let's block leases on non-existing paths.

siscia:
I want to add another check in such a way that this error is managed higher.

I simply want to check if the parent directory of which we take a lease exists, I am thinking of doing it in the bash code.

siscia:
Sup by #2480

PR #2457, opened 2019-12-03
cvmfs_cache_posix: packaging, test & minor bugfix

janpriessnitz:


test/src/089-external_cache_plugin/main
```diff
@@ -0,0 +1,65 @@
...
+  fi
+}
+
+cvmfs_run_test() {
+  logfile=$1
+
+  trap cleanup EXIT HUP INT TERM
+
+  mount_opts="CVMFS_CACHE_PRIMARY=externalplugin"
+  mount_opts="$mount_opts CVMFS_CACHE_externalplugin_TYPE=external"
```
jblomer: Isn't there a linebreak missing?

jblomer:
The integration tests should be excluded from the ram cache "platform"

test/src/089-external_cache_plugin/main
janpriessnitz: it's delimited by space. These opts are later passed to cvmfs_mount()

jblomer: Ah, thanks

PR #2461, opened 2019-12-11
Add note about rebooting the system after installation in Catalina

siscia:
I add a note, twice in the welcome and in the conclusion about the necessity of rebooting the system.

The note is:

```
Please note that from MacOS 10.15 Catalina is necessary to reboot your system after the installation in order to create the /cvmfs directory.
```


jblomer:
Very good!  Let's reopen against the 2.7 branch so that it gets into a future 2.7.1 patch release

PR #2463, opened 2019-12-11
iterate over PANIC design and ergonomics

siscia:
Follow up from #2424 

I believe is wiser to accept also the mask, this will allow us to use the PANIC more widely in the codebase.

### DIFF
--- a/cvmfs/catalog_diff_tool_impl.h
+++ b/cvmfs/catalog_diff_tool_impl.h
@@ -113,13 +113,13 @@ void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
     catalog::DirectoryEntry new_entry = new_listing[i_to];
 
     if (old_entry.linkcount() == 0) {
-      PANIC(kLogCvmfs, kLogStderr,
+      PANIC(kLogStderr,
             "CatalogDiffTool - Entry %s in old catalog has linkcount 0. "
             "Aborting.",
             old_entry.name().c_str());
     }
     if (new_entry.linkcount() == 0) {
-      PANIC(kLogCvmfs, kLogStderr,
+      PANIC(kLogStderr,
             "CatalogDiffTool - Entry %s in new catalog has linkcount 0. "
             "Aborting.",
             new_entry.name().c_str());
--- a/cvmfs/catalog_diff_tool_impl.h
+++ b/cvmfs/catalog_diff_tool_impl.h
@@ -12,6 +12,7 @@
 #include "download.h"
 #include "hash.h"
 #include "logging.h"
+#include "util/exception.h"
 #include "util/posix.h"
 
 const uint64_t kLastInode = uint64_t(-1);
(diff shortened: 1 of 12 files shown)


siscia:
Maybe we should also accept the paramenter for the LogSource? (this one: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/logging_internal.h#L71)

test/quickcheck/qc_catalog_merge_tool.cc
```diff
@@ -104,8 +105,7 @@ DirSpec ModifySpec(const DirSpec& in) {
         // TODO(radu): Implement file content modifications
         break;
       default:
-        LogCvmfs(kLogCvmfs, kLogStderr, "Unknown change type. Aborting.");
-        abort();
+        PANIC("Unknown change type. Aborting.");
```
jblomer: Mask is missing

cvmfs/catalog_diff_tool_impl.h
```diff
@@ -112,18 +113,16 @@ void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
     catalog::DirectoryEntry new_entry = new_listing[i_to];
 
     if (old_entry.linkcount() == 0) {
-      LogCvmfs(kLogCvmfs, kLogStderr,
-                "CatalogDiffTool - Entry %s in old catalog has linkcount 0. "
-                "Aborting.",
-                old_entry.name().c_str());
-      abort();
+      PANIC(kLogStderr,
+          "CatalogDiffTool - Entry %s in old catalog has linkcount 0. "
```
jblomer: Indentation is (I think) accidentally off by one

cvmfs/catalog_diff_tool_impl.h
```diff
@@ -112,18 +113,16 @@ void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
...
+          old_entry.name().c_str());
     }
     if (new_entry.linkcount() == 0) {
-      LogCvmfs(kLogCvmfs, kLogStderr,
-                "CatalogDiffTool - Entry %s in new catalog has linkcount 0. "
-                "Aborting.",
-                new_entry.name().c_str());
-      abort();
+      PANIC(kLogStderr,
+          "CatalogDiffTool - Entry %s in new catalog has linkcount 0. "
```
jblomer: Indentation

siscia:
Overall it seems smarter to allow the LogSource as well.

So we keep the interface the same and we can wide the scope of PANIC.

siscia:
@jblomer do we like this final interface?

This will change in several places and it will be quite a bit of work, so let's make sure that we agree on this one :)

jblomer:
I'm not sure about the log source.  The log source actually only makes sense for the debug log, when you try to parse a long log and quickly want to grep for things or skip others.  For `PANIC`, things come to a full stop, so we simply can use `kLogCvmfs` as the default source and not bother repeating it for every `PANIC` call.

siscia:
I added because there are other places that use the same pattern: Log+Abort but they have a different source.

One example is the cache: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/cache_transport.cc#L510

But I believe I found others...

jblomer:
Yep, `PANIC` and `LogCvmfs` would not be interchangeable function calls.  But in this case, I think the bigger benefit is not adding a log source to every `PANIC` call that, at the end of the day, is not needed.

siscia:
I removed the source from the PANIC call, but I left it in the underline code.
This will allow us the flexibility to add other PANIC call if necessary tomorrow.

PR #2465, opened 2019-12-13
remove death code

siscia:
Remove death code.

Everything seems to compile just fine without.


cvmfs/compat.h
```diff
@@ -647,10 +647,6 @@ class StringHeap : public SingleCopy {
  private:
   void AddBin(const uint64_t size) { assert(false); }
 
-  uint64_t size_;
```
jblomer: This is in fact not dead code.  It's necessary for the hot patch in order to establish the memory layout of that class in a previous version of cvmfs.

siscia: All the file or just `size_` ?

jblomer: All of them

siscia: Removed that file!

PR #2468, opened 2019-12-19
Remove log on different source

siscia:
This is the very last PR, built on top of #2467 

Here we use PANIC also for logs that are not directed to `kLogCvmfs`.

This completely address [CVM-1697](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1697)

There is this line here: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/options.cc#L194 where I would like a second opinion.

I believe it is fine as it is, without need to be modified.

PR #2471, opened 2020-01-07
Parallel catalog traversal in GC

janpriessnitz:
Implement CatalogTraversalParallel, a child of CatalogTraversal
class, which offers the functionality, but traverses the catalog
tree in parallel using multiple threads.
The interface of CatalogTraversalParallel is almost identical to
the one of CatalogTraversal and the two classes can be used almost
interchangeably.

Use CatalogTraversalParallel in garbage collection.

### DIFF
--- a/cvmfs/catalog_traversal_parallel.h
+++ b/cvmfs/catalog_traversal_parallel.h
@@ -1,0 +1,415 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#ifndef CVMFS_CATALOG_TRAVERSAL_PARALLEL_H_
+#define CVMFS_CATALOG_TRAVERSAL_PARALLEL_H_
+
+#include <stack>
+#include <string>
+#include <vector>
+
+#include "catalog_traversal.h"
+#include "ingestion/tube.h"
+#include "util/exception.h"
+
+namespace swissknife {
+
+/**
+ * This class implements the same functionality as CatalogTraversal, but in parallel.
+ * For common functionality, see the documentation of CatalogTraversal.
+ * Differences:
+ * - can choose number of threads
+ *
+ *  - traversal types change meaning:
+ * - depth-first -> parallelized post-order traversal (parents are processed
+ * after all children are finished)
+ * - breadth-first -> same as original, but parallelized
+ */
+template<class ObjectFetcherT>
+class CatalogTraversalParallel : public CatalogTraversalBase<ObjectFetcherT> {
+ public:
+  typedef CatalogTraversalBase<ObjectFetcherT> Base;
+  typedef ObjectFetcherT                      ObjectFetcherTN;
+  typedef typename ObjectFetcherT::CatalogTN  CatalogTN;
+  typedef typename ObjectFetcherT::HistoryTN  HistoryTN;
+  typedef CatalogTraversalData<CatalogTN>     CallbackDataTN;
+  typedef typename CatalogTN::NestedCatalogList NestedCatalogList;
+  typedef typename Base::Parameters Parameters;
+  typedef typename Base::TraversalType TraversalType;
+  typedef std::vector<shash::Any> HashList;
+
+  explicit CatalogTraversalParallel(const Parameters &params)
+    : CatalogTraversalBase<ObjectFetcherT>(params)
+    , num_threads_(params.num_threads)
+    , serialize_callbacks_(params.serialize_callbacks) {
+    atomic_init32(&num_errors_);
+    shash::Any null_hash;
+    null_hash.SetNull();
+    catalogs_processing_.Init(1024, null_hash, hasher);
+    catalogs_done_.Init(1024, null_hash, hasher);
+    pthread_mutex_init(&catalog_callback_lock_, NULL);
+    pthread_mutex_init(&catalogs_lock_, NULL);
+    effective_history_depth_ = this->default_history_depth_;
+    effective_timestamp_threshold_ = this->default_timestamp_threshold_;
+  }
+
+ protected:
+  struct CatalogJob : public CatalogTraversal<ObjectFetcherT>::CatalogJob,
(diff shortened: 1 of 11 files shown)


jblomer:
Please fix the linter errors.

cvmfs/catalog_traversal_parallel.h
```diff
@@ -0,0 +1,457 @@
...
+          continue;
+        }
+
+        CatalogJob *root_job = new CatalogJob("", *i, 0, 0);
+        PushJobUnlocked(root_job);
+        has_pushed = true;
+      }
+    }
+    // noop: no catalogs to traverse
+    if (!has_pushed) {
```
jblomer: Is `TraverseList` always called with an initially empty job list? In this case, we would not need the `has_pushed` variable but we could check whether the job list is empty.

cvmfs/catalog_traversal_parallel.h
```diff
@@ -0,0 +1,457 @@
...
+  }
+
+  bool DoTraverse() {
+    // Optimal number of threads is yet to be determined. The main event loop
+    // contains a spin-lock, so it should not be more than number of cores.
+    threads_process_ = reinterpret_cast<pthread_t *>
+                        (smalloc(sizeof(pthread_t)*num_threads_));
+    for (unsigned int i = 0; i < num_threads_; ++i) {
+      int retval = pthread_create(&threads_process_[i], NULL,
+                                  ProcessQueue, this);
```
jblomer: By convention, other parts of cvmfs call their thread function `Main...`, e.g. `MainProcessQueue`

cvmfs/ingestion/tube.h
```diff
@@ -65,17 +66,34 @@ class Tube : SingleCopy {
...
     while (size_ == limit_)
       pthread_cond_wait(&cond_capacious_, &lock_);
 
     Link *link = new Link(item);
-    link->next_ = tail_;
-    link->prev_ = tail_->prev_;
-    tail_->prev_->next_ = link;
-    tail_->prev_ = link;
-    tail_ = link;
+    link->next_ = head_->next_;
```
jblomer: Isn't this putting the link to position two, instead of at the back of the list?

cvmfs/garbage_collection/garbage_collector_impl.h
```diff
@@ -250,19 +251,17 @@ bool GarbageCollector<CatalogTraversalT, HashFilterT>::SweepReflog() {
...
-                                        CatalogTraversalT::kDepthFirstTraversal;
         std::vector<shash::Any>::const_iterator i    = catalogs.begin();
   const std::vector<shash::Any>::const_iterator iend = catalogs.end();
   for (; i != iend && success; ++i) {
     if (!hash_filter_.Contains(*i)) {
-      success =
-        success                                         &&
-        traversal_.TraverseRevision(*i, traversal_type) &&
-        RemoveCatalogFromReflog(*i);
+      success = success && RemoveCatalogFromReflog(*i);
```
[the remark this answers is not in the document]
janpriessnitz: Of course, that's right. Sorry, my bad.

cvmfs/ingestion/tube.h
```diff
@@ -65,17 +66,34 @@ class Tube : SingleCopy {
    * Push an item to the back of the queue.  Block if queue is currently full.
    */
   Link *Enqueue(ItemT *item) {
-    assert(item != NULL);
```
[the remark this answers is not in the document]
janpriessnitz: There's no comparison with sentinel in the code, so it should still work, but I agree that it's not the best way to do this. I'll fix it.

cvmfs/ingestion/tube.h
janpriessnitz: I don't think so. head_ is sentinel, head_->prev is the first element, head_->next is the last element.

cvmfs/catalog_traversal_parallel.h
janpriessnitz: There is a different problem here: the list can be non-empty, but only filled with root catalogs which were already traversed in the past. All of them will be skipped by the 'continue' on line 132. Even with non-empty list, you'll end up with an empty queue.

cvmfs/catalog_traversal_parallel.h
```diff
@@ -0,0 +1,457 @@
...
+      assert(retval == 0);
+    }
+
+    for (unsigned int i = 0; i < num_threads_; ++i) {
+      pthread_join(threads_process_[i], NULL);
```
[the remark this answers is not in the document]
janpriessnitz: Is assert sufficient?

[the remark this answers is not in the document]
jblomer: Yes, although then let's wait for #2468 and use `PANIC`

jblomer:
One more thing I forgot: let's also add some unit tests with generated large catalog trees in order to stress the parallelism.

janpriessnitz:
@cernvm-bot cpplint

jblomer:
@cernvm-bot unittest

jblomer:
@janpriessnitz coming (finally) back to this, we seem to have compile errors on old compilers.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest full

jblomer:
Replaced by #2558 (letting intrgration tests finish before closing)

jblomer:
There seems to be an issue with test case 559: https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/650/CVMFS_PLATFORM=ubuntu1804,CVMFS_PLATFORM_CONFIG=x86_64_s3,label=trampoline/testReport/junit/(root)/S3ServerIntegrationTests/Garbage_Collection_With_Named_Tag_Awareness/

PR #2469, opened 2020-01-06
Change geodb update to use license key

DrDaveD:
See [CVM-1833](https://sft.its.cern.ch/jira/browse/CVM-1833)

PR #2472, opened 2020-01-07
add CVMFS_GEO_DB_FILE

DrDaveD:
See [CVM-1833](https://sft.its.cern.ch/jira/browse/CVM-1833).  Makes most sense to merge after #2469.

### DIFF
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -976,10 +976,22 @@ _update_geodb() {
 
   # sanity checks
   [ -w "$dbdir"  ]   || { echo "Directory '$dbdir' doesn't exist or is not writable by $(whoami)" >&2; return 1; }
-  [ ! -f "$dbfile" ] || [ -w "$dbfile" ] || { echo "GeoIP database '$dbfile' is not writable by $(whoami)" >&2; return 2; }
 
   # check if an update/installation needs to be done
-  if [ ! -f "$dbfile" ]; then
+  if [ -n "$CVMFS_GEO_DB_FILE" ]; then
+    # This overrides the update/install; link to the given file instead.
+    if [ ! -L "$dbfile" ] || [ "`readlink $dbfile`" != "$CVMFS_GEO_DB_FILE" ]; then
+      if [ "$CVMFS_GEO_DB_FILE" != "NONE" ] && [ ! -r "$CVMFS_GEO_DB_FILE" ]; then
+        echo "CVMFS_GEO_DB_FILE $CVMFS_GEO_DB_FILE doesn't exist or is not readable" >&2
+        return 1
+      fi
+      rm -f $dbfile
+      echo "Linking GeoIP Database"
+      _to_syslog_for_geoip "linking db from $CVMFS_GEO_DB_FILE"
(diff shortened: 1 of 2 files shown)


cvmfs/server/cvmfs_server_util.sh
```diff
@@ -976,10 +976,22 @@ _update_geodb() {
 
   # sanity checks
   [ -w "$dbdir"  ]   || { echo "Directory '$dbdir' doesn't exist or is not writable by $(whoami)" >&2; return 1; }
-  [ ! -f "$dbfile" ] || [ -w "$dbfile" ] || { echo "GeoIP database '$dbfile' is not writable by $(whoami)" >&2; return 2; }
 
   # check if an update/installation needs to be done
-  if [ ! -f "$dbfile" ]; then
+  if [ -n "$CVMFS_GEO_DB_FILE" ]; then
```
siscia: Maybe

```
if [ x"$CVMFS_GEO_DB_FILE" != x"" ]
```

siscia:
I really like this solution, but how it works when we are creating a new replica? Is the variable `CVMFS_GEO_DB_FILE` define and read in that case? Or we don't really care?

If I understand correctly we would:
1. first create a replica,
2. then set `CVMFS_GEO_DB_FILE` and
3. finally call the update-geodb

I was trying to avoid step 2 and 3 and it was getting quite complex. If this is enough I am more than happy with the result.

On similar notes, with Jakob we were thinking how to avoid the user to set the `CVMFS_GEO_DB_FILE` variable at all and use the system provided one.
Do you think it would be a good idea to check if the file is present in the default locations? And if it is present use the most recently updated one?

The default location at the moment are `/usr/share/GeoIP` for debs packages and and `/var/lib/GeoIP` for rpms.

DrDaveD:
On a new machine, people won't be able to create a replica without setting either CVMFS_GEO_DB_FILE or CVMFS_GEO_LICENSE_KEY first.  There should be no need to call update-geodb once they have set one of those, because add-replica will either create the link or do the download.

I don't think it's worth looking in the default rpm/deb locations, because I don't believe they will continue to exist.  On the other hand it might be worth looking in the default location from Maxmind's [geoipupdate](https://dev.maxmind.com/geoip/geoipupdate/) tool.

cvmfs/server/cvmfs_server_util.sh
DrDaveD: No, I think that's worse style and just reflects a lack of knowledge of both test -n and the fact that double quotes makes it unnecessary to use the double-x pattern.  That pattern is only necessary when not using quotes.

siscia: I was under the impression that it was needed to comply with old/weird shell version. I would keep it just for consistency with the whole codebase at this point.

siscia:
Indeed, in my understanding, geoipupdate updates the files in /usr/share/GeoIP (at least reading from the source and reading the default configuration).

I was hoping they would coordinate the location between debs and rpms, but ok.

cvmfs/server/cvmfs_server_util.sh
DrDaveD: There are other uses of -n in cvmfs_server.

PR #2474, opened 2020-01-14
Add option to upload stats.db to storage backend

janpriessnitz:
Add option CVMFS_UPLOAD_STATS_DB to server config. When enabled,
CVMFS will upload the updated stats.db into the storage backend
after a publish or GC.

+ rework of storing statistics into stats.db: move the code from
swissknife_main into 'ingest', 'sync' and 'gc' commands.


cvmfs/statistics_database.cc
```diff
@@ -331,45 +375,123 @@ StatisticsDatabase::~StatisticsDatabase() {
...
-  } else if (command_name == "gc") {
-    insert_statement = PrepareStatementIntoGc(statistics, start_time,
-                                              finish_time, repo_name);
+StatisticsDatabase *StatisticsDatabase::OpenStandardDB(
+  const std::string repo_name)
+{
+  StatisticsDatabase *db;
+  std::string db_file_path = GetDBPath(repo_name);
+  if (FileExists(db_file_path)) {
+    db = StatisticsDatabase::Open(db_file_path, kOpenReadWrite);
```
jblomer: Perhaps we should check here whether reponame in the database is what we expect it to be.

cvmfs/statistics_database.cc
```diff
@@ -331,45 +375,123 @@ StatisticsDatabase::~StatisticsDatabase() {
...
+
+bool StatisticsDatabase::StoreGCStatistics(
+  const perf::Statistics *statistics,
+  const std::string &start_time,
+  const bool success)
+{
+  std::string finish_time = GetGMTimestamp();
+  std::string statement =
+    PrepareStatementIntoGc(statistics, start_time, finish_time,
+                           repo_name_, success);
```
jblomer: Perhaps this can be made symmetric to the `StorePublishStatistics`, i.e. without the repo_name parameter

cvmfs/statistics_database.cc
```diff
@@ -331,45 +375,123 @@ StatisticsDatabase::~StatisticsDatabase() {
...
+  return StoreEntry(statement);
+}
+
+
+bool StatisticsDatabase::StoreEntry(const std::string &insert_statement) {
   sqlite::Sql insert(this->sqlite_db(), insert_statement);
 
+  std::string error_message = "";
   if (!this->BeginTransaction()) {
```
jblomer: We can probably drop the transaction because we put it around only a single statement

cvmfs/statistics_database.h
```diff
@@ -42,15 +47,48 @@ class StatisticsDatabase : public sqlite::Database<StatisticsDatabase> {
...
+ * configured in spooler.
+ * 
+ * @return true on success, false otherwise
+ */
+  bool UploadStatistics(upload::Spooler *spooler,
+                        std::string local_path = "");
+
+/**
+ * Upload the statistics database into the storage backend
+ * configured in uploader.
```
jblomer: Let's clarify the context in which one or the other of the overloads is used

cvmfs/swissknife_gc.cc
```diff
@@ -169,13 +173,21 @@ int CommandGc::Main(const ArgumentList &args) {
     }
   }
 
+  StatisticsDatabase *stats_db = StatisticsDatabase::OpenStandardDB(repo_name);
```
jblomer: This should always succeed, shouldn't it?  We could crash if `stats_db == NULL` and then wouldn't need to test later on.  Same for the other commands.

cvmfs/upload.h
```diff
@@ -264,6 +264,14 @@ class Spooler : public Observable<SpoolerResult> {
    */
   bool Peek(const std::string &path) const;
 
+  /**
+   * Make directory in upstream storage. Noop if directory already present.
```
jblomer: Let's mention that its current only use is creation of the stats sub directory

cvmfs/upload_gateway.cc
```diff
@@ -117,6 +117,11 @@ void GatewayUploader::DoRemoveAsync(const std::string& /*file_to_delete*/) {
 
 bool GatewayUploader::Peek(const std::string& /*path*/) { return false; }
 
+// noop
```
jblomer: Rather a TODO than a noop.

test/src/675-statsupload/main
```diff
@@ -0,0 +1,53 @@
...
+  mkdir /cvmfs/$CVMFS_TEST_REPO/dir
+  echo "first 1" > /cvmfs/$CVMFS_TEST_REPO/dir/first
+  publish_repo $CVMFS_TEST_REPO || return $?
+
+  echo "*** check that stats.db was not uploaded"
+  curl -s --fail --output /dev/null "${repo_url}/stats/stats.db" && return 2
+
+  echo "*** enable statistics DB upload"
+  mv $server_cfg_file "${server_cfg_file}.old"
+  sed -e '/CVMFS_UPLOAD_STATS_DB/d' "${server_cfg_file}.old" > $server_cfg_file
```
jblomer: That's not needed, the later value will overwrite the earlier one

test/src/675-statsupload/main
```diff
@@ -0,0 +1,53 @@
...
+  curl -s --fail -o stats1.db "${repo_url}/stats/stats.db" || return 3
+  diff stats1.db "/var/spool/cvmfs/${CVMFS_TEST_REPO}/stats.db" || return 4
+
+  echo "*** perform basic garbage collection"
+  cvmfs_server gc -r1 -f $CVMFS_TEST_REPO || return 5
+
+  echo "*** check that stats.db was uploaded"
+  curl -s --fail -o stats2.db "${repo_url}/stats/stats.db" || return 6
+  diff stats2.db "/var/spool/cvmfs/${CVMFS_TEST_REPO}/stats.db" || return 7
+
```
jblomer: Perhaps in addition, check that the publish and the gc operation are in fact listed in the downloaded copy

PR #2478, opened 2020-01-15
Mention server.local in comment about update-geodb variables

DrDaveD:


DrDaveD:
I changed my mind on this PR, because I think I should update some other defaults and I'll include this comment change with that.  I'm going to close this one.

PR #2479, opened 2020-01-15
Reduce the minimum and maximum days for geo db updates

DrDaveD:
This is to help comply with the new license requirement of 30 days for updates.  Also adds mention in the comment of overriding via /etc/cvmfs/server.local.  See [CVM-1833](https://sft.its.cern.ch/jira/browse/CVM-1833).

### DIFF
--- a/cvmfs/server/cvmfs_server_coda.sh
+++ b/cvmfs/server/cvmfs_server_coda.sh
@@ -5,7 +5,8 @@
 ################################################################################
 
 # Configuration variables for update-geodb -l.  May be overridden in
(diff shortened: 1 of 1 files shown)


DrDaveD:
@jblomer I intended this to go into cvmfs-2.7.1 but it looks I messed up by submitting it against devel instead of the cvmfs-2.7 branch.  I don't think it's worth cutting a 2.7.2 just for this, but if a 2.7.2 is released, please include this.

jblomer:
@DrDaveD Backported.

PR #2483, opened 2020-01-27
add flag to opt-out of docker conversion

siscia:
WIP, still need to add test for this.

But it should completely address [CVM-1786](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1786?filter=allopenissues)

### DIFF
--- a/ducc/cmd/convert.go
+++ b/ducc/cmd/convert.go
@@ -19,13 +19,14 @@ var (
 )
 
 var (
-	convertAgain, overwriteLayer, convertSingularity bool
+	convertAgain, overwriteLayer, convertDocker, convertSingularity bool
 )
 
 func init() {
 	convertCmd.Flags().BoolVarP(&overwriteLayer, "overwrite-layers", "f", false, "overwrite the layer if they are already inside the CVMFS repository")
 	convertCmd.Flags().BoolVarP(&convertAgain, "convert-again", "g", false, "convert again images that are already successfull converted")
 	convertCmd.Flags().BoolVarP(&convertSingularity, "convert-singularity", "s", true, "also create a singularity images")
+	convertCmd.Flags().BoolVarP(&convertDocker, "convert-docker", "d", true, "unpacking the layers into the repository")
 	rootCmd.AddCommand(convertCmd)
 }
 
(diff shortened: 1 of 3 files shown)


siscia:
@jblomer this is good to review!

@DrDaveD Let me know if this address completely your jira! If it is the case feel free to close it, marking it as fixed! :)

jblomer:
In principle, looks good to me.  But I'd change the naming of variables from `Docker` and `Singularity` to `Layers` and `Flat`.  We have in fact three options for conversion:

1. Produce the flat unpacked image
2. Produce the individual unpacked layer directories
3. Upload the thin image

Options 1 and 2 are orthogonal, option 3 depends on option 2.

siscia:
I completely agree!

However I would like to keep changes small. I would prefer to get this merged. Along with the next one for custom storage directory.

Then we can move to the new terminology. Maybe waiting 2.8?

jblomer:
Sounds fine. Please just create a ticket then for the terminology change so that we don't forget.

DrDaveD:
I built the current version of this PR and tried it and am not having good luck.  Here's my wishlist.yaml:
```
version: 1
user: dwd
cvmfs_repo: test.ducc
input:
    - 'https://registry.hub.docker.com/davedykstra/cvmfs-fuse3:latest'
    - 'https://registry.hub.docker.com/davedykstra/cvmfs-fuse3:testing'
```
First with `cvmfs_ducc convert --convert-docker=false wishlist.yaml ` I got the error
```
ERRO[0000] No password provide to upload the docker images  error="Env variable (DUCC_DOCKER_REGISTRY_PASS) storing the password to access the docker registry is not set"
```
The registry is readable to the world so I don't see why that was needed, so I set DUCC_DOCKER_REGISTRY_PASS to `xx` and then got these errors:
```
WARN[0000] Error in creating the wish                    error="Impossible to identify the registry of the image:  | Error in parsing the output image"
WARN[0000] Error in creating the wish                    error="Impossible to identify the registry of the image:  | Error in parsing the output image"
INFO[0000] No recipe to convert  
```
Aren't I supposed to be able to do without an output_format line with `--convert-docker=false`?  If I add the output_format line from the [documentation](https://cvmfs.readthedocs.io/en/stable/cpt-ducc.html#ducc-commands) then it seems to do the right thing.  However if I run it again I thought it would notice that there's no change, but I see it publishing both images again.  Also it leaves behind a /tmp/conversionNNNN directory for every image each time it is run.

siscia:
Can you make sure that it publish both image again?
It is running 24/7 here and it is not publishing the same images again...

Please, understand that DUCC was born to manage Singularity and Docker thin images, with a lot of focus on the Docker thin images that use to be very complex to make. It seems like docker and docker thin images are loosing a bit of momentum and that they might not be so relevant in our future. But it will take more time on DUCC to adjust for it.

DrDaveD:
Yes, it is doing a singularity build and republishing all the files.  The timestamps get updated.  It should know the hash before calling singularity build so I would think it would be able to skip that step completely.

siscia:
Many thanks! I will have a look into this!

siscia:
So, this is already in via another PR addressing the comments of @DrDaveD as well.

Dave also suggested to change the flags options, to use something more standard. 

So no anymore `--convert-docker=false` but something like `--skip-docker-coversion`, that eventually would be `--skip-layer-ingestion` and `--skip-thin-image-creation`.

Which I believe is a great idea even though I can't find where this suggestion was made.

I am closing this issue since there is no reason for this to stay open.

PR #2485, opened 2020-01-29
Fix rpm build

DrDaveD:
This rm in the rpm spec was causing rpm builds to fail.


packaging/rpm/cvmfs-universal.spec
```diff
@@ -384,7 +384,6 @@ rm -f $RPM_BUILD_ROOT%{_libdir}/libcvmfs_fuse3*
 %endif
 
 # Remove experimental server binaries
-rm -f "$RPM_BUILD_ROOT%{_bindir}/cvmfs_publish"
 rm -f "$RPM_BUILD_ROOT%{_libdir}/libcvmfs_server*"
```
jblomer: Many thanks!  Actually, the entire `rm` block should go away.

PR #2487, opened 2020-02-05
Add feature to clone directory subtrees

jblomer:
Adds a `CloneTree()` method to the catalog manager that can copy an entire subtree to a new location. This is a meta-data only operation.  A first test with cloning /cvmfs/cms.cern.ch/slc7_ppc64le_gcc530 results in at least 50k file system entries per second.

Cloning is supposed to provide the core functionality for "fast flattening" of container images. Given that the chain directory for layers 1..n-1 already exists, one can create the chain for layers 1..n by cloning the chain[1..n-1] and then publishing only the last delta layer. See CVM-1801

### DIFF
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -146,7 +146,11 @@ class AbstractCatalogManager : public SingleCopy {
   bool ListCatalogSkein(const PathString &path,
                         std::vector<PathString> *result_list);
 
-  bool Listing(const PathString &path, DirectoryEntryList *listing);
+  bool Listing(const PathString &path, DirectoryEntryList *listing,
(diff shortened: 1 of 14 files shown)


jblomer:
ok to cloudtest

jblomer:
ok to cloudtest

jblomer:
@cernvm-bot help

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2494, opened 2020-02-17
integration test 675: minor bugfix

janpriessnitz:


janpriessnitz:
@cernvm-bot cloudtest

PR #2497, opened 2020-02-21
Ducc gc better version

siscia:
A rewrite of the GC in DUCC.

The new version is a little less smart than the old one, however it is much simpler and more complete in functionality.

It scans the whole tree and delete the layers and paths that are not in use anymore.

An upgrade to this would be to provide a grace period before to eliminate the files.

A possible implementation of this grace period would be to touch a `.last_used` file in the metadata folders of the layers and of the images once a day.

Eventually we would delete only the layers/images whose `.last_used` file is older than the grace period.

### DIFF
--- a/ducc/cmd/garbage_collection.go
+++ b/ducc/cmd/garbage_collection.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


siscia:
If nobody has anything against this I would merge it tomorrow!

jblomer:
Just curious, in which way was the old implementation smarter?

siscia:
It was keeping a file to know what directory to eliminate directly, without traversing the filesystem.

Actual test show that traversing the filesystem need between 30 seconds (parallel) and 1 minutes (sequential) on unpacked.cern.ch .

Given that the performance penalty is very low, I opted for a simpler and more robust implementation.

PR #2498, opened 2020-02-23
Fix compile and packaging errors

jblomer:


jblomer:
@cernvm-bot build

jblomer:
@cernvm-bot help

PR #2499, opened 2020-02-24
Fix unittests on newly tested platforms

janpriessnitz:
- 32-bit images on 64-bit hosts: T_MountPoint: explicit RAM cache size
- SLES11, SLES12: geoapi unittests: IPv4 fallback on old systems


janpriessnitz:
@cernvm-bot unittest

PR #2500, opened 2020-02-25
Ducc change naming

siscia:
This address [CVM-1838](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1838)

It changes the external naming interface from `docker` to `layers` and from `singularity` to `flat`. I don't see many reasons at the moment to change also the internal naming.

Since we are changing the interface anyway I followed the suggestion of @DrDaveD and change the boolean flags (`--convert-docker` and `--convert-singularity`) into their negative (`--skip-layers` and `--skip-flat`)

There is also some enhancement to the CLI, making it more clear what arguments are necessary.

Finally we enhance the `convert-singularity-image` and upgrade it to `convert-single-image` that by default convert both the flat filesystem and layers, but of course accept the flags to skip them.


siscia:
I will get this in tomorrow, unless someone has anything to say about it.

PR #2508, opened 2020-03-12
BUGFIX: GW SessionContext: more flexible GW reply JSON evaluation

janpriessnitz:
Instead of comparing the reply JSON to a hardcoded value,
decode the reply and look for the "status" field value.


janpriessnitz:
@jblomer what is an appropriate base branch for this PR?

jblomer:
Very good! Let's apply this to the 2.7 branch.

PR #2511, opened 2020-03-17
Packaging: fix broken path for Fedora 31

janpriessnitz:
/usr/sbin/hardlink -> /usr/bin/hardlink in cvmfs-universal.spec


jblomer:
Here, I'd rather introduce an RPM variable "hardlink" at the preamble of the spec file. This variable would contain the correct path depending on the OS.

janpriessnitz:
ok, sounds good

janpriessnitz:
@jblomer ready to merge.

PR #2513, opened 2020-03-20
Fix catalogs create by ingestion command

siscia:
The ingestion command accept an argument `--catalog` to create a nested catalog in the directory where the tarball is ingested.

The implementation seems wrong, and, sometimes it creates catalogs of size zero but with content.

I believe the problem is in the use of the libarchive library, but I was unable to actually understand completely what is going on.

This PR takes a much cleaner approach to the creation of nested catalogs.

The error was found because trying to open a `.cvmfscatalog` file resulted in an `I/O error`, moreover, the file pointed to an hash that actually contains content.

At the moment I don't see anymore `I/O error`s while trying to open the `.cvmfscatalog`, but it is true that the error was transient. Given the PR,  however, I don't expect the `.cvmfscatalog` file to have any content.

Maybe we could add a runtime check to make sure that the declared size of files is actually correct.


cvmfs/sync_item_dummy.h
```diff
@@ -10,9 +10,56 @@
...
+    return expected_type == kItemFile;
+  }
+
+  catalog::DirectoryEntryBase CreateBasicCatalogDirent() const {
+    catalog::DirectoryEntryBase dirent;
+    std::string name(".cvmfscatalog");
+    dirent.inode_ = catalog::DirectoryEntry::kInvalidInode;
+    dirent.linkcount_ = 1;
+    dirent.mode_ =
+        S_IFREG | S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
```
jblomer: Should this be rather 0664 instead of 0666?

cvmfs/sync_item_dummy.h
```diff
@@ -10,9 +10,56 @@
...
+    dirent.uid_ = getuid();
+    dirent.gid_ = getgid();
+    dirent.size_ = 0;
+    dirent.mtime_ = time(NULL);
+    dirent.checksum_ = this->GetContentHash();
+    dirent.is_external_file_ = this->IsExternalData();
```
jblomer: I think this should always be false (.cvmfscatalog files are not part of an external data set)

jblomer:
Actually, this should be opened against the 2.7 branch and only then merged into devel

jblomer:
But let's run the integration tests first while it is still open against devel

cvmfs/sync_item_dummy.h
siscia: Fixed!

cvmfs/sync_item_dummy.h
siscia: Right!

siscia:
@cernvm-bot cloudtest

siscia:
@jblomer @janpriessnitz are this our failure? Should I just re-run in cases like this?

janpriessnitz:
Rerun the tests. Those were unrelated errors which should not be there anymore.

siscia:
@cernvm-bot cloudtest

PR #2521, opened 2020-03-26
[DO NOT MERGE][DUCC] Ducc remove singularity keep deps

siscia:
Just like #2515 but smaller without touching the dependencies.


jblomer:
Expired, replaced by other PRs

PR #2523, opened 2020-03-30
fix panic macro

siscia:
it use to print the literal __FILE__ and __LINE__ instead of the
actual file and line of the panic.
Double macro from: https://stackoverflow.com/questions/19343205/c-concatenating-file-and-line-macros

Found this while working on the GC

### DIFF
--- a/cvmfs/util/exception.h
+++ b/cvmfs/util/exception.h
@@ -20,9 +20,9 @@ class ECvmfsException : std::runtime_error {
       : std::runtime_error(what_arg) {}
 };
 
-#define S1(x) #x
-#define S2(x) S1(x)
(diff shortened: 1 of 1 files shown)


cvmfs/util/exception.h
```diff
@@ -20,7 +20,10 @@ class ECvmfsException : std::runtime_error {
       : std::runtime_error(what_arg) {}
 };
 
-#define PANIC(...) Panic("PANIC: __FILE__ : __LINE__", kLogCvmfs, __VA_ARGS__);
+#define S1(x) #x
```
jblomer: Just to avoid accidental name clash, let's call this `__CVMFS_S1` and `__CVMFS_S2`

PR #2525, opened 2020-04-03
PayloadProcessor: copy data buffer before passing to uploader

janpriessnitz:


janpriessnitz:
@cernvm-bot cloudtest

janpriessnitz:
@cernvm-bot cloudtest

cvmfs/receiver/payload_processor.cc
```diff
@@ -165,8 +165,12 @@ void PayloadProcessor::ConsumerEventCallback(
   FileInfo& info = pending_files_[event.id];
 
   if (!info.skip) {
-    upload::AbstractUploader::UploadBuffer buf(event.buf_size, event.buf);
-    uploader_->ScheduleUpload(info.handle, buf);
+    void *buf_copied = smalloc(event.buf_size);
```
siscia, line 168: isn't worth to check that `buf_copied != null` and if it is just panic.

jblomer: Yep, that's what `smalloc` is doing on top of `malloc`: checks that the memory could be allocated (the "s" in `smalloc` was meant for "safe")

PR #2526, opened 2020-04-03
[WIP] CVM-1861 Gateway support for GC

siscia:
This is to be intended as a sketch of the real implementation.

Nevertheless I would appreciate feedback on it.


siscia:
We might keep the integration test, but we can close this one.

Ref to CVM-1861

PR #2527, opened 2020-04-06
Watchdog process for cvmfs_receiver - rebased on 2.7 

janpriessnitz:


janpriessnitz:
@jblomer ready to merge. I forgot that the unittests do not work work on 2.7 branch.

cvmfs/receiver/receiver.cc
```diff
@@ -7,16 +7,25 @@
...
 
 #include "../logging.h"
+#include "../monitor.h"
 #include "../swissknife.h"
+#include "../util/posix.h"
+#include "../util/string.h"
 
 #include "reactor.h"
 
+static const char *kDefaultReceiverLogDir = "var/log/cvmfs_receiver/";
```
jblomer, line 17: I think the last commit(s) from the PR against the devel branch are missing

PR #2531, opened 2020-04-07
Check if fix for CVM-1848 fixes CVM-1859 as well

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2532, opened 2020-04-07
Change cvmfs_server transaction return code

bencouturier:
…quired

### DIFF
--- a/test/src/808-lease_failure/main
+++ b/test/src/808-lease_failure/main
@@ -1,0 +1,74 @@
+#
+# Test for the cvmfs_server transaction command returns codes
+# 
+# This sets up a repository with gateway and simulates the case
+# when another release manager has already taken the lock for 
+# the path we want.
+
+cvmfs_test_name="Repository gateway"
+cvmfs_test_autofs_on_startup=false
+cvmfs_test_suites="quick"
+
+clean_up() {
+    echo "Cleaning up"
+
+    echo "  Removing output directories"
+    rm -rvf /tmp/cvmfs_out_*
+}
+
+
+run_transactions() {
+    set_up_repository_gateway
+
+    ## Transaction 1 (Creating a new deep directory sub-tree in place of a former file)
+    echo "  Starting transaction 1"
+    cvmfs_server transaction test.repo.org
+    if [ $? -ne 0 ]; then
+        echo "Did not manage to  start transaction"
+        return 1
+    fi
+
+    mkdir -p /cvmfs/test.repo.org/new/c/d
+    echo "New file" > /cvmfs/test.repo.org/new/c/d/new_file.txt
+
+    echo "  Checking that we cannot open a transaction again"
+    cvmfs_server transaction test.repo.org
+    retcode=$?
+    if [ $retcode -ne 17 ]; then
+        echo "cvmfs_server transaction should tell us that we have a transaction already"
+        return 1
+    fi
+
+    cvmfs_server publish test.repo.org
+    if [ $? -ne 0 ]; then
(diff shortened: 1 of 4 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

bencouturier:
This pull request proposes to change the return code for cvmfs_server transaction in several cases. I would find it useful to know when the failure is due to user errors:
- when a transaction already exists
- when a lock already was taken for path we want to lease
I tried to find meaningful return code, and looking at the standard error numbers, I thought that EEXISTS and EBUSY could make sense.

jblomer:
@cernvm-bot cloudtest

test/src/808-lease_failure/main
```diff
@@ -0,0 +1,74 @@
+#
+# Test for the cvmfs_server transaction command returns codes
+# 
+# This sets up a repository with gateway and simulates the case
+# when another release manager has already taken the lock for 
+# the path we want.
+
+cvmfs_test_name="Repository gateway"
```
jblomer, line 8: Perhaps better: "Lease contention with repository gateway"

test/src/808-lease_failure/main
```diff
@@ -0,0 +1,74 @@
+#
```
jblomer, line 1: Let's call it 809-lease_failure (number 808 is already taken by another test)

jblomer:
@cernvm-bot cloudtest cc7 nodestroy

test/src/808-lease_failure/main
```diff
@@ -0,0 +1,74 @@
...
+
+    mkdir -p /cvmfs/test.repo.org/new/c/d
+    echo "New file" > /cvmfs/test.repo.org/new/c/d/new_file.txt
+
+    echo "  Checking that we cannot open a transaction again"
+    cvmfs_server transaction test.repo.org
+    retcode=$?
+    if [ $retcode -ne 17 ]; then
+        echo "cvmfs_server transaction should tell us that we have a transaction already"
+        return 1
```
jblomer, line 39: We should use different codes for every return

test/src/808-lease_failure/main
```diff
@@ -0,0 +1,74 @@
...
+    fi
+
+    echo "  Checking that the lease is taken and that cvmfs_transaction fails"
+    cvmfs_server transaction test.repo.org/new/c
+    retcode=$?
+    if [ $retcode ]; then
+        echo "Return code should be 16 EBUSY"
+        return 1
+    fi
+    clean_up
```
jblomer, line 62: Let's wrap that in a trap at the beginning of the test

jblomer:
@cernvm-bot cloudtest

test/src/808-lease_failure/main
```diff
@@ -0,0 +1,77 @@
...
+    fi
+
+    echo "  Checking that the lease is taken and that cvmfs_transaction fails"
+    cvmfs_server transaction test.repo.org/new/c
+    retcode=$?
+    if [ $retcode -ne 16 ]; then
+        echo "Return code should be 16 EBUSY instead of ${retcode}"
+        return 24
+    fi
+    clean_up
```
jblomer, line 65: We actually don't need the call because the trap will run it anyway

test/src/808-lease_failure/main
bencouturier: Done

test/src/808-lease_failure/main
bencouturier: Removed

jblomer:
@cernvm-bot cloudtest

PR #2534, opened 2020-04-09
Add quantiles computation to Log2Histogram

siscia:
Quantiles are useful for performance and regression tests on the
client side.

This was suppose to be one of the first tasks of the summer student. Nevertheless I believe it is important to keep this development line open as well.

It would be interesting to have a command to spit out just this numbers so that we can compare the performance on the client programmatically from different installation and user bases.

At the moment the output produced is somehow like this:

```
Readdir                                                                                                                                  
                    nsec |   count |           distribution |                                                                            
         0 ->          1 :       0 |                        |                                                                            
         2 ->          3 :       0 |                        |                                                                            
         4 ->          7 :       0 |                        |                                                                            
         8 ->         15 :       0 |                        |                                                                            
        16 ->         31 :       0 |                        |                                                                            
        32 ->         63 :       0 |                        |                                                                            
        64 ->        127 :       0 |                        |                                                                            
       128 ->        255 :       0 |                        |                                                                            
       256 ->        511 :       0 |                        |                                                                            
       512 ->       1023 :       0 |                        |                                                                            
      1024 ->       2047 :       8 |                        |                                                                            
      2048 ->       4095 :     748 | ********************** |                                                                            
      4096 ->       8191 :     242 |                ******* |                                                                            
      8192 ->      16383 :      48 |                      * |                                                                            
     16384 ->      32767 :     188 |                  ***** |                                                                            
     32768 ->      65535 :       1 |                        |                                                                            
     65536 ->     131071 :       1 |                        |                                                                            
    131072 ->     262143 :       2 |                        |                                                                            
    262144 ->     524287 :       0 |                        |                                                                            
    524288 ->    1048575 :       0 |                        |                                                                            
   1048576 ->    2097151 :       0 |                        |                                                                            
   2097152 ->    4194303 :       0 |                        |                                                                            
   4194304 ->    8388607 :       0 |                        |                                                                            
   8388608 ->   16777215 :       0 |                        |                                                                            
  16777216 ->   33554431 :       0 |                        |                                                                            
  33554432 ->   67108863 :       0 |                        |                                                                            
  67108864 ->  134217727 :       0 |                        |                                                                            
 134217728 ->  268435455 :       0 |                        |                                                                            
 268435456 ->  536870911 :       0 |                        |
 536870912 -> 1073741823 :       0 |                        |
                overflow :       0 |                        |
                   total :    1238


Quantiles
q,0.10,2362
q,0.20,2702
q,0.25,2872
q,0.30,3041
[...]

PR #2536, opened 2020-04-14
Show latency for clients 

siscia:
Same as #2534


jblomer:
Needs to be opened against the 2.7 branch

siscia:
Aw sorry, I was sure I opened against the correct branch.

PR #2543, opened 2020-04-20
Add 'chunks_access' xattr listing regular file chunks

janpriessnitz:


cvmfs/cvmfs.cc
```diff
@@ -1523,6 +1523,33 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
       fuse_reply_err(req, ENOATTR);
       return;
     }
+  } else if (attr == "user.chunks_access") {
```
jblomer, line 1526: The work `access` sounds like permission control, how about `chunks_info`?

cvmfs/cvmfs.cc
```diff
@@ -1523,6 +1523,33 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+    if (d.IsRegular()) {
+      attribute_value = "hash,offset,size\n";
+      if (d.IsChunkedFile()) {
+        FileChunkList chunks;
+        if (!catalog_mgr->ListFileChunks(path, d.hash_algorithm(), &chunks) ||
```
jblomer, line 1531: We should get the chunk list earlier, within the catalog manager fence.  Because this is potentially expensive, we probably would also want to fill the chunk list only if asked for (like we do with the catalog counters).

cvmfs/cvmfs.cc
janpriessnitz: How about chunks_list?

jblomer: Yep, I think that's fine.  Perhaps `chunk_list`.

PR #2551, opened 2020-04-29
[CVM-1697] Receiver error propagate to client and gateway

siscia:
This PR should propagate the exception from the the receiver to the gateway, for logging, and to the client to give feedback to the user.

Should address: [CVM-1697](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1697?filter=myopenissues)


siscia:
Should also address [CVM-1881](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1881) and also [CVM-1882](https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1882)

cvmfs/util/exception.cc
```diff
@@ -39,7 +39,7 @@ void Panic(const char* coordinates, const LogSource source, const int mask,
   // From now on we deal only with `msg`
 
   // Either throw the exception of log + abort
-#ifdef LIBCVMFS_SERVER
+#ifdef RAISE_EXCEPTIONS
```
jblomer, line 42: We should prefix our defines with `CVMFS`, e.g. `CVMFS_RAISE_EXCEPTIONS`

cvmfs/util/exception.cc
```diff
@@ -39,7 +39,7 @@ void Panic(const char* coordinates, const LogSource source, const int mask,
   // From now on we deal only with `msg`
 
   // Either throw the exception of log + abort
```
jblomer, line 41: typo: `or log + abort`

cvmfs/receiver/reactor.cc
```diff
@@ -28,7 +28,7 @@ Reactor::Request Reactor::ReadRequest(int fd, std::string* data) {
   using namespace receiver;  // NOLINT
 
   // First, read the command identifier
-  int32_t req_id = 0;
+  int32_t req_id = kQuit;
   int nb = SafeRead(fd, &req_id, 4);
```
jblomer, line 32: Let's replace the `4` by `sizeof(req_id)`, and also in the following lines of the code

cvmfs/receiver/reactor.cc
```diff
@@ -463,40 +462,60 @@ CommitProcessor* Reactor::MakeCommitProcessor() {
...
-      ok &= WriteReply(fdout_, reply);
-      break;
-    case kError:
-      LogCvmfs(kLogReceiver, kLogSyslogErr,
-               "Reactor: unknown command received.");
-      ok = false;
-      break;
-    default:
-      break;
+  try {
```
jblomer, line 465: Do we actually still need the return code?  `HandleRequests()` is only used in `Reactor::Run`.  We can make the method `void` and put the `try {} catch {}` block around the use of the method.

cvmfs/receiver/reactor.h
```diff
@@ -35,7 +35,8 @@ class Reactor {
     kCheckToken,
     kSubmitPayload,
     kCommit,
-    kError
+    kError,
+    kTestCrash  // use to test the gateway
```
jblomer, line 39: Let's be more specific: to test what exactly in the gateway

cvmfs/receiver/reactor.cc
siscia: I believe this is 4 to reflect the one on the gateway.

So it will be `4` on both size. Insetad of `sizeof()` on the c++ part and 4 on the go part.

cvmfs/receiver/reactor.cc
siscia: No we don't need the return code, but it is the reactor that can send back the reply to the user.

In this way we catch the error, we format it, we make a json out of it, we send it to the gateway, the gateway understand it, and send feedback to the final user.

I put this extra layer for the catch part that does the error formatting and then re-raise the exception.

cvmfs/receiver/reactor.cc
jblomer: Fair enough, let's add this as a code comment then.

cvmfs/receiver/reactor.h
siscia: Here we are testing what happens when the receiver throw an exception (hopefully there are not false assert in the codepaths).

Before this PR (and related PR) if the receiver crash the gateway just hang waiting.

cvmfs/receiver/reactor.cc
siscia: That it is just a guess!!!

I didn't write this code :laughing:

siscia:
Yeah the trigger will happen in the test of the gateway, I am still playing around with that one.

siscia:
@cernvm-bot unittest

siscia:
@cernvm-bot unittest

cvmfs/receiver/reactor.cc
siscia: @jblomer , do we agree on this one?

We catch the code internally, we write the message to the output socket, and then we re-throw up the error for manage it at higher level.

siscia:
For what concerns the mix of error handling. I agree that it will look nicer, but it would require some effort.

I would suggest to do it in a separate PR so to keep things clean here.

cvmfs/json_document_write.h
```diff
@@ -53,12 +53,50 @@ class JsonStringGenerator {
...
+    entry.val = escape(val);
     entry.is_quoted = quoted;
     entries.push_back(entry);
   }
 
+  // this escape procedure is not as complete as it should be.
+  // we should manage ALL control chars from '\x00' to '\x1f'
+  // however this are the one that we can expect to happen
+  // More info: https://stackoverflow.com/a/33799784/869271
+  std::string escape(const std::string input) const {
```
jblomer, line 66: Can we use the whitelists from the cvmfs sanitizer here?

jblomer, line 66: --> `Escape`

jblomer: I guess not, never mind

cvmfs/receiver/receiver.cc
```diff
@@ -124,11 +125,25 @@ int main(int argc, char** argv) {
 
   receiver::Reactor reactor(fdin, fdout);
 
-  if (!reactor.Run()) {
+  try {
+    if (!reactor.Run()) {
+      LogCvmfs(kLogReceiver, kLogSyslogErr,
+               "Error running CVMFS Receiver event loop");
+      delete watchdog;
```
jblomer, line 132: Let's use a `UniquePtr` for `watchdog`

siscia: Great idea!

cvmfs/json_document_write.h
siscia: I am not sure I get this comment, but we should be good now...

jblomer: For consistency, the function name should start with a capital `E`.

jblomer: Eh, wrong reply

jblomer:
Something's still not quite right in the unit tests

siscia:
Hummm, I see. Weird though... It was working, let me understand what is going on...

siscia:
Awwww, I see the test was added here: https://github.com/cvmfs/cvmfs/pull/2563/commits/c109fd8c873e425cb53146dbbe7ca199f9930791

Glad we caught it!

siscia:
Hummm it is not an easy fix.

We need to escape strings because the stacktrace will contains characters that need escaping, however, when we create nested object, we just pass a string, that get escaped twice, which is wrong and generate the wrong JSON.

The option that I see at the moment are:
1. hack it around, adding another parameter to the JSON creator to decide if escape or not the strings. 
2. Create a hierarchical structure to manage complex JSON 
3. Include some external library that generate JSON for us

I want to explore how complex is 2 first, but maybe we can just go for 1.

siscia:
Continue on #2580

PR #2554, opened 2020-05-04
Create container integration tests

jblomer:
Adds a new EL8 based platform to test various container runtime integration scenarios with a new suite of tests.  This should serve both as a regular integration test suite and as a documentation on how to use cvmfs with containers.


jblomer:

PR #2555, opened 2020-05-05
Compile libcvmfs.a with -fPIC option

janpriessnitz:
Compile libcvmfs with -fPIC option (position independent code)
to enable users to build shared libraries on top of it.

libcvmfs.a needs to be compiled with -fPIC
(position independent code)


janpriessnitz:
@cernvm-bot cloudtest cc7

PR #2557, opened 2020-05-07
Fix generation of JSON snippets

siscia:
The gateway tests showed a small mistake on the statistics manager. It didn't handle correctly the case of no statistics while generating the JSON.

The JSON was generated by hand, but we do have an interface to generate valid JSON, hence I decided to let the statistics use the already present interface.

In doing so, however I would have introduced some dependencies (`util/pointer.cc` and `json_document.cc`) to  all the code that use the statistic class.

So I refactor our own JSON generation in a separate, header only, file.

I transition from a struct + function to a simple class, and rename in what I believe is more appropriate.

Also I modified the interface in something more ergonomic.

In modifying the interface, however, I broke the `receiver/reactor.h` interface. It was relying on raw pointer and checking them. I am not sure I understand why.

Hence I also re-work the interface of the reactor in something that seems better.

Overall, this PR fix a small error, add a test, make some interface more ergonomics in -7 lines of code! Yeaaah! :)

@janpriessnitz does it LGTY? Are the JSON code equivalent?

@jblomer I am making some mistake in refactoring the reactor to accept a reference to a string instead of a raw pointer? I really don't understand why that choice was made.

### DIFF
--- a/cvmfs/json_document_write.h
+++ b/cvmfs/json_document_write.h
@@ -23,20 +23,12 @@ class JsonStringGenerator {
   };
 
  public:
-  void PushBackQuoted(std::string key, std::string val) {
-    this->PushBack(key, val, true);
+  void AddQuoted(std::string key, std::string val) {
+    this->Add(key, val, true);
   }
 
-  void PushBackUnquoted(std::string key, std::string val) {
-    this->PushBack(key, val, false);
-  }
-
-  void PushBack(std::string key, std::string val, bool quoted = true) {
-    JsonStringEntry entry;
-    entry.key = key;
-    entry.val = val;
-    entry.is_quoted = quoted;
-    entries.push_back(entry);
+  void AddUnquoted(std::string key, std::string val) {
+    this->Add(key, val, false);
   }
 
   std::string GenerateString() const {
@@ -59,6 +51,14 @@ class JsonStringGenerator {
   }
 
  private:
+  void Add(std::string key, std::string val, bool quoted = true) {
+    JsonStringEntry entry;
+    entry.key = key;
+    entry.val = val;
+    entry.is_quoted = quoted;
+    entries.push_back(entry);
+  }
+
   std::vector<JsonStringEntry> entries;
 };
 
--- a/cvmfs/json_document_write.h
+++ b/cvmfs/json_document_write.h
@@ -1,0 +1,55 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#ifndef CVMFS_JSON_DOCUMENT_WRITE_H_
+#define CVMFS_JSON_DOCUMENT_WRITE_H_
+
+// This class is used for marshalling JSON objects to strings.
+// When adding an object, use quoted = true for strings and
+// quoted = false for numbers, nested objects, etc.
+
+class JsonStringGenerator {
+  struct JsonStringEntry {
+    std::string key;
+    std::string val;
+    bool quoted;
+
+    JsonStringEntry() {
+      quoted = true;
+    }
+  };
+
+ public:
+  void PushBack(std::string key, std::string val, bool quoted = true) {
+    JsonStringEntry entry;
+    entry.key = key;
+    entry.val = val;
+    entry.quoted = quoted;
+    entries.push_back(entry);
+  }
+
+  std::string GenerateString() const {
+    std::string output;
+
+    output += "{";
+    for (size_t i = 0u; i < this->entries.size(); ++i) {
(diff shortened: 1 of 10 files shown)


siscia:
Aw I see, we want to follow the google standard...

No line saving then :(

cvmfs/statistics.cc
```diff
@@ -90,19 +90,18 @@ string Statistics::PrintList(const PrintOptions print_options) {
...
+  JsonStringGenerator json_generator;
   for (map<string, CounterInfo *>::const_iterator i = counters_.begin(),
-       iEnd = counters_.end(); i != iEnd; ++i)
-  {
-    result +=
-      std::string("\"") + i->first + "\":\"" +
-      i->second->counter.ToString() + "\",";
+                                                  iEnd = counters_.end();
+       i != iEnd; ++i) {
+    json_generator.PushBack(i->first, i->second->counter.ToString());
```
janpriessnitz, line 101: You need to PushBack with quoted = false, since they need to be interpreted as numbers by the gateway.

cvmfs/receiver/reactor.cc
```diff
@@ -217,13 +217,13 @@ bool Reactor::HandleGenerateToken(const std::string& req, std::string* reply) {
...
-  JsonStringInput input;
+  JsonStringGenerator input;
   input.PushBack("token", session_token.c_str());
   input.PushBack("id", public_token_id.c_str());
   input.PushBack("secret", token_secret.c_str());
+  std::string json = input.GenerateString();
 
-  ToJsonString(input, reply);
-
+  reply = &json;
```
janpriessnitz, line 226: If you want to fill the `reply` pointer with data for later use outside the function, I think you need to do `*reply = json`. This way, the `HandleCheckToken` won't modify the `reply` outside of the function. 5x times in the code.

siscia: I am fixing it now. It was actually something like assign(), I believe...

cvmfs/statistics.cc
siscia: Thanks!

cvmfs/receiver/reactor.cc
janpriessnitz: Just `=` would surely be enough, but assign is good as well.

siscia: Yes, you are right!

I thought I tried it... Oh well...

janpriessnitz: Interesting! https://stackoverflow.com/questions/34196053/stdstringassign-vs-stdstringoperator

cvmfs/json_document_write.h
```diff
@@ -0,0 +1,58 @@
...
+    std::string val;
+    bool quoted;
+
+    JsonStringEntry() {
+      quoted = true;
+    }
+  };
+
+ public:
+  void PushBack(std::string key, std::string val, bool quoted = true) {
```
jblomer, line 27: I think I would prefer two public interfaces: `AddQuoted` and `AddUnquoted`, otherwise one always needs to lookup what the `true` or `false` actually means.

test/unittests/t_json.cc
```diff
@@ -63,14 +64,13 @@ TEST(T_Json, SearchInObject) {
   EXPECT_EQ(NULL, result);
 }
 
-TEST(T_Json, ToJsonString) {
-  JsonStringInput input;
+TEST(T_Json, GenerateValidJsonString) {
+  JsonStringGenerator input;
   input.PushBack("f1", "v1");
   input.PushBack("f2", "v2");
   input.PushBack("f3", "v3");
```
jblomer, line 71: Let's add one unquoted entry, too

cvmfs/json_document_write.h
siscia: Hummm honestly I didn't write this code. Changing this will require to change all the invocation.

If we really want to go that route, the I would prefer to have a PushBack overloaded, one for string, one for int, one for float, one for boolean and one for null, leaving the management of array and object to the caller.
Anyway, this already grow to much so I will just push back for string and numbers, leaving the implementation of bools and null to when it will be necessary.

siscia: Nah this is a bad idea within our constraints.

siscia:
I don't want to take any credit, I just moved this struct. It is not my code!

siscia:
@cernvm-bot unittest

siscia:
@cernvm-bot unittest

siscia:
@cernvm-bot unittest

siscia:
@jblomer I believe this is ready :)

PR #2559, opened 2020-05-11
Fix reading of an item from a repository

DrDaveD:
Fix variable name bug that caused .cvmfs_status.json updates to not preserve old entries.  [CVM-1887](https://sft.its.cern.ch/jira/browse/CVM-1887).


jblomer:
@cernvm-bot cloudtest

PR #2561, opened 2020-05-17
Enable GCC 10 to pass version check

beojan:
The version check wouldn't pass with GCC 10 because the `.*` necessary when using `gcc -v` eats the 1. By using `-dumpfullversion`, the `.*`s can be avoided.

### DIFF
--- a/cmake/Modules/cvmfs_compiler.cmake
+++ b/cmake/Modules/cvmfs_compiler.cmake
@@ -38,12 +38,12 @@ set (CVMFS_OPT_FLAGS "-Os")
 if (CMAKE_COMPILER_IS_GNUCC)
   message (STATUS "checking gcc version...")
   execute_process (
-    COMMAND ${CMAKE_C_COMPILER} -dumpfullversion
+    COMMAND ${CMAKE_C_COMPILER} --version
     OUTPUT_VARIABLE CVMFS_GCC_VERSION
     ERROR_VARIABLE  CVMFS_GCC_VERSION
   )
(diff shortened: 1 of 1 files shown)


beojan:
It turns out GCC 4 doesn't have `-dumpversion`, so this uses `--version` now.

jblomer:
@cernvm-bot unittest

jblomer:
Thanks a lot for the fix!

Something must still go wrong on certain GCC compilers, e.g. on EL7

    CMake Error at cmake/Modules/cvmfs_compiler.cmake:47 (if):
      if had incorrect arguments: ${CVMFS_GCC_MAJOR} LESS 4 (Unknown arguments
      specified).
    Call Stack (most recent call first):
      CMakeLists.txt:145 (include)
    
    
    CMake Error at cmake/Modules/cvmfs_compiler.cmake:48 (message):
      GCC < 4.1 unsupported
    Call Stack (most recent call first):
      CMakeLists.txt:145 (include)

beojan:
So it turns out Ubuntu and SuSE fiddle with the output of `gcc -version` to insert the distro name in the parentheses.

`CMAKE_C_COMPILER_VERSION` is available back to at least [CMake 2.8.10](https://cmake.org/cmake/help/v2.8.10/cmake.html#variable:CMAKE_LANG_COMPILER_VERSION), which is older than what's in SLC6. Is there any reason cvmfs can't use that?

jblomer:
@beojan The reason is SLES11, our oldest supported platform, which is on an even older cmake.

beojan:
Let's try this. I've changed the regex so it should accomodate SuSE and Ubuntu.

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

jblomer:
The test failures are unrelated, merging

PR #2562, opened 2020-05-19
S3Fanout: set correct Content-Type when uploading HTML files

janpriessnitz:


janpriessnitz:
I tested this locally using `cvmfs_swissknife upload` and everything seems to be working fine.

cvmfs/s3fanout.cc
```diff
@@ -773,7 +773,12 @@ string S3FanoutManager::GetContentType(const JobInfo &info) const {
     case JobInfo::kReqDelete:
       return "";
     case JobInfo::kReqPutCas:
-      return "application/octet-stream";
+    // Set correct content type for HTML files.
```
jblomer, line 776: I think I'd prefer adding another constant `kPutReqHtml` so that we don't need to look into the object details at this point.

janpriessnitz: Yeah, that's much better. Thanks.

PR #2566, opened 2020-05-26
add watchdog to the swissknife sync

siscia:


siscia:
@cernvm-bot cloudtest

cvmfs/monitor.cc
```diff
@@ -36,7 +36,7 @@
 #include <string>
 #include <vector>
 
-#ifndef CVMFS_LIBCVMFS
+#if !(defined(CVMFS_LIBCVMFS) || defined(CVMFS_SWISSKNIFE))
```
jblomer, line 39: I think we should use that expression to define a new name, e.g.

    #if !(defined(CVMFS_LIBCVMFS) || defined(CVMFS_SWISSKNIFE))
    #define CVMFS_FUSE_MODULE
    #endif 

and subsequently only use the `CVMFS_FUSE_MODULE`.  Or perhaps even better, let the build system single out the fuse module rather than everything else but the fuse module.

cvmfs/swissknife_sync.cc
```diff
@@ -559,6 +560,17 @@ bool swissknife::CommandSync::ReadFileChunkingArgs(
 int swissknife::CommandSync::Main(const swissknife::ArgumentList &args) {
   string start_time = GetGMTimestamp();
 
+  // Spawn monitoring process (watchdog)
+  char watchdog_path[128];
```
jblomer, line 564: Let's be generous and take `PATH_LIMIT` number of characters

cvmfs/swissknife_sync.cc
```diff
@@ -559,6 +560,17 @@ bool swissknife::CommandSync::ReadFileChunkingArgs(
...
+  std::string timestamp = GetGMTimestamp("%Y.%m.%d-%H.%M.%S");
+  int path_size = snprintf(watchdog_path, sizeof(watchdog_path),
```
jblomer, line 566: We should also create `watchdog_path` and make sure it's writable (otherwise crash right away).

siscia: Hummm, we should not create the file, otherwise we will create a lot of files every time we use this code path.

We should create the file, iff some crash happens, not always. Am I wrong?

If we agree on that, then, it should be responsibility of the Watchdog to create the file, not of us invoking the watchdog.

Maybe, a sensible solution would be to check if the watchdog can create files and write in the stacktrace directory.

We can add it, but I am not sure what happens if the watchdog cannot write in the directory, do we crash? I am afraid we might break some installation.

I am going ahead and just create the directory that will contains the stacktrace file, it seems the most sensible solution.

siscia:
I honestly can't reproduce the failure, it seems completely unrelated to the changes in this PR, but it is persistent.

cvmfs/swissknife_sync.cc
jblomer: Looks good to me know!  I think the important part was creating the parent path, which is now done.

jblomer:
I agree, failures are unrelated.  It's in the new publish code.

jblomer:
@cernvm-bot cloudtest

siscia:
@cernvm-bot cloudtest

siscia:
@cernvm-bot cloudtest

siscia:
I am thinking what it would mean to use a temp file for the stacktrace, and it is not as trivial as it seems.

The approach I considered are:
1. Create a temp file, I need to delete those files if not needed (most of the times) and I would rely on some RAII, that eventually will not work when we switch to exceptions.
2. Create a temp file, save the path, delete immediately the file. I don't like the idea of making the system call only to get the path of the file and immediately delete it. But it is a strong candidate.
3. Write on `os.getenv("TMP")` some configuration won't have the variable set. Especially weird one that run inside jenkins or other CI
4. Just write on `/tmp` it is the less correct option, but the most pragmatic. Everybody understand it and it will work on most systems and it is simple to understand also for operators.

At the moment I am going for 4. eventually it will be simple to change for 2.

jblomer:
I agree, let's go for 4 for the time being.

siscia:
@cernvm-bot cloudtest

siscia:
@cernvm-bot cloudtest

siscia:
@cernvm-bot unittests

siscia:
@cernvm-bot unittest

cvmfs/swissknife_sync.cc
```diff
@@ -559,6 +560,18 @@ bool swissknife::CommandSync::ReadFileChunkingArgs(
...
+  std::string watchdog_dir = "/tmp";
+  char watchdog_path[PATH_MAX];
+  std::string timestamp = GetGMTimestamp("%Y.%m.%d-%H.%M.%S");
+  int path_size = snprintf(watchdog_path, sizeof(watchdog_path),
+                           "%s/cvmfs-swissknife-sync-stacktrace.%s.%d",
+                           watchdog_dir.c_str(), timestamp.c_str(), getpid());
+  assert(path_size > 0);
+  assert(path_size < PATH_MAX);
+  Watchdog *watchdog = Watchdog::Create(std::string(watchdog_path));
+  watchdog->Spawn();
```
jblomer, line 573: I think we need to use a unique pointer here so that the watchdog gets deleted at the end.  Otherwise all terminations, even successful ones, will raise an alarm.

PR #2567, opened 2020-05-26
Introduce template transactions

jblomer:
Template transactions clone a directory tree in the process of opening the transaction.  They are foreseen to speed-up the generation of chain id directories in DUCC.

Still to be done:
- [x] Stratum 0 layout migration
- [x] Set UID and GID of `cvmfs_talk` socket
- [x] Unit test for `SendTalkCommand()`
- [x] Unit test for template transaction (to be completed later)
- [x] Split `template` parameter to `cvmfs_server transaction`


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest full

DrDaveD:
Could you please give an example of how to use this?  It's not obvious from the code how to do that.

Regarding the use of colon as a delimeter between source and destination paths, I think that's not a good idea because docker uses colon to distinguish image names from version numbers, and both unpacked.cern.ch and singularity.opensciencegrid.org as a result has colon in many container image names.  I suggest perhaps comma, or semi-colon if we don't mind asking people to quote it from the shell.

jblomer:
The usage is a bit [hidden in the integration test](https://github.com/cvmfs/cvmfs/pull/2567/files#diff-a1c97dfdc6f2c431af48281b0c075b86R22), it works with 

    cvmfs_server transaction -T /src:/dest

where /src and /dest are relative to the repository root.  I agree, the parameter mangling into a single string is problematic.  I think we should just turn it into two parameters, `--template-from` and `--template-to` or so.

DrDaveD:
It's telling me that cvmfs_server transaction -T is an illegal option.  I do not see its implementation in the files; there are no changes to cvmfs_server_transaction.sh.  Maybe you forgot to commit a file?

jblomer:
Oh, then I don't know how the integration test could pass. The `cvmfs_server transaction` command is already migrated to the new code base and simply redirects to the transitional `cvmfs_server publish $@`. The `cvmfs_server publish` utility [should have received support for the `-T` parameter](https://github.com/cvmfs/cvmfs/pull/2567/files#diff-ca2c55fae59806069133c32f83f4ea51R47).

DrDaveD:
Oh, I'm sorry.  I had installed the cvmfs client compiled from the branch but forgot to install the cvmfs-server.  I tried it now with a centos7 container and it was blazing fast.  I want to try it also on an atlas container.

I noticed that there's no description of the -T in the cvmfs_server transaction help.

It is convenient to be able to have one short one-letter parameter.  Maybe '=' would be a good delimeter.  It's kind of meaningful too, since you're making the two paths equal, although in programming languages it's the left hand side that's made equal to the right.  '->' would be more meaningful, but it would have to be quoted from the shell so that would be annoying.

DrDaveD:
I noticed that if you do the upgrade and try to do a cvmfs_server transaction before you do cvmfs_server migrate, it fails in a very unfriendly way.  It crashes with a stack trace and leaves a /var/spool/cvmfs/<repo>/in_transaction.lock file that has to be manually removed.

jblomer:
Great idea! So it's now `-T /from-dir=/to-dir`.  Alternatively, one can use `--template-from` and `--template-to`.  The help text now shows the `-T` parameter.

DrDaveD:
I tested with a copy of /cvmfs/unpacked.cern.ch/registry.hub.docker.com/atlas/athena:21.0.77/ and on my little 1-core VM with 4GB RAM it took close to 20 seconds.  That's kind of long but hopefully with a bigger machine it will be reasonably fast.  At least publishing only has to deal with new files, which is a huge savings; I didn't time it exactly, but it took several minutes to publish the repository the first time.

jblomer:
The layout revision on the publisher node should now be properly checked.

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2568, opened 2020-06-01
Fix chksetup for squashed NFS alien cache

jblomer:


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest cc7

PR #2569, opened 2020-06-02
Add option to create pass-through replicas

jblomer:
The new `cvmfs_server replica -P` command to create a proxy passed replica.  Such a replica will be marked by `CVMFS_PASSTHROUGH=true` in the replica.conf file and by the `pass-through` boolean flag in the repositories.json


jblomer:
@cernvm-bot cloudtest cc7

PR #2571, opened 2020-06-04
RPM packaging: fix selinux policy files

janpriessnitz:
files_type() macro seems to be broken.

This commit rewrites the files_type() macro
in cvmfs.te into the expanded explicit form.
The commit should not change the original meaning
of the macro.


janpriessnitz:
@cernvm-bot help

janpriessnitz:
@jblomer This should fix the issue, although it is a bit dirty. Packages built from this PR are here http://ecsft.cern.ch/dist/cvmfs/nightlies/cvmfs-git-2289/ (if you want to test it yourself)

jblomer:
@traylenator Do you have an opinion on this cvmfs packaging change?

traylenator:
it's bit dirty as you say.

I was just looking at what the new kafs packages do which is presumably the same.

https://src.fedoraproject.org/rpms/kafs-client/blob/master/f/kafs-client.spec

requires:

https://src.fedoraproject.org/rpms/filesystem/blob/master/f/filesystem.spec#_29

traylenator:
Is this just so you can do chmod ? If so
```
%attr(555,root,root) /afs
```
Think I don't understand the problem.

janpriessnitz:
@traylenator The problem is that the RPM selinux plugin wants to set some attributes on every file listed in the '%files' section, even just on package upgrade. With autofs mounted on '/cvmfs', this is not possible and results in error. I'm currently trying to find out how to disable this plugin hook for '/cvmfs'.

janpriessnitz:
@jblomer I think having a separate package, let alone a subpackage in the 'filesystem' package, is not a good option for us. The hook which causes problems is 'fsm_file_prepare'. Here it says it's run on every file in the '%files' section: https://rpm.org/devel_doc/plugins.html . That means '/cvmfs' cannot be in the '%files' section. I think the current dirty fix is the best we can do.

PR #2572, opened 2020-06-08
GetSingleAlias returns only the alias.

siscia:
GetSingleAlias use to return the whole path to directory, with the
last element containing the alias we are after.


siscia:
@cernvm-bot cloudtest

cvmfs/publish/settings.cc
```diff
@@ -268,7 +268,8 @@ std::string SettingsBuilder::GetSingleAlias() {
     throw EPublish("no repositories available in " + config_path_);
   if (repositories.size() > 1)
     throw EPublish("multiple repositories available in " + config_path_);
-  return repositories[0];
+  // the + 1 remove the / separator
+  return repositories[0].substr(config_path_.size() + 1, std::string::npos);
```
jblomer, line 272: Let's use `GetFileName` from posix.cc here

PR #2573, opened 2020-06-11
Remove perl dependency

jblomer:


jblomer:
@cernvm-bot cloudtest cc7

jblomer:
@cernvm-bot cloudtest

PR #2575, opened 2020-06-18
Manage case when we don't find the server.conf configuratio file.

siscia:
I tried to address the issue of that Valetin brought up today.

I am definitely not sure if this is enough.

### DIFF
--- a/cvmfs/publish/cmd_transaction.cc
+++ b/cvmfs/publish/cmd_transaction.cc
@@ -36,18 +36,16 @@ int CmdTransaction::Main(const Options &options) {
   }
 
   SettingsBuilder builder;
-  const std::string repo = fqrn.empty() ? builder.GetSingleAlias() : fqrn;
   SettingsPublisher* settings;
   try {
-    SettingsPublisher s = builder.CreateSettingsPublisher(repo, true /* needs_managed */);
+    SettingsPublisher s =
+        builder.CreateSettingsPublisher(fqrn, true /* needs_managed */);
     settings = &s;
-  } catch (const EPublish &e) {
-    if (e.failure() == EPublish::kFailRepositoryNotFound) {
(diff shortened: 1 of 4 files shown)


jblomer:
LGTM

siscia:
Still looks quite ugly...

```
$ cvmfs_server transaction bar.ch
(unexpected termination) Unable to find the configuration file `server.conf` for the cvmfs publisher

Stacktrace:
cvmfs_publish(+0x379c7) [0x563ae04049c7]
cvmfs_publish(+0x307a3) [0x563ae03fd7a3]
cvmfs_publish(+0x39e81) [0x563ae0406e81]
cvmfs_publish(+0x348a9) [0x563ae04018a9]
cvmfs_publish(+0x124dc) [0x563ae03df4dc]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7) [0x7ff9975eab97]
cvmfs_publish(_start+0x2a) [0x563ae03df5da]
```

jblomer:
Already better than before :)
For a nicer output, we should give this issue a well-defined error code and catch the exception in the transaction command. Like here: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/publish/cmd_transaction.cc#L77

siscia:
I don't like this at all. 

We are coupling how we get the repository name in `CreateSettingsPublisher` and in `CmdTransaction` that feels very wrong. Moreover we are making the  method `GetSingleAlias` public. That feels somehow wrong.

Maybe it is better to leave just the thrown of the exceptions.

siscia:
Alright, I like this version much better.

The methods are private again, and everything seems to flow quite nicely.

However it costed us the fact that I started a hierarchy of errors, and we kinda agree that it is not ideal. However in this case it seems as the most sensible strategy... Thoughts?

siscia:
@jblomer I would have liked to return the UniquePtr directly, but I wasn't able to do so.

PR #2576, opened 2020-06-19
CMake build system updates

amadio:
These are some patches I had to apply in Gentoo to build against external versions of the builtins.
I think that most are fine to be used upstream. Please let me know. More information can be found at https://packages.gentoo.org/packages/net-fs/cvmfs.


jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

amadio:
I will have a look. You may want to have a different fix for GCC 10. I just removed the checks because they consider 10 < 4, and then configuration fails because of that. See https://bugs.gentoo.org/709696 for more details. I can also try to break the changes into smaller pieces to test them individually if needed.

amadio:
BTW, I see warnings from newer versions of protobuf about deprecated things being used in CVMFS. You may want to update your builtin to check. I'm using protobuf 3.11.4.

amadio:
The Debian builds fail because of checks I removed for GCC 10. I will drop that patch from this PR.

amadio:
@jblomer Can you please restart the tests? (Or add me to the whitelist so I can try to fix what shows up)

jblomer:
@cernvm-bot cloudtest

jblomer:
@amadio Strangely, building still fail on all of the platforms.  Perhaps this error message from one of the platforms helps:

     *** No rule to make target 'externals/sha3/src/libcvmfs_sha3.a', needed by 'cvmfs/cvmfs_fsck'.  Stop.

amadio:
Yes, it works for me. It's probably a problem with older CMake versions. I probably need to remove the `$<TARGET_FILE:>` from `${SHA3_LIBRARIES}`.

amadio:
@cernvm-bot cloudtest

amadio:
Let's try again. CMake is a pain to support when developing with newer versions to support old ones... If this doesn't work still I may install an old version of CMake to test or test in a container.

jblomer:
@cernvm-bot cloudtest

amadio:
I also see problems like this
```
/usr/bin/cc -fno-strict-aliasing -g -O2 -fdebug-prefix-map=/home/sftnight/jenkins/workspace/CvmfsFullBuildDocker/CVMFS_BUILD_ARCH/docker-x86_64/CVMFS_BUILD_PLATFORM/ubuntu1804/build/wd_src=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include/python2.7 -c maxminddb/extension/maxminddb.c -o build/temp.linux-x86_64-2.7/maxminddb/extension/maxminddb.o -Wall -Wextra -fno-strict-aliasing
maxminddb/extension/maxminddb.c:2:10: fatal error: maxminddb.h: No such file or directory
 #include <maxminddb.h>
          ^~~~~~~~~~~~~
```
which should not have been created by my changes.

jblomer:
That message is present in all the builds and correctly handled by the build system.

amadio:
I will try on a container first. I understand why it fails, but I need to run with the same CMake version to be able to fix. I may also split the changes to make it easier to test.

amadio:
Alright, this new version was tested with CMake 2.8.12.2. Should work in CI now.

siscia:
@cernvm-bot unittest

amadio:
@siscia Can you try the cloudtest as well? Thanks!

siscia:
@cernvm-bot cloudtest

siscia:
Sure!

amadio:
CMake sucks, I don't know why it fails in the CI build now. It worked for me inside a CentOS 7 container in docker using CMake 2.8 as I mentioned above... @jblomer @siscia Is there an easy way to run the same docker build locally?

siscia:
Yes, we build it in dcoker containers.

We use the images from here: https://gitlab.cern.ch/cernvm/build-images/container_registry

amadio:
More important than the image is the configuration command to build. I tried to run the same script, but I cannot find it and I cannot see the Jenkins configuration. I am trying to build by copying the cmake configuration line from the build output. Let's see if that works.

amadio:
I think I tracked down the failure. I was missing adding back dependencies to the custom_command after removing the libraries sha3 and vjson from the variables. Let's try one more time. Thanks for the patience!

siscia:
@cernvm-bot cloudtest

amadio:
@jblomer I believe the last build failure is not my fault anymore. Could you please take a look at the changes and let me know what you think?

externals/sha2/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+project(SHA2)
```
jblomer, line 1: Actually, I think we don't need that dependency at all anymore.  We should just get rid of it.

externals/sha3/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+project(SHA3)
```
jblomer, line 1: The `configureHook` and `makeHook` from the src subdirectory should be removed

externals/vjson/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+project(VJSON)
```
jblomer, line 1: The `configureHook` and `makeHook` from the src subdirectory should be removed

bootstrap.sh
```diff
@@ -207,19 +207,6 @@ build_lib() {
       do_copy "sqlite3"
       do_build "sqlite3"
       ;;
-    vjson)
-      do_copy "vjson"
-      patch_external "vjson"       "missing_include.patch"
```
jblomer, line 212: I think the patch is now missing

CMakeLists.txt
```diff
@@ -314,34 +314,28 @@ endif (NOT MACOSX)
...
+  set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${GTEST_INCLUDE_DIRS})
 
-find_package (GMock REQUIRED)
-set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${GMOCK_INCLUDE_DIRS})
+  find_package (GMock REQUIRED)
+  set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${GMOCK_INCLUDE_DIRS})
+endif(BUILD_TESTING OR BUILD_UNITTESTS OR BUILD_UNITTESTS_DEBUG)
 
-find_package (VJSON REQUIRED)
+add_subdirectory(externals/vjson)
```
jblomer, line 325: I guess that the headers and static libraries are not ending up anymore in the externals directory.  Where will they end up?

externals/sha3/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+
+add_library(cvmfs_sha3 STATIC
```
jblomer, line 3: Is the 32bit/64bit header selection from `makeHook` taken care of?

externals/sha3/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+  src/64opt/KeccakF-1600-opt64.c
+  src/KeccakHash.c
+  src/KeccakSponge.c
+  src/SnP-FBWL-default.c
+  )
+
+set_property(TARGET cvmfs_sha3 PROPERTY POSITION_INDEPENDENT_CODE TRUE)
```
jblomer, line 10: In the current build process, we take the upstream Makefile and add the cvmfs compiler flags.  Do we here compile with equivalent options?

externals/sha3/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+
+target_include_directories(cvmfs_sha3 PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}/src
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/64opt
+  )
+
+set(SHA3_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/src ${CMAKE_CURRENT_SOURCE_DIR}/src/64opt PARENT_SCOPE)
+set(SHA3_LIBRARY $<TARGET_FILE:cvmfs_sha3> PARENT_SCOPE)
```
jblomer, line 18: Is the library stripped?

externals/vjson/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
+
+add_library(cvmfs_vjson STATIC src/block_allocator.cpp src/json.cpp)
+set_property(TARGET cvmfs_vjson PROPERTY POSITION_INDEPENDENT_CODE TRUE)
+target_include_directories(cvmfs_vjson PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/src)
+set(VJSON_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/src PARENT_SCOPE)
+set(VJSON_LIBRARY $<TARGET_FILE:cvmfs_vjson> PARENT_SCOPE)
```
jblomer, line 7: Is the library stripped?

externals/vjson/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+project(VJSON)
+
+add_library(cvmfs_vjson STATIC src/block_allocator.cpp src/json.cpp)
+set_property(TARGET cvmfs_vjson PROPERTY POSITION_INDEPENDENT_CODE TRUE)
```
jblomer, line 4: Is this using the `CVMFS_BASE_CXX_FLAGS` options?

CMakeLists.txt
```diff
@@ -314,34 +314,28 @@ endif (NOT MACOSX)
 find_package (OpenSSL REQUIRED)
 set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${OPENSSL_INCLUDE_DIR})
 
-find_package (GTest REQUIRED)
-set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${GTEST_INCLUDE_DIRS})
+if (BUILD_TESTING OR BUILD_UNITTESTS OR BUILD_UNITTESTS_DEBUG)
```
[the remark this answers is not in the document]
amadio: Which targets do you mean? Everything compiled here, but it may be that it's just because headers are in `/usr/include`. I will uninstall GTest to make sure that `find_package(GTest)` is not needed when not compiling tests.

externals/sha3/CMakeLists.txt
amadio: Sure, will do.

externals/sha2/CMakeLists.txt
amadio: I will try to just remove this then.

externals/vjson/CMakeLists.txt
amadio: Will do.

bootstrap.sh
amadio: Yes, but it doesn't seem to be needed. Should I patch the file in place and drop the patch file?

CMakeLists.txt
amadio: They end up in `${CMAKE_BINARY_DIR}/externals/vjson` during the build (i.e. not in `${CMAKE_SOURCE_DIR}/externals_build/vjson`).

externals/sha3/CMakeLists.txt
amadio: We should, as they are built as part of the cvmfs project itself, but it's something I can check to be sure if you want.

externals/sha3/CMakeLists.txt
amadio: Good point, I forgot about 32bit. I will make an equivalent selection in the CMakeLists.txt.

externals/sha3/CMakeLists.txt
amadio: That will probably depend on the configuration. Let me know if they should/should not be stripped by force and I can make sure that this will be the case.

externals/vjson/CMakeLists.txt
amadio: I will check that it does.

externals/sha3/CMakeLists.txt
jblomer: Yes, please enforce stripping the library (we don't expect any meaningful stack trace information in this library)

bootstrap.sh
jblomer: I think that would work

externals/sha3/CMakeLists.txt
amadio: I guess setting visibility of symbols to hidden achieves what you want, right?
See https://cmake.org/cmake/help/latest/prop_tgt/LANG_VISIBILITY_PRESET.html.

amadio: Or you mean that we should strip debugging symbols?

jblomer:
@amadio Do you still have time to work on it?

amadio:
Sorry, haven't had time lately. I will probably have some time soon, though. I want to update Gentoo to the latest CVMFS and then I will have to revisit the patches.

jblomer:
@amadio Should we close this for the time being and you can open a new PR at a later point?

amadio:
Yes, let's do that. I have not had time to work on this. When I manage, I can open a new pr.

PR #2581, opened 2020-06-25
Manage case of ingest with base directory root (/).

siscia:
It seems to me that this case was not manage yet.


siscia:
@cernvm-bot cloudtest

siscia:
Do not merge yet, it still need tests to cover the new feature.

cvmfs/sync_union_tarball.cc
```diff
@@ -146,10 +146,12 @@ void SyncUnionTarball::Traverse() {
 
       case ARCHIVE_EOF: {
         if (create_catalog_on_root_) {
```
jblomer, line 148: Perhaps better `if (create_catalog_on_root_ && (base_directory_ != "/"))`

siscia:
I added a simple test and, doing so I decided to slide other integration tests.

I tried to keep related tests all together, close to each other. I am not sure I succeeded.

cvmfs/sync_union_tarball.cc
siscia: I agree!

siscia:
@cernvm-bot cloudtest

siscia:
@cernvm-bot cloudtest

siscia:
@jblomer I believe we can merge this. The failures seems unrelated.

PR #2582, opened 2020-06-25
Create additional image store for podman

Mohitty:
This is intended to create additional image store directory (in unpacked.cern.ch by ducc) to be used by podman as part of GSoC 2020 project.

The library creates the following directory structure: [podmanStore](https://github.com/Mohitty/PodmanAdditonalStore/tree/master/additionalstore_overlay)

And is described in detail [here](https://github.com/Mohitty/PodmanAdditonalStore/blob/master/README.md) and [here](https://medium.com/@mohit2501tyagi/lets-walk-through-podman-37636cf223c5)

@siscia


ducc/lib/conversion.go
```diff
@@ -213,15 +213,70 @@ func ConvertWishDocker(wish WishFriendly, convertAgain, forceDownload, createThi
...
+		var firstError error
+		for expandedImgTag := range wish.ExpandedTagImagesLayer {
+			tag := expandedImgTag.Tag
+			outputWithTag := *outputImage
+			if inputImage.TagWildcard {
+				outputWithTag.Tag = tag
+			} else {
+				outputWithTag.Tag = outputImage.Tag
+			}
+			err = convertInputOutput(expandedImgTag, outputWithTag, wish.CvmfsRepo, false, false, false, true)
```
jblomer: Long lists of `false`/`true` are usually prone to mistakes.  Can we instead pass a config struct with named parameters or at least comment the parameter names for each of the booleans?

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,502 @@
...
+
+	layerlockpath := filepath.Join(rootPath, layerMetadataDir, "layers.lock")
+	err = img.CreateLockFiles(CVMFSRepo, layerlockpath)
+	if err != nil {
+		LogE(err).Error("Unable to create layers lock file in podman store")
+		return err
+	}
+
+	return nil
+}
```
jblomer: Missing newline

ducc/lib/util.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		if ch != '.' && !(ch >= '0' && ch <= '9') && !(ch >= 'a' && ch <= 'z') {
+			break
+		}
+	}
+	if reader.Len() > 0 {
+		fname = "=" + base64.StdEncoding.EncodeToString([]byte(digest))
+	}
+	return
+}
```
jblomer: Missing newline

ducc/lib/util.go
```diff
@@ -0,0 +1,86 @@
...
+			// EPERM represents an entropy pool exhaustion, a condition under
+			// which we backoff and retry.
+			return true
+		}
+	}
+
+	return false
+}
+
+func generateConfigFileName(digest string) (fname string, err error) {
```
jblomer: Add comment

ducc/lib/util.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"syscall"
+	"time"
+	"golang.org/x/sys/unix"
+)
+
+func generateID(l int) string {
```
jblomer: Please add a function comment.  Shouldn't there be a standard library routine that generates UUIDs?

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,502 @@
...
+
+	err = CreateSymlinkIntoCVMFS(CVMFSRepo, symlinkPath, targetPath)
+	if err != nil {
+		LogE(err).Error("Error in creating the symlink for manifest.json")
+		return err
+	}
+	return nil
+}
+
+func (img Image) CreateLockFiles(CVMFSRepo, fpath string) (err error) {
```
jblomer: Why do we need lock files?  Shouldn't cvmfs transactions make sure that images are imported in an orderly fashion?

siscia:
Many thanks!

From a tactical point of view I would ask to modify the podman flags to be negative flags, like the two others.

On an higher level, do you think we can come up with a design that does not force `ConvertWishDocker` to know if we are converting podman as well?

There are other things I would like to have clarified before to merge, but I do want to get the overall big design right at first.

ducc/cmd/convert.go
```diff
@@ -28,6 +28,7 @@ func init() {
 	convertCmd.Flags().BoolVarP(&skipFlat, "skip-flat", "s", false, "do not create a flat image (compatible with singularity)")
 	convertCmd.Flags().BoolVarP(&skipLayers, "skip-layers", "d", false, "do not unpack the layers into the repository, implies --skip-thin-image")
 	convertCmd.Flags().BoolVarP(&skipThinImage, "skip-thin-image", "i", false, "do not create and push the docker thin image")
+	convertCmd.Flags().BoolVarP(&convertPodman, "convert-podman", "p", true, "also create a podman image store")
```
siscia: This flag, can it be a negative flag?

Mohitty:
Thanks @siscia , @jblomer for the comments. I will incorporate the changes.

ducc/lib/podman_store.go
Mohitty: Hi @jblomer , these lock files are meant for podman to read. Podman expects lock files in the image store directories. I will add comments for the functions. Their role will be more clearer then

Mohitty:
> Many thanks!
> 
> From a tactical point of view I would ask to modify the podman flags to be negative flags, like the two others.
> 
> On an higher level, do you think we can come up with a design that does not force `ConvertWishDocker` to know if we are converting podman as well?
> 
> There are other things I would like to have clarified before to merge, but I do want to get the overall big design right at first.

Hi @siscia , I needed to process some extra layer info while downloading it. This info is not needed for converting the wish for docker. Hence, I sent a flag to avoid extra computing time if we are not converting for podman!. 
Still, I will take a look if I can get that info after it is place in .layers dir.

ducc/lib/util.go
Mohitty: updated the function to use github.com/google/uuid to generate uuid.

Mohitty:
@jblomer @siscia patch updated!

Mohitty:
> On an higher level, do you think we can come up with a design that does not force ConvertWishDocker to know if we are converting podman as well?

@siscia done!
I separated the code for downloading layers and converting for docker.

ducc/lib/conversion.go
```diff
@@ -32,6 +32,7 @@ const (
 )
 
 var subDirInsideRepo = ".layers"
+var	layerLocations map[string]string
```
siscia: Why this one is declared here?

siscia:
@Mohitty I left some comment.

But please, do not just fix it. Let's move forward some form of design document so that we can surface problems together, it will be simpler to review the code and overall it will be faster to get this done.

siscia:
Some other questions that I didn't manage to explore through.

What happens to the image.json file when we run DUCC multiple times, maybe with different wishlist? What if the image in the repository changed and a new image is being pushed over an old tag?

Is this managed?

Mohitty:
> Some other questions that I didn't manage to explore through.
> 
> What happens to the image.json file when we run DUCC multiple times, maybe with different wishlist? What if the image in the repository changed and a new image is being pushed over an old tag?
> 
> Is this managed?

@siscia They are not covered yet. Will do that after writing the document. Also, need to improve the AlreadyConverted check with the new design.

siscia:
Thanks! I got a better feeling about this PR now.

Do you run the go formatter on the code?

ducc/lib/conversion.go
siscia: @Mohitty This need to be declared here?

ducc/lib/utils.go
```diff
@@ -1,19 +1,107 @@
...
+}
+//wrapper around bytes.Buffer which implements close methods too.
+type ReadCloserBuffer struct {
+	*bytes.Buffer
+}
+
+func (cb ReadCloserBuffer) Close() (err error) {
+	return
+} 
+
```
siscia: It is not clear to me why we need all this. There is some interface that need to be Closer? Which one?

ducc/lib/image.go
```diff
@@ -626,7 +644,7 @@ func (img *Image) GetLayers(layersChan chan<- downloadedLayer, manifestChan chan
 	}
 }
 
-func (img *Image) downloadLayer(layer da.Layer, token, rootPath string) (toSend downloadedLayer, err error) {
+func (img *Image) downloadLayer(layer da.Layer, token, rootPath string, skipPodman bool) (toSend downloadedLayer, err error) {
```
siscia: Can we be any smarter here? To **download** the layer, I need to know if the layer will be used by the podman store?

I would tend to say that, "No, to **download** the layer, I don't need to know how the layer will be used".

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,443 @@
...
+		CompressedSize: layersize,
+		UncompressedDigest: "sha256:" + diffID,
+		UncompressedSize: size,
+	}
+	podmaninfo.LayerDigestMap[digest] = diffID
+	return
+}
+
+//creates images.json and layers.json file in podman store
+func (img Image) IngestImageInfo(CVMFSRepo string) (err error) {
```
siscia: All the method of `Image` take a pointer. This allow to modify internally the `Image` structure, which, in turn, is very useful for caching, especially for the manifest.

It is a good idea to have all the method on the same structure use the same calling convention if possible/appropriate.

Mohitty:
> Do you run the go formatter on the code?

Just did. There are still few things to implement before merging. Like feature for removal of image from podman store.

siscia:
@Mohitty please let me know when I should take another look to the PR?

Mohitty:
@siscia Please take a look. I think all the features are complete now.

ducc/lib/utils.go
```diff
@@ -1,19 +1,108 @@
...
 
 // this flag is populated in the main `rootCmd` (cmd/root.go)
 var (
 	TemporaryBaseDir string
 )
 
+//function that satisfies io.Closer.
+type CloseFunc func() error
+
+type readCloser struct {
```
siscia: I don't really get why we need this

ducc/lib/utils.go
```diff
@@ -1,19 +1,108 @@
...
+	io.Reader
+	c CloseFunc
+}
+
+func (r readCloser) Close() error {
+	return r.c()
+}
+
+//wrapper around bytes.Buffer which implements close methods too.
+type ReadCloserBuffer struct {
```
siscia: Similarly, why do we need this? A simple io.ReadCloser is not enough?

ducc/lib/conversion.go
```diff
@@ -32,6 +32,7 @@ const (
 )
 
 var subDirInsideRepo = ".layers"
+var layerLocations map[string]string
```
siscia: Why this is declared here? Can we do without it?

ducc/lib/image.go
siscia: @Mohitty Can you please address this comment as well?

ducc/lib/conversion.go
siscia: @Mohitty

ducc/lib/image.go
```diff
@@ -661,8 +679,15 @@ func (img *Image) downloadLayer(layer da.Layer, token, rootPath string) (toSend
...
 			}
-
-			toSend = downloadedLayer{Name: layer.Digest, Path: gread}
+			var path io.ReadCloser
+			path = gread
+			if !skipPodman {
+				var buf bytes.Buffer
+				path = TeeReadCloser(path, &buf)
+				podmaninfo := img.GetPodmanInfo()
+				podmaninfo.LayerReaderMap[layer.Digest] = ReadCloserBuffer{&buf}
```
siscia: @Mohitty hummm, I would say that we want something different.

Look at the code, we invoke `GetPodmanInfo()` and we create (or get a previously allocated) pointer to a map. 
Then, throught that pointer, we update the map.

Wouldn't it be simpler a method like: `StoreDigestLayerMap` (awful name, so please pick something a little better) that would get the layer.Digest and the buffer, and it will do the same, but hiddend for this piece of bussines logic?

What happens if we do this operation ALWAYS, no matter if we are storing podman stuff or not? It is a problem? Do we incur in a performance penality?

ducc/lib/conversion.go
```diff
@@ -174,18 +182,181 @@ func ConvertWishSingularity(wish WishFriendly) (err error) {
...
+		LogE(err).WithFields(log.Fields{"input image": wish.InputName}).
+			Error("Null image, should not happen")
+		return
+	}
+
+	manifest, err := inputImage.GetManifest()
+	if err != nil {
+		return
+	}
+	imageID := strings.Split(manifest.Config.Digest, ":")[1]
```
siscia: We introduced a `calculateID` function, didn't we? Either use it, or remove it altogether.

ducc/lib/image.go
siscia: Also, the use of TeeReadCloser against a bytes.Buffer implies that we keep everything in memory. Unfortunately I am not sure it is a suitable solution. We might have quite large layers... Like 1GB layers...

siscia: For instance, consider this snippet: https://play.golang.org/p/YrPv33KRqgi

ducc/lib/utils.go
```diff
@@ -1,19 +1,108 @@
...
+}
+
+//Trim the digest algorithm name from digest
+func calculateId(digest string) string {
+	return strings.Split(digest, ":")[1]
+}
+
+//generates the file name for link dir in podman store
+func generateID(l int) string {
+	id := uuid.New()
```
siscia: Can we do the same without an external dependency?

ducc/lib/utils.go
```diff
@@ -1,19 +1,108 @@
...
+		}
+	}
+	if reader.Len() > 0 {
+		fname = "=" + base64.StdEncoding.EncodeToString([]byte(digest))
+	}
+	return
+}
+
+//writes data to file and ingest in cvmfs repo path
+func writeDataToCvmfs(CVMFSRepo, path string, data []byte) (err error) {
```
siscia: This function is not more appropriate in cvmfs.go?

It will make simpler to split the huge `lib` package in a smaller `cvmfs` package eventually.

ducc/lib/utils.go
```diff
@@ -1,19 +1,108 @@
...
+	tmpFile, err := ioutil.TempFile("", "write_data")
```
siscia: Unfortunately not all systems have a big and ready to use /tmp directory.

We create above the function `UserDefinedTempDir` just for this case.

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,501 @@
...
+		LogE(err).Warn("Error in getting the image manifest")
+		return err
+	}
+	podmaninfo := img.GetPodmanInfo()
+	for _, layer := range manifest.Layers {
+		layerdir := podmaninfo.LayerDigestMap[layer.Digest]
+
+		//generate the link id
+		lid := generateID(26)
+		podmaninfo.LayerIdMap[layer.Digest] = "l/" + lid
```
siscia: `filepath.Join` maybe?

siscia:
@Mohitty I left several comments. Can you seem them? Let me know if something is unclear.

Mohitty:
> @Mohitty I left several comments. Can you seem them? Let me know if something is unclear.

@siscia No, I don't see any.

siscia:
@Mohitty which explains why you were not addressing those! Sorry my bad.

Can you seem them now?

Mohitty:
@siscia  yeah visible now!

ducc/lib/conversion.go
siscia: Also maybe `calculateID` is a bad naming... it is not really calculating anything...

ducc/lib/image.go
Mohitty: @siscia Now that we store relevant information in image itself, we can download layers independently of podman. We can just store the podman related info in all cases. I'll update it.

ducc/lib/utils.go
Mohitty: I will take a look again. I don't think this is the best way. I will try to change the implementation a bit. Will update accordingly

ducc/lib/conversion.go
Mohitty: Yes, we can now. Will update!

ducc/lib/image.go
Mohitty: > What happens if we do this operation ALWAYS, no matter if we are storing podman stuff or not? It is a problem? Do we incur in a performance penality?

We can store this information always. This won't cause much performance penalty. Infact, with this we can download layers and convert for docker independently of podman. Will incorporate this.
How do you suggest using TeeReader?

ducc/lib/utils.go
Mohitty: There is no implementation of this in standard lib. We can generate one using rand.Reader.
Like here: https://play.golang.org/p/4FkNSiUDMg
But isn't using a library better?

Mohitty: Well the implementation in github.com/google/uuid is also similar to above code snippet. Since we don't use any other library feature, I think we can go ahead without it!

siscia: I don't believe we need a whole UUID, but just some reasonably random characters. If we can avoid pulling in another library it would be better.

ducc/lib/image.go
siscia: Do we need a TeeReader at all?

If I understand the code correctly, we are using the TeeReader to create two stream of data, one is serialized into cvmfs and ingested, the other is used to compute the size and the hash. Am I wrong?

If we just override the Read method (or better, implement), so that it keep track of the size and compute the hash, one piece at the time we can do without the TeeReader.
Basically we are re-implementing it, but we manage only one buffer.

Definitely open to more suggestions on this. We could TeeReader inside the hash function for instance. But then we need some way to keep track of the size...

ducc/lib/utils.go
Mohitty: Done.

ducc/lib/utils.go
Mohitty: Done

ducc/lib/conversion.go
Mohitty: Done

ducc/lib/image.go
Mohitty: We can use TeeReader to directly write to hash.Hash . We can add functionality to ReadAndHash for calculating the size as well. Something like this: https://play.golang.org/p/sZ-cHoWm2fr

siscia: Love it! Go for it!

ducc/lib/podman_store.go
Mohitty: Done

ducc/lib/utils.go
Mohitty: Done

ducc/lib/image.go
Mohitty: Done

ducc/lib/conversion.go
Mohitty: Done

ducc/lib/utils.go
Mohitty: Updated.

ducc/lib/utils.go
Mohitty: Removed this.

ducc/lib/image.go
Mohitty: Done

ducc/lib/utils.go
Mohitty: Updated!

ducc/lib/utils.go
```diff
@@ -11,9 +18,77 @@ var (
...
+
+func NewReadAndHash(r io.ReadCloser) *ReadAndHash {
+	return &ReadAndHash{r: r, hash: sha256.New()}
+}
+
+func (rh *ReadAndHash) Read(b []byte) (n int, err error) {
+	reader := io.TeeReader(rh.r, rh.hash)
+	tr := io.TeeReader(reader, rh)
+	return tr.Read(b)
+}
```
siscia: @Mohitty  For each .Read() we are creating a couple of `TeeReader`s. Maybe it is worth to create them once in the constructors.

ducc/lib/utils.go
```diff
@@ -11,9 +18,77 @@ var (
...
+
+func (rh *ReadAndHash) Write(p []byte) (int, error) {
+	n := len(p)
+	rh.size += int64(n)
+	return n, nil
+}
```
siscia, line 44: Very interesting approach! :) :+1:

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,499 @@
...
+	//layerMetadataDir contains the metadata of layers
+	layerMetadataDir = "overlay-layers"
+)
+
+//Computes uncompressed digest and size for the given layer
+func (img *Image) ComputeLayerInfo(digest, parent string, layersize int) (layerinfo *LayerInfo, err error) {
+	podmaninfo := img.GetPodmanInfo()
+	r := podmaninfo.LayerReaderMap[digest]
+	diffID := fmt.Sprintf("%x", r.Sum256(nil))
+	size := r.GetSize()
```
siscia: For this two to be correct, we must be sure that we have consumed all the buffers.

Maybe we should call Read until it doesn't return `0, eof`, what you think?

siscia: We should also make sure that they somehow coordinates, right?

siscia:
@Mohitty I removed the old / addressed comments and I added a couple.

Also DUCC support the `*` glob. I believe also the podman integration should support it.

ducc/lib/utils.go
Mohitty: Donr

ducc/lib/utils.go
Mohitty: Thanks :)

ducc/lib/podman_store.go
Mohitty: Well, these two methods are called once the layers are completely read and ingested. The call to read function is managed where it's called!
Am I missing something?

ducc/lib/cvmfs.go
```diff
@@ -520,3 +520,28 @@ func RepositoryExists(CVMFSRepo string) bool {
...
+	if err != nil {
+		LogE(err).Error("Error in writing data to file")
+		return err
+	}
+	err = tmpFile.Close()
+	if err != nil {
+		LogE(err).Error("Error in closing temporary file")
+		return err
+	}
+	err = IngestIntoCVMFS(CVMFSRepo, path, tmpFile.Name())
```
siscia: here you can just:
```suggestion
	return IngestIntoCVMFS(CVMFSRepo, path, tmpFile.Name())
```

ducc/lib/wish.go
```diff
@@ -23,8 +23,8 @@ type WishFriendly struct {
 	UserOutput             string
 	InputImage             *Image
 	OutputImage            *Image
-	ExpandedTagImagesLayer <-chan *Image
```
siscia, line 26: Why this change?

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,511 @@
...
+	err = img.CheckImageChanged(CVMFSRepo)
+	if err != nil {
+		LogE(err).Error("error while checking if older image version with same tag present in store")
+		return err
+	}
+
+	Log().WithFields(log.Fields{"action": "Ingest the image into podman store"}).Info(img.GetSimpleName())
+	createCatalogIntoDirs := []string{rootPath, filepath.Join(rootPath, rootfsDir), filepath.Join(rootPath, imageMetadataDir), filepath.Join(rootPath, layerMetadataDir)}
+	for _, dir := range createCatalogIntoDirs {
+		err = CreateCatalogIntoDir(CVMFSRepo, dir)
```
siscia, line 481: Does this check if the catalog is already there?

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,511 @@
...
+		if err != nil {
+			LogE(err).WithFields(log.Fields{
+				"directory": dir}).Error(
+				"Impossible to create subcatalog in the directory.")
+		}
+	}
+
+	err = img.IngestImageManifest(CVMFSRepo)
```
siscia: Does it checks before to ingest?

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,511 @@
...
+	if err != nil {
+		LogE(err).Error("Unable to create manifest file in podman store")
+		return err
+	}
+
+	err = img.IngestConfigFile(CVMFSRepo)
```
siscia: Does it check before to ingest?

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,511 @@
...
+				if name == img.GetSimpleName() {
+					Log().WithFields(log.Fields{"image": img.GetSimpleName()}).Info("older image version present, cleaning and ingesting newer version")
+					present = true
+					directory := filepath.Join("/cvmfs", CVMFSRepo, rootPath, imageMetadataDir, image.ID)
+					err := RemoveDirectory(directory)
+					if err != nil {
+						LogE(err).Error("error while removing older image version from podman store")
+						return err
+					}
+					break
```
siscia, line 448: How do we know it is not the same image? I mean, why the reference we found must refer to an old one?

If I run the conversion twice, would I call the `RemoveDirectory` the second time?

ducc/lib/conversion.go
```diff
@@ -174,60 +182,99 @@ func ConvertWishSingularity(wish WishFriendly) (err error) {
...
+		}
+		err = CreateThinImage(manifest, layerLocations, *expandedImgTag, *outputImage)
+		if err != nil {
+			return err
+		}
+		err = PushImageToRegistry(*outputImage)
+		if err != nil {
+			return err
+		}
+	}
```
siscia: Here, if some image somehow fail, we stop processing all the other images with the same tag.

It is not unreasonable, but sometimes an image just fail to convert for reason outside our control.
I believe it is better to store the first error and continue the loop.

ducc/lib/conversion.go
```diff
@@ -174,60 +182,99 @@ func ConvertWishSingularity(wish WishFriendly) (err error) {
...
+		if err != nil {
+			return err
+		}
+		imageID := strings.Split(manifest.Config.Digest, ":")[1]
+		manifestPath := filepath.Join("/", "cvmfs", wish.CvmfsRepo, rootPath, imageMetadataDir, imageID, "manifest")
+		alreadyConverted := AlreadyConverted(manifestPath, manifest.Config.Digest)
+		if alreadyConverted == ConversionMatch {
+			if convertAgain == false {
+				Log().Info("Image already present in podman store, moving on")
+				return nil
```
siscia: Nope, if the image is already in the podman store, we want to move to the next image in the range, not return.

ducc/lib/conversion.go
```diff
@@ -174,60 +182,99 @@ func ConvertWishSingularity(wish WishFriendly) (err error) {
...
+			}
+		}
+		err = expandedImgTag.CreatePodmanImageStore(wish.CvmfsRepo, subDirInsideRepo)
+		if err != nil {
+			LogE(err).Warning("Unable to create Podman additional image store")
+			return err
+		}
+		Log().Info("Image successfully ingested into podmanStore")
+	}
+	return nil
```
siscia: Similar comment of above, an early return force the `for` loop to don't finish. I would prefer that we keep track of the first error but we try to convert all the images.

ducc/lib/conversion.go
```diff
@@ -334,6 +372,11 @@ func convertInputOutput(inputImage *Image, outputImage Image, repo string, conve
 					cleanup(TrimCVMFSRepoPrefix(layerPath))
 					return
 				}
+
+				err = StoreLayerInfo(inputImage, repo, layer.Name, layer.Path.(*ReadAndHash))
+				if err != nil {
```
siscia, line 392: Do not rely on reflection here: `layer.Path.(*ReadAndHash)` change directly the type.

After all we want `layer.Path` to be something that has the Read and the Hash method.

ducc/lib/conversion.go
```diff
@@ -395,48 +422,6 @@ func convertInputOutput(inputImage *Image, outputImage Image, repo string, conve
...
-			// here is possible to use the result of the above ReadAll to have
-			// informantion about the status of the upload.
-			_, errReadDocker := ioutil.ReadAll(res)
-			if err != nil {
-				LogE(errReadDocker).Warning("Error in reading the status from docker")
-			}
-			Log().Info("Finish pushing the image to the registry")
-		}
 		manifestPath := filepath.Join(".metadata", inputImage.GetSimpleName(), "manifest.json")
 		errIng := IngestIntoCVMFS(repo, manifestPath, <-manifestChanell)
```
siscia: Here it is a little problematic.

If there is some error in pushing the image, we don't save the manifest.

Let's keep this point on hold for a while though.

siscia:
Another thing, in the `ReadAndHash` interface. We should somehow make sure that we call the `Sum256` and the `GetSize` are called after the whole reader underneath was actually processed.

Maybe we can detect the EOF? Maybe we can introduce the invariant that the return of those methods are correct only after we call the `.Close()` method. And `Close()` return an error if we didn't finish to process the underneath reader...

ducc/lib/wish.go
Mohitty: I have explained it in the design doc

ducc/lib/podman_store.go
Mohitty: yes. CreateCatalogIntoDir function does before it creates one.

ducc/lib/podman_store.go
Mohitty: yes, we only ingest the image if the manifest is not already present

ducc/lib/podman_store.go
Mohitty: Config file is ingested only if manifest is ingested, and we check for manifest file. But still I will add a check.

ducc/lib/podman_store.go
Mohitty: No, we first check if the same image is present in store in ConvertWishPodman(). If we get past that check, it means the image is changed, since the name and tag of image are same but the digest/manifest is not the same.

ducc/lib/conversion.go
Mohitty: Ok, will do

ducc/lib/conversion.go
Mohitty: yes, will fix it. Btw this is the check I am referring to in the other comment of calling RemoveDirectory

Mohitty:
> In the `podman_store.go`, if I understand correctly, we keep creating symlinks without checking if those symlinks are really new. Can this be fixed?

Yes, I will fix it..

ducc/lib/conversion.go
Mohitty: Done

ducc/lib/conversion.go
Mohitty: Done

ducc/lib/conversion.go
Mohitty: Done

ducc/lib/conversion.go
Mohitty: Done

ducc/lib/podman_store.go
Mohitty: Done

ducc/lib/cvmfs.go
Mohitty: Done

siscia:
It looks great! 

Let me test it and see if there are any issue!

siscia:
Good job! I still need to run some images but it seems to work fine.

Do you think we can also work on the GC? Like delete the files that are not needed anymore when a new image take over? Is this case actually managed?

Mohitty:
> Do you think we can also work on the GC? Like delete the files that are not needed anymore when a new image take over? Is this case actually managed?

Done.

ducc/lib/conversion.go
```diff
@@ -77,14 +78,21 @@ func ConvertWishSingularity(wish WishFriendly) (err error) {
 	if err != nil {
 		return err
 	}
+	err = CreateCatalogIntoDir(wish.CvmfsRepo, ".flat")
+	if err != nil {
+		LogE(err).WithFields(log.Fields{
+			"directory": ".flat"}).Error(
+			"Impossible to create subcatalog in super-directory.")
```
jblomer: What's the `super-directory`? The parent directory?

ducc/lib/conversion.go
```diff
@@ -174,60 +182,114 @@ func ConvertWishSingularity(wish WishFriendly) (err error) {
...
+	}
+	return firstError
+}
+
+func ConvertWish(wish WishFriendly, convertAgain, forceDownload bool) (err error) {
+	err = CreateCatalogIntoDir(wish.CvmfsRepo, subDirInsideRepo)
+	if err != nil {
+		LogE(err).WithFields(log.Fields{
+			"directory": subDirInsideRepo}).Error(
+			"Impossible to create subcatalog in super-directory.")
```
jblomer: see above

ducc/lib/conversion.go
```diff
@@ -518,6 +518,91 @@ func CreateThinImage(manifest da.Manifest, layerLocations map[string]string, inp
 	return nil
 }
 
+func PushImageToRegistry(outputImage Image) (err error) {
+	// is necessary this mechanism to pass the authentication to the
```
jblomer: I think I don't understand the comment.

ducc/lib/cvmfs.go
```diff
@@ -520,3 +520,24 @@ func RepositoryExists(CVMFSRepo string) bool {
 		return false
 	}
 }
+
+//writes data to file and ingest in cvmfs repo path
```
jblomer: Rather "publish" than "ingest" since the ingestion term is charged with publishing tarballs.

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,534 @@
...
+	err = writeDataToCvmfs(CVMFSRepo, TrimCVMFSRepoPrefix(imageInfoPath), imgInfo)
+	if err != nil {
+		LogE(err).Error("Error in writing images.json")
+		return err
+	}
+	return nil
+}
+
+// Ingest the exploded rootfs in podman store.
+func (img *Image) IngestRootfsIntoPodmanStore(CVMFSRepo, subDirInsideRepo string, digestMap map[string]string) (err error) {
```
jblomer: I think `ingest` is a misleading name.  I'd rather call it `Link...`

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,534 @@
...
+			}
+			layerIdMap[layer.Digest] = string(data)
+		}
+	}
+	return nil
+}
+
+// Create the lower files for diff dirs to be used by podman.
+func (img *Image) CreateLowerFiles(CVMFSRepo string, digestMap, layerIdMap map[string]string) (err error) {
+	Log().WithFields(log.Fields{"action": "Creating Lower files for diff dirs in podman store"}).Info(img.GetSimpleName())
```
jblomer, line 267: Is it the lower _layers_? I think that would make more sense than calling it `lower files`

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,534 @@
...
+		return err
+	}
+
+	imageID := strings.Split(manifest.Config.Digest, ":")[1]
+	configFilePath := filepath.Join(rootPath, imageMetadataDir, imageID, fname)
+	if _, err := os.Stat(configFilePath); os.IsNotExist(err) {
+		user := img.User
+		pass, err := GetPassword()
+		if err != nil {
+			LogE(err).Warning("Unable to get the credential for downloading the configuration blog, trying anonymously")
```
jblomer: blog --> blob (or rather `snippet` as I guess it is not binary data.  Of course it's still under the "blob" url, so unsure which one's better)

ducc/lib/podman_store.go
```diff
@@ -0,0 +1,534 @@
...
+		paths = append(paths, imagelockpath)
+	}
+	for _, file := range paths {
+		if _, err := os.Stat(file); os.IsNotExist(err) {
+			tmpFile, err := ioutil.TempFile("", "lock")
+			tmpFile.Close()
+			if err != nil {
+				return err
+			}
+			err = IngestIntoCVMFS(CVMFSRepo, TrimCVMFSRepoPrefix(file), tmpFile.Name())
```
jblomer: `IngestInto` --> `PublishTo`?

ducc/lib/podman_store.go
jblomer: Let's add that as a code comment here

ducc/lib/wish.go
jblomer: It would be good to add a link to the design doc somewhere in the code, perhaps best in `podman_store.go`

jblomer (changes requested):
Very good!  Just a few nits.  We should add a DUCC integration tests that ingests an image and then starts it with podman.

ducc/lib/conversion.go
Mohitty: This and the above two comments refers to part of legacy code, I didn't modify them. They just needed restructuring to incorporate podman integration.

ducc/lib/podman_store.go
Mohitty: No, actually they store the directory names of the lower layers. Podman names these files as "lower" , hence this name.

ducc/lib/podman_store.go
Mohitty: I think, for now we can go with blob.

ducc/lib/podman_store.go
jblomer: I see, but then it's "lower directories" instead of "lower files", no?

ducc/lib/conversion.go
jblomer: Ok, let's still use the opportunity to improve the code (@siscia perhaps you can help)

ducc/lib/podman_store.go
Mohitty: No, they are files which contain a string (name of the parent layers) as its content.  Podman names these files as "lower"

jblomer: I see. Thanks!

ducc/lib/conversion.go
siscia: The docker documentation says that it is not necessary to pass the authentication credential. If we don't pass the credentials it does not work. It is actually necessary to pass the credentials.

jblomer: Understood, thanks!  Let's take this as a code comment :)

Mohitty:
> We should add a DUCC integration tests that ingests an image and then starts it with podman.

@jblomer done. Also edited the code comments.

siscia:
Mohit, if you force push, then I am forced to review everything again...

Mohitty:
> Mohit, if you force push, then I am forced to review everything again...

I had to rebase, sorry. I'll keep that in mind.

siscia:
I am getting around this! Just a little more time.

siscia:
Good to merge for me.

If no objections I merge it tomorrow!

PR #2584, opened 2020-07-09
Fix Debian/Ubuntu packaging

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2546, opened 2020-04-23
Update cvmfs-config-default for a config-repo based configuration

DrDaveD:
As discussed in [CVM-1085](https://sft.its.cern.ch/jira/browse/CVM-1085).  The corresponding configuration repository is the [default branch of cvmfs-contrib/config-repo](https://github.com/cvmfs-contrib/config-repo/tree/default).  I tested it by putting that contents into my own test repository, then setting CVMFS_CONFIG_REPOSITORY to point to it in a `/etc/cvmfs/default.d/95-dwd.conf.`  The configuration repository contents will need to be published before this is released.

I have seen one problem with this on Debian, in that when I upgrade from the current version of cvmfs-config-default to this version, it does not remove the files under /etc/cvmfs that got removed even though they hadn't been modified.  I don't know if that's just my test machine, something in the package definition, or what.  That will need further investigation before this is merged but I wanted to get it started being reviewed, and maybe someone else knows the answer to this problem.

PR #2586, opened 2020-07-13
Update default config, use cvmfs-config repository

jblomer:


jblomer:
Replaces #2546

rptaylor:
Instead of CVMFS_USE_OPENHTC , perhaps the variable name could be CVMFS_USE_CDN ?
This way we may have a chance to consolidate and standardize somewhat, for example Compute Canada may use Cloudflare but not openhtc.io, while other organizations might use some other CDN, or there could be other changes in CDN implementation details. But it would be unfortunate and could lead to confusion if different communities or organizations had different names of the special variable that users have to know in order to enable CDN use.

For that matter, it would be nice if there was a keyword like "CDN" so you could do something like
CVMFS_HTTP_PROXY="CDN;DIRECT" but that would sort of bend the concept of the proxy list and the server list ; it could be hard to implement that behaviour in a way that maintains consistency.

DrDaveD:
CVMFS_USE_CDN is a good idea.  I was thinking that at least for OSG & EGI we could use unset/blank to mean default (which is enabled when CVMFS_HTTP_PROXY is "DIRECT" or "auto;DIRECT") and when set to "yes" would be always enable or when set to "no" would be always disable.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

mount/default.conf
```diff
@@ -21,16 +21,34 @@
 # Use cvmfs_config showconfig to get the effective parameters.
 #
 
+# The client profile selects configuration options tuned for certain use cases
+# Overwrite this parameter in /etc/cvmfs/default.local
```
DrDaveD: Isn't this comment redundant, since the first line of the file already says to do that for everything in this file?

mount/default.conf
```diff
@@ -21,16 +21,34 @@
...
+# Set CVMFS_CLIENT_PROFILE=single if you run cvmfs on a laptop or a small
```
DrDaveD, line 25: I'm not sure it is appropriate to refer to a variable in the cvmfs-config-default package in /etc/cvmfs/default.conf.

mount/default.conf
```diff
@@ -53,20 +68,34 @@ CVMFS_KEYS_DIR=/etc/cvmfs/keys
...
 CVMFS_USE_GEOAPI=no
+
 # If the transfer rate falls below the given bytes/s for longer than
 # CVMFS_TIMEOUT[_DIRECT], treat the connection like a timeout.
 CVMFS_LOW_SPEED_LIMIT=1024
 
-# CA and CRL files used to verify repository signatures
-# EXPERIMENTAL!
-# CVMFS_TRUSTED_CERTS=/etc/grid-security/certificates
+# Add the next line to the default.local file in order to use the OpenHTC
```
DrDaveD: This too I think doesn't belong in /etc/cvmfs/default.local, because its behavior only applies to the cvmfs-config-default configuration.  I intend to include the same control in cvmfs-config-osg & cvmfs-config-egi, but it will sometimes be enabled by default.

mount/default.conf
jblomer: I'd like to establish the `CVMFS_CLIENT_PROFILE` variable for the benefit of all the config repositories.  I think it would be generally useful to have a hint how the client is intended to be used.

jblomer:
@cernvm-bot cloudtest

mount/default.conf
DrDaveD: If you'd like to establish it as a default for other config repos, ok, we can update those, but then please remove "from the cvmfs-config-default package" from the comment.  You could say instead "if supported by the cvmfs-config package".

Should CVMFS_USE_CDN be similarly documented then, for consistency?

Note for the record that I think that CVMFS_USE_CDN=yes should be the default (at least in config repo) when CVMFS_CLIENT_PROFILE=single.  That's effectively what the OSG & EGI config repos already do (implied when CVMFS_HTTP_PROXY isn't set) and what I intend for them to continue to do.  It protects the stratum 1s and gives a better user experience.  When we discussed it you disagreed.

jblomer:
@cernvm-bot cloudtest

PR #2588, opened 2020-07-23
S3 over HTTPS

siscia:
As the S3 is becoming always more and more widespread, some provider started to accept only HTTP**S** connections.

While this is not an immediate problem for us, it would be nice to stay up to date.

The main objective of this PR is to allow to create a CVMFS repo over HTTPS on S3 backend storage.

While reaching for this objective we also tidy up a little bit several pieces of code.

More checks are necessary, especially on the client side, but this PR might accidentally have add support for serving CVMFS over HTTPS.

## Progression

We started by introducing another option `CVMFS_S3_USE_HTTPS` (default to false) that force CVMFS to communicate to the S3 backend over HTTPS.

The next problem was to detect all the time the string "http" was hardcoded. 
They were mostly checks if the files is local or remote, we allow remote URI to start also with `https` and not only with `http`.

At this point we could start to make HTTPS requests, however they all failed since no CA authority was recognized.
There was some checks already in place, on the client part, to add CA authorities. We use to check either the directory pointed by `X509_CERT_DIR` or the directory in `/etc/grid-security/certificates`.
I refactor this check to work in:
1. the directory pointed by `X509_CERT_DIR`
2. `/etc/grid-security/certificates`
3. `/etc/ssl/certs/`
4. other less used directories

We try to be smart about which directory to load. Hopefully it will just work in all systems.

Then we make sure to enable the HTTPS checks and to load the certificates only if necessary.

During the development we also enable the errorbuffer provided by libcurl. It is not strictly necessary but it can provide useful information when something failed and I would suggest to leave it on.

One last touch was to enable by default the signature V4 instead of the V2 for S3. It seems that some S3 provider support only the V4.

## What is missing

- [x] We miss documentation for the `CVMFS_S3_USE_HTTPS` (added with: https://github.com/cvmfs/doc-cvmfs/pull/150)


siscia:
@cernvm-bot cloudtest

siscia:
@cernvm-bot cloudtest

siscia:
@jblomer what you think?

jblomer:
@siscia Excellent!  Let me just have a closer look on the certificate handling.

cvmfs/util/posix.cc
```diff
@@ -197,14 +197,14 @@ std::string GetAbsolutePath(const std::string &path) {
 
 
 bool IsHttpUrl(const std::string &path) {
-  if (path.length() < 7) {
```
jblomer: There's an asymmetry here: It will recognize "https://" as an (empty) URL but not "http://".  Let's just keep the 7 here.  The `path.substr(0, 8)` gracefully deals with strings that are only 7 characters long.

test/unittests/t_util.cc
```diff
@@ -316,15 +316,23 @@ TEST_F(T_Util, String2Uint64) {
 
 TEST_F(T_Util, IsHttpUrl) {
   EXPECT_TRUE(IsHttpUrl("http://cvmfs-stratum-one.cern.ch/cvmfs/cms.cern.ch"));
-  EXPECT_TRUE(IsHttpUrl("http://"));
+  EXPECT_TRUE(IsHttpUrl("https://cvmfs-stratum-one.cern.ch/cvmfs/cms.cern.ch"));
+  EXPECT_TRUE(IsHttpUrl("http://a"));
```
jblomer: Test also for "http://" (expect true)

cvmfs/upload_s3.h
```diff
@@ -81,6 +81,7 @@ class S3Uploader : public AbstractUploader {
 
  private:
   static const unsigned kDefaultPort = 80;
+  static const unsigned kHTTPSPort = 443;
```
jblomer: Naming convention: `kHttpsPort`

cvmfs/s3fanout.h
```diff
@@ -172,8 +178,9 @@ class S3FanoutManager : SingleCopy {
 
   struct S3Config {
     S3Config() {
-      authz_method = kAuthzAwsV2;
+      authz_method = kAuthzAwsV4;
```
jblomer: This should be correct on AwsV2.  It will be set to the V4 scheme if a region is set (see upload_s3.cc)

cvmfs/s3fanout.cc
```diff
@@ -636,8 +637,9 @@ int S3FanoutManager::InitializeDnsSettings(
   }
 
   // Remove port number if such exists
-  if (!HasPrefix(host_with_port, "http://", false /*ignore_case*/))
-    host_with_port = "http://" + host_with_port;
+  if (!HasPrefix(host_with_port, "http://", false /*ignore_case*/) ||
```
jblomer: I think that should be `&&` instead of `||`  Or perhaps: if `!IsHttpUrl(host_with_port)`

cvmfs/s3fanout.cc
```diff
@@ -636,8 +637,9 @@ int S3FanoutManager::InitializeDnsSettings(
   }
 
   // Remove port number if such exists
```
jblomer: The comment seems to be wrong.  We want to make sure that the protocol is part of the URL.

cvmfs/download.cc
```diff
@@ -973,9 +974,7 @@ void DownloadManager::SetUrlOptions(JobInfo *info) {
 
   curl_easy_setopt(curl_handle, CURLOPT_SSL_VERIFYPEER, 1L);
   if (url.substr(0, 5) == "https") {
-    const char *cadir = getenv("X509_CERT_DIR");
-    if (!cadir || !*cadir) {cadir = "/etc/grid-security/certificates";}
-    curl_easy_setopt(curl_handle, CURLOPT_CAPATH, cadir);
+    AddSSLCertificates(curl_handle);
     const char *cabundle = getenv("X509_CERT_BUNDLE");
```
jblomer: Let's make this part of `AddSSLCertificates()`, too

cvmfs/ssl.h
```diff
@@ -0,0 +1,61 @@
+/**
```
jblomer, line 1: Let's make it a proper `ssl.h` and `ssl.cc`

cvmfs/ssl.h
```diff
@@ -0,0 +1,61 @@
...
+#include <vector>
+
+#include "duplex_curl.h"
+#include "util/posix.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
+namespace CVMFS_NAMESPACE_GUARD {
+#endif
+
+inline unsigned int count_ssl_certificates(std::string directory) {
```
jblomer: Remove inline but make it a local static `CountSslCertificates`

cvmfs/ssl.h
```diff
@@ -0,0 +1,61 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#ifndef CVMFS_SSL_H_
+#define CVMFS_SSL_H_
```
jblomer, line 10: A top level comment should explain that this is about CA certificates of servers.  There is some other code in authz that deals with client certificates.

cvmfs/ssl.h
```diff
@@ -0,0 +1,61 @@
...
+
+  std::vector<std::string> dotpem = FindFilesBySuffix(directory, ".pem");
+  std::vector<std::string> dotcrt = FindFilesBySuffix(directory, ".crt");
+
+  return dotpem.size() + dotcrt.size();
+}
+
+inline bool AddSSLCertificates(CURL *handle) {
+  std::vector<std::string> cadirs;
+  const char *cadir = getenv("X509_CERT_DIR");
```
jblomer: Include `cstdlib` for `getenv`

cvmfs/ssl.h
```diff
@@ -0,0 +1,61 @@
...
+  if (cadir != NULL) {
+    cadirs.push_back(std::string(cadir));
+  }
+  cadirs.push_back("/etc/grid-security/certificates");
+
+  // most systems store the certificates here
+  cadirs.push_back("/etc/ssl/certs/");
```
jblomer: We should not change the client default behavior.  We should have another bool parameter for this function that specifies if we look in system default directories.  We set this to false for the client and to true for the s3fanout.  We could check another environment variable in download.cc to optionally set this to true.

cvmfs/ssl.h
```diff
@@ -0,0 +1,61 @@
...
+
+  cadirs.push_back("/etc/pki/tls/certs/");
+  cadirs.push_back("/etc/ssl/");
+  cadirs.push_back("/etc/pki/tls/");
+  cadirs.push_back("/etc/pki/ca-trust/extracted/pem/");
+  cadirs.push_back("/etc/ssl/");
+
+  for (std::vector<std::string>::const_iterator cadir = cadirs.begin();
+       cadir != cadirs.end(); ++cadir) {
+    if (count_ssl_certificates(*cadir) > 0) {
```
jblomer: Perhaps it makes more sense to call it `HasCertificates`.  Then we can back out if we find at least one, we don't need to count them all.

jblomer (changes requested):
Very nice!  Some comments inline.

cvmfs/s3fanout.h
siscia: The V2 is being deprecated everywhere. I believe it is a good opportunity to avoid running in problems in a future.

https://docs.aws.amazon.com/AmazonS3/latest/dev/auth-request-sig-v2.html

jblomer: The deprecation schedule might be different for 3rd party products (MinIO, Ceph, etc). In any case, let's handle the question of the default authz scheme in another PR.  I think we can reuse the `CVMFS_S3_FLAVOR` settings for this instead of guessing based on the region setting.

cvmfs/ssl.h
siscia: I am not sure why we want to make this code more complex? The change in behavior is backward compatible and it helps the client to work on HTTPS servers.

Are we afraid of loosing performance? I believe we could measure the difference that this PR makes, but I believe it won't be that bad.

This will bring us to introduce another variable, if somebody will use cvmfs outside the grid they might prefer to use https and it means that the user will need to manage a more complex configuration.

cvmfs/ssl.h
siscia: I believe we can discard this now.

cvmfs/ssl.h
jblomer: With the new code we are looking in more places for SSL certificates than we did previously.  This might eventually be what we want, but first it is a change in behavior for currently running clients.

cvmfs/ssl.h
siscia: I did change the name but the implementation is the same. Even if we want to check just one, is still is the easiest thing to do.

Should we change also the implementation? It will anyway be a couple of stat calls against the directory, we might save some string comparison with an early return.

jblomer: Probably OK for now because libcurl anyway has to read them all.  Still, there can be potentially a lot of certificate files, so we should at least add a `TODO` comment to the implementation to keep note of a possible improvement.

siscia: Hummm, I prefer to do it now. But I am not sure I see any potential improvement.

jblomer: I'm thinking of this: suppose we look into a directory with 10,000 files.  If the first of them is a .pem or a .crt file, we don't need to stat and check file names on the 9999 other files.

siscia: Alright, let's wait anyway to merge this.

siscia:
I am not sure this is ready yet. I am done some other tests.

siscia:
Somehow it seems to break the v4 authentication.

jblomer:
@cernvm-bot cloudtest

siscia:
@jblomer  As I was mentioning there is a last issue with the signing.

The sign is called inside `runuser` ( [here the call](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/server/cvmfs_server_common.sh#L665) ) and this reset the environmental variables that should signal to the download manager to use the system certificates. The system certificates are not loaded, and we cannot reach the HTTPS storage.

If I short-circuit the check (setting the variable [here](https://github.com/cvmfs/cvmfs/pull/2588/files#diff-737a1267e0ad3fb8ebb770759de2662aR1492) to `true`) for using or not the system certificates everything seems to work fine.

Not sure what is the best solution here.

siscia:
@jblomer I am not so convinced of the last couple of commits. Let me know what you think!

I was hoping that this work would be enough to work also over backblaze B2 but their S3 interface is not quite there. I opened a ticket with them as well.

If we like this implementation I will add some tests!

jblomer:
@cernvm-bot cloudtest

cvmfs/upload_s3.cc
```diff
@@ -162,6 +164,18 @@ bool S3Uploader::ParseSpoolerDefinition(
   if (options_manager.GetValue("CVMFS_S3_PEEK_BEFORE_PUT", &parameter)) {
     peek_before_put_ = options_manager.IsOn(parameter);
   }
+  if (options_manager.GetValue("CVMFS_S3_USE_HTTPS", &parameter)) {
+    if (parameter == "true") {
```
jblomer: Use `options_manager.IsOn(parameter)`, which also catches variants `on`, `1`, etc.

cvmfs/server/cvmfs_server_common.sh
```diff
@@ -981,6 +982,7 @@ CVMFS_HIDE_MAGIC_XATTRS=yes
 CVMFS_FOLLOW_REDIRECTS=yes
 CVMFS_SERVER_CACHE_MODE=yes
 CVMFS_NFILES=65536
+CVMFS_USE_SSL_SYSTEM_CA=true
```
jblomer, line 985: That should be added in the migration routine, too.  I think we already have the migration for repositories from 2.7 --> 2.8 in place, so it's only a matter of adding this particular parameter to existing server.conf files.

cvmfs/ssl.cc
```diff
@@ -0,0 +1,106 @@
...
+
+bool HasCertificates(std::string directory) {
+  if (!DirectoryExists(directory)) return false;
+
+  DIR *dirp = opendir(directory.c_str());
+  if (!dirp) return false;
+
+  platform_dirent64 *dirent;
+  while ((dirent = platform_readdir(dirp))) {
+    if ((dirent->d_type == DT_REG) || (dirent->d_type == DT_UNKNOWN)) {
```
jblomer: Unfortunately, we cannot rely on `d_type`. Some file systems don't fill it.  One has to make an explicit `stat` call given `dirent->d_name` to keep the code portable.

cvmfs/ssl.cc
```diff
@@ -0,0 +1,106 @@
...
+        return true;
+      }
+    }
+  }
+
+  closedir(dirp);
+  return false;
+}
+
+bool Add_CVMFS_SSL_Certificates(CURL *handle) {
```
jblomer, line 46: Let's put this function in an anonymous namespace

cvmfs/ssl.cc
```diff
@@ -0,0 +1,106 @@
...
+    CURLcode res = curl_easy_setopt(handle, CURLOPT_CAINFO, cabundle);
+    if (CURLE_OK == res) {
+      certificate_already_added = true;
+    }
+  }
+
+  return certificate_already_added;
+}
+
+bool AddSystemSSLCertificates(CURL *handle) {
```
jblomer, line 69: Let's put this function in an anonymous namespace.

cvmfs/s3fanout.cc
```diff
@@ -635,9 +641,9 @@ int S3FanoutManager::InitializeDnsSettings(
     return 0;
   }
 
-  // Remove port number if such exists
-  if (!HasPrefix(host_with_port, "http://", false /*ignore_case*/))
-    host_with_port = "http://" + host_with_port;
+  // Add protocol informantion for extraction of fields for DNS
```
jblomer: informantion --> information

cvmfs/download.cc
```diff
@@ -1646,6 +1641,15 @@ void DownloadManager::Init(const unsigned max_pool_handles,
       SetProxyChain(string(getenv("http_proxy")), "", kSetProxyRegular);
     }
   }
+
+  const char *use_cert_var = getenv("CVMFS_USE_SSL_SYSTEM_CA");
```
jblomer: Let's move this to the `MountPoint` class where the download managers for the client is initialized.  This way, we'd have the option handling for the download manager more centralized.  As a follow-up to this PR, it might anyway be a good idea to check why `X509_CERT_DIR` and `IPV4_ONLY` parameters are checked directly in the download manager and not in the code that initialized it.

jblomer (changes requested):
Excellent!  I like this change a lot, very clear code.

Couple of minor comments, and as you suggested, please add some more test.

Apologies for having left this PR unhandled for way too long!

siscia:
For what concerns the test, it is a bit of a mess in our infrastructure. Or we add a matrix column, or I am not sure how it is better to handle them.

siscia:
@jblomer I have never touched the migration routine, let me know if the change is a sensible one.

cvmfs/download.cc
siscia: No, this is wrong, we are not able to create the repository if we move this into the `MountPoint` as done with: https://github.com/cvmfs/cvmfs/pull/2588/commits/01799e9a69363f710439edc308d7680efad5f7f2

siscia: Fixed! Was a problem with the Clone method that was not taking into consideration the new use of system CA

cvmfs/server/cvmfs_server_migrate.sh
```diff
@@ -473,6 +473,7 @@ _migrate_141() {
   fi
 
   echo "--> updating server.conf"
+  echo "CVMFS_USE_SSL_SYSTEM_CA=true" >> $server_conf
```
jblomer, line 491: It's unfortunately more difficult because r141 is already released.  We need to introduce a `_migrate_142` together with corresponding changes at the bottom of `cvmfs_server_migrate` and in `cvmfs_layout_revision()`

cvmfs/ssl.cc
```diff
@@ -0,0 +1,109 @@
...
+#include <cstdlib>
+#include <string>
+#include <vector>
+
+#include "duplex_curl.h"
+#include "platform.h"
+#include "util/posix.h"
+#include "util/string.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
```
jblomer: The `CVMFS_NAMESPACE_GUARD` is defined only when building the loader.  The idea is to use the same code in the loader (cvmfs2 binary) and in the libcvmfs_fuse library without name clashes.

Here, we would simply open an anonymous namespace containing everything except the public `AddSSLCertificates` function

cvmfs/ssl.h
```diff
@@ -0,0 +1,30 @@
...
+#include <vector>
+
+#include "duplex_curl.h"
+#include "util/posix.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
+namespace CVMFS_NAMESPACE_GUARD {
+#endif
+
+bool HasCertificates(std::string directory);
```
jblomer: Does not need to be part of the header, I think

cvmfs/ssl.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef CVMFS_SSL_H_
+#define CVMFS_SSL_H_
+
+#include <string>
+#include <vector>
+
+#include "duplex_curl.h"
+#include "util/posix.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
```
jblomer: See above, in the header we would not create an anonymous namespace

cvmfs/ssl.cc
```diff
@@ -0,0 +1,109 @@
...
+bool HasCertificates(std::string directory) {
+  if (!DirectoryExists(directory)) return false;
+
+  DIR *dirp = opendir(directory.c_str());
+  if (!dirp) return false;
+
+  platform_dirent64 *dirent;
+  while ((dirent = platform_readdir(dirp))) {
+    platform_stat64 stat;
+    if (0 == platform_stat(dirent->d_name, &stat)) {
```
jblomer: Easier to read if reversed: 

    if (platform_stat(dirent->d_name, &stat) != 0)
      continue;

cvmfs/ssl.cc
```diff
@@ -0,0 +1,109 @@
...
+      if (S_ISREG(stat.st_mode)) {
```
jblomer: As well:

    if (!S_ISREG(stat.st_mode))
      continue;

cvmfs/upload_s3.cc
```diff
@@ -162,6 +164,16 @@ bool S3Uploader::ParseSpoolerDefinition(
   if (options_manager.GetValue("CVMFS_S3_PEEK_BEFORE_PUT", &parameter)) {
     peek_before_put_ = options_manager.IsOn(parameter);
   }
+  if (options_manager.IsOn("CVMFS_S3_USE_HTTPS")) {
```
jblomer: `IsOn` works slightly different:

    if (options_manager.GetValue("CVMFS_S3_USE_HTTPS", &parameter) &&
        options_manager.IsOn(parameter))

jblomer (changes requested):
Many thanks!  Just a few more minor things.

cvmfs/server/cvmfs_server_migrate.sh
```diff
@@ -479,6 +479,21 @@ _migrate_141() {
   load_repo_config $name
 }
 
+_migrate_142() {
```
jblomer, line 482: Actually, now that template transactions are merged the `_migrate_142` from there can be reused.

siscia:
@cernvm-bot cloudtest cc7

siscia:
@cernvm-bot cloudtest cc7

cvmfs/server/cvmfs_server_common.sh
jblomer: This is still missing, I think

test/src/820-https_s3/main
```diff
@@ -120,35 +119,52 @@ EOF
...
+    # we create the bucket and we make it public
     /usr/local/bin/mc -C $SETTING_DIR/minio_client mb cvmfs/https-bucket
     /usr/local/bin/mc -C $SETTING_DIR/minio_client policy public cvmfs/https-bucket
 }
 
 cvmfs_run_test() {
+
+    # lets run this test only in the S3 suite
+    if [ x"$CVMFS_TEST_S3_CONFIG" == x"" ]; then
+        return 0
```
jblomer: I think we should fail here (we exclude the test anyway on non-S3 platforms)

test/src/820-https_s3/main
```diff
@@ -8,24 +8,21 @@ cvmfs_test_name="HTTPS S3 implementation"
 cvmfs_test_autofs_on_startup=false
 cvmfs_test_suites="quick"
 
-set -x
-
 SETTING_DIR="/tmp/cvmfs_testing_out"
```
jblomer: This should become the current working directory in the `main()` function as the canonical place to store test case output, config, etc.

jblomer: Better `CVMFS_TEST820_SETTINGS_DIR`

test/src/820-https_s3/main
```diff
@@ -8,24 +8,21 @@ cvmfs_test_name="HTTPS S3 implementation"
...
-MINIO_PID=""
 
 clean_up() {
```
jblomer, line 13: Should also `unset X509_CERT_BUNDLE`

test/src/820-https_s3/main
```diff
@@ -8,24 +8,21 @@ cvmfs_test_name="HTTPS S3 implementation"
...
     sudo kill -9 $(cat $SETTING_DIR/minio_pid)
-    
+
     # remove the repository
-    sudo cvmfs_server rmfs -f https-test.cvmfs.ch    
-    
+    sudo cvmfs_server rmfs -f https-test.cvmfs.ch
+
     # remove all the configuration files
     sudo rm -rf "/tmp/cvmfs_testing_out"
```
jblomer: If the files are created in the working directory of the test, the test driver takes care of removal on test success.

test/src/820-https_s3/main
```diff
@@ -41,44 +38,44 @@ create_https_certificates() {
...
 
     # we are making minio client use the new root CAs
     mkdir -p $SETTING_DIR/minio_client/certs/CAs
     cp $SETTING_DIR/rootCA.* $SETTING_DIR/minio_client/certs/CAs
 
-    sudo cp $SETTING_DIR/rootCA.key /etc/ssl/certs/cvmfs_https_test.key
-    sudo cp $SETTING_DIR/certs/public.crt /etc/ssl/certs/cvmfs_https_test1.crt
+    # we are making the whole system accept the new key
+    # useful for when we call curl in the bash script, libcurl does not care of this
+    sudo cp $SETTING_DIR/rootCA.crt /etc/ssl/certs/cvmfs_https_test.crt
```
jblomer: I think that works only on Ubuntu, on EL the certificate directory is rather /etc/pki/tls/certs

test/src/820-https_s3/main
```diff
@@ -87,9 +84,11 @@ CVMFS_S3_SECRET_KEY=important
 CVMFS_S3_BUCKETS_PER_ACCOUNT=1
```
jblomer: Let's remove this deprecated option

jblomer (changes requested):
Nice! Some minor comments.

siscia:
@cernvm-bot cloudtest cc7

siscia:
@cernvm-bot cloudtest cc7

siscia:
@cernvm-bot cloudtest cc7

jblomer:
Replaced by #2751

PR #2590, opened 2020-07-29
Azure Blob support

hmeiland:
Adding the authentication headers for Azure Blob and using them by selecting the CVMFS_S3_REGION=Azure

### DIFF
--- a/cvmfs/s3fanout.cc
+++ b/cvmfs/s3fanout.cc
@@ -625,21 +625,20 @@ bool S3FanoutManager::MkAzureAuthz(const JobInfo &info, vector<string> *headers)
     "/" + config_.access_key + "/" + config_.bucket + "/" + info.object_key;
 
   string string_to_sign;
-  //  if ((info.request == JobInfo::kReqPutDotCvmfs) ||
-  //   (info.request == JobInfo::kReqPutCas)) {
-  if (info.origin.IsValid()) {
-    string_to_sign =
-      GetRequestString(info) +
-      string("\n\n\n") +
-      string(StringifyInt(info.origin->GetSize())) + "\n\n\n\n\n\n\n\n\n" +
-      canonical_headers + "\n" +
-      canonical_resource;
-  } else {
-  // if (info.request == JobInfo::kReqHeadPut) {
+  if ((info.request == JobInfo::kReqHeadOnly) ||
+     (info.request == JobInfo::kReqHeadPut) ||
+     (info.request == JobInfo::kReqDelete)) {
     string_to_sign =
       GetRequestString(info) +
       string("\n\n\n") +
       "\n\n\n\n\n\n\n\n\n" +
+      canonical_headers + "\n" +
+      canonical_resource;
+  } else {
+    string_to_sign =
+      GetRequestString(info) +
+      string("\n\n\n") +
+      string(StringifyInt(info.origin->GetSize())) + "\n\n\n\n\n\n\n\n\n" +
       canonical_headers + "\n" +
       canonical_resource;
   }
--- a/cvmfs/s3fanout.cc
+++ b/cvmfs/s3fanout.cc
@@ -874,7 +874,6 @@ Failures S3FanoutManager::InitializeRequest(JobInfo *info, CURL *handle) const {
     assert(retval == CURLE_OK);
 
     if (info->request == JobInfo::kReqDelete)
-    //    (info->request == JobInfo::kReqHeadPut))
     {
       retval = curl_easy_setopt(handle, CURLOPT_CUSTOMREQUEST,
                                 GetRequestString(*info).c_str());
--- a/cvmfs/s3fanout.cc
+++ b/cvmfs/s3fanout.cc
@@ -609,6 +609,88 @@ bool S3FanoutManager::MkV4Authz(const JobInfo &info, vector<string> *headers)
   return true;
 }
 
+/**
+ * The Azure Blob authorization header according to
+ * tbd 
+ */
+bool S3FanoutManager::MkAzureAuthz(const JobInfo &info, vector<string> *headers)
+  const
+{
+  //string payload_hash;
+  char payload_size [256];
+  snprintf(payload_size, 256, "%lu", info.origin->GetSize());
+  //bool retval = MkPayloadHash(info, &payload_hash);
+  //if (!retval)
+    //return false;
+  string content_type = GetContentType(info);
+  //string timestamp = IsoTimestamp();
+  string timestamp = RfcTimestamp();
+  string date = timestamp.substr(0, 8);
+  vector<string> tokens = SplitString(complete_hostname_, ':');
+  assert(tokens.size() <= 2);
+  string canonical_hostname = tokens[0];
+  if (tokens.size() == 2 && String2Uint64(tokens[1]) != kDefaultHTTPPort)
+    canonical_hostname += ":" + tokens[1];
+
+  string signed_headers;
+  string canonical_headers;
+  string canonical_resource;
+  //if (!content_type.empty()) {
+    //signed_headers += "content-type;";
+    //headers->push_back("Content-Type: " + content_type);
+    //canonical_headers += "content-type:" + content_type + "\n";
+  //}
+  signed_headers += "host;x-amz-acl;x-amz-content-sha256;x-amz-date";
+  canonical_headers = "x-ms-blob-type:BlockBlob\nx-ms-date:" + timestamp + "\nx-ms-version: 2011-08-18";
+    //"host:" + canonical_hostname + "\n" +
+    //"x-amz-acl:public-read\n"
+    //"x-amz-content-sha256:" + payload_hash + "\n" +
+    //"x-ms-date:" + timestamp + "\n";
+
+  canonical_resource = "/" + config_.access_key + "/" + config_.bucket + "/" + info.object_key;
+  printf("\nheaders: %s\n", canonical_headers.c_str());
+  printf("\nresource: %s\n", canonical_resource.c_str());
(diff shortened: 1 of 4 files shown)


jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

cvmfs/s3fanout.cc
```diff
@@ -609,6 +609,59 @@ bool S3FanoutManager::MkV4Authz(const JobInfo &info, vector<string> *headers)
...
+    "x-ms-blob-type:BlockBlob\nx-ms-date:" +
+    timestamp +
+    "\nx-ms-version:2011-08-18";
+  string canonical_resource =
+    "/" + config_.access_key + "/" + config_.bucket + "/" + info.object_key;
+
+  string string_to_sign;
+  if ((info.request == JobInfo::kReqPutDotCvmfs) ||
+     (info.request == JobInfo::kReqPutCas)) {
+    snprintf(payload_size, sizeof(payload_size), "%lu", info.origin->GetSize());
```
jblomer: We can use the utility function `StringifyInt()` instead

cvmfs/s3fanout.cc
```diff
@@ -609,6 +609,59 @@ bool S3FanoutManager::MkV4Authz(const JobInfo &info, vector<string> *headers)
...
+  string timestamp = RfcTimestamp();
+  string canonical_headers =
+    "x-ms-blob-type:BlockBlob\nx-ms-date:" +
+    timestamp +
+    "\nx-ms-version:2011-08-18";
+  string canonical_resource =
+    "/" + config_.access_key + "/" + config_.bucket + "/" + info.object_key;
+
+  string string_to_sign;
+  if ((info.request == JobInfo::kReqPutDotCvmfs) ||
```
jblomer: What we are actually want to know is if the request has content or not.  We should probably write it as

    if (info.origin) {
        // sign with content length
    } else {
        // sign without content length
    }

(Note to ourselves: we should change `MkPayloadHash` in a similar way)

cvmfs/s3fanout.cc
```diff
@@ -609,6 +609,59 @@ bool S3FanoutManager::MkV4Authz(const JobInfo &info, vector<string> *headers)
...
+     (info.request == JobInfo::kReqPutCas)) {
+    snprintf(payload_size, sizeof(payload_size), "%lu", info.origin->GetSize());
+    headers->push_back("cvmfs_request: " + GetRequestString(info));
```
jblomer: I guess this has been there for testing?  It doesn't seem necessary.

cvmfs/s3fanout.cc
```diff
@@ -609,6 +609,59 @@ bool S3FanoutManager::MkV4Authz(const JobInfo &info, vector<string> *headers)
...
+    string_to_sign =
+      GetRequestString(info) +
+      string("\n\n\n") +
+      "\n\n\n\n\n\n\n\n\n" +
+      canonical_headers + "\n" +
+      canonical_resource;
+  }
+
+  string signing_key;
+  int retval = Debase64(config_.secret_key, &signing_key);
```
jblomer, line 647: That's only a small optimization but still we could do this only once per S3Fanout instance.  We can add a `GetAzureSigningKey()` function similar to `GetAwsV4SigningKey()`. The `last_signing_key_` map can be used in both functions because we'll only ever use the AWSv4 or the Azure authz scheme with a single fanout instance.

cvmfs/s3fanout.cc
```diff
@@ -609,6 +609,59 @@ bool S3FanoutManager::MkV4Authz(const JobInfo &info, vector<string> *headers)
...
+  if (!retval)
+    return false;
+
+  string signature = shash::Hmac256(signing_key, string_to_sign, true);
+  string signature64 = Base64(signature);
+
+  headers->push_back("x-ms-date: " + timestamp);
+  headers->push_back("x-ms-version: 2011-08-18");
+  headers->push_back(
+    "Authorization: SharedKey " + config_.access_key + ":" + Base64(signature));
```
jblomer, line 656: `Base64(signature)` --> `signature64`

cvmfs/s3fanout.cc
```diff
@@ -718,6 +771,9 @@ bool S3FanoutManager::MkPayloadHash(const JobInfo &info, string *hex_hash)
         *hex_hash =
           "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855";
         break;
+      case kAuthzAzure:
+        // no payload hash required for Azure signature
```
jblomer, line 770: For consistency, let's use

    hex_hash->clear();
    break;

cvmfs/s3fanout.cc
```diff
@@ -744,6 +800,9 @@ bool S3FanoutManager::MkPayloadHash(const JobInfo &info, string *hex_hash)
       *hex_hash =
         shash::Sha256Mem(data, nbytes);
       return true;
+    case kAuthzAzure:
+        // no payload hash required for Azure signature
+      return true;
```
jblomer, line 802: hex_hash->clear();
    return true;

cvmfs/s3fanout.cc
```diff
@@ -816,16 +875,18 @@ Failures S3FanoutManager::InitializeRequest(JobInfo *info, CURL *handle) const {
     assert(retval == CURLE_OK);
     retval = curl_easy_setopt(handle, CURLOPT_NOBODY, 1);
     assert(retval == CURLE_OK);
-    info->http_headers =
-      curl_slist_append(info->http_headers, "Content-Length: 0");
 
-    if (info->request == JobInfo::kReqDelete) {
+    if ((info->request == JobInfo::kReqDelete) ||
+        (info->request == JobInfo::kReqHeadPut))
```
jblomer: This should not be necessary, isn't it?

cvmfs/s3fanout.cc
```diff
@@ -816,16 +875,18 @@ Failures S3FanoutManager::InitializeRequest(JobInfo *info, CURL *handle) const {
...
+      {
       retval = curl_easy_setopt(handle, CURLOPT_CUSTOMREQUEST,
                                 GetRequestString(*info).c_str());
       assert(retval == CURLE_OK);
     } else {
       retval = curl_easy_setopt(handle, CURLOPT_CUSTOMREQUEST, NULL);
       assert(retval == CURLE_OK);
+      info->http_headers =
```
jblomer: I think we can remove this, too.  We'd then simply not set a content-length for DELETE and HEAD requests.

cvmfs/upload_s3.cc
```diff
@@ -157,7 +157,11 @@ bool S3Uploader::ParseSpoolerDefinition(
     timeout_sec_ = String2Uint64(parameter);
   }
   if (options_manager.GetValue("CVMFS_S3_REGION", &region_)) {
-    authz_method_ = s3fanout::kAuthzAwsV4;
+    if (region_ == "Azure") {
```
jblomer: Perhaps it is cleaner to introduce a new config variable, e.g. `CVMFS_S3_FLAVOR=[azure|aws]`.  If it is set to `azure`, we can at the same time also validate / set other config options, namely turning of DNS style bucket addresses and perhaps enforce https entpoints.

jblomer (changes requested):
Many thanks!  I left some minor comments.  The test failures are not related to the patch.  I still have to do a manual test run against our Ceph S3 gateways.

cvmfs/s3fanout.cc
hmeiland: done

cvmfs/s3fanout.cc
hmeiland: yes, taken out

cvmfs/s3fanout.cc
hmeiland: taking out the creation of signature64; can be done in one step...

cvmfs/s3fanout.cc
hmeiland: done

cvmfs/s3fanout.cc
hmeiland: done

cvmfs/s3fanout.cc
hmeiland: done

cvmfs/s3fanout.cc
hmeiland: done

cvmfs/s3fanout.cc
hmeiland: to optimize later...

cvmfs/s3fanout.cc
hmeiland: info.origin.IsValid was not working; now short request for HEAD and DELETE and full request with filesize for all PUT variants

cvmfs/upload_s3.cc
hmeiland: done with simple CVMFS_S3_FLAVOR=azure; no tuning on bucket etc yet...

jblomer:
@cernvm-bot cloudtest

jblomer (approved):
Looks good to me!

I'll be off next week, just in order to not forget, some follow-up tasks

- [ ] Accept only "awsv2", "awsv4", "azure" as inputs for the flavor; adjust / check other configuration options according to the flavor
- [ ] Possibly a unit test for the `MkAzureAuthz`
- [ ] Integration test platform with azurite instead of minio
- [ ] Test with the S3 benchmarker against several implementations (Ceph, AWS, Azure)

PR #2593, opened 2020-08-17
Fix ingesting tarballs with leading slashes

siscia:
Fix the ingestion of tarballs when the file inside the tarball present leading slashes.


cvmfs/sync_union_tarball.cc
```diff
@@ -276,6 +276,13 @@ std::string SyncUnionTarball::SanitizePath(const std::string &path) {
       return to_return;
     }
   }
+  if (path.length() >= 1) {
+    if (path[0] == '/') {
+      std::string to_return(path);
+      to_return.erase(0, 1);
```
jblomer: Side note, why not simply return `path.substr(1)`.

jblomer (changes requested):
Looks in principal good to me, just one suggestion for an improvement.

jblomer:
Please also add a PR for the backport to the 2.7 branch

jblomer (approved):

siscia:
Still something isn't working quite alright in the test machine... Looking into it.

jblomer:
@cernvm-bot cloudtest

PR #2598, opened 2020-09-02
Debian Packaging: Only recommend apache2

ChristianTackeGSI:
cvmfs-server currently strictly depends on apache2.
On a publisher node, one does not need apache2.
So turn the "Depends: apache2" into a "Recommends: apache2".

Recommends means "You really should install this. If you don't, expect missing functionality". Which fits very well: If one does not install apache2, one can't run a stratum-0.

### DIFF
--- a/packaging/debian/cvmfs/control.in
+++ b/packaging/debian/cvmfs/control.in
@@ -21,7 +21,8 @@ Description: CernVM File System
 Package: cvmfs-server
 Architecture: i386 amd64 armhf arm64
 #Pre-Depends: ${misc:Pre-Depends}   (preparation for multiarch support)
-Depends: psmisc, curl, attr, openssl, apache2, libcap2, libcap2-bin, lsof, rsync, jq, usbutils, sqlite3, ${misc:Depends}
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer (changes requested):
Many thanks!  Looks in principle good to me.  Could you also add `apache2` as an explicit package to test/cloud_testing/platforms/ubuntu_setup.sh

That should fix the integration tests

ChristianTackeGSI:
done.

I have not tested this! So I hope your CI will do that now.

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

jblomer:
Thanks!  It won't make it in the 2.7.4 patch release that we are cutting just now but it will be part of the 2.8 release.

ChristianTackeGSI:
> Thanks! It won't make it in the 2.7.4 patch release that we are cutting just now but it will be part of the 2.8 release.

Thanks! 2.8 is fine. It will just simplify our setup a bit.

PR #2599, opened 2020-09-03
Extend integration test case

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2600, opened 2020-09-03
Fix concurrent hosting of repositories of certain names

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2601, opened 2020-09-06
Add support for WSL2

jblomer:
Version 2 of the Windows sub sytem for Linux starts a full, virtualized Linux kernel with fuse and autofs support.  Therefore, cvmfs can be used on the available distributions almost like on their native counterparts.  One difference is that the WSL2 Linux starts like being in singe-user mode.  The /init process is not systemd but a custom one.  So systemd, even though available as a binary, cannot be used as service manager.  This patch adds a `cvmfs_config wsl2_start` command to manually start the automounter.  Whenever opening a WSL2 Linux terminal, one needs to write `sudo cvmfs_config wsl2_start` to find the normal autofs-managed /cvmfs area.

Tested on Ubuntu 20.04 / WSL2.

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -129,27 +129,64 @@ compare_versions() {
 }
 
 cvmfs_config_usage() {
- echo "Common configuration tasks for CernVM-FS"
- echo "Usage: $0 <command>"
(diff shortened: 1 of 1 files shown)


jblomer:
@cernvm-bot cloudtest

PR #2604, opened 2020-09-16
remove build.sh and replace with self contained Dockerfile

siscia:


siscia:
This runs into problem when we want to build against commits that are not in the main repo, like PR.

Not sure there is a clean way around it thought.

jblomer:

packaging/container/Dockerfile
```diff
@@ -1,13 +1,61 @@
...
+COPY --from=builder /home/sftnight/root /
+COPY --from=builder /lib64/ld-linux-x86-64.so.2 /lib64
 
-# Either a release version or a pre-release tag generated from version and git hash
-ARG VERSION
-# The build platform from which base libraries are taken (fuse, ssl, etc)
-ARG PLATFORM
+RUN mkdir -p /var/log /var/run/cvmfs
 
-LABEL maintainer="jblomer@cern.ch"
```
jblomer, line 8: We should keep the labels for properness

packaging/container/Dockerfile
```diff
@@ -1,13 +1,61 @@
...
-LABEL version="$VERSION"
-LABEL platform="$PLATFORM"
+ADD mount_cvmfs.sh /usr/bin
+ADD check_cvmfs.sh /usr/bin
+ADD terminate.sh /usr/bin
 
 # Needs to be set to the site squid
 ENV CVMFS_HTTP_PROXY=
```
jblomer, line 61: Let's add `CVMFS_CLIENT_PROFILE` from https://github.com/cvmfs/cvmfs/pull/2603/files

ci/cvmfs/container.sh
```diff
@@ -64,7 +49,7 @@ else
 fi
 
 ${CVMFS_SOURCE_LOCATION}/packaging/container/build.sh \
-  ${CVMFS_SOURCE_LOCATION} ${CVMFS_RESULT_LOCATION} ${CVMFS_BUSYBOX} ${CVMFS_TAG} \
```
jblomer, line 67: I think we want both the `CVMFS_TAG` and the `git_hash`.  The `CVMFS_TAG` provides us a linear version number, which is nice, and the git hash allows us to clone the corresponding sources.

packaging/container/Dockerfile
```diff
@@ -1,13 +1,61 @@
+FROM gitlab-registry.cern.ch/cernvm/build-images/centos_x86_64:8 as build-base
+
+FROM build-base as builder
+ARG CVMFS_COMMIT=cvmfs-2.7.4
```
jblomer, line 4: This should be unset so that we enforce a provided argument.  Otherwise we might get "silently" the wrong version built.

packaging/container/Dockerfile
```diff
@@ -1,13 +1,61 @@
...
+        -DBUILD_GEOAPI=no \
+        -DBUILD_LIBCVMFS=no \
+        -DBUILD_LIBCVMFS_CACHE=no \
+        -DINSTALL_BASH_COMPLETION=no \
+        ..
+RUN make -j4
+
+RUN mkdir ~/root && \
+        make DESTDIR=~/root install && \
+        rm -rf ~/root/usr/lib*/libcvmfs_fuse.* \
```
jblomer, line 20: `rm -rf` --> `rm -f`

jblomer, line 20: The mount helper should be removed, too

packaging/container/build.sh
```diff
@@ -10,130 +10,28 @@ set -e
...
-
-mkdir ${CVMFS_RESULT_LOCATION}/rootfs/bin
-cp $CVMFS_BUSYBOX ${CVMFS_RESULT_LOCATION}/rootfs/bin/
-${CVMFS_RESULT_LOCATION}/rootfs/bin/busybox --install ${CVMFS_RESULT_LOCATION}/rootfs/bin
-
-cd ${CVMFS_RESULT_LOCATION}/build
-cmake -DBUILD_SERVER=no -DBUILD_RECEIVER=no -DBUILD_GEOAPI=no \
-  -DBUILD_LIBCVMFS=no -DBUILD_LIBCVMFS_CACHE=no \
-  -DINSTALL_BASH_COMPLETION=no \
-  -DEXTERNALS_PREFIX=${CVMFS_RESULT_LOCATION}/externals \
```
jblomer, line 43: That's of course a pity that we have to give it up.  Now when we want to debug building the container, we always have to build the full cvmfs including the externals.  Previously, the external builds could stick around and we did not need to rebuild them. Hm, any ideas?

packaging/container/Dockerfile
jblomer, line 20: The `cvmfs_config` script should be removed, too

packaging/container/Dockerfile
```diff
@@ -1,13 +1,61 @@
...
+RUN mkdir ~/deps && \
+        find ~/root/usr/ -type f | \
+        xargs -n 1 -I {} ldd {} 2>/dev/null | \
+        awk '{print $3}' | \
+        grep /lib | sort | uniq | \
+        rsync -aL --files-from=- / ~/deps/
+
+# build our own custom busybox (BB)
+FROM build-base as BB-builder
+ARG BUSYBOX_URL=https://ecsft.cern.ch/dist/cvmfs/builddeps/busybox
```
jblomer, line 40: Let's leave that empty, too

packaging/container/Dockerfile
```diff
@@ -1,13 +1,61 @@
...
+COPY --from=BB-builder /home/sftnight/BB/bin /bin
+COPY --from=builder /home/sftnight/deps /
+COPY --from=builder /home/sftnight/root /
+COPY --from=builder /lib64/ld-linux-x86-64.so.2 /lib64
 
-# Either a release version or a pre-release tag generated from version and git hash
-ARG VERSION
-# The build platform from which base libraries are taken (fuse, ssl, etc)
-ARG PLATFORM
+RUN mkdir -p /var/log /var/run/cvmfs
```
jblomer, line 54: The container config is missing

packaging/container/build.sh
```diff
@@ -10,130 +10,28 @@ set -e
...
-#  --tag cvmfs/service:$CVMFS_TAG
+        --tag ${IMAGE_NAME} \
+        --build-arg CVMFS_COMMIT=$CVMFS_TAG \
+        --build-arg BUSYBOX_URL=$CVMFS_BUSYBOX_URL \
+        $SCRIPT_LOCATION
 
-ARCHIVE_NAME="cvmfs-service-${CVMFS_TAG}.$(uname -m)"
-IMAGE_NAME="cvmfs/service:$CVMFS_TAG"
-
-docker inspect $IMAGE_NAME
```
jblomer, line 126: That provided at some point helpful debugging output

jblomer (changes requested):
Thanks! Couple of smaller issues inline.

At least optionally, we need to be able to take the cvmfs sources from the build host.  Perhaps this can be done with another Dockerfile argument that would point to a sources directory (or alternatively clone them from github).  If we can take a source tree from the build host, we solve two issues

1. We get the sticky externals back
2. We can build PRs where the sources live in a fork

siscia:
We should restructure this as a build step not a packaging one.

The packaging step should follow and generate the tar from the docker container.

PR #2605, opened 2020-09-17
install osxfuse via Travis Homebrew addon

boegel:
I noticed that the Travis tests often fail on macOS, so I'd like to suggest to use the recommended approach for installing packages in the macOS test environment, by using the "Homebrew addon", see https://docs.travis-ci.com/user/reference/osx#homebrew and https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-macos .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,10 @@ addons:
       - valgrind
       - voms-dev
       - gdb
-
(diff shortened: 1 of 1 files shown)


siscia:
Thanks!

LGTM!

PR #2607, opened 2020-09-21
Branch001

aandvalenzuela:


jblomer:
Excellent!  I'll close this and then we can continue with [CVM-1868](https://sft.its.cern.ch/jira/browse/CVM-1868)

PR #2608, opened 2020-09-24
Fix inode tracker when path store comes close to 4G

jblomer:


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2610, opened 2020-09-29
[publish] use full path in invocation from cvmfs_server

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2611, opened 2020-09-29
[publish] improve error reporting for bad command invocations

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2612, opened 2020-10-02
CVM-1868

aandvalenzuela:
A new check has been added: The sync_mediator.cc file now checks if the .cvmfscatalog has a correct name and also if it is a regular file before creating the nested catalog.


aandvalenzuela:
Code review failed

PR #2613, opened 2020-10-06
Check if the .cvmfscatalog file is a regular file

aandvalenzuela:


CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
-#
```
jblomer, line 1: Please revert the test change

cvmfs/sync_item.h
```diff
@@ -61,6 +61,7 @@ class SyncItem {
   SyncItem();
   virtual ~SyncItem();
 
+
```
jblomer, line 64: Remove the additional line

test/src/681-nonregularfile_cvmfscatalog/main
```diff
@@ -0,0 +1,21 @@
...
+cvmfs_run_test() {
+  logfile=$1
+
+  echo "*** Create a new repository"
+  create_empty_repo $CVMFS_TEST_REPO $CVMFS_TEST_USER || return 1
+  echo "*** Start a transaction"
+  start_transaction $CVMFS_TEST_REPO || return 2
+  echo "*** Create the folder /nested"
+  mkdir /cvmfs/$CVMFS_TEST_REPO/nested || return 3
+  echo "*** Create non-regular file"
```
jblomer, line 14: --> `*** Create non-regular .cvmfscatalog file`

test/src/681-nonregularfile_cvmfscatalog/main
```diff
@@ -0,0 +1,21 @@
...
+  mkfifo /cvmfs/$CVMFS_TEST_REPO/nested/.cvmfscatalog || return 4
+  echo "*** A non-regular file should not pass"
+  publish_repo $CVMFS_TEST_REPO && return 5
+
```
jblomer, line 18: As an additional test, you can now remove the fifo and instead create a directory named `.cvmfscatalog`, which should also fail.

cvmfs/util/posix.cc
```diff
@@ -715,6 +715,16 @@ bool FileExists(const std::string &path) {
...
 }
 
+bool PathExists(const std::string &path) {
+  platform_stat64 info;
+  return ((platform_lstat(path.c_str(), &info) == 0));
+}
+
+bool RegularExists(const std::string &path) {
+  platform_stat64 info;
+  return (S_ISREG(info.st_mode));
```
jblomer, line 725: `info` is not initialized here.  In any case, I don't think we actually need `RegularExists`.  We should be able to work only with `PathExists`

cvmfs/util/posix.cc
```diff
@@ -715,6 +715,16 @@ bool FileExists(const std::string &path) {
           S_ISREG(info.st_mode));
 }
 
+bool PathExists(const std::string &path) {
```
jblomer, line 718: Let's add a unit test for this new function in test/unittest/t_util.cc

jblomer, line 718: Let's add a comment to this function, like

```
Returns true if the provided path exists independent of the file type (regular, directory, symbolic link, etc.)
```

cvmfs/sync_mediator.cc
```diff
@@ -920,9 +920,16 @@ void SyncMediator::AddDirectory(SharedPtr<SyncItem> entry) {
   }
 
   if (entry->HasCatalogMarker() &&
-      !catalog_manager_->IsTransitionPoint(entry->GetRelativePath())) {
+    !catalog_manager_->IsTransitionPoint(entry->GetRelativePath())) {
```
jblomer, line 923: Let's keep the original white spacing

cvmfs/sync_mediator.cc
```diff
@@ -920,9 +920,16 @@ void SyncMediator::AddDirectory(SharedPtr<SyncItem> entry) {
...
     CreateNestedCatalog(entry);
-  }
+    } else if (!entry->HasCatalogMarker()) {
```
jblomer, line 925: The indentation seems off here

cvmfs/sync_mediator.cc
```diff
@@ -920,9 +920,16 @@ void SyncMediator::AddDirectory(SharedPtr<SyncItem> entry) {
...
+      if (entry->IsPathPresent()) {
```
jblomer, line 926: This logic is not quite correct: we are in `AddDirectory`, so `entry` represents a directory.  Now we go into the `else` branch, so the directory `entry` does _not_ contain a .cvmfscatalog regular file.  Then we test if the directory itself is present (it should) and in the next line if it is a regular file (it isn't, we know already that it is a directory).

So this block would panic whenever we try to add a directory that does not contain a .cvmfscatalog regular file.

Let's not put the changes here, directly, but let's instead patch `SyncItem::CheckCatalogMarker()`

cvmfs/sync_item.h
```diff
@@ -282,6 +288,8 @@ class SyncItem {
   bool opaque_;                       /**< SyncUnion marked this as opaque dir*/
   bool masked_hardlink_;              /**< SyncUnion masked out the linkcount */
   bool has_catalog_marker_;           /**< directory containing .cvmfscatalog */
+  bool is_path_present_;
```
jblomer, line 291: We should be able to fix the issue without adding any new members to `SyncItem`.  I think it is sufficient to path `CheckCatalogMarker()`

jblomer (changes requested):
Thanks a lot, I think we are getting closer!

Now, I think the change we should be doing is patching `SyncItem::CheckCatalogMarker`.  There, we currently check more or less `FileExists(".cvmfscatalog")`.  Instead, we should check `PathExists(".cvmfscatalog")` and, if true, `FileExists` must also be true or else panic.

PR #2614, opened 2020-10-08
Check if the .cvmfscatalog file is a regular file

aandvalenzuela:


cvmfs/sync_item.cc
```diff
@@ -263,7 +263,18 @@ void SyncItem::CheckMarkerFiles() {
 }
 
 void SyncItem::CheckCatalogMarker() {
```
aandvalenzuela, line 265: Finally, there is no need of declaring a new `PathExists()` function since the functionality of checking the path is already implemented in `StatGeneric()`.

cvmfs/sync_item.cc
```diff
@@ -263,7 +263,18 @@ void SyncItem::CheckMarkerFiles() {
...
-  has_catalog_marker_ = FileExists(GetUnionPath() + "/.cvmfscatalog");
+  std::string path(GetUnionPath() + "/.cvmfscatalog");
+  EntryStat stat;
```
aandvalenzuela, line 267: @siscia has found that the structure `EntryStat` should also check the `error_code` just as it is done with the  `obtained` in `assert(obtained)`. Should we open a new PR?

siscia: She is referring to this code here: https://github.com/cvmfs/cvmfs/blob/e3fa33ad1543076bf5d8f9322e0c34522330d42b/cvmfs/sync_item.h#L230-L250

Where we don't check if there was an error during the stat call.

test/src/681-nonregularfile_cvmfscatalog/main
```diff
@@ -0,0 +1,35 @@
...
+  mkdir /cvmfs/$CVMFS_TEST_REPO/nested/.cvmfscatalog || return 7
+  echo "*** A non-regular file should not pass"
+  publish_repo $CVMFS_TEST_REPO && return 8
+
+  echo "*** Remove .cvmfscatalog directory"
+  rm -r /cvmfs/$CVMFS_TEST_REPO/nested/.cvmfscatalog || return 9
+  echo "*** Create non-regular .cvmfscatalog file (symbolic link)"
+  ln -s /dev/null /cvmfs/$CVMFS_TEST_REPO/nested/.cvmfscatalog || return 10
+  echo "*** A non-regular file should not pass"
+  publish_repo $CVMFS_TEST_REPO && return 11
```
aandvalenzuela: Additional tests with other non-regular .cvmfscatalog files.

jblomer:
@cernvm-bot unittest

cvmfs/sync_item.cc
```diff
@@ -263,7 +263,18 @@ void SyncItem::CheckMarkerFiles() {
...
+  StatGeneric(path, &stat, false);
+  if (stat.error_code) {
+    has_catalog_marker_ = false;
+    return;
+  }
+  if (stat.GetSyncItemType() == kItemFile) {
+    has_catalog_marker_ = true;
+    return;
+  }
+  PANIC(kLogStderr, "Error: The .cvmfscatalog file is not a regular one.");
```
jblomer: Let's use `path` here instead of `.cvmfscatalog` in order to tell the user more precisely where the error is.

test/src/681-nonregularfile_cvmfscatalog/main
```diff
@@ -0,0 +1,35 @@
...
+
+  echo "*** Remove .cvmfscatalog fifo file"
+  rm /cvmfs/$CVMFS_TEST_REPO/nested/.cvmfscatalog || return 6
+  echo "*** Create non-regular .cvmfscatalog file (directory)"
+  mkdir /cvmfs/$CVMFS_TEST_REPO/nested/.cvmfscatalog || return 7
+  echo "*** A non-regular file should not pass"
+  publish_repo $CVMFS_TEST_REPO && return 8
+
+  echo "*** Remove .cvmfscatalog directory"
+  rm -r /cvmfs/$CVMFS_TEST_REPO/nested/.cvmfscatalog || return 9
```
jblomer: `rm -r` --> `rmdir`

test/src/681-nonregularfile_cvmfscatalog/main
jblomer: Very good.  Now, for completion let's add one more case where `.cvmfscatalog` actually _is_ a regular file, in which case we want the test to succeed.

jblomer (changes requested):
Very good!  I think we're almost there!

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot unittest

jblomer (approved):

jblomer:
@cernvm-bot cloudtest

PR #2619, opened 2020-10-20
[WIP] Fix for catalogs from dirs to symlink

siscia:
Ref to: https://sft.its.cern.ch/jira/browse/CVM-1927

PR #2622, opened 2020-10-27
Fix gateway uploader finalization

jblomer:
@siscia That should fix the hang you saw when working on #2619 

(Note to ourselves: there is unfortunately quite a bit of disorder in the uploader / spooler / session context set of classes.  Tracked in CVM-1932)

### DIFF
--- a/cvmfs/session_context.cc
+++ b/cvmfs/session_context.cc
@@ -84,7 +84,8 @@ SessionContextBase::SessionContextBase()
       current_pack_mtx_(),
       objects_dispatched_(0),
       bytes_committed_(0),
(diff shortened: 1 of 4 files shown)


jblomer:
@cernvm-bot cloudtest

siscia:
Thanks! I will rebase #2619 on top of this one.

jblomer:
The standard gateway tests now go through, the "gateway - missing reflog" test still seems sick. Looking into it.

siscia:
Yes, the test keep failing. But even before the commit that I pointed out.

jblomer:
@cernvm-bot cloudtest

PR #2623, opened 2020-10-28
Allow enforcing hard timeout in integration tests

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2626, opened 2020-10-29
[test] Fix container platform setup

jblomer:
Makes podman and docker available at the same time.  Upgrades to containerd 1.4


jblomer:

PR #2629, opened 2020-10-30
importing s3 backed cmvfs and be able to publish to the cvmfs from new master node

hmeiland:
this patch will add s3 backed import capability to **cvmfs_server import** , and adds the remove_reflog_hash option to **cvmfs_server publish**; so that the first publish from the new master node can be forced.


cvmfs/server/cvmfs_server_publish.sh
```diff
@@ -59,6 +60,9 @@ cvmfs_server_publish() {
       f)
         open_fd_dialog=0
       ;;
+      r)
+        remove_reflog_hash=1
```
jblomer, line 69: Can you explain when / why this option is necessary?

jblomer (changes requested):
Many thanks, that's very useful!

We should have an integration test before merging.  In principle like `test/src/591-importrepo/main` but it can be much simpler.  I'd suggest to create a new test that would have `cvmfs_test_suites="quick"` in its preamble and that only does the following steps:

  - Create a new repository
  - Create one file in that repository
  - Remove the repository without removing the storage area using `cvmfs_server rmfs -p`
  - Re-import the repository
  - Verify that the file created at the beginning is still there

jblomer:
@cernvm-bot cloudtest

cvmfs/server/cvmfs_server_publish.sh
hmeiland: first publish after import:
[hpcadmin@testnode ~]$ sudo cvmfs_server publish my.test.repo
Using auto tag 'generic-2020-11-03T09:38:01Z'
Processing changes...
Waiting for upload of files before committing...
Committing file catalogs...
Wait for all uploads to finish
Exporting repository manifest
Statistics stored at: /var/spool/cvmfs/my.test.repo/stats.db
Tagging my.test.repo
Flushing file system buffers
Signing new manifest
cvmfs_swissknife: /home/hpcadmin/cvmfs/cvmfs/signing_tool.cc:192: SigningTool::Result SigningTool::Run(const string&, const string&, const string&, const string&, const string&, const string&, const string&, const string&, const string&, const string&, bool, bool, bool, std::vector<shash::Any>): Assertion `!reflog_chksum_path.empty()' failed.
Aborted
Signing failed

with underlying error being:
[hpcadmin@testnode ~]$ sudo /sbin/runuser -m hpcadmin -c 'cvmfs_swissknife sign           -c /etc/cvmfs/keys/my.test.repo.crt                 -k /etc/cvmfs/keys/my.test.repo.key
              -n my.test.repo                                       -u http://cvmfs.blob.core.windows.net/mpi/my.test.repo                             -m /var/spool/cvmfs/my.test.r
epo/tmp/manifest                          -t /var/spool/cvmfs/my.test.repo/tmp                      -r S3,/var/spool/cvmfs/my.test.repo/tmp,my.test.repo@/etc/cvmfs/azure.conf '
data corruption in .cvmfsreflog for my.test.repo, remove for automatic recreation or verify reflog.chksum file
reflog hash specified but reflog not present

after deletion of .cvmfsreflog in blob storage:
[hpcadmin@testnode ~]$ sudo cvmfs_server publish my.test.repo
Using auto tag 'generic-2020-11-03T09:44:04Z'
Processing changes...
Waiting for upload of files before committing...
Committing file catalogs...
Wait for all uploads to finish
Exporting repository manifest
Statistics stored at: /var/spool/cvmfs/my.test.repo/stats.db
Tagging my.test.repo
Flushing file system buffers
Signing new manifest
reflog for 'my.test.repo' not found but reflog.chksum is present; remove reflog.chksum to recreate the reflog
reflog hash specified but reflog not present
Signing failed

no reflog.chksum present:
[hpcadmin@testnode ~]$ sudo find /var/spool/cvmfs/my.test.repo/ | grep reflog

there is a reflog checksum in the manifest:
[hpcadmin@testnode ~]$ cat /var/spool/cvmfs/my.test.repo/tmp/manifest
C61b81a6da4f8062da49c11335989c4dc8de8ff6f
B16384
Rd41d8cd98f00b204e9800998ecf8427e
D240
S3
Gno
Ano
Nmy.test.repo
Xe6dae054b4de159da086e61745d6d0b49b8ff056
Hdd90a82bb48ff2d5579ad98447a064dce0ed44d9
T1604394147
M85e4f386da0d5d77c7f66cf2f58e10951f24b653
Y9c5e5e2e8383d93e4f38bab317375b49ae0eab04

since initial transaction was started with the .cvmfsreflog still present: transaction was aborted and system rebooted; this does not change the issue; 
Removing the Y line from the manifest before signing does solve the issue (since the manifest is being created dynamically, changing the file manually is not enough...

[hpcadmin@testnode ~]$ sudo cvmfs_server publish -r my.test.repo
Using auto tag 'generic-2020-11-03T09:52:59Z'
Processing changes...
Waiting for upload of files before committing...
Committing file catalogs...
Wait for all uploads to finish
Exporting repository manifest
Statistics stored at: /var/spool/cvmfs/my.test.repo/stats.db
Tagging my.test.repo
Flushing file system buffers
Signing new manifest
Remounting newly created repository revision

jblomer: Apologies for the delayed response!  Let me try to reproduce the issue.  I'm surprised that the .cvmfsreflog hash in the `Y` key doesn't fit anymore after the import.  Neither the reflog nor the manifest should have changed.  Temporarily removing the reflog is certainly a workaround but it seems to me there is an underlying issue that requires a fix.

jblomer: I can reproduce the issue. The actual problem is that the publishing fails to work with the reflog checksum provided as `Y` key in the manifest -- the publish always wants the local /var/spool/cvmfs/$reponame/reflog.chksum file (that overwrites the value from the manifest).  A quick fix would be creating that local file during import.  A better fix would be in the C++ code in the `SigningTool` to work correctly with the has from the manifest.

jblomer:
@hmeiland Did you have a chance to look into creating a test case?  Please let me know if you have any question.

jblomer (commented):
There is an additional issue with S3 import at `if [ $recreate_whitelist -eq 0 ]; then`, which subsequently checks explicitly local storage.

hmeiland:
closing pr, new pr created;

PR #2630, opened 2020-10-30
read credentials for different docker registry from env

siscia:
This change should be enough to allow us to work with docker hub after the implement their rate limiting, providing a set of configuration variables.

It is necessary to add the environment variables:

```
DUCC_DOCKERHUB_USER
DUCC_DOCKERHUB_PASS
DUCC_GITLAB_REGISTRY_USER
DUCC_GITLAB_REGISTRY_PASS
```

### DIFF
--- a/ducc/lib/image.go
+++ b/ducc/lib/image.go
@@ -505,7 +505,7 @@ func getDockerHubCredentials() (Credentials, error) {
 }
 
 func getGitlabContainersCredentials() (Credentials, error) {
-	return getCredentialsFromEnv("DUCC_GITLAB_USER", "DUCC_GITLAB_PASS")
+	return getCredentialsFromEnv("DUCC_GITLAB_REGISTRY_USER", "DUCC_GITLAB_REGISTRY_PASS")
 }
 
 func GetAuthToken(url string, credentials []Credentials) (token string, err error) {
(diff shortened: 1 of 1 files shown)


ducc/lib/image.go
```diff
@@ -483,7 +483,56 @@ func getManifestWithUsernameAndPassword(img *Image, user, pass string) ([]byte,
...
+func GetAuthToken(url string, credentials []Credentials) (token string, err error) {
+	docker, err := getDockerHubCredentials()
+	if err == nil {
+		credentials = append(credentials, docker)
+	}
+	gitlab, err := getGitlabContainersCredentials()
+	if err == nil {
+		credentials = append(credentials, gitlab)
+	}
+	for _, c := range credentials {
```
jblomer, line 520: If I understand correctly, we would probably try with wrong credentials to one of the services.  Not sure if this is harmful.  Does it make sense to look at the URL to decide whether to use Docker or Gitlab credentials?

jblomer (commented):
Thanks for the quick fix!  Looks in general good to me.  If in the future we have to support more than two registries with authentication, we should start thinking about a more structured approach to store their authentication information.

ducc/lib/image.go
siscia: I had the same thought.

I don't believe this to be harmful, and in general we avoid making a lot of those requests.

My take on this would be to try, and then, if something break, we fix it and we opt for something more structured.

PR #2632, opened 2020-11-02
Exclude integration test 015 on Ubuntu 16 / 32bit

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2634, opened 2020-11-02
Add platform name to the xml reports

aandvalenzuela:
This is just a first trial to build the CvmfsFullBuildDocker with a small test case.


jblomer (changes requested):
Many thanks!  Please separate the PR from the previous one, i.e. checkout a fresh branch starting from devel and make only the XML report related changes.

PR #2635, opened 2020-11-02
Add platform name to the xml report name

aandvalenzuela:
This is just a first trial to build the CvmfsFullBuildDocker with a small test case.


test/cloud_testing/platforms/centos7_x86_64_test.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+CVMFS_PLATFORM_NAME = "centos7_x86_64"
```
jblomer, line 3: Let's spell it `centos7-x86_64` with a dash as a separator between OS and architecture, so that we distinguish it from the underscore intrinsic to x86_64

PR #2636, opened 2020-11-02
Install nscd on integration test nodes

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2637, opened 2020-11-03
Implement wait & retry for transactions

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2640, opened 2020-11-09
[publish] detect outdated r/o layer on remote publisher

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2641, opened 2020-11-11
Cvm 1927 cvmfs dir 2 symlink unittests

siscia:
Still investigating CVM-1927


siscia:
We close this and we keep workin on #2642 which is in a cleaner state.

PR #2645, opened 2020-11-17
CVM-1928 Add platform name to the xml reports

aandvalenzuela:


test/cloud_testing/platforms/centos7_x86_64_s3_test.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+export CVMFS_PLATFORM_NAME="centos7-x86_64_s3"
```
jblomer: Let's rename to `centos7-x86_64_S3` (capital `S3`) to be consistent with the others

test/cloud_testing/platforms/centos7_x86_64_test.sh
```diff
@@ -29,7 +31,9 @@ CVMFS_TEST_CLASS_NAME=ClientIntegrationTests                                  \
                                  src/007-testjobs                             \
                                  src/084-premounted                           \
                                  --                                           \
-                                 src/0*                                       \
+                                 src/000                                      \
```
jblomer: The changes to the test case selection need to be reverted.

test/cloud_testing/platforms/fedora_test.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+export CVMFS_PLATFORM_NAME="fedora"
```
jblomer: Here it gets more tricky. We use the same test script for all Fedora platforms so we must add the fedora version dynamically. Let try with

    export CVMFS_PLATFORM_NAME="fedora$(. /etc/os-release && echo "$VERSION_ID")-$(uname -m)"

test/cloud_testing/platforms/ubuntu_s3_test.sh
```diff
@@ -1,4 +1,5 @@
 
+export CVMFS_PLATFORM_NAME="ubuntu-s3"
```
jblomer: --> `ubuntu_S3`

jblomer: Same as Fedora, we need to include version and architecture

test/cloud_testing/platforms/ubuntu_test.sh
```diff
@@ -1,4 +1,5 @@
 
+export CVMFS_PLATFORM_NAME="ubuntu"
```
jblomer: Same as Fedora, we need to include version and architecture

test/cloud_testing/platforms/yubikey_test.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+export CVMFS_PLATFORM_NAME="yubikey"
```
jblomer: Perhaps better: `centos7-x86_64-YUBIKEY`

jblomer (changes requested):
Great!  I have a few comments inline.

jblomer (approved):
Very good! Let's switch it later today.

PR #2646, opened 2020-11-18
make copy inside DUCC preserve the mtime and atime

siscia:
Address: https://sft.its.cern.ch/jira/browse/CVM-1950

Eventually, if https://github.com/otiai10/copy/pull/31 get merged, we should roll back to use that implementation.


siscia:
This change was deployed, and 5 images were ingested by the new code fixing the problem in the JIRA ticket.

Believe it is safe to merge.

I will wait tomorrow and then merge.

jblomer (approved):

PR #2648, opened 2020-11-20
Improve reflog repair when zombie hash is present in the manifest

jblomer:
Better deals with the situation where .cvmfsreflog and reflog.chksum are missing but the `Y` key is set in the manifest.


jblomer:
@cernvm-bot cloudtest cc7

PR #2650, opened 2020-11-30
Investigating test while fixing small identation issue.

siscia:
Investigating issues with this test.

The problem appears on test agains the HTTPS S3 backend, that might be correlated.


siscia:
@cernvm-bot cloudtest cc7

siscia:
The test are marked as failed, but they seems to have worked...

PR #2651, opened 2020-12-02
[libcvmfs] implement cvmfs_remount(), add cvmfs_get_revision()

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2656, opened 2020-12-10
Ducc refactor

siscia:
Overdue refactoring on DUCC code.


jblomer:
@cernvm-bot cloudtest

DrDaveD:
I got a build error when building this PR:
```
go: inconsistent vendoring in /cloud/login/dwd/work/RPMBUILD/BUILD/cvmfs-2.8.0/ducc:
	github.com/otiai10/copy@v1.3.0: is explicitly required in go.mod, but vendor/modules.txt indicates github.com/otiai10/copy@v1.2.0
	github.com/otiai10/copy@v1.2.0: is marked as replaced in vendor/modules.txt, but not replaced in go.mod

run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory
```
I see this was changed in a commit to go.mod today.  If I run `go mod vendor` as it suggests it updates the needed files and the build succeeds.

siscia:
@DrDaveD thanks a lot! In my  local workstation it was getting the dependencies from somewhere else! Fixed now!

siscia:
The process of applying one layer on top of another is still very slow.

[x] Make parallel apply (DONE)

siscia:
Singularity does not seem to set special bit files. With this refactor, we do. Are there security concerns on setting those bits? 

I feel somehow more comfortable on recreating the image like docker exposes it, so with the bit sets, but I am unsure if it is a good idea...

@jblomer @DrDaveD

jblomer:
I think we should ingest as accurately as possible, including setuid-bit, file capabilities, and special files.  We can still decide to hide/disable certain files client-side. Without the `cvmfs_suid` mount flag, for instance, suid bit and file capabilities have no effect (this is the default).

DrDaveD:
I don't think it makes much of a difference, since the end result is the same.  I suppose it might make a difference for docker someday.

siscia:
Sup by #2671

PR #2658, opened 2020-12-14
Release lock if opening the geo database fails

DrDaveD:
See [CVM-1956](https://sft.its.cern.ch/jira/browse/CVM-1956)

### DIFF
--- a/cvmfs/webapi/cvmfs_geo.py
+++ b/cvmfs/webapi/cvmfs_geo.py
@@ -47,24 +47,26 @@ def lookup_geoinfo(now, addr):
 
     if gireader is None or now > gichecktime + geo_cache_secs:
         gilock.acquire()
-        # gichecktime might have changed before acquiring the lock, look again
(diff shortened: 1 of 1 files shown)


cvmfs/webapi/cvmfs_geo.py
```diff
@@ -47,24 +47,26 @@ def lookup_geoinfo(now, addr):
...
+                if modtime != gimodtime:
+                    # database was modified, reopen it
+                    oldgireader = gireader
+                    gireader = open_geodb(gidb)
+                    gimodtime = modtime
+                    print 'cvmfs_geo: opened ' + gidb
+        finally:
+            gilock.release()
 
     return gireader.get(addr)
```
jblomer, line 71: What happens now if the file cannot be opened (I guess `gireader` is then `None`)?

DrDaveD: It throws an exception, which gets into the log.  Everything cleans up properly if the lock is released.

PR #2659, opened 2020-12-14
Log when geosort ends up switching a proxy

DrDaveD:
See [CVM-1920](https://sft.its.cern.ch/jira/browse/CVM-1920)


jblomer:
@cernvm-bot cloudtest

jblomer (changes requested):
Many thanks!

I think that we cannot be sure that `opt_proxy_groups_` is not `NULL`, so we should check that when setting `old_proxy` and `new_proxy`.

DrDaveD:
> I think that we cannot be sure that `opt_proxy_groups_` is not `NULL`, so we should check that when setting `old_proxy` and `new_proxy`.

Good catch.  Done.

PR #2663, opened 2020-12-22
Several fixes for mac and reloading

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2664, opened 2021-01-05
Improve network error handling during GC

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2665, opened 2021-01-06
Fix instantiation of Ubuntu 20 test platform

jblomer:


jblomer:
@cernvm-bot cloudtest cc8

PR #2666, opened 2021-01-06
Improve cvmfs_config status for broken mountpoints

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2668, opened 2021-01-08
Fix cvmfs_config status for inaccessible mountpoints

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2671, opened 2021-01-14
Ducc sneaky layers

siscia:
This PR contains 3 main objective. In order.

1. Overdue refactor of the DUCC code
2. Implementation of the chain for fast image/user code ingestion
3. Updates of the libraries

While from a code hygiene perspective it would have been better to have 3 separate PR, none of the steps make sense without the preceding one.

Moreover, step 3 was kind of necessary for step 2.

I also write code that I ended up not using, but that could be of some interest, so I packed it up in a separate go libraries.

https://github.com/siscia/gooverlay

If code review get difficult due to the sheer amount of changes, I can split the PR. If that does not endup being an issue, I will leave it as it is.


siscia:
Close #2656

siscia:
@cernvm-bot unittest

siscia:
@cernvm-bot unittest

PR #2672, opened 2021-02-03
Add an option for the fill ratio

dvandok:
We've been running this sensor at Nikhef for some time with a minor change. Hope this could be integrated upstream.


jblomer (changes requested):
Many thank!  Looks good to me.  Let's just make it a proper new version 1.11.

jblomer:
Being replaced by a new PR.

PR #2673, opened 2021-02-06
Fix service container config

jblomer:
@DrDaveD As discussed at the workshop, this fix makes OpenHTC servers the default for the service container

### DIFF
--- a/mount/default.d/42-container.conf
+++ b/mount/default.d/42-container.conf
@@ -2,3 +2,4 @@
 
 CVMFS_USYSLOG=/var/log/cvmfs.log
 CVMFS_CLAIM_OWNERSHIP=no
+CVMFS_USE_CDN=yes


DrDaveD:
So ... when does this go into effect?  Is a new cvmfs service container built with each new cvmfs release?

Does this make the behavior with respect to CDN and WPAD the same as the default in CernVM?

jblomer:
> So ... when does this go into effect? Is a new cvmfs service container built with each new cvmfs release?

Yes, so with 2.8.1, probably in March.

> Does this make the behavior with respect to CDN and WPAD the same as the default in CernVM?

It's the other way round, `CVMFS_USE_CDN=yes` is yet to become the default in CernVM.

DrDaveD:
I just had a chance to look at this more carefully, and I tried running the currently released cvmfs/service container with `-e CVMFS_USE_CDN=yes -e CVMFS_HTTP_PROXY=auto`.  That failed with 
```
CernVM-FS: loading Fuse module... failed to discover HTTP proxy servers (23 - proxy auto-discovery failed)
```
Would you have expected that to work?  I didn't have any luck with guessing how to get more debug info.  Can you reproduce it?

DrDaveD:
Note that it failed on the first mount, when mounting the config repo.  I think the problem is that the config repo does not support WPAD.  Indeed, if I change it to `-e CVMFS_HTTP_PROXY="auto;DIRECT"` then it succeeds.  It results still in using the openhtc.io servers for all repositories, but the local proxy is found only for the non config-repo repositories.

DrDaveD:
What we do in the EGI and OSG configurations (rpm and config repo) is the following:
```
if [ -z "$CVMFS_USE_CDN" ]; then
  if [ "$CVMFS_HTTP_PROXY" = "DIRECT" ] || [ "$CVMFS_HTTP_PROXY" = "auto;DIRECT" ]; then
    # If an explicit site proxy and CDN use/nonuse are not set, use the CDN
    CVMFS_USE_CDN=yes
  fi
fi
```
The advantage of that is that it only uses the CDN by default when no explicit squid is specified.  What do you think about doing that here instead of always enabling CVMFS_USE_CDN?

jblomer:
For the cvmfs-config.cern.ch repository, it should work with `-e CVMFS_CLIENT_PROFILE=single`

DrDaveD:
I'm not sure which question you were answering.  I don't see how setting `-e CVMFS_CLIENT_PROFILE=single` helps anything with the cvmfs/service container.

DrDaveD:
Oh it does avoid needing to set `-e CVMFS_HTTP_PROXY='auto;DIRECT'` but it doesn't affect the use of the CDN.

PR #2677, opened 2021-02-18
[ducc] Ingest images using sneaky layers

siscia:
Updates on DUCC code to support sneaky layers and chains ingestion.


jblomer (approved):
Cool!

PR #2681, opened 2021-02-25
upgrade DUCC version to follow cvmfs one

siscia:
I was looking for a better way to do this, without any success.


jblomer:
Thanks! Please also edit the top-level `bump_version.sh` script so that we don't forget it next time.

siscia:
@jblomer  This should do!

PR #2682, opened 2021-03-01
CVM-1928 Add timestamp and modify default reports

aandvalenzuela:


aandvalenzuela:
@cernvm-bot cloudtest

jblomer (approved):
Very good!

PR #2684, opened 2021-03-15
Simplify reading r/o xattrs when getting server settings

jblomer:


jblomer:
@cernvm-bot cloudtest

aandvalenzuela (commented):
I think it is a good generalization for getting the necessary extended attributes by using the same method. Occam's razor wins haha Thank you!

PR #2686, opened 2021-03-16
Fix cache hitrate reporting

jblomer:
Adds a new magic extended attribute "hitrate"


jblomer:
@cernvm-bot cloudtest

PR #2687, opened 2021-03-16
Fix several compilation warnings

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2689, opened 2021-03-17
Do not escape '@' in URI strings

mcb30:
RFC 3986 section 3.2.1 defines inline user information within a URI in
the form of "user:password@host".

The '@' character is currently escaped unconditionally when setting
CURLOPT_URL in DownloadManager::SetUrlOptions().  This causes any such
URIs to be treated as invalid and hence requests will eventually fail
with kFailBadUrl.

Fix by adding '@' to the list of characters that do not require URI
escaping.  The '@' character has no special meaning outside of the
user information field, and so this will not alter the interpretation
of any other part of the URI.

### DIFF
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -70,6 +70,7 @@ static inline bool EscapeUrlChar(char input, char output[3]) {
       ((input >= 'A') && (input <= 'Z')) ||
       ((input >= 'a') && (input <= 'z')) ||
       (input == '/') || (input == ':') || (input == '.') ||
+      (input == '@') ||
       (input == '+') || (input == '-') ||
       (input == '_') || (input == '~') ||
       (input == '[') || (input == ']') || (input == ','))


jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cpplint

mcb30:
> building for cloudtests finished: SUCCESS
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsFullBuildDocker/2767/

This seems to be reporting a success but the "cloudtest" check is still showing as "pending"?

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1049/

@jblomer I don't seem to be able to view the test results (with my CERN lightweight account): could they please be made available so that I can see the failure and fix the problem?

jblomer:
@mcb30 Ah, the Jenkins login doesn't accept your account.  That's on us, I'm unfortunately not sure if we can change that easily.  I'll get in touch with person managing the Jenkins server.  Checked the failure, it's unrelated to the patch (node down).

jblomer (approved):
Thanks a lot!

PR #2690, opened 2021-03-18
Allow for user information in DNS-resolved URI strings

mcb30:
RFC 3986 section 3.2.1 defines inline user information within a URI in
the form of "user:password@host".

The host name extraction logic in dns::PinpointHostSubstr() does not
currently understand inline user information, and will therefore end
up taking the username as the hostname.

Fix by searching for and skipping inline user information before
extracting the hostname.

This allows the CVMFS_HTTP_PROXY list to include proxies that require
authentication via the Proxy-Authorization header.

### DIFF
--- a/cvmfs/dns.cc
+++ b/cvmfs/dns.cc
@@ -68,10 +68,22 @@ static void PinpointHostSubstr(
     }
   }
 
+  // Search '@' within the hostname part and jump behind if present
+  if (*pos_begin > 0) {
+    for (i = *pos_begin; i < len; ++i) {
+      if (url[i] == '/') {
+        break;
+      } else if (url[i] == '@') {
+        *pos_begin = ++i;
+        break;
+      }
+    }
+  }
+
   // Find the end of the hostname part
   if (*pos_begin > 0) {
-    bool in_ipv6 = (url[i] == '[');
-    for (; i < len; ++i) {
+    bool in_ipv6 = (url[*pos_begin] == '[');
+    for (i = *pos_begin; i < len; ++i) {
       if (in_ipv6) {
         if (url[i] != ']')
           continue;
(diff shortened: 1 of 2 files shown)


cvmfs/dns.cc
```diff
@@ -68,10 +68,22 @@ static void PinpointHostSubstr(
     }
   }
 
+  // Search '@' within the hostname part and jump behind if present
+  if (*pos_begin > 0) {
+    for (i = *pos_begin; i < len; ++i) {
+      if (url[i] == '/') {
+	break;
```
jblomer: The white-spacing seems a bit off

cvmfs/dns.cc
```diff
@@ -68,10 +68,22 @@ static void PinpointHostSubstr(
...
+      } else if (url[i] == '@') {
+	*pos_begin = ++i;
```
jblomer: Same here

jblomer: Do we need to check here that `pos_begin` does not go beyond the string? I think we should have a test with `http://user:pw@` as an input.

jblomer (changes requested):
Many thanks!  I added a few minor comments.

jblomer:
@cernvm-bot cpplint

cvmfs/dns.cc
mcb30: Looks like a tabs-vs-spaces issue

cvmfs/dns.cc
mcb30: The character at `url[i]` is guaranteed here to be `'@'` (and therefore not the string terminator), so the subsequent character at `url[++i]` is guaranteed not to be beyond the end of the string.  But I'll add a test case for `http://user:pw@` anyway

mcb30:
@jblomer Comments should all now be addressed; thanks for reviewing!

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot cloudtest

mcb30:
I'm still not able to view the cloudtests log: is this another case of a failure unrelated to the change being tested?

jblomer (approved):
Looks good to me!

(We have a few flaky integration tests)

PR #2692, opened 2021-03-19
add spec file for build on opensuse build servcie OBS

skluth:
experts pls have a look, there are still multiple packaging issues
flagged by OBS which I am not able to resolve, because they
need moving files or renaming directories


jblomer:
Many thanks for the PR!

Could you provide a little more context on how this spec file differs from the existing `cvmfs-universal.spec`?

skluth:
do a 

diff cvmfs.spec cvmfs-universal.spec

to see the changes. I started from cvmfs-universal.spec and had to put in naming directories etc to follow rpm spec
rules of OBS. This is not a finished work, as you can see if you look at the build logs I have on OBS. I think a cvmfs developer should take a look at the remaining issues in OBS, and of course also if your local builds still work. OBS does a lot of spec file quality and sanity checks which tend to get overlooked in local rpm packaging.

jblomer:
If it is supposed to replace the `cvmfs-universal.spec` file, please overwrite the file in the PR.  We can then see the diff directly in github and we can also run the usual build & test procedure to check for errors.

skluth:
One problem was that OBS expects the spec file name to be <project>.spec, i.e. cvmfs.spec. So the renaming is part of the proposed changes. 

There are other issues: the cvmfs.tar.gz with the packaging subfolder unpacks in an unusual way to cvmfs/cvmfs which leads to build failure on OBS. I had to get the cvmfs.tar.gz from https://ecsft.cern.ch/dist/cvmfs, which lacks the packaging subfolder though, and upload that together with the renamed spec file from this repo to have a start on OBS. 

It would be a great help for OBS (or COPR) or any other rpm builds to post (a fully working) cvmfs.spec on https://ecsft.cern.ch/dist/cvmfs

I don't understand how the cvmfs.tar.gz on https://ecsft.cern.ch/dist/cvmfs (w/o packaging and with regular unpacking) is generated from the repo, but at that point the cvmfs.spec should also be transported to https://ecsft.cern.ch/dist/cvmfs .

jblomer:
I see, thanks for the additional information!

I'd like to keep only one spec file in the sources. If we change the name from `cvmfs-universal.spec` to `cvmfs.spec`, it would need further adjustments to keep our build and tests from breaking.  We might still decide to change the name but for this PR, please use the `cvmfs-universal.spec` name.

The source tarball is created by [a script in the ci folder](https://github.com/cvmfs/cvmfs/blob/devel/ci/common.sh#L67). Perhaps we could extend the PR to copy the `cvmfs-universal.spec` file as a `cvmfs.spec` file in the source tarball.

skluth:
ok I did update the PR pls have a look
I think for the installation script changes we should open another PR, to keep things separate.

skluth:
I'll create an issue for the script changes since it is not clear at all to me which scripts to change and how, and for what purpose.

DrDaveD:
> One problem was that OBS expects the spec file name to be \<project\>.spec, i.e. cvmfs.spec. So the renaming is part of the proposed changes.

I don't believe that is true.  In my [OBS attempt to build cvmfs](https://build.opensuse.org/package/show/home:drdaved:cvmfs/cvmfs) I used cvmfs-universal.spec and although the CentOS7 build [failed](https://build.opensuse.org/package/live_build_log/home:drdaved:cvmfs/cvmfs/CentOS_7/x86_64) it clearly used cvmfs-universal.spec.  It got as far as the end of the rpm build where it complained about unpackaged files.  I strarted with your cvmfs.spec and simply renamed it to cvmfs-universal.spec.

DrDaveD:
> I don't understand how the cvmfs.tar.gz on https://ecsft.cern.ch/dist/cvmfs (w/o packaging and with regular unpacking) is generated from the repo, but at that point the cvmfs.spec should also be transported to https://ecsft.cern.ch/dist/cvmfs .

I don't think that is necessary either.  It would be best if a single cvmfs-universal.spec worked for OBS and non-OBS, and if we get that then at that point OBS can download it from github.

skluth:
Ok, seems to work for you ... I still don't fully grasp all of OBSs automatisms ... I looked at other packages and they all had <package>.spec and with that is worked. 

AFAICT there is at least a convention that the spec file name should be "package".spec 

https://docs.fedoraproject.org/en-US/packaging-guidelines/#_spec_files

skluth:
> > I don't understand how the cvmfs.tar.gz on https://ecsft.cern.ch/dist/cvmfs (w/o packaging and with regular unpacking) is generated from the repo, but at that point the cvmfs.spec should also be transported to https://ecsft.cern.ch/dist/cvmfs .
> 
> I don't think that is necessary either. It would be best if a single cvmfs-universal.spec worked for OBS and non-OBS, and if we get that then at that point OBS can download it from github.

yes I fully agree "there can only be one", that is what I wanted to say. But due to e.g. fedora naming convention is should be cvmfs.spec, and yes, one should be able to link to it as e.g. a "OBS service" directly out of the github repo, or from the ecsft repo.

Note that currently the tarball from github (if you click https://github.com/cvmfs/cvmfs/releases) is broken (unpacks to cvmfs/cvmfs) for automated build systems. This is greatly confusing to non-insiders ...

DrDaveD:
>> It would be best if a single cvmfs-universal.spec worked for OBS and non-OBS, and if we get that then at that point OBS can download it from github.
> 
> yes I fully agree "there can only be one", that is what I wanted to say. But due to e.g. fedora naming convention is should be cvmfs.spec, and yes, one should be able to link to it as e.g. a "OBS service" directly out of the github repo, or from the ecsft repo.

It might be nice to change the name, but I don't think there's a compelling need to stick with the Fedora naming convention.  cvmfs-universal.spec works too.  

> Note that currently the tarball from github (if you click https://github.com/cvmfs/cvmfs/releases) is broken (unpacks to cvmfs/cvmfs) for automated build systems. This is greatly confusing to non-insiders ...

That's true, and it's because cvmfs uses tag names that start with 'cvmfs' even though github always prepends  repository names to the directory names inside tarballs that it dynamically creates from the archives/refs/tags/<repo>-<version>.tar.gz.  It's a github convention, and cvmfs hasn't used those downloads so far so the issue hasn't come up.  It's not clear to me that it is worth changing, however, since there is an alternate download location.

You haven't mentioned it in this ticket, but where you first brought this up on the [CernVM forum](https://cernvm-forum.cern.ch/t/build-cvmfs-packages-on-obs-howto/72) you said your motivation was to get an OpenSUSE Tumbleweed build.  So far the CernVM project hasn't tried to use OBS because they instead have a large set of VMs in their own build system. It may be less effort for the CernVM project to set up an OpenSUSE Tumbleweed VM in their build system than to try to make the complicated cvmfs build work on OBS with all its different constraints.

skluth:
> > > It would be best if a single cvmfs-universal.spec worked for OBS and non-OBS, and if we get that then at that point OBS can download it from github.
> > 
> > 
> > yes I fully agree "there can only be one", that is what I wanted to say. But due to e.g. fedora naming convention is should be cvmfs.spec, and yes, one should be able to link to it as e.g. a "OBS service" directly out of the github repo, or from the ecsft repo.
> 
> It might be nice to change the name, but I don't think there's a compelling need to stick with the Fedora naming convention. cvmfs-universal.spec works too.
>
> > Note that currently the tarball from github (if you click https://github.com/cvmfs/cvmfs/releases) is broken (unpacks to cvmfs/cvmfs) for automated build systems. This is greatly confusing to non-insiders ...
> 
> That's true, and it's because cvmfs uses tag names that start with 'cvmfs' even though github always prepends repository names to the directory names inside tarballs that it dynamically creates from the archives/refs/tags/-.tar.gz. It's a github convention, and cvmfs hasn't used those downloads so far so the issue hasn't come up. It's not clear to me that it is worth changing, however, since there is an alternate download location.
> 
> You haven't mentioned it in this ticket, but where you first brought this up on the [CernVM forum](https://cernvm-forum.cern.ch/t/build-cvmfs-packages-on-obs-howto/72) you said your motivation was to get an OpenSUSE Tumbleweed build. So far the CernVM project hasn't tried to use OBS because they instead have a large set of VMs in their own build system. It may be less effort for the CernVM project to set up an OpenSUSE Tumbleweed VM in their build system than to try to make the complicated cvmfs build work on OBS with all its different constraints.

There are two objectives:
1) cvmfs can become a package in fedora or opensuse (or debian) and thus become available to a much broader audience. For this cvmfs packaging has to follow the distro rules of course. A clean OBS or COPR build proves that to the distro. This also requires the spec file to have the proper name cvmfs.spec.

2) If a cvmfs build based on OBS "service links" (or equivalent in COPR) exists and works it is practically zero work for the cvmfs team to support TW, and various other opensuse variants, Fedora, Centos8 etc pp. OBS (or COPR) automatically rebuilds for any upstream changes, and also for any dependency package changes in the distro. If you do it yourself, you'll have to maintain and update a TW VM frequently and manually, and that is guaranteed to fall behind and become inconsistent.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -26,7 +26,7 @@
 %endif
 
 # List of platforms that require systemd/autofs fix as described in CVM-1200
-%if 0%{?rhel} >= 7 || 0%{?fedora} || 0%{?sle12}
+%if 0%{?rhel} >= 7 || 0%{?fedora} || 0%{?sle12} || 0%{?suse_version} > 1500
```
jblomer, line 29: Let's have a preamble check that looks at `suse_version` and sets the `sle15` dist tag, along the lines of SLES12

packaging/rpm/cvmfs-universal.spec
```diff
@@ -36,7 +36,7 @@
 %endif
 
 %define cvmfs_python python
-%if 0%{?el8} || 0%{?fedora} >= 31
+%if 0%{?el8} || 0%{?fedora} >= 31 || 0%{?suse_version} > 1500 || 0%{?rhel} >= 8
```
jblomer: The `0%{?rhel} >= 8` should have been covered by `0%{?el8}`, no?

packaging/rpm/cvmfs-universal.spec
```diff
@@ -53,7 +53,8 @@
 
 Summary: CernVM File System
 Name: cvmfs
-Version: 2.9.0
+URL: https://cernvm.cern.ch/fs/
+Version: 2.8.0
```
jblomer, line 58: For the current devel branch, the version should be 2.9.0 (i.e. the next release)

packaging/rpm/cvmfs-universal.spec
```diff
@@ -93,6 +94,11 @@ BuildRequires: pkgconfig
 BuildRequires: %{cvmfs_python}-devel
 BuildRequires: unzip
 
+BuildRequires: zlib-devel
+
+BuildRequires: systemd
+
```
jblomer: Let's just keep one empty line between the `BuildRequires` and the `Requires` block

packaging/rpm/cvmfs-universal.spec
```diff
@@ -93,6 +94,11 @@ BuildRequires: pkgconfig
 BuildRequires: %{cvmfs_python}-devel
 BuildRequires: unzip
 
+BuildRequires: zlib-devel
+
+BuildRequires: systemd
```
jblomer, line 100: Interesting that this is required. This would break the SLES11 build though so should be conditional.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -103,12 +109,12 @@ Requires: autofs
 Requires: fuse
 Requires: curl
 Requires: attr
-Requires: zlib
+# Requires: zlib
```
jblomer: You can just remove the line if not needed

packaging/rpm/cvmfs-universal.spec
```diff
@@ -103,12 +109,12 @@ Requires: autofs
...
 Requires: gdb
 # Account for different package names
 %if 0%{?suse_version}
 Requires: aaa_base
-Requires: libfuse2
+# Requires: libfuse2
```
jblomer: I guess the fuse libraries are automatically found by rpmbuild? In this case, just remove the line. (In case of name change, we need it working for SLES 11 and SLES 12, too)

packaging/rpm/cvmfs-universal.spec
```diff
@@ -183,11 +189,9 @@ BuildRequires: %{cvmfs_python}-devel
 BuildRequires: libcap-devel
 BuildRequires: unzip
 BuildRequires: %{cvmfs_python}-setuptools
-%if 0%{?suse_version}
-Requires: insserv
```
jblomer, line 187: Requires for older SLES

packaging/rpm/cvmfs-universal.spec
```diff
@@ -198,7 +202,7 @@ Requires: gzip
 Requires: attr
 Requires: openssl
 Requires: httpd
-Requires: libcap
+# Requires: libcap
```
jblomer: Remove if unnecessary

packaging/rpm/cvmfs-universal.spec
```diff
@@ -353,12 +357,13 @@ exit 0
   if [ $? -ne 0 ]; then
     /usr/sbin/groupadd -r cvmfs
   fi
-%endif
 
-/usr/bin/getent group fuse | grep -q cvmfs
-if [ $? -ne 0 ]; then
-  /usr/sbin/usermod -aG fuse cvmfs > /dev/null 2>&1 || :
-fi
+  /usr/bin/getent group fuse | grep -q cvmfs
```
jblomer, line 359: Did SLES15 remove the fuse group?

packaging/rpm/cvmfs-universal.spec
```diff
@@ -443,13 +449,14 @@ restorecon -R /var/lib/cvmfs
 if [ -d /var/run/cvmfs ]; then
   /usr/bin/cvmfs_config reload
 fi
-:
+
```
jblomer, line 449: Let's remove the line

packaging/rpm/cvmfs-universal.spec
```diff
@@ -499,10 +508,13 @@ if [ $1 -eq 0 ]; then
   /usr/sbin/semanage port -d -t http_port_t -p tcp 8000 2>/dev/null || :
 fi
 %endif
+/sbin/ldconfig
+
```
jblomer, line 505: Let's remove the empty line

packaging/rpm/cvmfs-universal.spec
```diff
@@ -499,10 +508,13 @@ if [ $1 -eq 0 ]; then
   /usr/sbin/semanage port -d -t http_port_t -p tcp 8000 2>/dev/null || :
 fi
 %endif
+/sbin/ldconfig
```
jblomer, line 504: Good catch.  Shouldn't we also have a call to `ldconfig` for the client `%postun`?

packaging/rpm/cvmfs-universal.spec
```diff
@@ -513,26 +525,37 @@ fi
...
 /usr/libexec/cvmfs/authz/cvmfs_allow_helper
 /usr/libexec/cvmfs/authz/cvmfs_deny_helper
+%dir /usr/libexec/cvmfs/cache
 /usr/libexec/cvmfs/cache/cvmfs_cache_ram
 /usr/libexec/cvmfs/cache/cvmfs_cache_posix
+
 %{_sysconfdir}/auto.cvmfs
-%{_sysconfdir}/cvmfs/config.sh
+%dir %{_sysconfdir}/cvmfs
+%config %{_sysconfdir}/cvmfs/config.sh
```
jblomer, line 533: This is not a configuration file but a bash source code file, so it should fall under the normal update policy

packaging/rpm/cvmfs-universal.spec
```diff
@@ -513,26 +525,37 @@ fi
 %{_bindir}/cvmfs_talk
 %{_bindir}/cvmfs_fsck
 %{_bindir}/cvmfs_config
+
+%dir /usr/libexec/cvmfs
```
jblomer, line 522: For this and the following `%dir` directives, they should not be required, rpmbuild automatically creates the parent directories of the specified files

packaging/rpm/cvmfs-universal.spec
```diff
@@ -513,26 +525,37 @@ fi
...
+%{_sysconfdir}/cvmfs/default.d
+
 %if 0%{?selinux_cvmfs}
 %{_datadir}/selinux/mls/cvmfs.pp
 %{_datadir}/selinux/strict/cvmfs.pp
 %{_datadir}/selinux/targeted/cvmfs.pp
 %endif
+
 %if 0%{?systemd_autofs_patch}
+/usr/lib/systemd/system/autofs.service.d
```
jblomer, line 543: Can be removed, I think

packaging/rpm/cvmfs-universal.spec
```diff
@@ -513,26 +525,37 @@ fi
...
 /usr/lib/systemd/system/autofs.service.d/50-cvmfs.conf
 %endif
+
 /sbin/mount.cvmfs
 %dir %{_sysconfdir}/cvmfs/config.d
 %dir %{_sysconfdir}/cvmfs/domain.d
 %attr(700,cvmfs,cvmfs) %dir /var/lib/cvmfs
-%{_sysconfdir}/cvmfs/default.d/README
+%config %{_sysconfdir}/cvmfs/default.d/README
```
jblomer, line 551: It seems strange to mark the `README` as something the user would further edit

packaging/rpm/cvmfs-universal.spec
```diff
@@ -572,18 +595,22 @@ fi
 %{_libdir}/libcvmfs_server.so.%{version}
 %{_libdir}/libcvmfs_server_debug.so
 %{_libdir}/libcvmfs_server_debug.so.%{version}
-%{_sysconfdir}/cvmfs/cvmfs_server_hooks.sh.demo
+%config %{_sysconfdir}/cvmfs/cvmfs_server_hooks.sh.demo
 %dir %{_sysconfdir}/cvmfs/repositories.d
+%dir /var/www/
```
jblomer, line 593: Can be removed, I think

packaging/rpm/cvmfs-universal.spec
```diff
@@ -572,18 +595,22 @@ fi
 %{_libdir}/libcvmfs_server.so.%{version}
 %{_libdir}/libcvmfs_server_debug.so
 %{_libdir}/libcvmfs_server_debug.so.%{version}
-%{_sysconfdir}/cvmfs/cvmfs_server_hooks.sh.demo
+%config %{_sysconfdir}/cvmfs/cvmfs_server_hooks.sh.demo
```
jblomer, line 591: Probably also not a "config" file in the sense of RPM

packaging/rpm/cvmfs-universal.spec
```diff
@@ -572,18 +595,22 @@ fi
...
 %dir %{_sysconfdir}/cvmfs/repositories.d
+%dir /var/www/
+/var/www/wsgi-scripts
```
jblomer, line 594: Can be removed, I think

packaging/rpm/cvmfs-universal.spec
```diff
@@ -572,18 +595,22 @@ fi
...
 /var/www/wsgi-scripts/cvmfs-server/cvmfs-api.wsgi
 /usr/share/cvmfs-server/
 /var/lib/cvmfs-server/
+%dir /var/spool/cvmfs
```
jblomer, line 598: Can be removed, I think

DrDaveD:
If one of us can get OBS to work it would be good to support it.  You and I have more experience on OBS than the CernVM team does so it's probably up to us to work through the problems.  It seems like the build is not far off from working.  The major problem remaining seems to be that it is generating /usr/src/debug files despite the fact that there is `%define debug_package %{nil}` which is supposed to disable that.  I haven't been able to find anything with google to give a hint on what to do about it.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -572,18 +595,22 @@ fi
...
 /var/spool/cvmfs/README
 %doc COPYING AUTHORS README.md ChangeLog
 
 %files shrinkwrap
 %defattr(-,root,root)
 %{_bindir}/cvmfs_shrinkwrap
-/usr/libexec/cvmfs/shrinkwrap/spec_builder.py*
+%dir /usr/libexec/cvmfs/shrinkwrap
+/usr/libexec/cvmfs/shrinkwrap/spec_builder.py
```
jblomer, line 606: I guess we should have written down _why_ we want the bytecode packaged back at the time.  There is an [extra commit](https://github.com/cvmfs/cvmfs/pull/2231/commits/6135728d9d152d91fcf4d65248b880117fb06ce7) for this, so I guess there was a reason. Is it important to leave the byte code out? If it is, we can just see what happens and if we hit the original issue again.

jblomer (changes requested):
Thanks a lot for the patch!  See comments inline.  Please also add a change log entry.

I would like to provide from our infrastructure a SLES15 build (see [JIRA ticket](https://sft.its.cern.ch/jira/browse/CVM-1656)). Assuming that such a build works fine on Tumbleweed, too, I don't think we would add an extra build & test node for the rolling release.

Making 3rd party builds easy is great to figure out potential problems early on and to enable cvmfs on less used platforms. However, we'll continue to provide our prebuilt packages on a set of core / production platforms in order to facilitate the support. In case of problems, we give priority to the issues that come from binaries that were built in a well-controlled environment and where we have the infrastructure to reproduce problems readily available.

packaging/rpm/cvmfs-universal.spec
skluth: sure

packaging/rpm/cvmfs-universal.spec
skluth: the OBS build VMs are very minimal and don't need systemd to function ... I was surprised as well but it fixed an error about missing systemd. I'll try to add SLE11 on OBS to make sure conditionals work.

packaging/rpm/cvmfs-universal.spec
skluth: ok

packaging/rpm/cvmfs-universal.spec
skluth: adding sle11 and sle12 to the OBS build should tell us

packaging/rpm/cvmfs-universal.spec
skluth: trying with sle11, sle12

packaging/rpm/cvmfs-universal.spec
skluth: ok

packaging/rpm/cvmfs-universal.spec
skluth: group fuse is not there on TW, not sure about sle15.

packaging/rpm/cvmfs-universal.spec
skluth: ok

packaging/rpm/cvmfs-universal.spec
skluth: ok

packaging/rpm/cvmfs-universal.spec
skluth: not sure, OBS complained about the missing ldconfig which I insertetd

packaging/rpm/cvmfs-universal.spec
skluth: OBS complained ... I think they don't want non-config files in %{_sysconfdir} and search for e.g. .sh files. This is one of the examples where for a proper solution the file should be moved out of %{_sysconfdir} into a place for bash source files. Marking it %config is a hack.

packaging/rpm/cvmfs-universal.spec
skluth: OBS wants them: by rpm rules you can't use directories which the package doesn't own. This is about making sure packages don't interfere.

packaging/rpm/cvmfs-universal.spec
skluth: ok lets try

packaging/rpm/cvmfs-universal.spec
skluth: see above, its a hack to make OBS happy. Should be moved to a place for READMEs, or leave the hack in so OBS knows we know what we are doing

packaging/rpm/cvmfs-universal.spec
skluth: OBS wants it ... and it is just another hack: by FHS a package can't own /var/www. We should have sth.l. /var/www/cvmfs/... to make it unique. I don't know if that breaks other things in the cvmfs universe though.

packaging/rpm/cvmfs-universal.spec
skluth: see above, a hack. sh files should properly live in other places, or marked.

packaging/rpm/cvmfs-universal.spec
skluth: see above, /var/www cannot belong to a package and we should have /var/www/cvmfs/...

packaging/rpm/cvmfs-universal.spec
skluth: OBS wants it for ownership bookkeeping.

packaging/rpm/cvmfs-universal.spec
skluth: I think I just removed the * because I didn't know what it was doing there ... it seems strange to ship pyc files right? python will create them the first time it gets executed. I would really worry if that doesn't work. Lets try.

skluth:
> Thanks a lot for the patch! See comments inline. Please also add a change log entry.
> 
> I would like to provide from our infrastructure a SLES15 build (see [JIRA ticket](https://sft.its.cern.ch/jira/browse/CVM-1656)). Assuming that such a build works fine on Tumbleweed, too, I don't think we would add an extra build & test node for the rolling release.

Thats a long shot, TW is meant to integrate upstream changes, so by construction you can't rely on TW being compatible with sle15. Quite often it is, but you can't rely on it. TW really needs a constantly updated build tracking TW changes. That is actually my original motivation for doing the OBS build. 

> Making 3rd party builds easy is great to figure out potential problems early on and to enable cvmfs on less used platforms. However, we'll continue to provide our prebuilt packages on a set of core / production platforms in order to facilitate the support. In case of problems, we give priority to the issues that come from binaries that were built in a well-controlled environment and where we have the infrastructure to reproduce problems readily available.

Yes sure. But it isolates you from your user community, because nobody can do the build in exactly the same way as you do it. When I worked with opensuse on the fuse crashes they gave us kernels built on OBS with their patches, that was very quick and transparent.

packaging/rpm/cvmfs-universal.spec
skluth: not sure how to do that https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto does not have a specific tag for sle15, but suse_version > 1500 should do it.

packaging/rpm/cvmfs-universal.spec
skluth: but on ecsft.cern.ch/cvmfs there is no 2.9.0 tar.gz so I can't test that but I'll change it before updating.

packaging/rpm/cvmfs-universal.spec
skluth: I'll try

skluth: ok centos8 still works

packaging/rpm/cvmfs-universal.spec
skluth: the sle11 build works, but there is a packaging error at the end. The sle12 build breaks with sth in protobuf. Anyway, we can resolve this issue.

packaging/rpm/cvmfs-universal.spec
skluth: looks like they work

packaging/rpm/cvmfs-universal.spec
skluth: sle11 build actually works so ok

packaging/rpm/cvmfs-universal.spec
skluth: oh its again directory ownership policies, packages are expected to name everything.

jblomer:
@cernvm-bot cloudtest

packaging/rpm/cvmfs-universal.spec
jblomer: Without reverting to 2.9.0, we cannot test the PR.

jblomer:
Some of the fixes were taken into #2704 
Two more things came to my mind
  - Regarding the URL: The stable branch points to sources that are already released, so there the URL in the spec file should work
  - To reproduce the packages we build, you can use our build images which are available here: https://gitlab.cern.ch/cernvm/build-images/container_registry

jblomer:
I think the key points of the PR are addressed, closing. Please feel free to reopen if necessary.

PR #2694, opened 2021-03-22
Do not include an explicit default port number within S3 upload URI

mcb30:
The S3 uploader will currently always include an explicit port number
within URIs (e.g. "http://mybucket.s3.us-east-1.amazonaws.com:80")
even when using the default HTTP port 80.

This is not incorrect, but unfortunately triggers a bug within libcurl
(see https://github.com/curl/curl/issues/6769) that causes it to
construct requests that will be rejected if they happen to pass
through HAProxy.

Work around this libcurl bug by omitting an explicit port number from
the constructed URI when the default port is used.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot unittest

jblomer (approved):
Interesting catch on the curl side!

PR #2695, opened 2021-03-22
[shrinkwrap] Fix spec file parsing

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2699, opened 2021-04-08
Fail readconfig when parsing config fails

DrDaveD:
I noticed that if there's any problem parsing a repository's configuration, cvmfs_config showconfig does show a failure but instead shows a bunch of empty variables.  This changes the cvmfs_readconfig function to return an error if cvmfs2 -o  parse returns an error.

### DIFF
--- a/mount/config.sh
+++ b/mount/config.sh
@@ -28,6 +28,10 @@ cvmfs_readconfig() {
   local domain; domain=`cvmfs_getdomain $fqrn`
 
   CVMFS_PARMS=$(cvmfs2 -o parse "$fqrn" /)
+  local ret=$?
+  if [ $ret != 0 ]; then
+    return $ret
+  fi
   unset CVMFS_CACHE_DIR
   eval "$CVMFS_PARMS"
 
--- a/test/test_functions
+++ b/test/test_functions
@@ -490,9 +490,13 @@ cvmfs_mount() {
   sudo sh -c "echo CVMFS_REPOSITORY_TAG=TESTING > /etc/cvmfs/config.d/cvmfs-config.cern.ch.local" || return 100
 
   # add additional parameters
+  local proberunner
   while [ $# -gt 0 ]; do
     local param="$1"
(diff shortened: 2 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
The patch causes integration test 083 to fail.  There is not much output

```
      -- Testing Run suid binaries (Fri Apr  9 00:00:37 CEST 2021 / test number 083)
restarting autofs...
using systemd
*** run suid binary without special mount option
My persona is 1000/1000
*** Persona: My persona is 1000/1000
*** unmount cernvm-prod.cern.ch
*** re-mount and try again with suid support enabled
execution took 3.977 seconds
Testcase failed with RETVAL 20
```

Can you run the integration tests locally?  Otherwise I'll have a look.

DrDaveD:
The problem is is that the `cvmfs_config probe` in the [cvmfs_mount test function](https://github.com/cvmfs/cvmfs/blob/devel/test/test_functions#L507) is failing with an error
```
PANIC: /cloud/login/dwd/work/RPMBUILD/BUILD/cvmfs-2.8.1/cvmfs/loader.cc : 786
must be root to mount with suid option
Failed to read CernVM-FS configuration
```
If I put a sudo ahead of that, it succeeds.

Without the change in this PR, the cvmfs_config probe still prints the 'must be root to mount with suid option' but it does not print 'Failed to read CernVM-FS configuration' and it returns a success code instead of an error code.

Do you want me to change cvmfs_mount to always put in sudo, or only when CVMFS_SUID=yes?

DrDaveD:
I went ahead and put it in only when CVMFS_SUID=yes.  If you think it should instead always use sudo, I will back it out.

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

DrDaveD:
Should this be merged also into 2.8?

jblomer:
Yep, I cherry-picked the commits.  Same for #2700

PR #2700, opened 2021-04-08
Skip attempting to parse variable names with spaces in them

DrDaveD:
This makes BashOptionsManager::ParsePath() behave like TryParsePath() in skipping attempting to read back variable names with spaces in them.  This prevents cvmfs2 -o parse from crashing on such lines.

### DIFF
--- a/cvmfs/options.cc
+++ b/cvmfs/options.cc
@@ -78,7 +78,7 @@ string OptionsManager::SanitizeParameterAssignment(string *line,
     if (comment_idx != string::npos)
       *line = line->substr(0, comment_idx);
     *line = Trim(*line);
-    if (line->empty() || line->find("if ") == 0)
+    if (line->empty())
       return "";
     *tokens = SplitString(*line, '=');
     if (tokens->size() < 2)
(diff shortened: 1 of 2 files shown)


DrDaveD:
Come to think of it, there probably ought to be more common code between TryParsePath and ParsePath so a problem like this difference between the two doesn't crop up in the future.

jblomer:
@cernvm-bot cloudtest

jblomer:
I think that is a good idea. We could factor out a function `SanitizeParameterAssignment`. So you want to amend the PR?

DrDaveD:
I created the common function.  There were slight differences in both, and I included checks from both.  I see now though that the check for "if " is redundant because it will be caught later by the check for any blank; I'll take that out.

The OptionsManager::TrimParameter protected function is now only called from OptionsManager::SanitizeParameterAssignment.  It could be deleted if you like, and just incorporated into SanitizeParameterAssignment.

jblomer:
@cernvm-bot cloudtest

DrDaveD:
I tried changing TrimParameter into private as we discussed but it causes a compilation error:
```
'TrimParameter' was not declared in this scope
```

jblomer:
I actually meant making `SanitizeParameterAssignment` private, but now I realize that it is fine as is.

jblomer (approved):

PR #2705, opened 2021-04-23
Register redundant bulk hashes in filestats db

jblomer:


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2706, opened 2021-04-24
Add missing install() block for cvmfs_receiver_debug

mcb30:


jblomer:
Many thanks for the patch!  This addresses [https://sft.its.cern.ch/jira/browse/CVM-1988](https://sft.its.cern.ch/jira/browse/CVM-1988)

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

jblomer (changes requested):
I approved to early.  I realized that with this change, also the RPM spec file and the debian package sources should be updated as well

mcb30:
> I approved to early. I realized that with this change, also the RPM spec file and the debian package sources should be updated as well

Pushed two further commits (untested) to update the spec file and Debian package

jblomer:
@cernvm-bot cloudtest

Issue #2707, opened 2021-04-27
Nested catalog merge optimisation


mcb30:
I have done some investigation of using `cvmfs-gateway` to handle large volumes of grafted files.  My experiments suggest that we are currently missing a key optimisation that could increase the throughput by a factor of a hundred or more.

In `catalog_diff_tool_impl.h` there exists an optimisation:
https://github.com/cvmfs/cvmfs/blob/8e0712b0230afe859fc7b5bd421878b4d64677f8/cvmfs/catalog_diff_tool_impl.h#L173-L178

The above optimisation eliminates the need to recurse into nested catalog mountpoints to find differences when the nested catalogs to be compared are known to be identical.  With a well-structured repository, this can eliminate 99% of the work required to find the difference between old and new catalogs.

A similar optimisation is available when performing a merge operation.  At the point that the `ReportModification()` callback is called: if both `entry1` and `entry2` are nested catalog mountpoints, then we could just call `UpdateNestedCatalog()` to record the new hash value for the nested catalog.  Having done this, there would be no need to recurse into the nested catalog mountpoints, because the act of updating the hash would have already performed all of the required changes.  Something along the lines of patching:
https://github.com/cvmfs/cvmfs/blob/8e0712b0230afe859fc7b5bd421878b4d64677f8/cvmfs/receiver/catalog_merge_tool_impl.h#L189-L201
to become:
```c++
  if (entry1.IsDirectory() && entry2.IsDirectory()) {
    // From directory to directory
    const catalog::DirectoryEntryBase* base_entry =
        static_cast<const catalog::DirectoryEntryBase*>(&entry2);
    output_catalog_mgr_->TouchDirectory(*base_entry, xattrs, rel_path.c_str());
    if (entry1.IsNestedCatalogMountpoint() &&
        entry2.IsNestedCatalogMountpoint()) {
      // ... get new hash and size from entry2 ...
      output_catalog_mgr_->UpdateNestedCatalog(std::string(rel_path.c_str()),
                                               new_hash, new_size);
      return true; // skip recursion into nested catalog mountpoints
    } else if (!entry1.IsNestedCatalogMountpoint() &&
        entry2.IsNestedCatalogMountpoint()) {
      output_catalog_mgr_->CreateNestedCatalog(std::string(rel_path.c_str()));
    } else if (entry1.IsNestedCatalogMountpoint() &&
               !entry2.IsNestedCatalogMountpoint()) {
[...]

PR #2709, opened 2021-05-15
Nested catalog merge optimisation

mcb30:
This is an untested first draft of the nested catalog merge optimisation described in #2707.  It is compile-tested only.

### DIFF
--- a/cvmfs/catalog_mgr_rw.cc
+++ b/cvmfs/catalog_mgr_rw.cc
@@ -886,6 +886,106 @@ void WritableCatalogManager::RemoveNestedCatalog(const string &mountpoint,
 
 
 /**
+ * Swap in a new nested catalog
+ *
+ * The old nested catalog must not have been already attached to the
+ * catalog tree.  This method will not attach the new nested catalog
+ * to the catalog tree.
+ *
+ * @param mountpoint - the path of the nested catalog to be removed
+ * @param new_hash - the hash of the new nested catalog
+ * @param new_size - the size of the new nested catalog
+ */
+void WritableCatalogManager::SwapNestedCatalog(const string &mountpoint,
+                                               const shash::Any &new_hash,
+                                               const uint64_t new_size) {
+  const string nested_root_path = MakeRelativePath(mountpoint);
+  const string parent_path = GetParentPath(nested_root_path);
+  const PathString nested_root_ps = PathString(nested_root_path);
+
+  SyncLock();
+
+  // Find the immediate parent catalog
+  WritableCatalog *parent = NULL;
+  if (!FindCatalog(parent_path, &parent)) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': could not find parent '%s'",
+          nested_root_path.c_str(), parent_path.c_str());
+  }
+
+  // Get old catalog hash
+  shash::Any old_hash;
+  uint64_t old_size;
+  const bool old_found = parent->FindNested(nested_root_ps, &old_hash,
+                                            &old_size);
+  if (!old_found) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': not found in parent",
+          nested_root_path.c_str());
+  }
+
+  // Check that old catalog was not already attached
+  if (parent->FindChild(nested_root_ps)) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': already attached",
+          nested_root_path.c_str());
+  }
+
+  // Load freely attached old catalog
+  UniquePtr<Catalog> old_catalog(LoadFreeCatalog(nested_root_ps, old_hash));
+  if (!old_catalog) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': failed to load old catalog",
+          nested_root_path.c_str());
+  }
+
+  // Load freely attached new catalog
+  UniquePtr<Catalog> new_catalog(LoadFreeCatalog(nested_root_ps, new_hash));
+  if (!new_catalog) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': failed to load new catalog",
+          nested_root_path.c_str());
+  }
+
+  // Get new catalog root directory entry
+  DirectoryEntry dirent;
+  XattrList xattrs;
+  const bool dirent_found = new_catalog->LookupPath(nested_root_ps, &dirent);
+  if (!dirent_found) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': missing dirent in new catalog",
+          nested_root_path.c_str());
+  }
+  if (dirent.HasXattrs()) {
+    const bool xattrs_found = new_catalog->LookupXattrsPath(nested_root_ps,
+                                                            &xattrs);
+    if (!xattrs_found) {
+      PANIC(kLogStderr,
+            "failed to swap nested catalog '%s': missing xattrs in new catalog",
+            nested_root_path.c_str());
+    }
+  }
+
+  // Swap catalogs
+  parent->RemoveNestedCatalog(nested_root_path, NULL);
+  parent->InsertNestedCatalog(nested_root_path, NULL, new_hash, new_size);
+
+  // Update parent directory entry
+  dirent.set_is_nested_catalog_mountpoint(true);
+  dirent.set_is_nested_catalog_root(false);
+  parent->UpdateEntry(dirent, nested_root_path);
+  parent->TouchEntry(dirent, xattrs, nested_root_path);
+
+  // Update counters
+  DeltaCounters delta = Counters::Diff(old_catalog->GetCounters(),
+                                       new_catalog->GetCounters());
+  delta.PopulateToParent(&parent->delta_counters_);
+
+  SyncUnlock();
+}
+
+
+/**
  * Checks if a nested catalog starts at this path.  The path must be valid.
  */
 bool WritableCatalogManager::IsTransitionPoint(const string &mountpoint) {
--- a/cvmfs/receiver/catalog_merge_tool_impl.h
+++ b/cvmfs/receiver/catalog_merge_tool_impl.h
@@ -167,7 +167,7 @@ void CatalogMergeTool<RwCatalogMgr, RoCatalogMgr>::ReportRemoval(
 }
 
 template <typename RwCatalogMgr, typename RoCatalogMgr>
-void CatalogMergeTool<RwCatalogMgr, RoCatalogMgr>::ReportModification(
+bool CatalogMergeTool<RwCatalogMgr, RoCatalogMgr>::ReportModification(
     const PathString& path, const catalog::DirectoryEntry& entry1,
     const catalog::DirectoryEntry& entry2, const XattrList& xattrs,
     const FileChunkList& chunks) {
@@ -180,7 +180,7 @@ void CatalogMergeTool<RwCatalogMgr, RoCatalogMgr>::ReportModification(
    *       The correct course of action is to ignore this change here.
    * */
   if (!IsPathInLease(lease_path_, rel_path)) {
-    return;
+    return true;
   }
 
   const std::string parent_path =
@@ -276,6 +276,7 @@ void CatalogMergeTool<RwCatalogMgr, RoCatalogMgr>::ReportModification(
     perf::Xadd(counters_->sz_removed_bytes, entry1.size());
     perf::Xadd(counters_->sz_added_bytes, entry2.size());
   }
+  return true;
 }
 
 template <typename RwCatalogMgr, typename RoCatalogMgr>
--- a/cvmfs/receiver/catalog_merge_tool_impl.h
+++ b/cvmfs/receiver/catalog_merge_tool_impl.h
@@ -180,7 +180,9 @@ bool CatalogMergeTool<RwCatalogMgr, RoCatalogMgr>::ReportModification(
    *       The correct course of action is to ignore this change here.
    * */
   if (!IsPathInLease(lease_path_, rel_path)) {
-    return true;
+    // All child paths will similarly be outside of the lease path,
+    // and so there is no need to recurse any further.
+    return false;
   }
 
   const std::string parent_path =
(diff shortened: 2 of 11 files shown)


jblomer:
Many thanks for starting this effort!  Let's have a first look at the CI.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

cvmfs/catalog_mgr_rw.cc
```diff
@@ -56,7 +56,12 @@ WritableCatalogManager::WritableCatalogManager(
 {
   sync_lock_ =
     reinterpret_cast<pthread_mutex_t *>(smalloc(sizeof(pthread_mutex_t)));
-  int retval = pthread_mutex_init(sync_lock_, NULL);
+  pthread_mutexattr_t attr;
```
jblomer: I'm not sure why we need a recursive lock.

jblomer (changes requested):
Many thanks for working on this!  I think in principal the patch takes the right approach!

The tricky bit is the implementation of `WritableCatalogManager::SwapCatalog`. We have to select carefully when to use methods from `WritableCatalogManager` and when to prefer methods from `WritableCatalog`. The `WritableCatalogManager` can attach nested catalogs behind our back, which is not what we want in this case (e.g., `WritableCatalogManager::FindCatalog()`).

I think we should use `WritableCatalogManager` methods for the parent path string to make sure all nested catalogs up the `mountpoint` but not including `mountpoint` are attached.  The rest of the swapping logic should be implemented by calls to `WritableCatalog` methods of that parent.

We could walk through the code together if you like, let me know if you want to do a shared screen session.

cvmfs/catalog_mgr_rw.cc
mcb30: The recursive lock is there to allow `WritableCatalogManager::SwapNestedCatalog` to call `WritableCatalogManager::RemoveNestedCatalog`, to avoid code duplication.  It would not be required if the code is restructured to avoid calling `WritableCatalogManager` methods.

mcb30:
> We could walk through the code together if you like, let me know if you want to do a shared screen session.

@jblomer Thank you - that would be extremely useful!  As I mentioned in https://github.com/cvmfs/cvmfs/issues/2707#issuecomment-841705500, I found it difficult to understand the expected separation of responsibilities between `WritableCatalogManager` and `WritableCatalog`, so it would be very useful to have some guidance on that.  Do you have availability tomorrow (Tuesday)?

mcb30:
@jblomer I've rebased onto current `devel`, and refactored it to avoid using `WritableCatalogManager` methods (essentially by duplicating the logic from `WritableCatalogManager::RemoveNestedCatalog` within `WritableCatalogManager::SwapNestedCatalog`, so that there's no need to subsequently call `WritableCatalogManager::FindCatalog` in order to get the parent catalog object).  Hopefully this is along the lines you were thinking.

The previous version is still available within the [`mcb30:nested-merge-v1`](https://github.com/mcb30/cvmfs/tree/nested-merge-v1) branch.

mcb30:
@jblomer I've updated it with the new structure following this morning's discussion - thank you again!

cvmfs/catalog_mgr_rw.cc
```diff
@@ -885,6 +885,77 @@ void WritableCatalogManager::RemoveNestedCatalog(const string &mountpoint,
...
+
+  // Check that old catalog was not already attached
+  if (parent->FindChild(nested_root_ps)) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': already attached",
+          nested_root_path.c_str());
+  }
+
+  // Load freely attached old catalog
+  UniquePtr<Catalog> old_catalog(LoadFreeCatalog(PathString("", 0), old_hash));
```
jblomer: The `mountpoint` passed to `LoadFreeCatalog` should be the nested path, I think.

cvmfs/catalog_mgr_rw.cc
```diff
@@ -885,6 +885,77 @@ void WritableCatalogManager::RemoveNestedCatalog(const string &mountpoint,
...
+  if (!old_catalog) {
+    PANIC(kLogStderr,
+          "failed to swap nested catalog '%s': could not load old catalog",
+          nested_root_path.c_str());
+  }
+
+  // Load freely attached new catalog
+  UniquePtr<Catalog> new_catalog(LoadFreeCatalog(PathString("", 0), new_hash));
```
jblomer: See above

cvmfs/catalog_mgr_rw.cc
```diff
@@ -885,6 +885,77 @@ void WritableCatalogManager::RemoveNestedCatalog(const string &mountpoint,
 }
 
 
+/**
+ * Swap in a new nested catalog
```
jblomer, line 889: Let's add a comment on the pre- / post-condition, namely that none of the nested catalogs is or gets attached to the catalog tree.

jblomer (commented):
Very well - I think that can work nicely!  Let's add some tests to the PR before merging.

cvmfs/catalog_mgr_rw.cc
mcb30: Thanks, have updated.

cvmfs/receiver/catalog_merge_tool_impl.h
```diff
@@ -193,7 +193,24 @@ bool CatalogMergeTool<RwCatalogMgr, RoCatalogMgr>::ReportModification(
...
+      shash::Any new_hash;
+      uint64_t new_size;
+      const bool found = new_catalog_mgr->LookupNested(rel_path, &mountpoint,
+                                                       &new_hash, &new_size);
+      if (!found) {
+        PANIC(kLogSyslogErr,
+              "CatalogMergeTool - nested catalog %s not found. Aborting",
+              rel_path.c_str());
+      }
+      output_catalog_mgr_->SwapNestedCatalog(std::string(rel_path.c_str()),
```
jblomer: Simply `output_catalog_mgr_->SwapNestedCatalog(rel_path,` ?

mcb30:
Rebased onto current `devel` (including the merged #2718), and now includes unit tests.  Integration tests still to be done.

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot cloudtest

jblomer:
Test 802 seems to be unhappy

mcb30:
> Test 802 seems to be unhappy

The error I'm seeing locally is
```
PANIC: /home/mcb30/work/extcvs/cvmfs/cvmfs/catalog_mgr_rw.cc : 928
failed to swap nested catalog '/a': already attached
```

I think the problem is that the call to `output_catalog_mgr_->TouchDirectory()` ends up calling `FindCatalog()` and therefore attaching the old catalog:

https://github.com/cvmfs/cvmfs/blob/65dd0082ddeb2f3ba299f59ba5016191d6698799/cvmfs/catalog_mgr_rw.cc#L703-L707

This violates the precondition that the old catalog has not been previously attached, and (correctly) triggers the PANIC error.

It feels as though `SwapNestedCatalog()` ought to be allowed to handle the situation in which the old catalog happens to be already attached, with the overall effect being similar to first calling `RemoveNestedCatalog()` with `merge=false`.

@jblomer Is there a reason why `SwapNestedCatalog()` should not be allowed to just detach the old catalog?

jblomer:
Oh, I see. But in this case, let's just move down the `TouchDirectory` call in the merge tool.  We only need it if we don't swap nested catalogs.

I'm hesitant to detach because in general it's hard to figure out what exactly changed in the attached subtree.  We currently want to call `SwapNestedCatalog` only in a context where we didn't make changes to the tree, but the API would still allow for misuse.

mcb30:
> Oh, I see. But in this case, let's just move down the `TouchDirectory` call in the merge tool. We only need it if we don't swap nested catalogs.

Done.

I also had to modify the code to use
```c++
const bool found = new_catalog_mgr->LookupNested(path, ...)
```
rather than
```c++
const bool found = new_catalog_mgr->LookupNested(rel_path, ...)
```
It seems as though `MakeRelative()` strips off any leading slash, while `MakeRelativePath()` inserts a leading slash. :confused:

@jblomer Test 802 should now pass, and turns out to already test the ability to swap a nested catalog (that contains another nested catalog).  Do we still need a separate integration test?

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer:
@mcb30 Good point on the integration tests. Let me have a look.  We might indeed already be well covered.

jblomer:
@mcb30 The integration tests look good!  I'll take a final look over the code today.

test/unittests/t_catalog_mgr_rw.cc
```diff
@@ -118,4 +131,98 @@ TEST_F(T_CatalogMgrRw, CloneTree) {
...
+  EXPECT_TRUE(catalog_mgr->LookupPath("/dir/dir/dir/sub1/file2",
+                                      kLookupSole, &dirent));
+  EXPECT_STREQ(g_hashes[7], dirent.checksum().ToString().c_str());
+  EXPECT_TRUE(catalog_mgr->LookupPath("/dir/dir/dir/sub2/file1",
+                                      kLookupSole, &dirent));
+  EXPECT_STREQ(g_hashes[6], dirent.checksum().ToString().c_str());
+}
+
+
+TEST_F(T_CatalogMgrRw, SwapNestedCatalogFailSlow) {
```
jblomer, line 176: This test should not take longer than 1-2 seconds, does it? The `EXPECT_DEATH` are expensive but there are no network timeouts as far as I can see.  Let's make it a regular test then so that it runs routinely.

```suggestion
TEST_F(T_CatalogMgrRw, SwapNestedCatalogFail) {
```

test/unittests/t_catalog_mgr_rw.cc
```diff
@@ -118,4 +131,98 @@ TEST_F(T_CatalogMgrRw, CloneTree) {
...
+
+  // Create nonexistent nested catalog hash
+  shash::Any subX_hash(shash::kMd5,
+                       shash::HexPtr("3e25960a79dbc69b674cd4ec67a72c62"),
+                       shash::kSuffixCatalog);
+  uint64_t subX_size = 42;
+
+  // Fail if parent catalog does not exist
+  catalog_mgr->DetachNested();
+  EXPECT_DEATH(catalog_mgr->SwapNestedCatalog("/dir/dir/dir/sub1",
```
jblomer: It's currently subtle: the nested catalog does not exist due to the leading slash, right?

```suggestion
  EXPECT_DEATH(catalog_mgr->SwapNestedCatalog("/no/such/directory/sub1",
```

jblomer (changes requested):
I think we are almost there.  Two tiny suggestions, and in addition I think it would be good to slightly extend integration test 802.  Between transaction 2 and transaction 3, it would be good to exercise some changes, e.g. adding one or two files to /a and /a/b, ideally in such a way that the counts of files change wrt. the previous state.

Really cool improvement!

test/unittests/t_catalog_mgr_rw.cc
mcb30: It seems to take about 10 seconds.  I based the choice upon the existing `CloneTreeFailSlow` test within the same file, which AFAICT is "slow" only in the sense of using `EXPECT_DEATH()`.

I'll make the change - let me know if it should be "slow" after all.

test/unittests/t_catalog_mgr_rw.cc
mcb30: You're right: I think this is a relic from when I was still being surprised by the unexpected behaviours of `MakeRelative()`/`MakeRelativePath()` (see https://github.com/cvmfs/cvmfs/pull/2709#issuecomment-853102465).

Will update.

mcb30:
> I think we are almost there. Two tiny suggestions, and in addition I think it would be good to slightly extend integration test 802. Between transaction 2 and transaction 3, it would be good to exercise some changes, e.g. adding one or two files to /a and /a/b, ideally in such a way that the counts of files change wrt. the previous state.

Have updated as per the suggestions, and I'll add the extension of test 802 tomorrow.

> Really cool improvement!

Thank you, and thanks for all of your help with this change!

test/unittests/t_catalog_mgr_rw.cc
jblomer: Oh, it takes _that_ long.  Apologies, then the `Slow` suffix is of course appropriate.

mcb30:
Having updated test 802 to actually check for test failures, I'm seeing errors of the form:
```
linkcounts differ: 2 / 3
timestamps differ: 1622820061 / 1622820074
transition point and root entry differ (/a)
```
I assume the timestamp difference arises because something needs to explicitly update the timestamp as recorded in the parent catalog to match the swapped-in nested catalog?

mcb30:
> I assume the timestamp difference arises because something needs to explicitly update the timestamp as recorded in the parent catalog to match the swapped-in nested catalog?

As far as I can tell, all of the existing code that handles this kind of thing does so by mounting the nested catalog and updating the root directory entry in the nested catalog to match the transition point directory entry in the parent catalog.  Since we're trying to avoid touching the new nested catalog, I'm assuming that we should do basically the reverse: get the root directory entry from the freely attached new catalog object, and use that to update the transition point directory entry in the parent catalog?

@jblomer Is this the correct approach?

jblomer:
Good catch!  Yes, we forgot about updating the transition point in the parent catalog entry. It should work as you describe.

mcb30:
> Good catch! Yes, we forgot about updating the transition point in the parent catalog entry. It should work as you describe.

Updated - `SwapNestedCatalog()` now looks up the root directory entry from the freely attached new catalog object, and applies it using:
```c++
  dirent.set_is_nested_catalog_mountpoint(true);
  dirent.set_is_nested_catalog_root(false);
  parent->UpdateEntry(dirent, nested_root_path);
  parent->TouchEntry(dirent, xattrs, nested_root_path);
```
It seems to be necessary to use both `UpdateEntry()` and `TouchEntry()` - AFAICT the former is required to update the link count and the latter is required to update the extended attributes.

This now passes the updated test 802 and I have verified that the test code really is checking counters etc (by temporarily hacking out the call to `PopulateToParent()`).

@jblomer I believe this should now be ready to merge.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer (approved):

PR #2718, opened 2021-05-29
Fix use of EXPECT_DEATH()

### LINKED ISSUE #2717: Regexps in EXPECT_DEATH() do not function as expected
The `EXPECT_DEATH()` macro used in the unit tests seems not to allow for the use of a regexp to check for the cause of the (expected) failure.

Specifically: attempting to use any regexp other than `.*` fails with an error such as:
```
    Result: died but not with expected error.
  Expected: <regexp as provided to EXPECT_DEATH()>
Actual msg:
[  DEATH   ] cvmfs_unittests: /home/mcb30/work/extcvs/cvmfs/cvmfs/sql.h:499: void sqlite::Sql::LazyInit(): Assertion `NULL != query_string_' failed.
```

The root cause is currently unclear - I am creating this issue to track progress on debugging this problem.

mcb30:
The `EXPECT_DEATH()` macro will currently give false positive results on any catalog tests, since the act of running a catalog test via `EXPECT_DEATH()` will unconditionally die due to a failure to create the test catalog history database (rather than due to the expected failure).

This fix has the side benefit of making the `EXPECT_DEATH()` regexp argument usable, which allows the test to verify that the death occurred for the expected reason.

Fixes: #2717

### DIFF
--- a/cvmfs/sql.cc
+++ b/cvmfs/sql.cc
@@ -18,7 +18,8 @@ Sql::Sql(sqlite3 *sqlite_db, const std::string &statement)
   , query_string_(NULL)
   , last_error_code_(0)
 {
-  Init(sqlite_db, statement);
+  const bool success = Init(sqlite_db, statement);
+  assert(success);
 }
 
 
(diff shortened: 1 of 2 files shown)


aandvalenzuela:
@cernvm-bot all

jblomer (approved):

PR #2719, opened 2021-06-01
Avoid starting autofs for tests that require autofs to be disabled

mcb30:
The current logic in setup_environment() will always restart the
autofs service, and will then stop it if the test specifies
cvmfs_test_autofs_on_startup=false.  This causes a test failure if the
autofs service cannot be started (e.g. because the service has been
deliberately masked), even though autofs is not required for the test
to run.

Fix by attempting to restart autofs only if autofs is required by the
test being performed.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

PR #2721, opened 2021-06-03
Do not consume 100% CPU while waiting for upload jobs to complete

### LINKED ISSUE #2720: SessionContext::UploadLoop becomes a CPU spinloop
The existing code in `SessionContext::UploadLoop()`:
https://github.com/cvmfs/cvmfs/blob/8fc9e82c68d5810b6460dd0585704b7fe25c28ec/cvmfs/session_context.cc#L414-L429
uses a simple (atomic) integer flag to indicate that the loop should terminate:
https://github.com/cvmfs/cvmfs/blob/8fc9e82c68d5810b6460dd0585704b7fe25c28ec/cvmfs/session_context.cc#L434-L436

At the point that all jobs have been submitted (i.e. `jobs_processed` is equal to `ctx->NumJobsSubmitted()`) but the jobs have not yet completed (because uploading is still in progress), this loop becomes equivalent to:

```c++
while (1) {
  // do nothing
}
```
i.e. a CPU spinloop.  This behaviour can be easily observed by publishing anything via a gateway: the publisher will hit 100% CPU usage within `cvmfs_swissknife sync`, and attaching `gdb` shows that one thread is just spinning inside `SessionContext::UploadLoop()`.

mcb30:
The current logic in SessionContext::UploadLoop() degenerates into a
CPU spinloop while waiting for the final upload jobs to complete.

Fix by using an end-of-queue dummy terminator job instead of an atomic
flag, thereby allowing the loop to sleep instead of spinning.

There seems to be no need to use the atomic compare-and-swap in
FinalizeDerived() to guarantee that pthread_join() is called exactly
once, since the surrounding control flow already appears to require
that FinalizeDerived() is called exactly once.

Fixes: #2720

### DIFF
--- a/cvmfs/session_context.cc
+++ b/cvmfs/session_context.cc
@@ -278,17 +278,12 @@ void SessionContextBase::Dispatch() {
 SessionContext::SessionContext()
     : SessionContextBase(),
       upload_jobs_(),
-      worker_terminate_(),
       worker_()
 {
-  // Set to 0 in InitializeDerived()
-  atomic_write32(&worker_terminate_, 1);
 }
 
 bool SessionContext::InitializeDerived(uint64_t max_queue_size) {
   // Start worker thread
-  atomic_init32(&worker_terminate_);
-
   upload_jobs_ = new FifoChannel<UploadJob*>(max_queue_size, max_queue_size);
   upload_jobs_->Drop();
 
@@ -299,8 +294,16 @@ bool SessionContext::InitializeDerived(uint64_t max_queue_size) {
 }
 
 bool SessionContext::FinalizeDerived() {
-  if (atomic_cas32(&worker_terminate_, 0, 1))
-    pthread_join(worker_, NULL);
+  // Note: in FinalizedDerived, we know that the worker is running.  The
+  // SessionContext is called only from GatewayUploader::FinalizeSession(),
+  // which in turn is from Spooler::FinalizeSession().  The Spooler ensures
+  // that GatewayUploader::Initialize() is called on construction.
+  //
+  // TODO(jblomer): Refactor SessionContext (and Uploader*) classes to
+  // use a factory method for construction.
+  //
+  upload_jobs_->Enqueue(&terminator_);
+  pthread_join(worker_, NULL);
 
   return true;
 }
@@ -409,11 +412,11 @@ bool SessionContext::DoUpload(const SessionContext::UploadJob* job) {
 
 void* SessionContext::UploadLoop(void* data) {
   SessionContext* ctx = reinterpret_cast<SessionContext*>(data);
+  UploadJob *job;
 
-  int64_t jobs_processed = 0;
-  while (!ctx->ShouldTerminate()) {
-    while (jobs_processed < ctx->NumJobsSubmitted()) {
-      UploadJob* job = ctx->upload_jobs_->Dequeue();
+  do {
+    job = ctx->upload_jobs_->Dequeue();
+    if (job != &terminator_) {
       if (!ctx->DoUpload(job)) {
         PANIC(kLogStderr,
               "SessionContext: could not submit payload. Aborting.");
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

cvmfs/session_context.cc
```diff
@@ -299,8 +294,8 @@ bool SessionContext::InitializeDerived(uint64_t max_queue_size) {
 }
 
 bool SessionContext::FinalizeDerived() {
-  if (atomic_cas32(&worker_terminate_, 0, 1))
-    pthread_join(worker_, NULL);
+  upload_jobs_->Enqueue(&terminator_);
```
jblomer, line 295: I think we should add a comment here.  If I'm not mistaken, `Spooler::Construct` will ensure that `SessionContext::Initialize()` is called so that the worker thread is present when `Spooler::FinalizeSession()` calls `GatewayUploader::FinalizeSession()` calls `SessionContext::Finalize()`. The comment should explain why there has to be a worker thread.  I think we should also add a `TODO` to add a factory method to the `SessionContext` class so that we cannot construct it without also initializing it.

jblomer:
@cernvm-bot cloudtest

cvmfs/session_context.cc
```diff
@@ -409,30 +404,27 @@ bool SessionContext::DoUpload(const SessionContext::UploadJob* job) {
...
         PANIC(kLogStderr,
               "SessionContext: could not submit payload. Aborting.");
       }
       job->result->Set(true);
       delete job->pack;
       delete job;
-      jobs_processed++;
     }
-    if (ctx->queue_was_flushed_.IsEmpty()) {
+    if (ctx->upload_jobs_->IsEmpty() && ctx->queue_was_flushed_.IsEmpty()) {
```
jblomer: The logic `if (IsEmpty()) Enqueue();` smells like a race condition. I realize that it was there before, I need another moment to think it through.

jblomer (changes requested):

cvmfs/session_context.cc
mcb30: Agreed that the code needs more fixing.  I don't think this PR adds any race conditions that weren't already present, but it does at least remove a CPU spinloop.  Could fixing the race condition(s) be deferred for a later fix?

cvmfs/session_context.cc
mcb30: I'm happy to add a comment, but I don't think I understand the control flow here well enough to document why the worker thread needs to exist (which it already did prior to this commit).  What should the comment say?

jblomer, line 295:
```suggestion
  // Note: in FinalizedDerived, we know that the worker is running. The SessionContext is only called from
  // GatewayUploader::FinalizeSession(), which in turn is from Spooler::FinalizeSession().  The Spooler ensures that
  // GatewayUploader::Initialize() is called on construction.
  // TODO(jblomer): Refactor SessionContext (and Uploader*) classes to use a factory method for construction.
  upload_jobs_->Enqueue(&terminator_);
```

mcb30: Thank you very much - have added this to the commit

cvmfs/session_context.cc
jblomer: Boy, these classes need a rewrite.

If I'm not mistaken, we can get rid of `queue_was_flushed_` entirely.  It seems to me not to serve any useful purpose.  The way it is implemented, it blocks the `WaitForUpload()` call until a submitted object pack in flight is sent and submitted to the gateway.  However, that logic only starts as of the second `WaitForUpload()` call because the `queue_was_flushed_` is initially `true`.  Also, the upload queue works on entire object packs -- so as long as an object pack is not full (and hence not dispatched), `WaitForUpload()` doesn't wait.

The barrier for the gateway uploader is actually not `WaitForUpload()` but `FinalizeSession()`.  Finalizing the session will end up in `SessionContextBase::Finalize`, which looks at the currently open object pack and closes out the tab itself.

Let's try (in this PR) removing `queue_was_flushed_` and just returning from `WaitForUpload()`.

mcb30: Added an extra commit to remove `queue_was_flushed_` completely, as described.  I put you as the commit author since my contribution here is merely secretarial.

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

cvmfs/session_context.cc
mcb30: As a quick smoke test: running tests 800 and 802 still passes after making this change.

jblomer:
@cernvm-bot cpplint

mcb30:
Are there any more changes needed?  It looks as though the single cloudtest failure was unrelated.

jblomer (approved):

PR #2722, opened 2021-06-04
Fix gateway integration tests

jblomer:
It turns out there are a number of unchecked conditions in the gateway integration tests which need fixing.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2724, opened 2021-06-04
Limit lsof to searching only the specified mountpoint

mcb30:
When performing an operation such as "cvmfs_server publish", a
substantial length of time (around 10-30 seconds on a typical system)
is spent in the invocation of lsof.

Use the "+f -- <path>" option to limit lsof to listing only the
specified mountpoint, rather than listing every open file handle in
the system.  Experimentation shows that this cuts the execution time
for "lsof" by around 95%.

A side benefit is that any integration tests that involve publishing a
repository will also complete substantially faster.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

PR #2725, opened 2021-06-08
MagicXattrRAIIWrapper add "The rule of three"

FanFansfan:
otherwise will unlock pthread_mutex twice and crash

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1359,9 +1359,8 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     assert(retval);
   }
 
-  MagicXattrRAIIWrapper magic_xattr;
   bool magic_xattr_success = true;
-  magic_xattr = mount_point_->magic_xattr_mgr()->Get(attr, path, &d);
+  MagicXattrRAIIWrapper magic_xattr(mount_point_->magic_xattr_mgr()->Get(attr, path, &d));
   if (!magic_xattr.IsNull()) {
     magic_xattr_success = magic_xattr->PrepareValueFenced();
   }
--- a/cvmfs/magic_xattr.cc
+++ b/cvmfs/magic_xattr.cc
@@ -98,7 +98,7 @@ std::string MagicXattrManager::GetListString(catalog::DirectoryEntry *dirent) {
   return result;
 }
 
-MagicXattrRAIIWrapper MagicXattrManager::Get(const std::string &name,
+BaseMagicXattr* MagicXattrManager::Get(const std::string &name,
                                              PathString path,
                                              catalog::DirectoryEntry *d)
 {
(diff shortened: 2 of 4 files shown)


jblomer:
Many thanks for your contribution!  You are right, we should at least have a custom copy constructor and perhaps remove the copy assignment (so that we don't need to implement both). Just wondering because the code is like this for quite some time, did you actually experience a crash?

FanFansfan:
> Many thanks for your contribution! You are right, we should at least have a custom copy constructor and perhaps remove the assignment constructor (so that we don't need to implement both). Just wondering because the code is like this for quite some time, did you actually experience a crash?

Yes. With gdb breakpoint set I observe the following execution sequence.
ed9306257cfd1e85ddaceb6b4269591
And crash backtrace log.
3b71f82f1375be4060591130479fcca

I suggest that we should also make copy assignment operator custom/private, or compiler will still invoke the implicitly-declared copy assignment operator. See https://wandbox.org/permlink/fhhYAiVg5Alo8vYT

jblomer (changes requested):
In principle I agree.  What the code effectively does is implement the copy constructor as a move constructor. Given that the code base is C++98, we cannot use C++ move semantics.

So what I would propose is this:
  - Let's make the `MagicXattrRAIIWrapper` class inherit from `SingleCopy`. That makes copy assignment and copy construction private.
  - Let's change `MagicXattrManager::Get` to a void method `MagicXattrManager::GetIntoWrapper` that takes as a last parameter a pointer to a `MagicXattrRAIIWrapper` and fills its members.  For that to work, `MagicXattrManager` needs to be a friend of `MagicXattrRAIIWrapper`.

Thinking one more time about it, perhaps it's nicer to do like this:
  - We still let`MagicXattrRAIIWrapper` inherit from `SingleCopy`
  - We introduce an additional `MagicXattrRAIIWrapper` constructor that takes a `BaseMagicXattr *` with the understanding that the argument needs to be in a locked state
  - We add `BaseMagicXattr *MagicXattrRAIIWrapper::Move()`

This way we can return a  `BaseMagicXattr *` from `MagicXattrManager::Get()` and use it to construct an `MagicXattrRAIIWrapper` in `cvmfs_getxattr`.

FanFansfan:
done.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer (changes requested):
Many thanks for the update!  Please also adjust the unit tests accordingly.  You can build the unit tests with `cmake -DBUILD_UNITTESTS=yes`

FanFansfan:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer:
Many thanks!  I'll carry on with small adustments in #2739

PR #2727, opened 2021-06-09
Do not consume 100% CPU while waiting for S3 upload jobs to complete

### LINKED ISSUE #2726: S3Uploader::MainCollectResults is a CPU spinloop
The current logic in S3Uploader::MainCollectResults() is a pure CPU spinloop that lasts for the duration of any S3 upload.  The call to sched_yield() ensures that other threads are allowed to execute, but produces no blocking behaviour that will prevent the mostly-idle thread from consuming 100% of a CPU core.

Originally mentioned in https://github.com/cvmfs/cvmfs/issues/2720#issuecomment-853900921 but this turns out to be a more severe variant of the issue: this loop consumes 100% CPU at all times, not just while waiting for the final job to complete.

mcb30:
The current logic in S3Uploader::MainCollectResults() is a pure CPU
spinloop that lasts for the duration of any S3 upload.  The call to
sched_yield() ensures that other threads are allowed to execute, but
produces no blocking behaviour that will prevent the mostly-idle
thread from consuming 100% of a CPU core.

Fix by extending the existing pipe-based pointer queue in
S3FanoutManager to add a queue of pointers to completed jobs that can
be retrieved by S3Uploader::MainCollectResults(), with a NULL pointer
used to request termination of the thread.

Fixes: #2726 

### DIFF
--- a/cvmfs/upload_s3.cc
+++ b/cvmfs/upload_s3.cc
@@ -7,9 +7,6 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <inttypes.h>
-#ifdef _POSIX_PRIORITY_SCHEDULING
-#include <sched.h>
-#endif
 #include <unistd.h>
 
 #include <string>
@@ -47,7 +44,6 @@ S3Uploader::S3Uploader(const SpoolerDefinition &spooler_definition)
          spooler_definition.driver_type == SpoolerDefinition::S3);
 
   atomic_init32(&io_errors_);
-  atomic_init32(&terminate_);
 
   if (!ParseSpoolerDefinition(spooler_definition)) {
     PANIC(kLogStderr, "Error in parsing the spooler definition");
@@ -78,7 +74,8 @@ S3Uploader::S3Uploader(const SpoolerDefinition &spooler_definition)
 
 
 S3Uploader::~S3Uploader() {
-  atomic_inc32(&terminate_);
+  // Signal termination to our own worker thread
+  s3fanout_mgr_->PushCompletedJob(NULL);
   pthread_join(thread_collect_results_, NULL);
 }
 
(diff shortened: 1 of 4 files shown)


mcb30:
This is compile-tested only at this point.  @jblomer Could you check to see if the overall logic matches the design that you had in mind?  Thanks!

cvmfs/upload_s3.cc
```diff
@@ -220,13 +217,10 @@ void *S3Uploader::MainCollectResults(void *data) {
...
   S3Uploader *uploader = reinterpret_cast<S3Uploader *>(data);
 
-  std::vector<s3fanout::JobInfo *> jobs;
-  while (atomic_read32(&uploader->terminate_) == 0) {
-    jobs.clear();
-    uploader->s3fanout_mgr_->PopCompletedJobs(&jobs);
-    for (unsigned i = 0; i < jobs.size(); ++i) {
+  while (true) {
+    s3fanout::JobInfo *info = uploader->s3fanout_mgr_->PopCompletedJob();
+    if (info) {
```
jblomer: I think it's nicer to reverse the check:
```
if (!info)
  break;
...
```

jblomer (commented):
Many thanks!  That looks very good to me!  Please go ahead.

cvmfs/upload_s3.cc
mcb30: Sure - have updated to reverse the check (and reindented the rest of the loop accordingly).

jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1126/

Looks as though the S3 tests all passed.  :slightly_smiling_face: 

(I see only what looks like a transient failure on Fedora 31 test 669, and an overall failure status for x86_64_exlcache even though the log for that test seems to show no errors.)

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

PR #2728, opened 2021-06-11
Graft uncompressed files via gateway

mcb30:
Add the ability to flag grafted files as uncompressed, and tests to verify that uncompressed grafted files can be published via a gateway.

### DIFF
--- a/cvmfs/sync_union.cc
+++ b/cvmfs/sync_union.cc
@@ -36,7 +36,9 @@ SharedPtr<SyncItem> SyncUnion::CreateSyncItem(
   PreprocessSyncItem(entry);
   if (entry_type == kItemFile) {
     entry->SetExternalData(mediator_->IsExternalData());
-    entry->SetCompressionAlgorithm(mediator_->GetCompressionAlgorithm());
+    if (!(entry->IsValidGraft() && entry->HasCompressionAlgorithm())) {
+      entry->SetCompressionAlgorithm(mediator_->GetCompressionAlgorithm());
+    }
   }
   return entry;
 }
--- a/test/src/811-repository_gateway_graft/main
+++ b/test/src/811-repository_gateway_graft/main
@@ -1,0 +1,94 @@
+cvmfs_test_name="Grafted files via gateway"
+cvmfs_test_autofs_on_startup=false
+cvmfs_test_suites="quick"
+
+verify_sha1sum() {
+  local filename=$1
+  local expected=$2
+  local actual=$(sha1sum ${filename} | cut -d' ' -f1)
+  if [ "${actual}" != "${expected}" ]; then
+    echo "Checksum mismatch for ${filename}: expected ${expected} actual ${actual}"
+    return 101
+  fi
+}
+
+verify_compression() {
+  local filename=$1
+  local expected=$2
+  local actual=$(attr -qg compression ${filename})
+  if [ "${actual}" != "${expected}" ] ; then
+    echo "Compression mismatch for ${filename}: expected ${expected} actual ${actual}"
+    return 102
+  fi
+}
+
+cvmfs_run_test() {
+  local reponame=test.repo.org
+  local repodir=/cvmfs/${reponame}
+
+  echo "set up gateway"
+  set_up_repository_gateway || return 1
+
+  echo "parse configuration"
+  load_repo_config ${reponame}
+  local spooldir=${CVMFS_SPOOL_DIR}
+
+  echo "start transaction 1"
+  cvmfs_server transaction ${reponame} || return 11
+
+  echo "create test files"
+  echo "Hello uncompressed world!" > file1
+  echo "Hello compressed world!" > file2
+  cvmfs_publish zpipe < file2 > file2.z || return 12
+  local hash1=bfc7a4e4317c6601b632801b210d0359a6315a11
+  local hash2=f558929030c0b03ac8ee1dc087ca98ed8b2adf61
+  local hash2z=58356c914c5b602d889b45ca00c9e8b68e5809af
+  verify_sha1sum file1 ${hash1} || return $?
+  verify_sha1sum file2 ${hash2} || return $?
+  verify_sha1sum file2.z ${hash2z} || return $?
+
+  echo "create graft files"
+  cvmfs_swissknife graft -i file1 -o ${repodir}/file1 -Z none || return 13
+  cvmfs_swissknife graft -i file2 -o ${repodir}/file2 -Z zlib || return 14
+  touch ${repodir}/file3
+  grep -v compression= ${repodir}/.cvmfsgraft-file2 > ${repodir}/.cvmfsgraft-file3
+  echo "----- START GRAFT -----"
+  cat ${repodir}/.cvmfsgraft-file1
+  echo "----- END GRAFT -----"
+  echo "----- START GRAFT -----"
+  cat ${repodir}/.cvmfsgraft-file2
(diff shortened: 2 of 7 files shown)


jblomer:
Is this meant to support grafting compressed and uncompressed files in the same transaction? To graft all files uncompressed, it should be enough to set `CVMFS_COMPRESSION_ALGORITHM=none` on the publisher node.

mcb30:
> Is this meant to support grafting compressed and uncompressed files in the same transaction? To graft all files uncompressed, it should be enough to set `CVMFS_COMPRESSION_ALGORITHM=none` on the publisher node.

It does support grafting a mixture of compressed and uncompressed files in the same transaction, but my primary motivation was to make the process of using uncompressed files less confusing by recording the compression information within the graft metadata so that the publisher does the right thing without requiring the user to know about setting `CVMFS_COMPRESSION_ALGORITHM`.  (I personally messed up several times trying to get grafting of uncompressed files to work, and I suspect I'm not alone in this.)

jblomer:
Thanks for the clarification! That sounds reasonable and a useful improvement.

jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1122/

Looks as though `openssl zlib` doesn't exist as a command on some of the platforms - is there a standard way for CVMFS test scripts to construct compressed versions of files?

Also looks as though the S3 tests are failing because `/srv/cvmfs/<repo>` doesn't exist on those platforms, which is reasonable enough.  Is there a standard way for CVMFS test scripts to inject content-addressed files directly into whatever functions as the storage backend?

Thanks

DrDaveD:
> (I personally messed up several times trying to get grafting of uncompressed files to work, and I suspect I'm not alone in this.)

On the contrary you probably are alone on this, since grafting is not a very heavily used feature!  I think you are only the second user, the first being the authors @djw8605 and @bbockelm for osgstorage.org repositories.

jblomer:
For the compression: `cat foo | cvmfs_publish zpipe > foo.compressed` and with `cvmfs_publish zpipe -d` for decompression.

For planting a file in the backend, there is `upload_into_backend` in `test_functions`

mcb30:
> For the compression: `cat foo | cvmfs_publish zpipe > foo.compressed` and with `cvmfs_publish zpipe -d` for decompression.
> 
> For planting a file in the backend, there is `upload_into_backend` in `test_functions`

Wonderful; thank you!  Am updating now...

mcb30:
I'm hitting an error that I don't understand:
```
+ cvmfs_swissknife upload -i file1 -o data/bf/c7a4e4317c6601b632801b210d0359a6315a11 -r gw,/srv/cvmfs/test.repo.org/txn,http://localhost:4929/api/v1 -a sha1
Failed to configure gateway uploader. Missing session token file.
```
At the point that the error occurs, the file `/var/spool/cvmfs/test.repo.org/session_token` does exist, and a `cvmfs_server publish test.repo.org` at this point would succeed.

I can't find the portion of code that is supposed to construct the path to the `session_token` file.  The string `"session_token"` appears only twice across the whole C++ codebase, both of which are using it as a key name for `JsonDocument::SearchInObject()`.  Where within the code does anything access this file?

Thanks

jblomer:
That's right, the `cvmfs_swissknife upload` command only works with the local and the S3 backend.  It would be better to bail out with an error message when this is called with a gateway uploader.

In the integration tests, the gateway runs on the same machine that the publisher, so we should pick the backend configuration of the gateway.  You can get this original backend from `TEST_CVMFS_RECEIVER_UPSTREAM_STORAGE` in /etc/cvmfs/repositories.d/test.repo.org/server.conf.  It might be best to adjust `upload_into_backend` accordingly.

mcb30:
Thanks.  I've updated to use `cvmfs_publish zpipe` and `upload_into_backend` as appropriate, with a separate commit to update `upload_into_backend` (and related functions) to handle the case of a gateway test.

mcb30:
@cernvm-bot cloudtest

mcb30:
> @cernvm-bot cloudtest

Wasn't sure if that was going to work, and it looks as though @cernvm-bot indeed ignores requests from non-admin users.

@jblomer Could you possibly kick off cloudtests for this?  Thank you!

jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1124/

Looks like just a single transient failure in test 633 on Ubuntu 16.04 S3.  AFAICT this test should not be affected by these changes, and the same test passed on other S3 platforms and other Ubuntu 16.04 platforms in this test run.

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

test/src/811-repository_gateway_graft/main
```diff
@@ -0,0 +1,81 @@
...
+  upload_into_backend ${reponame} file2.z $(make_path ${hash2z}) || return 23
+
+  echo "publish and check 2"
+  cvmfs_server publish ${reponame} || return 24
+  cvmfs_server check -i ${reponame} || return 25
+  verify_compression ${spooldir}/rdonly/file1 none || return $?
+  verify_compression ${spooldir}/rdonly/file2 zlib || return $?
+  verify_sha1sum ${repodir}/file1 ${hash1} || return $?
+  verify_sha1sum ${repodir}/file2 ${hash2} || return $?
+}
```
jblomer, line 94: It's not really necessary but in line with the other tests it would be nice to have an explicit `return 0` at the end.

cvmfs/sync_union.cc
```diff
@@ -36,7 +36,9 @@ SharedPtr<SyncItem> SyncUnion::CreateSyncItem(
   PreprocessSyncItem(entry);
   if (entry_type == kItemFile) {
     entry->SetExternalData(mediator_->IsExternalData());
-    entry->SetCompressionAlgorithm(mediator_->GetCompressionAlgorithm());
+    if (!entry->IsValidGraft()) {
+      entry->SetCompressionAlgorithm(mediator_->GetCompressionAlgorithm());
```
jblomer, line 40: This might affect users who create graft files but do not use the `swissknife_graft` utility.  I'd prefer to make the `compression` key optionally so that graft files without the key would use `mediator_->GetCompressionAlgorithm()`.  Let's add a test for this case.

jblomer (changes requested):
Many thanks! This looks in principle good to me, I have just one concern regarding backwards compatibility of the graft file format.

test/src/811-repository_gateway_graft/main
mcb30: Will do

cvmfs/sync_union.cc
mcb30: Good point - I'll add this now.  Thank you!

mcb30: Added along with a test case

mcb30:
@jblomer I've updated as per your comments, and rebased onto current `devel`.  Thanks for the review!

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer (approved):

PR #2731, opened 2021-06-15
Handle zero-length uncompressed objects in object pack consumer

### LINKED ISSUE #2730: ObjectPackConsumer loops indefinitely on zero-length objects
The processing loop in `ObjectPackConsumer::ConsumePayload()`:
https://github.com/cvmfs/cvmfs/blob/f6095b55e94daa71577990e7828ca768f90233de/cvmfs/pack.cc#L338-L341
can terminate prematurely when dealing with a zero-length object.  Specifically: for a zero-length object at the end of the buffer, `pos_in_buf` will equal `buf_size` and so the loop will terminate, but `idx_` will not equal `index.size()` and so the method ends up erroneously returning `kStateContinue`.

Zero-length objects can occur only when using uncompressed objects, since a compressed object will never have a zero-length compressed representation.  The following integration test case triggers the issue:

```sh
cvmfs_test_name="Uncompressed files via gateway"
cvmfs_test_autofs_on_startup=false
cvmfs_test_suites="quick"

verify_compression() {
  local filename=$1
  local expected=$2
  local actual=$(attr -qg compression ${filename})
  if [ "${actual}" != "${expected}" ] ; then
    echo "Compression mismatch for ${filename}: expected ${expected} actual ${actual}"
    return 101
  fi
}

cvmfs_run_test() {
  local reponame=test.repo.org
  local repodir=/cvmfs/${reponame}

  echo "set up gateway"
  set_up_repository_gateway || return 1

  echo "parse configuration"
  load_repo_config ${reponame}
  local spooldir=${CVMFS_SPOOL_DIR}

  echo "start transaction"
  cvmfs_server transaction ${reponame} || return 11

  echo "create test files"
  echo "Hello world!" > ${repodir}/file1
  touch ${repodir}/file2

  echo "publish (uncompressed) and check"
  cvmfs_server publish -Z none ${reponame} || return 12
  cvmfs_server check -i ${reponame} || return 13

  echo "verify files"
  [ -f ${repodir}/file1 -a -s ${repodir}/file1 ] || return 14
  [ -f ${repodir}/file2 -a ! -s ${repodir}/file2 ] || return 15
  verify_compression ${spooldir}/rdonly/file1 none || return $?
  verify_compression ${spooldir}/rdonly/file2 none || return $?
}
```

mcb30:
The current logic in ObjectPackConsumer::ConsumePayload() behaves
incorrectly in the case of a zero-length object at the end of the
buffer: the loop terminates prematurely since the end of the buffer
has been reached, but the method then returns kStateContinue since not
all objects have been processed.  This causes the receiver to
erroneously wait for further data that will never arrive.

Fix by ensuring that the loop continues until either all objects have
been processed or an incomplete object is encountered.

Note that a zero-length object can occur only when using uncompressed
objects, since even a zero-length object will have a non-zero length
after compression.

Fixes: #2730

### DIFF
--- a/cvmfs/pack.cc
+++ b/cvmfs/pack.cc
@@ -338,7 +338,8 @@ ObjectPackBuild::State ObjectPackConsumer::ConsumeNext(
 ObjectPackBuild::State ObjectPackConsumer::ConsumePayload(
     const unsigned buf_size, const unsigned char *buf) {
   uint64_t pos_in_buf = 0;
-  while ((pos_in_buf < buf_size) && (idx_ < index_.size())) {
+  while ((idx_ < index_.size()) &&
+         ((pos_in_buf < buf_size) || (index_[idx_].size == 0))) {
     // Fill the accumulator or process next small object
     uint64_t nbytes;  // How many bytes are consumed in this iteration
     const uint64_t remaining_in_buf = buf_size - pos_in_buf;
--- a/test/src/812-repository_gateway_uncompressed/main
+++ b/test/src/812-repository_gateway_uncompressed/main
@@ -1,0 +1,44 @@
+cvmfs_test_name="Uncompressed files via gateway"
+cvmfs_test_autofs_on_startup=false
+cvmfs_test_suites="quick"
+
+verify_compression() {
+  local filename=$1
+  local expected=$2
+  local actual=$(attr -qg compression ${filename})
+  if [ "${actual}" != "${expected}" ] ; then
+    echo "Compression mismatch for ${filename}: expected ${expected} actual ${actual}"
+    return 101
+  fi
+}
+
+cvmfs_run_test() {
+  local reponame=test.repo.org
+  local repodir=/cvmfs/${reponame}
+
+  echo "set up gateway"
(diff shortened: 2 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

mcb30:
I've updated to fix the failure on test 803, and to add a `return 0` to the new test 812 as per https://github.com/cvmfs/cvmfs/pull/2728#discussion_r652399403, and rebased onto current `devel` branch.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1128/

Only transient unrelated failures, as far as I can tell.

jblomer (approved):

PR #2732, opened 2021-06-17
Fix checking whether doc target exists

amadio:
Fixes the error shown below:

```
-- Generating Doxygen configuration ...
CMake Error at cmake/Modules/UseDoxygen.cmake:139 (get_target_property):
  get_target_property() called with non-existent target "doc".
Call Stack (most recent call first):
  CMakeLists.txt:515 (include)
```


jblomer:
@cernvm-bot cloudtest

jblomer (approved):
Thanks!

PR #2734, opened 2021-06-17
Report modified nested catalog hashes as modifications when diffing

mcb30:
A change in a nested catalog hash value is not detected as a
difference via DirectoryEntry::CompareTo(), as exemplified by the
logic in CatalogDiffTool::DiffRec() to detect identical nested
catalogs.

The current logic will therefore not call ReportModification() for a
change in a nested catalog hash value unless there is some other
modification (e.g. mtime) to the corresponding directory entry.

Fix by treating a modified hash as a modification that should be
reported via ReportModification().

This ensures that any nested catalog swapping always occurs at the
highest possible level within the tree, thereby avoiding unnecessary
recursion.

### DIFF
--- a/cvmfs/catalog_diff_tool_impl.h
+++ b/cvmfs/catalog_diff_tool_impl.h
@@ -178,7 +178,7 @@ void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
       if (id_nested_from == id_nested_to) continue;
     }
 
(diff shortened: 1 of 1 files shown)


jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1132/

Just a single unrelated transient failure.

jblomer (approved):

PR #2735, opened 2021-06-18
Do not truncate recursion for a modification above the lease pat

mcb30:
Commit e9df405 ("Avoid recursing into paths outside the lease when
performing a merge") introduced a bug by omitting to consider the case
of a modification detected in a directory that is a parent of the
lease path.  Such a path is not within the lease path, but a child
path may still lie within the lease path.

Fix by continuing recursion if the current path is a parent of the
lease path, and add a test case that covers this scenario.

### DIFF
--- a/cvmfs/receiver/lease_path_util.cc
+++ b/cvmfs/receiver/lease_path_util.cc
@@ -7,26 +7,19 @@
 namespace receiver {
 
 bool IsPathInLease(const PathString& lease, const PathString& path) {
-  // If lease is "", any path falls within item
-  if (!strcmp(lease.c_str(), "")) {
+
+  // If lease is "", then any path is a subpath
+  if (lease.GetLength() == 0) {
     return true;
   }
 
-  // Is the lease string is the prefix of the path string and the next
-  // character of the path string is a "/".
-  if (path.StartsWith(lease) && path.GetChars()[lease.GetLength()] == '/') {
-    return true;
-  }
-
-  // The lease is a prefix of the path and the last char of the lease is a "/"
+  // If the lease string is the prefix of the path string and either
+  // the strings are identical or the separator character is a "/",
+  // then the path is a subpath
   if (path.StartsWith(lease) &&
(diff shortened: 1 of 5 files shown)


jblomer:
@cernvm-bot cloudtest

cvmfs/receiver/lease_path_util.cc
```diff
@@ -6,27 +6,20 @@
...
-  // If the path string is exactly the lease path return true (allow the
-  // creation of the leased directory during the lease itself)
-  if (lease == path) {
+  // If the parent string is the prefix of the path string and either
+  // the strings are identical or the separator character is a "/",
+  // then the path is a subpath
+  if (path.StartsWith(parent) &&
+      ((path.GetLength() == parent.GetLength()) ||
+       (path.GetChars()[parent.GetLength()] == '/') ||
+       (path.GetChars()[parent.GetLength() - 1] == '/'))) {
```
jblomer, line 22: I am not quite sure here why both tests are needed `parent.GetLength()` and `parent.GetLength() - 1`

jblomer (commented):
In principle looks good to me, just haven't yet fully understood the logic in `IsSubPath`

cvmfs/receiver/lease_path_util.cc
mcb30: That logic is retained from the previous implementation, which also checks both conditions (but duplicates the call to `StartsWith()`).

I believe it's to allow for the fact that the parent path (i.e. the lease path in the original version) may or may not include a terminating `/` character.

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1134/

Just a couple of unrelated transient failures in `Test pickup of changed files`.  (I've noticed that one failing intermittently quite often.)

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

PR #2738, opened 2021-06-22
Allow nested catalog to be swapped even if already attached

### LINKED ISSUE #2737: Swapping nested subcatalogs fails within a newly created parent catalog
Converting an existing directory to a nested catalog will cause any child catalogs of the new catalog to become attached.  This violates the assumption in `SwapNestedCatalog()` that the catalog about to be swapped out has not yet been attached to the tree.

The issue can be reproduced with a sequence such as:
```
# Initial setup
cvmfs_server transaction ${reponame}
mkdir ${repodir}/tree/branch
touch ${repodir}/tree/branch/.cvmfscatalog
cvmfs_server publish ${reponame}

# Trigger failure
cvmfs_server transaction ${reponame}
touch ${repodir}/tree/.cvmfscatalog                   # Convert directory to nested catalog
echo "hello world" > ${repodir}/tree/branch/hw.txt    # Modify child catalog in same transaction
cvmfs_server publish ${reponame}
```

This causes an assertion failure `failed to swap nested catalog '/tree/branch': already attached`

mcb30:
There exist situations in which SwapNestedCatalog() may encounter a catalog that has already been attached.  For example: if a directory is converted into a nested catalog then any child catalogs will be attached during the call to CreateNestedCatalog() on that directory.  A subsequent call to SwapNestedCatalog() for one of the child catalogs will currently hit an assertion failure due to the old catalog having already been attached.

Fix by calling RemoveNestedCatalog() to cleanly handle the removal of the old catalog whatever its current state, and then adding the new catalog as before.

Convert SyncLock() to a recursive lock, to allow for calling RemoveNestedCatalog() from within SwapNestedCatalog().

Add a test case to cover the issue as described in #2737 

Fixes: #2737 

### DIFF
--- a/cvmfs/catalog_mgr_rw.cc
+++ b/cvmfs/catalog_mgr_rw.cc
@@ -913,30 +913,43 @@ void WritableCatalogManager::SwapNestedCatalog(const string &mountpoint,
           nested_root_path.c_str(), parent_path.c_str());
   }
 
-  // Get old catalog hash
-  shash::Any old_hash;
-  uint64_t old_size;
-  const bool old_found = parent->FindNested(nested_root_ps, &old_hash,
-                                            &old_size);
-  if (!old_found) {
-    PANIC(kLogStderr,
-          "failed to swap nested catalog '%s': not found in parent",
-          nested_root_path.c_str());
-  }
+  // Get old nested catalog counters
+  Catalog *old_attached_catalog = parent->FindChild(nested_root_ps);
+  Counters old_counters;
+  if (old_attached_catalog) {
 
-  // Check that old catalog was not already attached
-  if (parent->FindChild(nested_root_ps)) {
-    PANIC(kLogStderr,
-          "failed to swap nested catalog '%s': already attached",
-          nested_root_path.c_str());
-  }
+    // Old catalog was already attached (e.g. as a child catalog
+    // attached by a prior call to CreateNestedCatalog()).  Ensure
+    // that it has not been modified, get counters, and detach it.
+    WritableCatalogList list;
+    if (GetModifiedCatalogLeafsRecursively(old_attached_catalog, &list)) {
+      PANIC(kLogStderr,
+            "failed to swap nested catalog '%s': already modified",
+            nested_root_path.c_str());
+    }
+    old_counters = old_attached_catalog->GetCounters();
+    DetachSubtree(old_attached_catalog);
 
-  // Load freely attached old catalog
-  UniquePtr<Catalog> old_catalog(LoadFreeCatalog(nested_root_ps, old_hash));
-  if (!old_catalog) {
-    PANIC(kLogStderr,
-          "failed to swap nested catalog '%s': failed to load old catalog",
-          nested_root_path.c_str());
+  } else {
+
+    // Old catalog was not attached.  Download a freely attached
+    // version and get counters.
+    shash::Any old_hash;
+    uint64_t old_size;
+    const bool old_found = parent->FindNested(nested_root_ps, &old_hash,
+                                              &old_size);
+    if (!old_found) {
+      PANIC(kLogStderr,
+            "failed to swap nested catalog '%s': not found in parent",
+            nested_root_path.c_str());
+    }
+    UniquePtr<Catalog> old_free_catalog(LoadFreeCatalog(nested_root_ps, old_hash));
+    if (!old_free_catalog) {
+      PANIC(kLogStderr,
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1137/

Just a single unrelated transient failure (on connecting to `cvmfs-stratum-one.cern.ch`)

jblomer (changes requested):
I think we can go a little bit different about this: currently we are already trying to find the nested catalog in the tree of attached catalogs.  We fail if we find it.  Instead, if we find it we can check with `GetModifiedCatalogLeafsRecursively()` if the nested tree has any changes recorded (it must not).

Since we have the catalog already open, we can then reuse that open file and don't need to redownload it.

I think from `RemoveNestedCatalog` we then only need the `DetachCatalog` call (actually better a `DetachSubtree` call), and we don't need to make the mutex recursive.

I'm actually surprised that `nested_catalog->RemoveFromParent()` does not cause any problems in the current implementation of `SwapNestedCatalog`.  Shouldn't it distort the catalog counters?

mcb30:
> I'm actually surprised that `nested_catalog->RemoveFromParent()` does not cause any problems in the current implementation of `SwapNestedCatalog`. Shouldn't it distort the catalog counters?

I believe `RemoveFromParent()` already has to successfully update the counters to compensate for the removed catalog, and the revised version of the patch then applies the modifications due solely to adding the new catalog.  This seems to work as expected, based on the integration test results.

> I think we can go a little bit different about this: currently we are already trying to find the nested catalog in the tree of attached catalogs. We fail if we find it. Instead, if we find it we can check with `GetModifiedCatalogLeafsRecursively()` if the nested tree has any changes recorded (it must not).
> 
> Since we have the catalog already open, we can then reuse that open file and don't need to redownload it.
> 
> I think from `RemoveNestedCatalog` we then only need the `DetachCatalog` call (actually better a `DetachSubtree` call), and we don't need to make the mutex recursive.

OK, will have a go at implementing this approach.

mcb30:
> > I think we can go a little bit different about this: currently we are already trying to find the nested catalog in the tree of attached catalogs. We fail if we find it. Instead, if we find it we can check with `GetModifiedCatalogLeafsRecursively()` if the nested tree has any changes recorded (it must not).
> > Since we have the catalog already open, we can then reuse that open file and don't need to redownload it.
> > I think from `RemoveNestedCatalog` we then only need the `DetachCatalog` call (actually better a `DetachSubtree` call), and we don't need to make the mutex recursive.
> 
> OK, will have a go at implementing this approach.

I've updated it to use this approach.  I find the code harder to read now, but it does still seem to pass the test cases.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

cvmfs/catalog_mgr_rw.cc
```diff
@@ -913,30 +913,42 @@ void WritableCatalogManager::SwapNestedCatalog(const string &mountpoint,
...
-                                            &old_size);
-  if (!old_found) {
-    PANIC(kLogStderr,
-          "failed to swap nested catalog '%s': not found in parent",
-          nested_root_path.c_str());
-  }
+  // Get old nested catalog counters
+  Catalog *old_attached_catalog = parent->FindChild(nested_root_ps);
+  Counters old_counters;
+  if (old_attached_catalog) {
```
jblomer, line 919: Perhaps one more thing: could you add the example from the PR description as a comment so that we remember why it might be possible that the catalog is already attached.

jblomer:
There is a problem without code signing service, looking into it...

cvmfs/catalog_mgr_rw.cc
mcb30: Have done

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1139/

Looks like just a couple of isolated unrelated failures

PR #2739, opened 2021-06-23
Fix potential crash when accessing extended attributes

jblomer:


jblomer:
Replaces #2725

jblomer:
@cernvm-bot cloudtest

PR #2740, opened 2021-06-24
Avoid diff recursion outside of the lease path during a catalog merge

mcb30:
The existing logic allows for recursion to be terminated early upon
encountering a reported modification that lies outside of the lease
path, in order to avoid wasting time comparing subdirectories that can
never be of interest.

Generalise this to terminate recursion as soon as any path that lies
outside of the lease path is encountered, even if such a path does not
contain a modification.

This generalisation requires centralising the concept of an ignored
path into the DiffRec() function.  For the sake of consistency, also
centralise the related concept of a reportable path, and test for
reportability prior to calling ReportModification() et al.

### DIFF
--- a/cvmfs/catalog_diff_tool.h
+++ b/cvmfs/catalog_diff_tool.h
@@ -52,6 +52,41 @@ class CatalogDiffTool {
   bool Run(const PathString& path);
 
  protected:
+  /**
+   * Check if a path (and, by implication, any subpath) should be
+   * ignored and not considered for comparison purposes.
+   *
+   * This can be used to avoid unnecessary work by avoiding recursion
+   * into paths that will not be of interest (e.g. paths that are
+   * neither above nor within the lease path, when using a gateway).
+   */
+  virtual bool IsIgnoredPath(const PathString& path) { return false; }
+
+  /**
+   * Check if a difference found on a path should be reported via
+   * ReportAddition(), ReportRemoval(), or ReportModification().
+   *
+   * This can be used to filter out differences that are not of
+   * interest (e.g. paths that are not within the lease path, when
+   * using a gateway).
+   *
+   * Note that an ignored path must necessarily be a non-reportable
+   * path, since an ignored path will never even be compared (and so
+   * can never be reported upon).  However, there do exist paths that
+   * are neither ignored nor reportable: when using a gateway, a
+   * parent of the lease path is not reportable (since it is not
+   * within the lease path) but must not be ignored (since we need to
+   * recurse into the parent path in order to reach the lease path).
+   *
+   * As a concrete example, with a lease path of /foo/bar:
+   *
+   * /foo             <-  not ignored, not reportable
+   * /foo/bar         <-  not ignored      reportable
+   * /foo/bar/thing   <-  not ignored      reportable
+   * /foo/baz         <-      ignored (and therefore not reportable)
+   */
+  virtual bool IsReportablePath(const PathString& path) { return true; }
+
   virtual void ReportAddition(const PathString& path,
                               const catalog::DirectoryEntry& entry,
                               const XattrList& xattrs,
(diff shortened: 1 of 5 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1140/

Looks as though all the MacOS tests are failing due to an unrelated missing `/etc/cvmfs/cvmfs_server_hooks.sh` file in the MacOS build

mcb30:
@jblomer Do you have any thoughts on the design approach in this patch?

Thanks in advance.

jblomer:
@cernvm-bot cpplint

cvmfs/catalog_diff_tool.h
```diff
@@ -52,6 +52,9 @@ class CatalogDiffTool {
   bool Run(const PathString& path);
 
  protected:
+  virtual bool IsIgnoredPath(const PathString& path) { return false; }
```
jblomer, line 63: Let's add comments that explain the semantics of an ignored part and a reportable part.  On first sight, they seem to be complementary but only in the merge tool it becomes apparent that there are subtle differences.

cvmfs/catalog_diff_tool_impl.h
```diff
@@ -178,8 +189,9 @@ void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
       if (id_nested_from == id_nested_to) continue;
     }
 
-    if ((diff != catalog::DirectoryEntryBase::Difference::kIdentical) ||
-        old_entry.IsNestedCatalogMountpoint()) {
+    if (IsReportablePath(old_path) &&
```
jblomer, line 192: Not sure about this condition: If we find a difference between old and new on a non-reportable path, we won't recurse further into the tree.  Is this the desired logic?

jblomer (changes requested):
Very nice improvement, thank you very much!

I'm not entirely sure yet if modifications of parent directories of the lease path are handled correctly wrt. recursing the tree.  Perhaps we could add a test for that case.

cvmfs/catalog_diff_tool.h
mcb30: Good idea.  I've added explanatory comments to the prototypes, along with a concrete example of the situation in which a path is not ignored but also not reportable.

cvmfs/catalog_diff_tool_impl.h
mcb30: No, that `if` statement guards only the call to `ReportModification()`.  If `IsReportablePath()` returns false then the call to `ReportModification()` will be skipped (along with the potential early termination that `ReportModification()` may request), but the control flow will continue to the normal recursion path that follows the `if` block.

I'll make sure there's a test case that covers this case.

mcb30: I've extended the integration tests to include situations where a path directly above the lease path (i.e. neither ignored nor reportable) is modified, and confirmed that the behaviour is correct.

jblomer:
@cernvm-bot cloudtest

jblomer (changes requested):
Great, many thanks for checking! Just one more thing, please use doxygen syntax for the `IsIgnoredPath` and `IsReportablePath` comments:

    /**
     * ...
     * /

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot unittest

mcb30:
> Great, many thanks for checking! Just one more thing, please use doxygen syntax for the `IsIgnoredPath` and `IsReportablePath` comments:
> 
> ```
> /**
>  * ...
>  * /
> ```

Will do, one sec...

mcb30:
> > Great, many thanks for checking! Just one more thing, please use doxygen syntax for the `IsIgnoredPath` and `IsReportablePath` comments:
> > ```
> > /**
> >  * ...
> >  * /
> > ```
> 
> Will do, one sec...

Done.

jblomer (approved):

jblomer:
@cernvm-bot cloudtest

PR #2736, opened 2021-06-22
Add support for cvfmsbundles file

RohitTopi:
Support for .cvmfsbundles files

To improve cold cache loading, a new content type, bundle, is introduced
([PR 2741](https://github.com/cvmfs/cvmfs/pull/2741)). The bundle has a binary format of an object pack

A bundle specification file, named `.cvmfsbundles`, allows the creation of such
bundles on the server-side. The `.cvfmsbundles` file is maintained by the
repository owners. During publication, the bundles are created and
stored according to the contents of this file.

This PR will
 * allow recognition of the bundle specification files on the server-side.
 * enforce the requirement that such files must be regular and not a hard link.
     Additionally, ensure that the file is in the root directory of a repository.
 * reserve the filename, `.cvmfsbundles`, for the bundle specification files. 

Corresponding JIRA issue: [CVM-1886](https://sft.its.cern.ch/jira/browse/CVM-1886)

PR #2741, opened 2021-06-25
Introduce bundles content type in the repository

RohitTopi:
Introduce bundles content type

To improve cold cache loading, several files which are known to be required in sets can be bundled together and streamed as an object pack. This is the case, for instance, for well-known Python libraries (e.g. Tensorflow) that upon start always load a certain set of inter-dependent python files.

To support this, a new content type *bundles* is introduced. A bundle can have the binary format of object packs used by the gateway as a form to concatenate multiple objects in a single BLOB. Such bundles would be created using a special hidden file (.cvmfsbundles in [draft PR 2736](https://github.com/cvmfs/cvmfs/pull/2736)). Bundle files can be chunked the same way regular large files get chunked. Chunked and non-chunked bundle files would get a B as a hash suffix.

Corresponding JIRA issue: [CVM-1886](https://sft.its.cern.ch/jira/browse/CVM-1886)


cvmfs/bundle.cc
```diff
@@ -0,0 +1 @@
+
```
jblomer, line 4: Let's add least add the usual cvmfs comment header.

jblomer (changes requested):
Let's add to this one the header file and a skeleton for the unit test. Let's also add the files to the CMakeLists.txt file.

cvmfs/bundle.cc
```diff
@@ -0,0 +1,87 @@
...
+#include <string>
+
+#include "json_document.h"
+#include "pack.h"
+#include "util/exception.h"
+#include "util/posix.h"
+#include "util/string.h"
+
+ObjectPack* Bundle::CreateBundle(const JSON* json_obj)
```
jblomer: Style: 

```suggestion
ObjectPack *Bundle::CreateBundle(const JSON *json_obj) {
```

cvmfs/bundle.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef CVMFS_BUNDLE_H_
+#define CVMFS_BUNDLE_H_
+
+#include <string>
+
+#include "json_document.h"
+#include "pack.h"
+
+/**
+ * A bundle has the binary format of object packs used by the gateway as
```
jblomer:
```suggestion
 * A bundle has the binary format of object packs used as
```

cvmfs/bundle.h
```diff
@@ -0,0 +1,39 @@
...
+ public:
+  const int64_t kMaxFileSize = 1024 * 1024 * 100;
+
+/**
+ * Creates a ObjectPack bundle of one or more regular files
+ * The bundle is not created if any of the files to be included is not found or
+ * if the file size of any of the files exceeds a maximum value kMaxFileSize
+ * The JSON Object json_obj passed has a bundle id and a JSON array of file
+ * paths
+ *
```
jblomer: Let's document the JSON structure that `CreateBundle` expects

cvmfs/bundle.h
```diff
@@ -0,0 +1,39 @@
...
+ * Returns a pointer to the created ObjectPack on success, NULL
+ * otherwise
+ */
+  ObjectPack* CreateBundle(const JSON* json_obj);
```
jblomer:
```suggestion
  ObjectPack *CreateBundle(const JSON *json_obj);
```

cvmfs/bundle.cc
```diff
@@ -0,0 +1,87 @@
...
+{
+  const JSON* value = json_obj;
+  if (value->type != JSON_OBJECT) {
+    PANIC(kLogStderr, "JSON object not found");
+  } else {
+    // create an ObjectPack
+    ObjectPack* op = new ObjectPack();
```
jblomer: Let's use a `UniquePtr` from pointer.cc. We then don't need the explicit `delete` calls in the error paths. We return it with `UniquePtr::Release()`

cvmfs/bundle.cc
```diff
@@ -0,0 +1,87 @@
...
+
+    value = (value->first_child);  // bundleid
+    value = (value->next_sibling);  // JSON array
+    if (value->first_child) {
+      value = value->first_child;
+      std::string filepath = std::string(value->string_value);
+
+      // if file doesn't exist then bundle is not created
+      if (!FileExists(filepath)) {
+        PrintWarning("File not found: " + filepath);
```
jblomer: We should decide here: either warn and skip (instead of returning) or we `PANIC`. I think we want to warn and skip.

cvmfs/bundle.cc
```diff
@@ -0,0 +1,87 @@
...
+    // create an ObjectPack
+    ObjectPack* op = new ObjectPack();
+
+    if (op == NULL) {
+      PANIC(kLogStderr, "Insufficient memory");
+    }
+
+    value = (value->first_child);  // bundleid
+    value = (value->next_sibling);  // JSON array
+    if (value->first_child) {
```
jblomer: Combine with while loop further down to remove the code duplication.

cvmfs/bundle.cc
```diff
@@ -0,0 +1,87 @@
...
+          delete op;
+          return NULL;
+        }
+
+        if (!AddFileToObjectPack(op, filepath)) {
+          delete op;
+          return NULL;
+        }
+      }
+    } else {
```
jblomer, line 162: Let's error out early if an empty bundle is found.

jblomer (changes requested):
It's taking shape!  You can actually add unit tests for `CreateBundle` right away.

vvolkl:
Superseded by #4002

PR #2742, opened 2021-06-25
Extend catalog table to support bundle files

RohitTopi:
Extend the catalog table

The catalog table in the file catalogs needs an additional column that specifies a bundle ID that can be optionally associated with a regular file.
An additional file catalog table will register the bundles (or the respective bundle chunks) and their IDs

Corresponding JIRA issue: [CVM-1886](https://sft.its.cern.ch/jira/browse/CVM-1886)


vvolkl:
Superseded by #4002

PR #2743, opened 2021-06-25
Add client-side support for bundle files

RohitTopi:
Bundle files would allow improving cold cache loading of certain sets files that are known to be required if any of the files in the set is accessed. The Bundle files would be created on the server-side according to `.cvmfsbundles` files

On the client-side when there is a cache miss for a file that is part of a bundle, the file catalog can be checked for the list of files part of that bundle. So the client can then decide based on the cache contents whether to download the individual requested file or the entire bundle. Being an object pack, the bundle would be directly streamed into the local cache.

For more background please refer to [CVM-1886](https://sft.its.cern.ch/jira/browse/CVM-1886)


vvolkl:
Superseded by #4002

PR #2746, opened 2021-07-02
Add support for direct I/O files

jblomer:
Allows to mark files as "direct I/O" during publish.  This can be a workaround for files that are known to be subject to in-place updates (see CVM-2001) at the cost of a performance penalty.  Also adds the `direct_io` extended attribute.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2745, opened 2021-07-01
Add 'device id' talk command

jblomer:
The device ID is stable across mount namespaces and thus allows to identify a mountpoint system-wide. It can be used to find "zombie" fuse modules that are only active in child mount namespaces (see [CVM-2004](https://sft.its.cern.ch/jira/browse/CVM-2004))

PR #2747, opened 2021-07-08
Add cvmfs_config fuser command

DrDaveD:
See [CVM-2004](https://sft.its.cern.ch/jira/browse/CVM-2004).  Based on the new `cvmfs_talk device id` command added in  #2745.

### DIFF
--- a/test/src/094-fuser/main
+++ b/test/src/094-fuser/main
@@ -1,0 +1,45 @@
+
+cvmfs_test_name="Using cvmfs_config fuser to find processes accessing a repository"
+cvmfs_test_suites="quick"
+
+cvmfs_run_test() {
+  logfile=$1
+
+  echo "Mount repository"
+  cvmfs_mount grid.cern.ch || return 1
+
+  echo "Start background process accessing it and check that it is being used"
+  set -m # enables job control handling
+  (cd /cvmfs/grid.cern.ch; sleep 30) &
+  backpid="$!"
+  pids="`sudo cvmfs_config fuser grid.cern.ch`"
+  [[ " $pids " == *" $backpid "* ]] || return 10
+  kill -15 -$backpid
+  wait $backpid 2>/dev/null || true
+  pids="`sudo cvmfs_config fuser grid.cern.ch`"
+  [ -z "$pids" ] || return 11
+
+  echo "Start unshared background process accessing it, and check that"
+  unshare -Um sh -c "cd /cvmfs/grid.cern.ch && /bin/sleep 60" & 
+  backpid="$!"
+  sleep 1
+  cvmfs_umount grid.cern.ch
+  # can't get error code from background proc before wait, so instead do
+  #   kill -0 on it to see if it is still running
+  if ! kill -0 $backpid; then
+    wait $backpid || true
+    echo "WARNING: could not unshare, skipping unshare test"
+    CVMFS_GENERAL_WARNING_FLAG=1
+  else
+    pids="`sudo cvmfs_config fuser grid.cern.ch`"
+    [[ " $pids " == *" $backpid "* ]] || return 20
+    kill -15 -$backpid
+    wait $backpid 2>/dev/null || true
+    echo "Checking once more, where it should be completely unmounted"
+    pids="`sudo cvmfs_config fuser grid.cern.ch`"
+    [ -z "$pids" ] || return 21
+  fi
+
+  return 0
+}
+
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -1274,7 +1274,7 @@ cvmfs_config_fuser() {
   fi
   [ -n "$deviceid" ] || die "Could not obtain device id for $fqrn"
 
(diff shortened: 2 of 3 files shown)


jblomer (changes requested):
Looks good to me!  Let's add an integration test for the new command.

test/src/094-fuser/main
```diff
@@ -0,0 +1,45 @@
...
+
+  echo "Mount repository"
+  cvmfs_mount grid.cern.ch || return 1
+
+  echo "Start background process accessing it and check that it is being used"
+  set -m # enables job control handling
+  (cd /cvmfs/grid.cern.ch; sleep 30) &
+  backpid="$!"
+  pids="`sudo cvmfs_config fuser grid.cern.ch`"
+  [[ " $pids " == *" $backpid "* ]] || return 10
```
jblomer: If the test fails, will $backpid survive? If so, we'd need a cleanup.

test/src/094-fuser/main
```diff
@@ -0,0 +1,45 @@
...
+
+cvmfs_run_test() {
+  logfile=$1
+
+  echo "Mount repository"
+  cvmfs_mount grid.cern.ch || return 1
+
+  echo "Start background process accessing it and check that it is being used"
+  set -m # enables job control handling
+  (cd /cvmfs/grid.cern.ch; sleep 30) &
```
jblomer:
```suggestion
  (cd /cvmfs/grid.cern.ch; sleep 10000) &
```
I think we should wait essentially infinitely. This should make the test less flaky and in case of a hang the test driver timeout cleans up.

test/src/094-fuser/main
```diff
@@ -0,0 +1,45 @@
...
+  backpid="$!"
+  pids="`sudo cvmfs_config fuser grid.cern.ch`"
+  [[ " $pids " == *" $backpid "* ]] || return 10
+  kill -15 -$backpid
+  wait $backpid 2>/dev/null || true
+  pids="`sudo cvmfs_config fuser grid.cern.ch`"
+  [ -z "$pids" ] || return 11
+
+  echo "Start unshared background process accessing it, and check that"
+  unshare -Um sh -c "cd /cvmfs/grid.cern.ch && /bin/sleep 60" &
```
jblomer:
```suggestion
  unshare -Um sh -c "cd /cvmfs/grid.cern.ch && /bin/sleep 10000" & 
```
see above

test/src/094-fuser/main
```diff
@@ -0,0 +1,45 @@
...
+  [[ " $pids " == *" $backpid "* ]] || return 10
+  kill -15 -$backpid
+  wait $backpid 2>/dev/null || true
+  pids="`sudo cvmfs_config fuser grid.cern.ch`"
+  [ -z "$pids" ] || return 11
+
+  echo "Start unshared background process accessing it, and check that"
+  unshare -Um sh -c "cd /cvmfs/grid.cern.ch && /bin/sleep 60" & 
+  backpid="$!"
+  sleep 1
```
jblomer: Is this necessary to wait for the unshared process to start?  If so, perhaps the unshared process could create a file and the main test waits for the file to appear. The sleeps have a tendency to fail every now and then (test VMs might be paused).

test/src/094-fuser/main
```diff
@@ -0,0 +1,45 @@
...
+  cvmfs_umount grid.cern.ch
+  # can't get error code from background proc before wait, so instead do
+  #   kill -0 on it to see if it is still running
+  if ! kill -0 $backpid; then
+    wait $backpid || true
+    echo "WARNING: could not unshare, skipping unshare test"
```
jblomer: I think I'd rather like to enable/disable the test based on platforms. Otherwise we risk overlooking the warnings.  We can run it once on all platforms to see where we need to disable it.

test/src/094-fuser/main
```diff
@@ -0,0 +1,45 @@
...
+    CVMFS_GENERAL_WARNING_FLAG=1
+  else
+    pids="`sudo cvmfs_config fuser grid.cern.ch`"
+    [[ " $pids " == *" $backpid "* ]] || return 20
```
jblomer: As above, let's make sure cleanup is done properly if the test fails

jblomer (changes requested):
Many thanks!  Some minor comments inline.

We have had a race in the test numbering, could you rename the test to `095-fuser`?

DrDaveD:
I think I addressed all your comments except the one about the warning.  We would could run it as-is to find out which platforms would need to be disabled.  On el7 it would need to depend on whether or or not user namespaces was enabled, I think.  I wonder if the unshare command could be run as root without that.  I would have to test that.

jblomer:
@cernvm-bot cloudtest

jblomer:
The integration test run looks quite good!

I think it is a good idea to run the unshare command as root.  For the test, it does not matter whether or not user namespaces work. We'd then get EL7 as a test platform.

Only other related failure is a missing `lsof` on the EL8/fuse3 platform. With the new command, we need to add `lsof` as a cvmfs package dependency.

jblomer:
Closing for a new PR

PR #2748, opened 2021-07-15
Add support for attaching mount to an existing fuse module

jblomer:
Allows the mount helper to ask an existing fuse module for its /dev/fuse file descriptor and use it to attach the mountpoint. That helps when the fuse module serves a process with another mount namespace and has been unmounted from the system-wide mount namespace.

It requires a [kernel patch](https://sourceforge.net/p/fuse/mailman/message/37317340/).

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1603,6 +1603,27 @@ void UnregisterQuotaListener() {
   }
 }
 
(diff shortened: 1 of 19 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2750, opened 2021-07-16
Use systemctl to unmount filesystems when available

mcb30:
When systemctl is available, cvmfs_suid_helper will use it to mount
filesystems in order to prevent the cvmfs2 process from being
terminated unexpectedly by systemd (see CVM-1398 for details).

Generalise this logic to use systemctl for unmounting as well as
mounting, to avoid ending up in a situation in which the systemd mount
unit is still marked as active even though cvmfs_suid_helper has
unmounted the corresponding filesystem.

### DIFF
--- a/cvmfs/cvmfs_suid_helper.cc
+++ b/cvmfs/cvmfs_suid_helper.cc
@@ -302,20 +302,20 @@ int main(int argc, char *argv[]) {
   const string command = argv[1];
   const string fqrn = argv[2];
 
-  // Verify if repository exists
-  platform_stat64 info;
-  retval = platform_lstat((string(kSpoolArea) + "/" + fqrn).c_str(), &info);
-  if (retval != 0) {
-    fprintf(stderr, "unknown repository: %s\n", fqrn.c_str());
-    return 1;
-  }
-  repository_uid = info.st_uid;
-
-  // Verify if caller uid matches
-  if ((calling_uid != 0) && (calling_uid != repository_uid)) {
-    fprintf(stderr, "called as %d, repository owned by %d\n",
-            calling_uid, repository_uid);
-    return 1;
+  // Verify if caller uid matches (or is root)
+  if (calling_uid != 0) {
+    platform_stat64 info;
+    retval = platform_lstat((string(kSpoolArea) + "/" + fqrn).c_str(), &info);
+    if (retval != 0) {
+      fprintf(stderr, "unknown repository: %s\n", fqrn.c_str());
+      return 1;
+    }
+    repository_uid = info.st_uid;
+    if (calling_uid != repository_uid) {
+      fprintf(stderr, "called as %d, repository owned by %d\n",
+              calling_uid, repository_uid);
+      return 1;
+    }
   }
 
   if (command == "lock") {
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

mcb30:
I'm seeing multiple failures during test setup, which seem to arise from the direct invocation of `umount` in `unmount_and_teardown_repository()`: https://github.com/cvmfs/cvmfs/blob/4793b85da3dbe62a20b0fc7012c4e43ec3ac5642/cvmfs/server/cvmfs_server_common.sh#L1258-L1267

All other unmounting logic in this file seems to use `run_suid_helper`, so I suspect that this logic should do the same.

@jblomer I'll update `unmount_and_teardown_repository()` to use `run_suid_helper` unless you think this would be the wrong thing to do?

mcb30:
> @jblomer I'll update `unmount_and_teardown_repository()` to use `run_suid_helper` unless you think this would be the wrong thing to do?

@jblomer I've added a commit to do this.  It still passes tests locally - could you please kick off the cloudtests again?

Thanks.

jblomer:
@cernvm-bot cloudtest

mcb30:
@jblomer Have force-pushed an update - could you please trigger cloudtests again?  Thanks!

jblomer:
@cernvm-bot cloudtest

mcb30:
Lots of the tests seem to be failing because `cvmfs_suid_helper` is reporting "Unknown repository":
```
Repository test.cern.ch is already present... removing it
Wiping local repository storage in gateway setup /srv/cvmfs/test.cern.ch
Unmounting CernVM-FS Area... unknown repository: test.cern.ch
done
Removing Spool Area... rmdir: failed to remove '/cvmfs/test.cern.ch': Device or resource busy
```
 I have unfortunately not been able to reproduce this failure locally.

I've pushed an extra commit that bypasses the repository existence check for the root user, on the assumption that something in the cloudtests setup might be deleting the `/var/spool/cvmfs/<reponame>` directory between tests.  @jblomer Could you retrigger cloudtests again, if this seems like a viable fix?

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot help

jblomer:
@cernvm-bot cloudtest

mcb30:
@jblomer Thanks for helping me establish a test environment that reproduces the CI builders.  I think I've found the root cause - `systemctl stop` doesn't work on any autofs mounts, since there is no generated mount unit for those.

I tried extending this PR to always perform both `systemctl stop` and `umount`, so as to be able to cope with both autofs and fstab mount points, but it's getting to be far too messy.

I still believe that avoiding `systemctl` altogether is the correct solution (which would require finding an alternative fix for CVM-1398).  However, an alternative change that at least fixes the symptoms on Fedora 34 is to use `systemctl restart` instead of `systemctl start` - that is sufficient to make the mount happen as expected even when mixed with manual `umount`s.

Closing this in favour of #2752, which implements the `systemctl restart` approach.

PR #2752, opened 2021-07-22
Ensure that filesystem is really mounted when using systemctl

mcb30:
On some versions of systemd (observed with systemd 248 on Fedora 34),
the systemd mount unit activity status seems not to track the mount
state.  In particular, the following sequence of events:

  - systemctl start cvmfs-repo.name.mount
  - umount /cvmfs/repo.name
  - systemctl start cvmfs-repo.name.mount

will fail to remount the filesystem, since systemd believe that the
mount unit is still in the active state after the manual "umount".

Work around this problem by using "systemctl restart" instead of
"systemctl start" in cvmfs_suid_helper.

### DIFF
--- a/cvmfs/cvmfs_suid_helper.cc
+++ b/cvmfs/cvmfs_suid_helper.cc
@@ -94,7 +94,7 @@ static void Mount(const string &path) {
       }
       systemd_unit = cvmfs_suid::EscapeSystemdUnit(resolved_path);
     }
-    ExecAsRoot("/bin/systemctl", "start", systemd_unit.c_str(), NULL);
+    ExecAsRoot("/bin/systemctl", "restart", systemd_unit.c_str(), NULL);
   } else {
     ExecAsRoot("/bin/mount", path.c_str(), NULL, NULL);
   }


jblomer:
@cernvm-bot cloudtest

cvmfs/cvmfs_suid_helper.cc
```diff
@@ -94,7 +94,7 @@ static void Mount(const string &path) {
       }
       systemd_unit = cvmfs_suid::EscapeSystemdUnit(resolved_path);
     }
-    ExecAsRoot("/bin/systemctl", "start", systemd_unit.c_str(), NULL);
+    ExecAsRoot("/bin/systemctl", "restart", systemd_unit.c_str(), NULL);
```
jblomer, line 97: I vaguely remember that "restart" does not do anything if the unit is not already started. But I might be wrong about that or perhaps systemd changed behavior.  Integration tests should tell.

mcb30: From the `systemctl` manpage:
> restart PATTERN...
>    Stop and then start one or more units specified on the command line. If the units are not running yet, they will be started.
>
> try-restart PATTERN...
>     Stop and then start one or more units specified on the command line if the units are running. This does nothing if units are not running.

so using `restart` should be OK, and it seems to work for me (on Fedora 34).

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1162/

From a quick sample of the test results, these failures all appear to relate to a timeout error on `cvmfschecksum.grid.cern.ch`.  As far as I can tell these are unrelated failures caused by an external resource not responding - @jblomer could you let me know if there is a real problem here that needs to be investigated.  Thanks!

jblomer:
The failures look unrelated.  Hm, I'll kick the tests off again, just to see if it is a systematic problem or not.

jblomer:
@cernvm-bot cloudtest

PR #2756, opened 2021-07-26
Reset restart counter after successfully starting a mount unit

mcb30:
Some test cases will perform a rapid sequence of mounts and unmounts.
Since mounts are performed by starting a systemd mount unit, this can
result in tripping the restart rate limit counter, with an error such
as:

  cvmfs-test.cern.ch.mount: Start request repeated too quickly

Fix by resetting the restart counter (via "systemctl reset-failed")
after a successful mount.

### DIFF
--- a/cvmfs/cvmfs_suid_helper.cc
+++ b/cvmfs/cvmfs_suid_helper.cc
@@ -59,6 +59,25 @@ static void ExecAsRoot(const char *binary,
   exit(1);
 }
 
+static void ForkAndExecAsRoot(const char *binary,const char *arg1,
+                              const char *arg2, const char *arg3)
+{
+  pid_t child = fork();
+  if (child == -1) {
+    fprintf(stderr, "failed to fork %s... (%d)\n", binary, errno);
+    exit(1);
+  } else if (child == 0) {
+    ExecAsRoot(binary, arg1, arg2, arg3);
+  } else {
+    int wstatus;
+    waitpid(child, &wstatus, 0);
+    if (WIFSIGNALED(wstatus)) {
+      exit(128 + WTERMSIG(wstatus));
+    } else if (WIFEXITED(wstatus) && WEXITSTATUS(wstatus)) {
+      exit(WEXITSTATUS(wstatus));
+    }
+  }
+}
 
 static void Remount(const string &path, const RemountType how) {
   string remount_option = "remount,";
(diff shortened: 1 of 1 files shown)


cvmfs/cvmfs_suid_helper.cc
```diff
@@ -59,6 +59,22 @@ static void ExecAsRoot(const char *binary,
   exit(1);
 }
 
+static void RunAsRoot(const char *binary,
+                      const char *arg1, const char *arg2, const char *arg3)
+{
+  pid_t child = fork();
+  if (child == 0) {
```
jblomer: Let's explicitly handle `child == -1`

cvmfs/cvmfs_suid_helper.cc
```diff
@@ -94,7 +110,8 @@ static void Mount(const string &path) {
       }
       systemd_unit = cvmfs_suid::EscapeSystemdUnit(resolved_path);
     }
-    ExecAsRoot("/bin/systemctl", "restart", systemd_unit.c_str(), NULL);
+    RunAsRoot("/bin/systemctl", "restart", systemd_unit.c_str(), NULL);
```
jblomer: Perhaps `ExecAsRootOrAbort`?

jblomer (changes requested):

jblomer:
@cernvm-bot cloudtest

cvmfs/cvmfs_suid_helper.cc
mcb30: Will do

cvmfs/cvmfs_suid_helper.cc
mcb30: I was trying to convey the non-`exec()` nature of the function `RunAsRoot()` (i.e. the fact that it actually returns, rather than replacing the current process).  Happy to switch to `ExecAsRootOrAbort()` if you prefer.

mcb30: I've switched to `ForkAndExecAsRoot()`: let me know if that's ok

mcb30:
@jblomer Have updated with requested changes

cvmfs/cvmfs_suid_helper.cc
jblomer: That makes more sense, thanks!

jblomer (approved):

jblomer:
@cernvm-bot cloudtest

PR #2757, opened 2021-07-26
Use "setcap -v" rather than "getcap" to verify capabilities

mcb30:
The format of the output from "getcap" was changed in libcap commit
177cd41 ("A more compact form for the text representation of
capabilities").  Test 585 currently expects the output:

  /cvmfs/test.cern.ch/dir/capabilities = cap_chown+ep

but any version of libcap from version 2.41 onwards will instead
produce:

  /cvmfs/test.cern.ch/dir/capabilities cap_chown=ep

Fix by using "setcap -v" to verify the capability set, since this
method does not depend on the precise "getcap" output format.


jblomer:
@cernvm-bot cloudtest

jblomer (approved):

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1168/

I can't see any pattern in these failures: it looks to me as though none of them are related to this PR.

PR #2758, opened 2021-07-26
Minor fixes to attach mount

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2760, opened 2021-07-27
Add clang-tidy default and a first set of fixes (for /mount)

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2764, opened 2021-07-30
Add cvmfs_config fuser command (PR version 2)

jblomer:
Replaces #2747


jblomer:
@cernvm-bot cloudtest

PR #2765, opened 2021-07-30
Use "Cache-Control: no-cache" instead of bypassing system proxy

mcb30:
The get_item() shell function has an optional "noproxy" parameter.
The purpose of this parameter is unclear, but the most plausible
explanation seems to be as a means of ensuring that the download
fetches the current upstream file, ignoring any versions that may have
previously been cached in intermediate proxy servers.

This fails on any system where use of a proxy server is required for
all downloads, since direct origin server fetches are not permitted on
such systems.

Unconditionally add the HTTP/1.1 standard request header
"Cache-Control: no-cache" to force an origin server download, thereby
eliminating the need to bypass the system proxy.

### DIFF
--- a/cvmfs/server/cvmfs_server_common.sh
+++ b/cvmfs/server/cvmfs_server_common.sh
@@ -503,21 +503,15 @@ is_due_auto_garbage_collection() {
 # download a given file from the backend storage
 # @param name  the name of the repository to download from
 # @param url   the url to download from
-# @param noproxy  (optional)
 get_item() {
   local name="$1"
   local url="$2"
-  local noproxy="$3"
 
   load_repo_config $name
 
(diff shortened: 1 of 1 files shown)


DrDaveD:
Where is this used from?  We have to be careful that it is never used from large numbers of nodes at a time.   It's generally much better to use something like `Cache-Control: max-age=N` where N is a number of seconds.

mcb30:
> Where is this used from? We have to be careful that it is never used from large numbers of nodes at a time. It's generally much better to use something like `Cache-Control: max-age=N` where N is a number of seconds.

This is all within the `cvmfs_server` script: it's not code that ever executes on the client nodes.

mcb30:
Closing this PR in favour of #2766, which builds upon this commit.

PR #2766, opened 2021-08-04
Allow use of proxy server by publisher and gateway machines

mcb30:
This patch series adds the ability to specify a proxy server to be used by a publisher or gateway machine in order to download files from the Stratum 0 repository.

This is particularly useful when the Stratum 0 files are held in cloud storage such as AWS S3: it allows for the use of an authenticating proxy (to avoid exposing the S3 reader credentials directly to every client) and opens up the possibility of using the proxy to cache files uploaded by a publisher so that the gateway does not need to waste time immediately re-downloading those same files from the cloud storage.

### DIFF
--- a/cvmfs/publish/settings.cc
+++ b/cvmfs/publish/settings.cc
@@ -280,6 +280,7 @@ SettingsRepository::SettingsRepository(
   const SettingsPublisher &settings_publisher)
   : fqrn_(settings_publisher.fqrn())
   , url_(settings_publisher.url())
+  , proxy_(settings_publisher.proxy())
   , tmp_dir_(settings_publisher.transaction().spool_area().tmp_dir())
   , keychain_(settings_publisher.fqrn())
 {
@@ -290,6 +291,11 @@ SettingsRepository::SettingsRepository(
 void SettingsRepository::SetUrl(const std::string &url) {
   // TODO(jblomer): sanitiation, check availability
   url_ = url;
+}
+
+
+void SettingsRepository::SetProxy(const std::string &proxy) {
+  proxy_ = proxy;
 }
 
 
@@ -313,6 +319,7 @@ SettingsPublisher::SettingsPublisher(
   const SettingsRepository &settings_repository)
   : fqrn_(settings_repository.fqrn())
   , url_(settings_repository.url())
+  , proxy_(settings_repository.proxy())
   , owner_uid_(0)
   , owner_gid_(0)
   , whitelist_validity_days_(kDefaultWhitelistValidity)
@@ -329,6 +336,11 @@ SettingsPublisher::SettingsPublisher(
 void SettingsPublisher::SetUrl(const std::string &url) {
   // TODO(jblomer): sanitiation, check availability
   url_ = url;
+}
+
+
+void SettingsPublisher::SetProxy(const std::string &proxy) {
+  proxy_ = proxy;
 }
 
 
@@ -448,6 +460,8 @@ SettingsRepository SettingsBuilder::CreateSettingsRepository(
     settings.GetKeychain()->SetKeychainDir(arg);
   if (options_mgr_->GetValue("CVMFS_STRATUM0", &arg))
     settings.SetUrl(arg);
+  if (options_mgr_->GetValue("CVMFS_SERVER_PROXY", &arg))
+    settings.SetProxy(arg);
   // For a replica, the stratum 1 url is the "local" location, hence it takes
   // precedence over the stratum 0 url
   if (options_mgr_->GetValue("CVMFS_STRATUM1", &arg))
@@ -492,6 +506,8 @@ SettingsPublisher* SettingsBuilder::CreateSettingsPublisherFromSession() {
 
   std::string arg;
   settings_publisher->SetUrl(settings_publisher->GetReadOnlyXAttr("user.host"));
+  settings_publisher->SetProxy(
+    settings_publisher->GetReadOnlyXAttr("user.proxy"));
   if (omgr.GetValue("CVMFS_KEYS_DIR", &arg))
     settings_publisher->GetKeychain()->SetKeychainDir(arg);
   settings_publisher->GetTransaction()->SetLayoutRevision(
--- a/cvmfs/publish/settings.h
+++ b/cvmfs/publish/settings.h
@@ -354,17 +354,20 @@ class SettingsRepository {
   explicit SettingsRepository(const std::string &fqrn)
     : fqrn_(fqrn)
     , url_(std::string("http://localhost/cvmfs/") + fqrn_())
+    , proxy_("")
     , tmp_dir_("/tmp")
     , keychain_(fqrn)
   {}
   explicit SettingsRepository(const SettingsPublisher &settings_publisher);
 
   void SetUrl(const std::string &url);
+  void SetProxy(const std::string &proxy);
   void SetTmpDir(const std::string &tmp_dir);
   void SetCertBundle(const std::string &cert_bundle);
 
   std::string fqrn() const { return fqrn_(); }
   std::string url() const { return url_(); }
+  std::string proxy() const { return proxy_(); }
   std::string tmp_dir() const { return tmp_dir_(); }
   std::string cert_bundle() const { return cert_bundle_(); }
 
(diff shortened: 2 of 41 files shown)


cvmfs/server/cvmfs_server_common.sh
```diff
@@ -503,21 +503,15 @@ is_due_auto_garbage_collection() {
...
   load_repo_config $name
 
-  if [ x"$noproxy" != x"" ]; then
-    unset http_proxy && curl -f $(get_x509_cert_settings) \
-      $(get_follow_http_redirects_flag) "$url" 2>/dev/null | tr -d '\0'
-  else
-    curl -f $(get_x509_cert_settings) $(get_follow_http_redirects_flag) \
-      "$url" 2>/dev/null | tr -d '\0'
-  fi
+  curl -f -H "Cache-Control: no-cache" $(get_curl_proxy) \
```
DrDaveD: I recommend using `Cache-Control: max-age=0` instead of `no-cache`.  They're close, but not exactly the same thing.  With `max-age=0` a proxy may use a cached copy if it has been proven to be valid with `If-Modified-Since`, where `no-cache` means it absolutely cannot reuse a cached copy.  `max-age=0` is gentler on caches.  See the user agent section of this [stackover flow answer](https://stackoverflow.com/a/1383359/10457761) for more details.

mcb30: Sure, that's an easy change and doesn't materially impact the code.  I'll update it now.

mcb30: Updated to use `max-age=0` instead

jblomer:
@cernvm-bot all

jblomer:
@cernvm-bot help

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot tidy

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

cvmfs/publish/settings.h
```diff
@@ -357,18 +357,21 @@ class SettingsRepository {
     : fqrn_(fqrn)
     , url_(std::string("http://localhost/cvmfs/") + fqrn_())
     , tmp_dir_("/tmp")
+    , proxy_("")
```
jblomer, line 357: Let's change the order of the members and put `url_` and `proxy_` together

cvmfs/publish/settings.cc
```diff
@@ -456,6 +468,8 @@ SettingsRepository SettingsBuilder::CreateSettingsRepository(
     settings.SetTmpDir(arg + "/tmp");
   if (options_mgr_->GetValue("X509_CERT_BUNDLE", &arg))
     settings.SetCertBundle(arg);
+  if (options_mgr_->GetValue("CVMFS_HTTP_PROXY", &arg))
```
jblomer: We should as well set the proxy in `CreateSettingsPublisherFromSession()` from the `user.proxy` extended attribute.

cvmfs/receiver/params.h
```diff
@@ -34,6 +34,7 @@ struct Params {
   size_t max_weight;
   size_t min_weight;
   bool upload_stats_db;
+  std::string proxy;
```
jblomer: Also here let's put `stratum0` and `proxy` together.

jblomer:
We are just getting used to clang-tidy checks. The warnings are not related to the patch, afaics. I'll open a separate PR to fix the code surrounding the changes.

cvmfs/server/cvmfs_server_common.sh
```diff
@@ -556,6 +550,19 @@ get_x509_cert_settings() {
   fi
 }
 
+# Parse proxy server for curl command
+get_curl_proxy() {
+  if [ x"$CVMFS_HTTP_PROXY" != x"DIRECT" ]; then
```
jblomer: I think we should also ignore an empty environment variable

cvmfs/server_tool.cc
```diff
@@ -34,6 +35,11 @@ bool ServerTool::InitDownloadManager(const bool follow_redirects,
   download_manager_->SetRetryParameters(kDownloadRetries, 2000, 5000);
   download_manager_->UseSystemCertificatePath();
 
+  if (proxy != "") {
```
jblomer, line 37: With the possibility to explicitly set a proxy server, I think we should remove the "use_system_proxy" makeshift mechanism.

jblomer (changes requested):
Many thanks!  I think an explicit proxy setting for the server side is quite useful!

For certain stratum 1 sites, we already had the possibility of using the system http proxy (`http_proxy` environment variable).  I think we should drop that in favor of administrators explicitly setting a proxy.

It might make sense to use a different parameter name, e.g. `CVMFS_SERVER_HTTP_PROXY`, because on the server side we only support a single proxy and not the client-side load-balancing configuration.

We would also need an integration test that creates a repository and makes a small publish (we can later extend the test if we need to). The test could for instance use Apache's built-in proxy, or setup a squid on the fly (or another forward proxy), whatever is simplest.

mcb30:
Thanks for the feedback!  I'll get on this now and have a new version ready within a couple of hours

cvmfs/receiver/params.h
mcb30: Done

cvmfs/publish/settings.cc
mcb30: Done

cvmfs/publish/settings.h
mcb30: Done

mcb30:
> It might make sense to use a different parameter name, e.g. `CVMFS_SERVER_HTTP_PROXY`, because on the server side we only support a single proxy and not the client-side load-balancing configuration.

Good idea.  I've switched to `CVMFS_SERVER_PROXY` in the `server.conf` file.

cvmfs/server/cvmfs_server_common.sh
mcb30: After switching to `CVMFS_SERVER_PROXY`, the test is now:
```
if [ x"$CVMFS_SERVER_PROXY" != x"" ]; then
```
since `CVMFS_SERVER_PROXY` will be defined only if it contains a proxy URL (i.e. it will never exist with the value `DIRECT` anyway)

cvmfs/server_tool.cc
mcb30: I've added a separate commit to remove the `use_system_proxy` mechanism.

mcb30:
@jblomer All comments should be addressed apart from the integration test, which I'm putting together now.  Please feel free to review the updated commits, since I'm not planning any further code changes at this point.

mcb30:
I've added a basic integration test.  Looking at the cloudtest failures now...

jblomer:
@cernvm-bot cloudtest

mcb30:
@jblomer I can't immediately see any reason for the consistent cloudtest failures across x86_64 (but passing on x86).  Does the cloudtest environment utilise any kind of system proxy server?

mcb30:
Found the problem - a missing `$` in one of the scripts.  Please abort the current cloudtest run

jblomer:
@mcb30 They are aborted.

mcb30:
I've force-pushed the fix - should be worth restarting the tests now, thanks.

jblomer:
@cernvm-bot cloudtest

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1181/

The few remaining cloudtest failures now all appear to be unrelated to this PR.

jblomer:
Could you rebase to the latest devel branch? I meanwhile tried to address the clang-tidy errors in the surroundings of the PR.

jblomer (approved):

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot tidy

mcb30:
Rebased onto latest devel and updated to fix the linter warning

jblomer:
@cernvm-bot tidy

jblomer:
@cernvm-bot cpplint

PR #2768, opened 2021-08-05
[tidy] Fix several clang-tidy detected issues

jblomer:
Follow up from #2766


jblomer:
@cernvm-bot cloudtest

PR #2769, opened 2021-08-06
[tidy] Resolve most issues in download.cc

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2773, opened 2021-08-23
Treat HTTP 503 Service Unavailable as a possible proxy error

mcb30:
Proxy servers such as Squid or Varnish will typically return an HTTP
503 Service Unavailable error when they are unable to connect to the
origin server.  CVMFS will currently treat HTTP 503 as a host error,
which will prevent the client from failing over to the next available
proxy server.

Fix by treating an HTTP 503 as a potential proxy error (as is
currently done for any unrecognised HTTP 4XX errors).

### DIFF
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -261,6 +261,17 @@ static size_t CallbackCurlHeader(void *ptr, size_t size, size_t nmemb,
   } else if (HasPrefix(header_line, "LOCATION:", true)) {
     // This comes along with redirects
     LogCvmfs(kLogDownload, kLogDebug, "%s", header_line.c_str());
+  } else if (HasPrefix(header_line, "X-SQUID-ERROR:", true)) {
+    // Reinterpret host error as proxy error
+    if (info->error_code == kFailHostHttp) {
+      info->error_code = kFailProxyHttp;
+    }
+  } else if (HasPrefix(header_line, "PROXY-STATUS:", true)) {
+    // Reinterpret host error as proxy error if applicable
+    if ((info->error_code == kFailHostHttp) &&
+        (header_line.find("error=") != string::npos)) {
+      info->error_code = kFailProxyHttp;
+    }
   }
 
   return num_bytes;
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -217,8 +217,8 @@ static size_t CallbackCurlHeader(void *ptr, size_t size, size_t nmemb,
       // libcurl will handle this because of CURLOPT_FOLLOWLOCATION
       return num_bytes;
     } else {
-      LogCvmfs(kLogDownload, kLogDebug, "http status error code: %s",
-               header_line.c_str());
+      LogCvmfs(kLogDownload, kLogDebug, "http status error code: %s [%d]",
+               header_line.c_str(), info->http_code);
       if (((info->http_code / 100) == 5) ||
           (info->http_code == 400) || (info->http_code == 404))
       {


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot tidy

mcb30:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1189/

Looks as though Fedora 32 x86_64 had a single transient failure, and Ubuntu 16.04 x86_64 had an unrelated failure that then broke all subsequent tests (almost all of which hit a `failed to clean environment` error during setup)

DrDaveD:
The current behavior is [documented](https://cvmfs.readthedocs.io/en/stable/cpt-configure.html#failover-rules) so at minimum if the behavior changes the document has to change.  In what cases is 503 Service Unavailable a proxy failure?  Why would we not want it to keep the same proxy and try the next server, the current behavior?  The cvmfs failover rules are based on the [frontier client retry strategy](https://twiki.cern.ch/twiki/bin/view/Frontier/ClientRetryStrategy) and it treats all 500 series errors as server errors.

DrDaveD:
Servers will also return 503 codes when they can't contact their back ends or when they are too busy and I would hate to cause extra loads on them by retrying them with every proxy.

DrDaveD:
If it's a matter of a proxy's outgoing network being down, Linux will typically try for longer than cvmfs does, so cvmfs will get a timeout and consider it a proxy error.  If it is connection refused, that happens right away but it indicates a server error, not a proxy error.  So are there any cases that a proxy will return a 503 Service Unavailable in less than 10 seconds that really is a problem on the proxy?

mcb30:
> If it's a matter of a proxy's outgoing network being down, Linux will typically try for longer than cvmfs does, so cvmfs will get a timeout and consider it a proxy error. If it is connection refused, that happens right away but it indicates a server error, not a proxy error. So are there any cases that a proxy will return a 503 Service Unavailable in less than 10 seconds that really is a problem on the proxy?

Yes, these situations do exist:
* A proxy in maintenance mode will return a 503 Service Unavailable
* An overloaded proxy will return a 503 Service Unavailable
* A proxy suffering a temporary DNS outage will return a 503 Service Unavailable

This is a case of type 1 vs type 2 errors:
* If the origin server returns 503 but CVMFS considers it to be a proxy error, then CVMFS will erroneously send some more requests to the same origin server.  This will degrade service slightly.
* If the proxy server returns 503 but CVMFS considers it to be an origin server error, then CVMFS will erroneously abandon the entire request and return EIO to the application.  This will cause a total failure of the service.  (We have observed this happening in real-world deployments.)

Given these impacts and since it is plausible that the 503 could be generated by either the proxy or the origin server, the safest course of action is to assume that it represents a proxy error.

DrDaveD:
You have observed these behaviors in squid, or only varnish?  I have never seen a squid in maintenance mode or returning a 503 for being overloaded.  I'll have to try simulating a DNS failure.

Does varnish actually do a decent job of being a forward proxy?  Last I checked it was only good as a reverse proxy, which cvmfs treats as a server.  If the problem is only in varnish perhaps there should be a non-default option in the cvmfs client.

DrDaveD:
I was able to reproduce a 503 error on a simulated DNS failure on squid.  The first time it happens for a particular host it takes longer the cvmfs timeout time, but after that squid negatively caches the error for a period of time.

This type of error can be distinguished between real server errors with squid, however, because squid inserts an X-Squid-Error header.  Does varnish have a similar header?  I would prefer to only treat it as a proxy failure by default if the extra header is there.  The 503 error is almost always a server error, and it's quite common.

mcb30:
The closest I can find to a standardised way to identify a response generated by a proxy is the draft RFC https://httpwg.org/http-extensions/draft-ietf-httpbis-proxy-status.html#name-the-proxy-status-http-field which defines a 'Proxy-Status' header with values such as:
```
Proxy-Status: SomeCDN; error=destination_unavailable
```

CVMFS could potentially apply a heuristic of assuming that if a response comes via a proxy (i.e. is not `DIRECT`) and includes a `Proxy-Status` header which includes an `error=` indication, then this response represents a proxy error rather than an origin server error.

We can configure our proxy to include a `Proxy-Status` header for any synthesised response, so this would work for our setup, and would hopefully become generally useful in future once the `Proxy-Status` header is more widely supported.

DrDaveD:
Good find.  I would support that, and have it also look for X-Squid-Error which would take care of the vast majority of existing deployments.

If you're using varnish as a forward proxy and can recommend a specific configuration for use with cvmfs I suggest that you also submit an update to the cvmfs documentation with that recommendation.

mcb30:
> Good find. I would support that, and have it also look for X-Squid-Error which would take care of the vast majority of existing deployments.

I'm happy to implement support for both `Proxy-Status` and `X-Squid-Error` if that is the preferred approach.  @jblomer do you have any comments?
 
> If you're using varnish as a forward proxy and can recommend a specific configuration for use with cvmfs I suggest that you also submit an update to the cvmfs documentation with that recommendation.

We are using Varnish configured as a pseudo-forward proxy: it has a limited set of predefined backends (one per CVMFS upstream) so appears similar to a normal proxy with a very strict access policy.  I'm not sure how much of our implementation is generally applicable, but we could think about pulling together some documentation.

jblomer:
> I'm happy to implement support for both `Proxy-Status` and `X-Squid-Error` if that is the preferred approach. @jblomer do you have any comments?

I think that makes sense! It was always a problem that we could not clearly distinguish between proxy and server errors and this approach hopefully gives us a good heuristics.
 
> We are using Varnish configured as a pseudo-forward proxy: it has a limited set of predefined backends (one per CVMFS upstream) so appears similar to a normal proxy with a very strict access policy. I'm not sure how much of our implementation is generally applicable, but we could think about pulling together some documentation.

That would be much appreciated!

mcb30:
@jblomer @DrDaveD I've updated to check for `X-Squid-Error` and `Proxy-Status` as discussed - please let me know if this looks sensible.

DrDaveD (approved):
Looks good.  This should be coupled with a corresponding change to the "latest" branch of the [documentation](https://cvmfs.readthedocs.io/en/stable/cpt-configure.html#failover-rules).

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot tidy

mcb30:
> Looks good. This should be coupled with a corresponding change to the "latest" branch of the [documentation](https://cvmfs.readthedocs.io/en/stable/cpt-configure.html#failover-rules).

Created as https://github.com/cvmfs/doc-cvmfs/pull/173

mcb30:
The unittest failures appear to be unrelated to the change: it looks as though the Mac environment is missing IPv4 configuration:
```
/Users/sftnight/jenkins/workspace/CvmfsUnittest/CVMFS_BUILD_ARCH/osx10-x86_64/CVMFS_BUILD_PLATFORM/mac/cvmfs/test/unittests/t_dns.cc:193
Value of: host.HasIpv4()
  Actual: false
Expected: true
```

mcb30:
> cloudtests finished: FAILURE https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1237/

As far as I can tell, these are unrelated failures occurring when trying to list extended attributes on `grid.cern.ch`

jblomer:
Thanks a lot!

PR #2778, opened 2021-09-04
[ducc] Fix authenticated requests to registries

jblomer:
Rename `DUCC_DOCKER_REGISTRY_PASS` --> `DUCC_OUTPUT_REGISTRY_PASS`.

Introduces `DUCC_AUTH_REGISTRIES=val1,val2,...` environment variable to
specify names of registries requiring user name and password. For each
$name from the list, there needs to be 3 environment variables

    DUCC_$name_USER
    DUCC_$name_PASS
    DUCC_$name_IDENT

The last one is a string that is used to identify URLs of that registry,
e.g. "registry.hub.docker.com".

### DIFF
--- a/ducc/CMakeLists.txt
+++ b/ducc/CMakeLists.txt
@@ -8,12 +8,12 @@ endif()
 
 add_custom_target(
   cvmfs_ducc ALL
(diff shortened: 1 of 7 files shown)


jblomer:
@cernvm-bot cloudtest

PR #2781, opened 2021-09-21
Use repository name as the fsname option for read-only mount points [v2]

jblomer:
Replaces #2755


jblomer:
@cernvm-bot cloudtest

PR #2783, opened 2021-09-21
Fix cvmfs_talk host info for empty host chain

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2784, opened 2021-09-22
Fix rpmbuild/cmake invocation for FC34

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2785, opened 2021-09-30
Remove version requirement from selinux-policy dependency

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2786, opened 2021-09-30
Add fuse version to bugreport tarball

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2788, opened 2021-10-04
Fix certificate store handling for explicitly set bundles

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2789, opened 2021-10-04
Disable test 095 on Ubuntu 16

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2790, opened 2021-10-05
Fixing the CMake config of the gateway component when using the Ninja generator

radupopescu:


jblomer:
@cernvm-bot cloudtest

PR #2791, opened 2021-10-06
Use UTC timestamp for .cvmfs_is_snapshotting

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2793, opened 2021-10-07
Fix-up integration tests after config repo changes

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2777, opened 2021-09-01
Fix import s3

hmeiland:
renewing pull request for fixing import of s3 backed cvmfs (closing old one due to merge conflicts)

PR #2795, opened 2021-10-12
Fix import of S3 repositories [v2]

jblomer:
Replaces #2777


jblomer:
@cernvm-bot cloudtest

PR #2796, opened 2021-10-12
[rpm] Cleanup creation of cvmfs system user and group

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2799, opened 2021-10-15
Kill any remaining cvmfs_receiver processes when stopping gateway service

radupopescu:


jblomer (commented):
I think `killall` is part of the psmisc package (at least on EL), which we should then make dependency of the cvmfs-gateway package.

radupopescu:
> I think `killall` is part of the psmisc package (at least on EL), which we should then make dependency of the cvmfs-gateway package.

Is there still a separate gateway package, or is it just `cvmfs-server` now?

jblomer:
> > I think `killall` is part of the psmisc package (at least on EL), which we should then make dependency of the cvmfs-gateway package.
> 
> Is there still a separate gateway package, or is it just `cvmfs-server` now?

It's still a separate package, but a sub package of the cvmfs package (i.e. it comes all from the same build process).  The sources are under packaging/rpm and packaging/deb.

PR #2800, opened 2021-10-19
[ducc] make output_format line in wish list optional

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2802, opened 2021-10-19
Webhook registry listener

aandvalenzuela:


aandvalenzuela:
@cernvm-bot cloudtest

aandvalenzuela:
@cernvm-bot cloudtest

ducc/webhook/registry-listener.go
```diff
@@ -0,0 +1,77 @@
...
+    file.Seek(0, os.SEEK_END)
+    bf := bufio.NewReader(file)
+
+    go func() {
+        for {
+            line, _, err := bf.ReadLine()
+            if len(line) != 0 {
+                changes <- string(line)
+            } else if err == io.EOF {
+                time.Sleep(1 * time.Millisecond)
```
radupopescu: I think a longer sleep duration is fine here, 1ms is a bit agressive.

`1 * time.Second` should be ok for general use. If there is a test depending on this sleep time, I would make it configurable and set it to a shorter time in the test context.

ducc/webhook/registry-listener.go
```diff
@@ -0,0 +1,77 @@
...
+        log.Fatalf("OpenFile: %s", err)
+        }
+
+    changes := ReadChanges(file)
+
+    for {
+        msg := <-changes
+
+        if msg == "xx|file rotation|xx" {
+                NotifyDucc(file_name, repository_name)
```
radupopescu: This recursive call means that this long-running process is leaking file descriptors. The file opened inside `ReadChanges` is opened but never closed. It also means the old (pre-rotation) version of the file is kept around until the process is finally stopped.

ducc/webhook/registry-webhook.py
```diff
@@ -46,8 +47,10 @@ def publish_message(notifications_file, action, image):
             if (last_line_id % int(args_dic['rotation']) == 0 and last_line_id != 0):
                 new_notifications_file = str(first_line_id)+'-'+str(last_line_id)+notifications_file.split('/')[-1]
                 os.rename(notifications_file, new_notifications_file)
+                open(notifications_file, 'a').close()
                 with open(new_notifications_file, 'a+') as f:
                     f.write('xx|file rotation|xx\n')
+                time.sleep(1)
```
radupopescu: This synchronization logic between `registry-webhook.py` and `registry-listener.go` looks quite fragile, especially given the file rotation process.

I think it would be easier having both sides in the same file. Wouldn't it be possible to call the `cvmfs_ducc ...` command line directly from the Python process?

radupopescu (commented):
Thanks for the PR and good job!

I left some notes and some minor comments.

aandvalenzuela:
@cernvm-bot cloudtest

aandvalenzuela:
@cernvm-bot cloudtest

jblomer:
The failures are unrelated, the VMs simply did not boot.

jblomer:
Should we convert from draft to "ready for review"? It looks good to me!

jblomer (approved):

radupopescu:
Looks good to me, too. The Go listener is now able to replay items from the beginning of the last notification file.

Great work!

jblomer:
@cernvm-bot cloudtest

radupopescu (approved):

PR #2803, opened 2021-10-19
Add cvmfs_server check -a option

DrDaveD:
This adds an cvmfs_server check -a option, covering points 1 through 3 in [CVM-1524](https://sft.its.cern.ch/jira/browse/CVM-1524).

### DIFF
--- a/cvmfs/server/cvmfs_server_check.sh
+++ b/cvmfs/server/cvmfs_server_check.sh
@@ -164,6 +164,42 @@ __check_repair_reflog() {
   fi
 }
 
+# This is a separate function because dash segfaults if it is inline :-(
+__get_checks_repo_times() {
+  set -- '*'
+  check_parameter_count_for_multiple_repositories $#
+  names=$(get_or_guess_multiple_repository_names "$@")
+  check_multiple_repository_existence "$names"
+
+  for name in $names; do 
+    # note that is_inactive_replica also does load_repo_config
+    if is_inactive_replica $name; then
+      continue
+    fi
+
+    local upstream=$CVMFS_UPSTREAM_STORAGE
+    if [ x$(get_upstream_type $upstream_storage) = "xgw" ]; then
+      continue
+    fi
+
+    local check_status="$(read_repo_item $name .cvmfs_status.json)"
+    local last_check="$(get_json_field "$check_status" last_check)"
+    local check_time=0
+    if [ -n "$last_check" ]; then
+      check_time="$(date --date "$last_check" +%s)"
+      local min_secs num_secs
+      min_secs="$((${CVMFS_CHECK_ALL_MIN_DAYS:-30}*60*60*24))"
+      num_secs="$(($(date +%s)-$check_time))"
+      if [ "$num_secs" -lt "$min_secs" ]; then
+        # less than $CVMFS_CHECK_ALL_MIN_DAYS has elapsed since last check
+        continue
+      fi
+    fi
+
+    echo "${check_time}:${name}"
+  done
+}
+
 __do_all_checks() {
   local log
   local repo
@@ -185,40 +221,7 @@ __do_all_checks() {
   log=/var/log/cvmfs/checks.log
 
   # Sort the active repositories on local storage by last check time
-  repos="$((
-    set -- '*'
-    check_parameter_count_for_multiple_repositories $#
-    names=$(get_or_guess_multiple_repository_names "$@")
-    check_multiple_repository_existence "$names"
-
-    for name in $names; do 
-      # note that is_inactive_replica also does load_repo_config
-      if is_inactive_replica $name; then
-        continue
-      fi
-
-      local upstream=$CVMFS_UPSTREAM_STORAGE
-      if [ x$(get_upstream_type $upstream_storage) = "xgw" ]; then
-        continue
-      fi
-
-      local check_status="$(read_repo_item $name .cvmfs_status.json)"
-      local last_check="$(get_json_field "$check_status" last_check)"
-      local check_time=0
-      if [ -n "$last_check" ]; then
-        check_time="$(date --date "$last_check" +%s)"
-        local min_secs num_secs
-        let min_secs="${CVMFS_CHECK_ALL_MIN_DAYS:-30}*60*60*24" 1
-        let num_secs="$(date +%s)-$check_time" 1
-        if [ "$num_secs" -lt "$min_secs" ]; then
-          # less than $CVMFS_CHECK_ALL_MIN_DAYS has elapsed since last check
-          continue
-        fi
-      fi
-
-      echo "${check_time}:${name}"
-
-    done)|sort -n|cut -d: -f2)"
+  repos="$(__get_checks_repo_times|sort -n|cut -d: -f2)"
 
   for repo in $repos; do
     (
--- a/cvmfs/server/cvmfs_server_check.sh
+++ b/cvmfs/server/cvmfs_server_check.sh
@@ -90,8 +90,6 @@ __do_check() {
                      $with_reflog                      \
                      -z /etc/cvmfs/repositories.d/${name}/trusted_certs"
   $user_shell "$check_cmd"
-
-  update_repo_status $name last_check "`date --utc`"
 }
 
 # Checks for mismatch between the reflog and the checksum and tries to fix them,
@@ -200,7 +198,7 @@ __do_all_checks() {
       fi
 
       local upstream=$CVMFS_UPSTREAM_STORAGE
-      if ! is_local_upstream $upstream; then
+      if [ x$(get_upstream_type $upstream_storage) = "xgw" ]; then
         continue
       fi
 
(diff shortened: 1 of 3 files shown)


DrDaveD:
Ok, I added a test, please verify & review.

jblomer:
@cernvm-bot cloudtest

cvmfs/server/cvmfs_server_check.sh
```diff
@@ -197,3 +166,142 @@ __check_repair_reflog() {
...
+
+  else
+    __do_check "$@"
+    retcode=$?
+  fi
+
+  release_lock $check_lock
+
+  return $retcode
+
```
jblomer: Let's remove this empty line

cvmfs/server/cvmfs_server_check.sh
```diff
@@ -197,3 +166,142 @@ __check_repair_reflog() {
...
+    esac
+  done
+  shift $(($OPTIND-1))
+
+  if [ $do_all -eq 1 ]; then
+    [ $# -eq 0 ] || die "no non-option parameters expected with -a"
+
+    __do_all_checks
+
+    # Always return success because this is used from cron and we
```
jblomer, line 301: Perhaps this should be an optional behavior. I think that a failed check indicates a severe issue and should raise an alarm, so cron sending an email does not sound right to me.

I think we can safely ignore skipped checks due to an active lock.

cvmfs/server/cvmfs_server_check.sh
```diff
@@ -197,3 +166,142 @@ __check_repair_reflog() {
...
+    check_multiple_repository_existence "$names"
+
+    for name in $names; do 
+      # note that is_inactive_replica also does load_repo_config
+      if is_inactive_replica $name; then
+        continue
+      fi
+
+      local upstream=$CVMFS_UPSTREAM_STORAGE
+      if ! is_local_upstream $upstream; then
```
jblomer: I think we should skip only gateway repos (which would be checked rather on the gateway) but include S3 repos.

cvmfs/server/cvmfs_server_check.sh
```diff
@@ -10,46 +10,13 @@
...
-  local tag=
-  local repair_reflog=0
-
-  # optional parameter handling
-  OPTIND=1
-  while getopts "cit:s:r" option
-  do
-    case $option in
-      c)
-        check_chunks=0
```
jblomer, line 30: Is this parameter and the following ones (`check_integrity`, `repair_reflog`) passed from the `cvmfs_server_check` to `__do_check`?

jblomer (changes requested):
Many thanks!  In general looks good to me!  I have a few minor comments.

cvmfs/server/cvmfs_server_check.sh
DrDaveD: > Perhaps this should be an optional behavior. I think that a failed check indicates a severe issue and should raise an alarm, so cron sending an email does not sound right to me.

That comment was copied from cvmfs_server_gc.sh; it is the same behavior there.  It sounds like your response here misunderstands the comment, because returning an error code to cron does send an email, as the comment says, and you seem to agree with the comment that it should not send an email.  I think an alarm should be designed to either analyze the log for errors or look for an old timestamp in "last_check" of .cvmfs_status.json; I plan to update cvmfs-servermon to do the latter.  Should I also have another field "check_status" in .cvmfs_status.json set to "pass" or "fail"?  That would be a good way to raise an alarm sooner.

> I think we can safely ignore skipped checks due to an active lock.

And so this does.  It just logs a message to /var/log/messages.

cvmfs/server/cvmfs_server_check.sh
DrDaveD: Ok, will include it in the next commit.

cvmfs/server/cvmfs_server_check.sh
DrDaveD: How exactly would I do that?  Is there an example I can follow? Here I need to read the "last_check" field of .cvmfs_status.json and I was looking at __do_all_snapshots which for the [non-local case](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/server/cvmfs_server_snapshot.sh#L285) looked only for the existence of .cvmfs_last_snapshot using "cvmfs_swissknife peek" and there is no corresponding file for checks.  So I didn't know what to do here and just skipped other cases, not knowing what should go there.

cvmfs/server/cvmfs_server_check.sh
DrDaveD: Yes, "local" variables are available to functions called in bash.  I verified there was different messages with -i and -r options. I didn't see a different message with -c but I saw it getting passed to cvmfs_swissknife with bash -x.

jblomer: Great, thanks for verifying!

cvmfs/server/cvmfs_server_check.sh
jblomer: I see, I did not realize that `cvmfs_server_gc.sh` behaves in the same way.  In both cases I find it strange that if one of the operations fails (here I mean a real failure, not purposefully skipped) the command would return a zero return code. Cron may send an email in this case but shouldn't this be fixed or configured on the cron side? If, for example, cvmfs finds a catalog inconsistency in one of the repositories, I'd certainly expect the command to return a non-zero return value.

cvmfs/server/cvmfs_server_check.sh
jblomer: I think there is an issue in the `__do_all_snapshots` code, too. Both local and S3 repositories maintain the `.cvmfs_status.json` file.  So it should work with checking like

    if [ `x$(get_upstream_type $upstream_storage) != "xgw" ] ; then

DrDaveD: Oh that's too easy then.  I'll change that.

cvmfs/server/cvmfs_server_check.sh
DrDaveD: Yeah it could require cron to always add `|| true` but chances are people would forget that.  I don't ever expect people to run this interactively.  In any case it's too late to change it for the other `-a` commands, and it would be strange to require this only for one.  I like indicating if the check failed in .cvmfs_status.json so it is easily accessible to alarms.  I'll try that.

DrDaveD: Ok, I added a "check_status" item to .cvmfs_status.json set to "failed" or "succeeded".  Please re-review.  I made it only update the status when using check -a, not every time somebody runs check by hand.

jblomer:
@cernvm-bot cloudtest

cvmfs/server/cvmfs_server_check.sh
jblomer: I agree that keeping the behavior of `cvmfs_server gc -a ` and `cvmfs_server check -a` is important.  So let's leave the return code logic as is for the 2.9 release.  For the following release, I created [a ticket](https://sft.its.cern.ch/jira/browse/CVM-2035) to change the behavior for both commands.

jblomer:
Many thanks, looks good to me; there are some failures still in the tests which would need to be addressed before we can merge.

jblomer:
@cernvm-bot cloudtest

DrDaveD:
That fixed the s3 test, the other failures seem unrelated.

jblomer:
The other failures affect all debian based platforms and all the server-side tests.  Let me check if this is related to the PR (which I wouldn't quite understand why) or if it is a problem with the current devel branch.

jblomer:
It must be something in this PR.  I just ran the comparison between `devel` branch and the patch in this PR.  This time I kept the integration node alive.  Looking into it.

jblomer:
It is related to `dash` vs. `bash`. With `dash`, the `cvmfs_server` script provokes a segfault on invocation.

DrDaveD:
I confirm that bizarre behavior happens also with dash installed from epel on centos7.  With gdb it shows the problem is in memcpy:
```
Program received signal SIGSEGV, Segmentation fault.
0x000000000040dba7 in memcpy (__len=91708, __src=0x7ffff7fb8b08,
    __dest=0x7fffff7f7550) at /usr/include/bits/string3.h:51
warning: Source file is more recent than executable.
51	  return __builtin___memcpy_chk (__dest, __src, __len, __bos0 (__dest));
(gdb) bt
#0  0x000000000040dba7 in memcpy (__len=91708, __src=0x7ffff7fb8b08,
    __dest=0x7fffff7f7550) at /usr/include/bits/string3.h:51
#1  readtoken1 (firstc=<optimized out>, syntax=<optimized out>,
    eofmark=eofmark@entry=0x0, striptabs=striptabs@entry=0) at parser.c:1377
#2  0x000000000040cd0d in xxreadtoken () at parser.c:836
#3  0x000000000040ce42 in xxreadtoken () at parser.c:792
#4  readtoken () at parser.c:714
#5  0x000000000040e3e5 in peektoken () at parser.c:691
#6  list (nlflag=2, nlflag@entry=0) at parser.c:185
#7  0x000000000040e4b6 in command () at parser.c:466
#8  0x000000000040ea9f in simplecmd () at parser.c:567
#9  command () at parser.c:472
#10 0x000000000040ebb3 in pipeline () at parser.c:275
#11 0x000000000040e2da in andor () at parser.c:239
#12 list (nlflag=3, nlflag@entry=1) at parser.c:189
#13 0x000000000040ecda in parsecmd (interact=<optimized out>) at parser.c:158
#14 0x000000000040a981 in cmdloop (top=top@entry=1) at main.c:216
#15 0x000000000040217a in main (argc=2, argv=0x7fffffffdda8) at main.c:174
```

If I do `print (char *) __src` I see that it is copying the section that I added in __do_all_checks beginning with `repos="$((`.  If I put that whole section into a function, it doesn't crash.  I'll make an update with the workaround.

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

PR #2804, opened 2021-10-20
Check if log file location is writable in cvmfs_config chksetup

aandvalenzuela:


aandvalenzuela:
@cernvm-bot cloudtest

cvmfs/cvmfs_config
```diff
@@ -374,6 +374,14 @@ cvmfs_chksetup() {
     num_errors=$(($num_errors+1))
   fi
 
+  # Check that if CVMFS_USYSLOG is set, the path is writable.
+  if [ x"$CVMFS_USYSLOG" != x"" ]; then
+    if runascvmfsuser "test -w $CVMFS_USYSLOG"; then
```
jblomer: I think we need to make one more turn here: the file may not yet exist, in which case `test -w` returns false.  So I think we should instead check if the parent directory is writable, i.e. check for `test -w $(dirname $CVMFS_USYSLOG)`

jblomer (changes requested):

cvmfs/cvmfs_config
aandvalenzuela: Oh, true! I didn't think about the file not yet existing. Thank you!

aandvalenzuela:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2806, opened 2021-10-23
Improvements to error logging and reporting

jblomer:
- Adds an in-memory circular buffer of the last 10 log messages, which can be queried through the `logbuffer` extended attribute. The attribute is throttled because printing the logbuffer is potentially costly (string copies). A new `kLogSensitive` flag can be used to hide messages from the log buffer.

- Adds the `timestamp_last_ioerr` extended attribute to report when the most recent i/o failure took place.  That can be used by monitoring to ignore old i/o errors.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2811, opened 2021-11-02
Make "inspecting catalog" check message verbose

DrDaveD:
This is idea 5 from [CVM-1524](https://sft.its.cern.ch/jira/browse/CVM-1524).

### DIFF
--- a/cvmfs/logging.cc
+++ b/cvmfs/logging.cc
@@ -88,7 +88,7 @@ pthread_mutex_t customlog_locks[] = {
   PTHREAD_MUTEX_INITIALIZER,
   PTHREAD_MUTEX_INITIALIZER};
 
-LogLevels min_log_level = kLogNormal;
+LogLevels max_log_level = kLogNormal;
 static void (*alt_log_func)(const LogSource source, const int mask,
                             const char *msg) = NULL;
 
@@ -255,9 +255,9 @@ void SetLogSyslogShowPID(bool flag) {
 }
 
 /**
- * Set the minimum verbosity level.  By default kLogNormal.
+ * Set the maximum verbosity level.  By default kLogNormal.
  */
-void SetLogVerbosity(const LogLevels min_level) { min_log_level = min_level; }
+void SetLogVerbosity(const LogLevels max_level) { max_log_level = max_level; }
 
 
 /**
@@ -423,7 +423,7 @@ void LogCvmfs(const LogSource source, const int mask, const char *format, ...) {
   int log_level = mask & ((2 * kLogNone - 1) ^ (kLogLevel0 - 1));
   if (!log_level) log_level = kLogNormal;
   if (log_level == kLogNone) return;
-  if (log_level < min_log_level) return;
+  if (log_level > max_log_level) return;
 #endif
 
   // Format the message string
--- a/cvmfs/logging_internal.h
+++ b/cvmfs/logging_internal.h
@@ -61,9 +61,9 @@ enum LogFlags {
 
 enum LogLevels {
   kLogLevel0   = 0x01000,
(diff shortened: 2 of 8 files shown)


jblomer (changes requested):
Looks good to me for the `check -a` use case.  For an interactive check, the `[inspecting catalog]` messages are still useful to indicate progress.  We can add a `-l <log level>` parameter to the `cvmfs_swissknife check` invocation for the interactive case.

DrDaveD:
But the verbose level is used to print ["[path]"](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/swissknife_check.cc#L341) too.  

If I could separate those two messages to different log levels, I could add a default -l setting by setting CVMFS_LOG_LEVEL if it is not already set to something else, only when invoking __do_check when there's no -a option.  What kLog level should I use for "[inspecting]", what level for "[path]", and what are the -l numbers corresponding to them?

jblomer:
Good point! The messages for [path] would then also need to be lowered to [`kLogLevel0`](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/logging_internal.h#L63)

DrDaveD:
I can't get that to work.  If I put in just `kLogLevel0` the message never comes out with any setting of CVMFS_LOG_LEVEL 0 to 3.  If I make it look like the definition of kLogVerboseMsg, that is, `kLogStdout | kLogShowSource | kLogLevel0` then it behaves just like kLogVerboseMsg, in that it is printed with any value of CVMFS_LOG_LEVEL.  Based on the implementation of '-l' I would expect kLogLevel0 to be a lower level than verbose, not higher, but it isn't behaving at all as I expect it.

jblomer:
Sorry, my bad!  The `kLogLevel0` is actually not itself a log level, it is used only as a lower marker.  We need to introduce an additional log level.  I'd suggest to add in `logging_internal.h` an additional level, e.g. `kLogGossipy` (feel free to pick a better name), and to shift all following constants by a factor of two.

DrDaveD:
Ok, I ended up having to do quite a lot for that.  The '-l' options to swissknife were broken, trying to shift the number 1 by a mask level (a power of 2) rather than a number of bits.  I added a kLogInfoMsg alias (with mask kLogInform) between the kLogWarning (kLogNormal) and kLogVerboseMsg (kLogVerbose) levels.  I reversed the order so the log levels increase as people expect, where level 1 corresponds to Warning/Normal, level 2 is Info, and level 3 is Verbose.  I removed kLogDiscrete because it was only used during one test to turn off all log levels, and kLogLevel0 does that now.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot tidy

jblomer:
@cernvm-bot cpplint

DrDaveD:
The failures seem unrelated to me.  Maybe run them again?

jblomer:
There is just one more issue: the messages are now formatted like

    (cvmfs) [inspecting catalog ...

This is because `kLogShowSource` is used.  I guess the code should use `kLogStdout | kLogVerbose` instead of `kLogVerboseMsg`.

DrDaveD:
I don't think that extra (cvmfs) hurts but anyway I changed it to `kLogStdout | kLogInform`, as ugly as that is.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot tidy

jblomer:
@cernvm-bot unittest

jblomer (approved):

PR #2819, opened 2021-11-11
Fix segfault in re-resolving proxy IPs

jblomer:


jblomer:
@cernvm-bot cloudtest

cvmfs/download.cc
```diff
@@ -965,7 +965,12 @@ void DownloadManager::SetUrlOptions(JobInfo *info) {
     info->proxy = "DIRECT";
     curl_easy_setopt(info->curl_handle, CURLOPT_PROXY, "");
   } else {
-    ValidateProxyIpsUnlocked(proxy->url, proxy->host);
+    // Note: inside ValidateProxyIpsUnlocked() we may change the proxy data
+    // structure, so we must not pass proxy->... (== current_proxy())
+    // parameters directly
+    std::string purl = proxy->url;
+    dns::Host phost = proxy->host;
+    ValidateProxyIpsUnlocked(purl, phost);
```
DrDaveD, line 973: Wouldn't it be better to change the parameters to be pass by value instead of by reference?

jblomer: That's probably true, for the release it doesn't matter.  I expect any way that we'll need to revisit the code, there are some changes ahead in this part.

PR #2835, opened 2022-01-03
Clean up server lock file handling

mcb30:
The Publisher model is largely transliterated directly from the original shell scripts.  This leads to C++ code that is procedural in nature and difficult to follow, and prevents the use of object-oriented programming techniques such as RAII.

As a first step towards updating the Publisher to a full object-oriented model, clean up the handling of the server lock files:

- Use an object model to represent a server lock file
- Separate the logically distinct lock files and flag files
- Clean up method names to more closely approximate C++11 standard library equivalents
- Implement and use an RAII lock guard to hold a server lock file
- Reimplement server lock files as regular POSIX lock files (as per the existing TODO comment)
- Remove duplicated and redundant code

Commits have been structured for easy review.  Please retain the individual commits when merging this PR.

### DIFF
--- a/cvmfs/publish/repository.h
+++ b/cvmfs/publish/repository.h
@@ -360,8 +360,6 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
   void TransactionRetry();
   void TransactionImpl();
 
-  void SyncImpl();
-
   SettingsPublisher settings_;
   UniquePtr<perf::StatisticsTemplate> statistics_publish_;
   /**
--- a/cvmfs/publish/repository.h
+++ b/cvmfs/publish/repository.h
@@ -11,6 +11,7 @@
 #include "gateway_util.h"
 #include "history.h"  // for History::Tag
 #include "publish/settings.h"
+#include "repository_util.h"
 #include "upload_spooler_result.h"
 #include "util/pointer.h"
 #include "util/single_copy.h"
@@ -313,8 +314,8 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
   void Migrate();
 
   const SettingsPublisher &settings() const { return settings_; }
-  bool in_transaction() const { return in_transaction_; }
-  bool is_publishing() const { return is_publishing_; }
+  const ServerLockFile &in_transaction() const { return in_transaction_; }
+  const ServerLockFile &is_publishing() const { return is_publishing_; }
   Session *session() const { return session_.weak_ref(); }
   const upload::Spooler *spooler_files() const { return spooler_files_; }
   const upload::Spooler *spooler_catalogs() const { return spooler_catalogs_; }
@@ -358,7 +359,6 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
 
   void TransactionRetry();
   void TransactionImpl();
-  void CheckTransactionStatus();
 
   void SyncImpl();
 
@@ -368,8 +368,8 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
    * The log level, set to kLogNone if settings_.is_silent() == true
    */
   int llvl_;
-  bool in_transaction_;
-  bool is_publishing_;
+  ServerLockFile in_transaction_;
+  ServerLockFile is_publishing_;
   gateway::GatewayKey gw_key_;
   /**
    * Only really used gateway mode when a transaction is opened. The session
--- a/cvmfs/publish/repository.h
+++ b/cvmfs/publish/repository.h
@@ -97,7 +97,8 @@ class __attribute__((visibility("default"))) Repository : SingleCopy {
 
   static std::string GetFqrnFromUrl(const std::string &url);
 
-  explicit Repository(const SettingsRepository &settings);
+  explicit Repository(const SettingsRepository &settings,
+                      const bool exists = true);
   virtual ~Repository();
 
   void Check();
@@ -126,7 +127,6 @@ class __attribute__((visibility("default"))) Repository : SingleCopy {
   std::string meta_info() const { return meta_info_; }
 
  protected:
-  Repository();
   void DownloadRootObjects(
     const std::string &url,
     const std::string &fqrn,
@@ -277,7 +277,8 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
 
   static Publisher *Create(const SettingsPublisher &settings);
 
-  explicit Publisher(const SettingsPublisher &settings);
+  explicit Publisher(const SettingsPublisher &settings,
+                     const bool exists = true);
   virtual ~Publisher();
 
   void UpdateMetaInfo();
@@ -319,8 +320,6 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
   const upload::Spooler *spooler_catalogs() const { return spooler_catalogs_; }
 
  private:
-  Publisher();  ///< Used by Create
-
   /**
    * Used just before a spooler is required, e.g. in Create()
    */
--- a/cvmfs/publish/repository.h
+++ b/cvmfs/publish/repository.h
@@ -314,7 +314,7 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
   void Migrate();
 
   const SettingsPublisher &settings() const { return settings_; }
-  const ServerLockFile &in_transaction() const { return in_transaction_; }
+  const ServerFlagFile &in_transaction() const { return in_transaction_; }
   const ServerLockFile &is_publishing() const { return is_publishing_; }
   Session *session() const { return session_.weak_ref(); }
   const upload::Spooler *spooler_files() const { return spooler_files_; }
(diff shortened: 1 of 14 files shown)


cvmfs/publish/repository_transaction.cc
```diff
@@ -34,8 +34,6 @@ void Publisher::CheckTransactionStatus() {
     settings_.transaction().spool_area().publishing_lock();
   is_publishing_ =
     ServerLockFile::IsLocked(publishing_lock, false /* ignore_stale */);
-
-  session_ = new Session(settings_, llvl_);
```
jblomer, line 38: The `CheckTransactionStatus()` is also called from `Publisher::TransactionRetry()`, so I guess there, too, we'd need to create a new session object.

radupopescu:
Happy New Year! Thanks for the work, this is great!

I'll get back with some feedback as soon as possible.

cvmfs/publish/repository_abort.cc
```diff
@@ -26,11 +26,7 @@ void Publisher::WipeScratchArea() {
 }
 
 void Publisher::Abort() {
-  if (is_publishing_.IsLocked()) {
-    throw EPublish(
-      "Repository " + settings_.fqrn() + " is currently publishing "
-      "(aborting abort)", EPublish::kFailTransactionState);
-  }
+  ServerLockFileGuard g(is_publishing_);
```
jblomer, line 29: I'm not sure this is correct: the new code would acquire the lock instead of checking if another process holds it, isn't it?

If another process holds it, it will throw an exception as the old code. But unlike the old code, the `abort` command now briefly holds the `is_publishing` lock, which is at least surprising.

cvmfs/publish/repository_managed.cc
```diff
@@ -83,6 +83,7 @@ void Publisher::ManagedNode::ClearScratch() {
 
 
 int Publisher::ManagedNode::Check(bool is_quiet) {
+  ServerLockFileCheck publish_check(publisher_->is_publishing_);
```
jblomer, line 86: Also here, we are taking the lock for a short time during the check, aren't we?

cvmfs/publish/repository_util.cc
```diff
@@ -61,70 +61,61 @@ void CheckoutMarker::SaveAs(const std::string &path) const {
...
+  int new_fd = TryLockFile(path_);
+  if (new_fd >= 0) {
+    assert(fd_ < 0);
+    fd_ = new_fd;
     return true;
-  }
-  unlink(tmp_path.c_str());
-  if (errno == EEXIST)
+  } else if (new_fd == -1) {
+    throw EPublish("Error while attempting to acquire lock " +path_);
```
jblomer:
```suggestion
    throw EPublish("Error while attempting to acquire lock " + path_);
```

jblomer (commented):
Great improvement, thanks a lot!

cvmfs/publish/repository_transaction.cc
mcb30: I was wondering about this.  There was nothing obvious within `Publisher::TransactionRetry()` that seemed to rely on the session object being recreated.  Is this required?

cvmfs/publish/repository_abort.cc
mcb30: Yes, this is correct and deliberate.  The old code has a race condition: it will check the lock at a point in time, but there is nothing that prevents another publisher action from starting while the abort is in progress.  The new code ensures that nothing else tries to modify the repository state mid-abort.

cvmfs/publish/repository_managed.cc
mcb30: Yes.  This is necessary when using a standard POSIX lock file: the only way to tell if something else holds the lock is to attempt to acquire the lock.

In general, this is correct behaviour for locking semantics: any code that does a test for "is this lock held?" rather than "can I acquire this lock?" is immediately suspicious since it suggests a race condition.

cvmfs/publish/repository_util.cc
mcb30: Sorry; typo.  Will fix now...

mcb30:
@jblomer Thanks for the quick review!

cvmfs/publish/repository_abort.cc
jblomer: Ok, thanks for the clarification!

mcb30:
Force-pushed a one-line update after noticing a potential race condition between the `assert` in `ServerLockFile::TryLock` and the clearing of `fd_` in `ServerLockFile::Unlock`.

cvmfs/publish/repository_transaction.cc
jblomer: I think so. The `session_` variable is a unique pointer.  By recreating the object, its destructor drops the lease so that the following `TransactionImpl()` call can re-acquire it.

jblomer:
I'm still not fully back at work but I had some time to spare to take a look at this.  I'll leave the rest of the review and merging to @radupopescu.

cvmfs/publish/repository_transaction.cc
mcb30: Thanks.  I'll update the PR, most likely by ensuring that `session_->Drop()` is called explicitly so that this intention becomes evident in the code.

mcb30: It looks as though `session_->Drop()` will already be called anyway for any exception other than `kFailTransactionState`: https://github.com/cvmfs/cvmfs/blob/f78763335b3a9c9060886d89929912c4bfd55ae4/cvmfs/publish/repository_transaction.cc#L55-L60

In the case of `kFailTransactionState`, either:
* there was no pre-existing lease file, in which case `has_lease_` would be false since the exception is raised before the call to `Acquire()` in `TransactionImpl()`
* there was a pre-existing lease file, in which case `keep_alive_` would be true

In both of these cases the `Session` destructor is a no-op, and so there is no need to reinitialise `session_`.

The overall logic here is fairly convoluted and would be improved by using RAII to automatically release the transaction lock and drop the session, rather than splitting the logic between `TransactionRetry()` and `TransactionImpl()`.  However, that cleanup can be the subject of a future PR.

For now, I am going to leave the code as-is, since reinitialising the `session_` object is unnecessary, and doing so would merely add to the confusion.

cvmfs/publish/repository.h
```diff
@@ -312,15 +314,13 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
   void Migrate();
 
   const SettingsPublisher &settings() const { return settings_; }
-  bool in_transaction() const { return in_transaction_; }
-  bool is_publishing() const { return is_publishing_; }
+  const ServerFlagFile &in_transaction() const { return in_transaction_; }
```
radupopescu, line 317:
```suggestion
  const ServerFlagFile& in_transaction() const { return in_transaction_; }
```

cvmfs/publish/repository.h
```diff
@@ -312,15 +314,13 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
...
+  const ServerLockFile &is_publishing() const { return is_publishing_; }
```
radupopescu, line 318:
```suggestion
  const ServerLockFile& is_publishing() const { return is_publishing_; }
```

radupopescu:
@cernvm-bot tidy

radupopescu:
@cernvm-bot help

radupopescu:
@cernvm-bot cpplint

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot cloudtest

cvmfs/publish/repository.h
mcb30: I chose `const Thing &method()` rather than `const Thing& method()` for consistency with the surrounding lines: https://github.com/cvmfs/cvmfs/blob/78b6e377b5ec65ee1714f21fd3c68cd4abc7f7f0/cvmfs/publish/repository.h#L316 and https://github.com/cvmfs/cvmfs/blob/78b6e377b5ec65ee1714f21fd3c68cd4abc7f7f0/cvmfs/publish/repository.h#L319-L321

If you prefer, I can insert a preliminary commit that reverses the spacing on those existing lines, and then match consistency with that?

cvmfs/publish/repository.h
mcb30: (as above)

cvmfs/publish/repository.h
radupopescu: I didn't realise there were these instances in this code base, I actually though the cosmetic checker would complain without these changes.

We can go with your version, to be consistent.

mcb30:
@radupopescu Looks as though the cloudtest build might have got stuck?

mcb30:
@radupopescu Thanks for merging.  Did the cloudtests eventually pass?

PR #2836, opened 2022-01-18
Fix integrity check with external chunked files

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2837, opened 2022-01-24
Fix broken unit test T_CatalogMgrRw.SwapNestedCatalogFailSlow

mcb30:
Commit d5fbe10 ("Allow nested catalog to be swapped even if already
attached") neglected to update the unit tests to match the new
functionality, in which SwapNestedCatalog() should fail for an
already-attached nested catalog only if the catalog has been modified.

Fix the unit test to match the intended behaviour as described in the
commit.

### DIFF
--- a/test/unittests/t_catalog_mgr_rw.cc
+++ b/test/unittests/t_catalog_mgr_rw.cc
@@ -211,12 +211,13 @@ TEST_F(T_CatalogMgrRw, SwapNestedCatalogFailSlow) {
                                               sub1_hash, sub1_size),
                "not found in parent");
 
-  // Fail for nested catalog that is already attached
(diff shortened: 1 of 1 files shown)


jblomer:
Many thanks!
@radupopescu As a follow-up, we should check why this went unnoticed in the testing. I think the full unit tests should run together with the weekly full integration tests.

jblomer:
@cernvm-bot help

jblomer:
@cernvm-bot unittest full

jblomer:
@cernvm-bot tidy

jblomer:
@cernvm-bot cpplint

jblomer (approved):

PR #2838, opened 2022-01-24
Remove concept of channels

mcb30:
The concept of a numeric update channel predates the introduction of named branches.  It is obsolete and unsupported: the command-line tools provide no sensible way to handle channels, and there is no way to configure a client to track a particular channel in its mountpoint.

Simplify the code by removing the entire concept of channels.  The "channel" column is left in the database to avoid potentially destroying existing metadata, since current versions of the code do allow a channel to be specified (even though the code does not allow a channel to be used in any meaningful way).


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot tidy

jblomer:
@cernvm-bot cpplint

radupopescu (approved):

radupopescu:
Looks good to me.

This changes the internal API between the server tools and the gateway, so both components should be deployed at once.

mcb30:
> Looks good to me.
> 
> This changes the internal API between the server tools and the gateway, so both components should be deployed at once.

Is it worth keeping the API unchanged to avoid this problem? We could have the newer version always provide a channel value of zero, and ignore any received channel value.

More generally: how does the project aim to handle this class of API change?

radupopescu:
> > Looks good to me.
> > This changes the internal API between the server tools and the gateway, so both components should be deployed at once.
> 
> Is it worth keeping the API unchanged to avoid this problem? We could have the newer version always provide a channel value of zero, and ignore any received channel value.
> 
> More generally: how does the project aim to handle this class of API change?

I'm not sure it would be worth it in this case, since we are discussing an internal API change. The API between the client and the server should be handled differently, maintaining backward compatibility - in our community, the server components are upgraded at a different rate than client components.

cvmfs/history_sql.cc
```diff
@@ -245,7 +244,7 @@ bool SqlInsertTag::BindTag(const History::Tag &tag) {
     BindTextTransient(2, tag.root_hash.ToString()) &&  // temporary (ToString)
     BindInt64(3, tag.revision) &&
     BindInt64(4, tag.timestamp) &&
-    BindInt64(5, tag.channel) &&
+    BindInt64(5, 0) &&
```
jblomer, line 250:
```suggestion
    BindInt64(5, 0) /* used to be kChannelTrunk, channels are now removed */ &&
```

jblomer (commented):
Many thanks! I linked [CVM-2061](https://sft.its.cern.ch/jira/browse/CVM-2061) to this PR.

I think the patch should also involve two updates to the documentation:
  - The `tag_channel` parameter in the [gateway API](https://cvmfs.readthedocs.io/en/stable/cpt-repository-gateway.html#post-leases-token)
  - A notice for the 2.10 release notes stating that the format of the tag list changed.  That's relevant, I think, for the machine-readable tag list (`cvmfs_server tag -lx`) which may be part of scripts. I think it's right to remove the channel from that list even though it may break scripts.  I found that column quite confusing in my own script. 

Regarding protocol compatibility between remote publishers and gateway: I agree that currently it's acceptable to ask users to update the two sides in lockstep.  In the light of possible future protocol changes, I think we should have a mechanism for proper error reporting. So if the two components are not compatible, the remote publisher (and the gateway) should say so rather than fail somewhere down the line of publishing. But that's out of the scope of this PR.

For this PR, what do you think about sending `0` as a channel from the remote publisher to the gateway (like it's done with the database) and ignoring the value on the gateway side?

jblomer:
@cernvm-bot unittest full

mcb30:
> For this PR, what do you think about sending `0` as a channel from the remote publisher to the gateway (like it's done with the database) and ignoring the value on the gateway side?

Sounds good to me.  I've force-pushed an update that:
* Applies an expanded version of your suggested comment for the `BindInt64(5, 0)`
* Sends `0` as channel to the gateway, with a similar comment
* Rebases on current `devel`

Will create the corresponding docs PR once this one is merged and final.

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot tidy

radupopescu:
@cernvm-bot cpplint

radupopescu:
@cernvm-bot cloudtest

mcb30:
@radupopescu I've force-pushed to remove the "Apply changes from code review" commit, since I'd already incorporated an expanded version of that comment into the latest revision: https://github.com/mcb30/cvmfs/blob/e509cc2106d0d085a6d1889eab864cfd24345fc5/cvmfs/history_sql.cc#L247-L249

radupopescu:
@cernvm-bot cpplint

radupopescu:
@cernvm-bot tidy

radupopescu:
@cernvm-bot cloudtest

PR #2840, opened 2022-01-31
Add possibility to hide magic xattr for root node only

jblomer:
Instead of the binary `CVMFS_HIDE_MAGIC_XATTRS=[yes|no]`, the new `CVMFS_MAGIC_XATTRS_VISIBILITY=[never,always,rootonly]` option allows for more fine grained control.  The new option, if present, takes precedent.


jblomer:
@cernvm-bot cloudtest

radupopescu:
The failures in the integration tests are not related to the PR. I will merge it.

cvmfs/mountpoint.cc
```diff
@@ -1761,13 +1761,29 @@ bool MountPoint::SetupBehavior() {
   LogCvmfs(kLogCvmfs, kLogDebug, "kernel caches expire after %d seconds",
            static_cast<int>(kcache_timeout_sec_));
 
-  bool hide_magic_xattrs = false;
+  MagicXattrManager::EVisibility xattr_visibility =
+    MagicXattrManager::kVisibilityAlways;
```
DrDaveD, line 1765: Are you sure you don't want the default to be kVisibilityRootOnly?

jblomer: The default is `kVisibilityRootOnly` only if the `CVMFS_MAGIC_XATTRS_VISIBILITY` option is present.  Otherwise cvmfs sticks to the old default behavior.  However, this can still be changed of course in any of the config files.

DrDaveD: Right, but it seems to me that kVisibilityRootOnly is almost all the time going to be the preferred behavior, so there is an argument for making it be default if neither environment variable is set.

jblomer: Fair enough, see #2842 that changes the default in the devel branch (i.e., for the 2.10 release). For the patch release, I feel uncomfortable changing the client's default behavior.

PR #2842, opened 2022-02-01
Make 'rootonly' the builtin default for xattr visibility

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2859, opened 2022-03-04
Improve error reporting for watchdog failures

jblomer:
- Let the watchdog report unexpected signals
- Fix an issue that prevented the watchdog to log to syslog
- In the main process: report the reason of a broken pipe to the watchdog


radupopescu:
@cernvm-bot cloudtest

PR #2863, opened 2022-03-15
Optionally adding extra information to meta.json.

terjekv:
This patch seeks to add some useful system information to the meta.json file for logging and monitoring. This feature is optional.

  - Adds the following fields to meta.json:
    - cvmfs_version : version_string
    - os_version : $(lsb_release -d | cut -f2)
  - Can be disabled by setting PUBLISH_VERSIONS_IN_META_FILE to anything but "true".

### DIFF
--- a/cvmfs/server/cvmfs_server_coda.sh
+++ b/cvmfs/server/cvmfs_server_coda.sh
@@ -25,7 +25,7 @@ LATEST_JSON_INFO_SCHEMA=1
 
 # Should we publish CVMFS and OS versions in meta.json?
 # Set to any non-true string to turn off this feature.
-PUBLISH_VERSIONS_IN_META_FILE=true
+CVMFS_PUBLISH_VERSIONS_IN_META_FILE=true
 
 if [ -f /etc/cvmfs/server.local ]; then
   if [ -r /etc/cvmfs/server.local ]; then
--- a/cvmfs/server/cvmfs_server_coda.sh
+++ b/cvmfs/server/cvmfs_server_coda.sh
@@ -23,6 +23,10 @@ DEFAULT_LOCAL_STORAGE="/srv/cvmfs"
 
 LATEST_JSON_INFO_SCHEMA=1
 
+# Should we publish CVMFS and OS versions in meta.json?
+# Set to any non-true string to turn off this feature.
+PUBLISH_VERSIONS_IN_META_FILE=true
+
 if [ -f /etc/cvmfs/server.local ]; then
   if [ -r /etc/cvmfs/server.local ]; then
     . /etc/cvmfs/server.local
--- a/cvmfs/server/cvmfs_server_json.sh
+++ b/cvmfs/server/cvmfs_server_json.sh
@@ -91,7 +91,7 @@ _render_info_file() {
     echo '  "last_geodb_update" : "'$modtime'",'
   fi
 
-  if [ "x${PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
+  if [ "x${CVMFS_PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
     echo '  "cvmfs_version" : "$(cvmfs_version_string)",'
     echo '  "os_version" : "' $(lsb_release -d | cut -f2) '", '
   fi
--- a/cvmfs/server/cvmfs_server_json.sh
+++ b/cvmfs/server/cvmfs_server_json.sh
@@ -91,6 +91,11 @@ _render_info_file() {
     echo '  "last_geodb_update" : "'$modtime'",'
   fi
 
+  if [ "x${PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
+    echo '  "cvmfs_version" : "$(cvmfs_version_string)",'
+    echo '  "os_version" : "' $(lsb_release -d | cut -f2) '", '
+  fi
+
   echo '  "repositories" : ['
 
   _render_repos $(_available_repos "stratum0")
@@ -101,6 +106,7 @@ _render_info_file() {
   _render_repos $(_available_repos "stratum1")
 
   echo '  ]'
+
   echo '}'
 }
 
--- a/cvmfs/server/cvmfs_server_json.sh
+++ b/cvmfs/server/cvmfs_server_json.sh
@@ -92,7 +92,7 @@ _render_info_file() {
   fi
 
   if [ "x${CVMFS_PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
-    echo '  "cvmfs_version" : "$(cvmfs_version_string)",'
+    echo '  "cvmfs_version" : "'$(cvmfs_version_string)'",'
     echo '  "os_id" : "'$(_os_id)'", '
     echo '  "os_version_id" : "'$(_os_version_id)'", '
     echo '  "os_pretty_name" : "'$(_os_pretty_name)'", '
--- a/cvmfs/server/cvmfs_server_json.sh
+++ b/cvmfs/server/cvmfs_server_json.sh
@@ -106,7 +106,6 @@ _render_info_file() {
   _render_repos $(_available_repos "stratum1")
 
   echo '  ]'
-
   echo '}'
 }
 
--- a/cvmfs/server/cvmfs_server_json.sh
+++ b/cvmfs/server/cvmfs_server_json.sh
@@ -93,7 +93,9 @@ _render_info_file() {
 
   if [ "x${CVMFS_PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
     echo '  "cvmfs_version" : "$(cvmfs_version_string)",'
-    echo '  "os_version" : "' $(lsb_release -d | cut -f2) '", '
+    echo '  "os_id" : "'$(_os_id)'", '
+    echo '  "os_version_id" : "'$(_os_version_id)'", '
+    echo '  "os_pretty_name" : "'$(_os_pretty_name)'", '
   fi
 
   echo '  "repositories" : ['
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -756,7 +756,7 @@ _os_set_etc_release_filename() {
 _os_etc_release_get_field() {
 	_os_set_etc_release_filename
 
-	fieldname=$1
+	local fieldname=$1
 	if [ "x${fieldname}" == "x" ]; then
 		die "Internal error: _os_etc_release_get_field expects a field name to search for!"
 	fi
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -733,6 +733,54 @@ has_selinux() {
   $GETENFORCE_BIN | grep -qi "enforc" || return 1
 }
 
+# Functions to retrieve information about the operating system.
+# Used in cvmfs_server_json.sh to populate meta.json.
+# Fails gracefully if no os-release file is found (values become empty strings)
+_os_set_etc_release_filename() {
+	_CVMFS_OS_RELEASE_FILENAME=''
+	if test -e "/etc/os-release"; then
+		_CVMFS_OS_RELEASE_FILENAME="/etc/os-release";
+	elif test -e "/usr/lib/os-release"; then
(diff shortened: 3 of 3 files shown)


terjekv:
As stated on Mattermost, I find it quite tricky to test this in a nice fashion. Suggestions on how to test this locally would be greatly appreciated. :)

radupopescu:
@cernvm-bot tidy

radupopescu:
@cernvm-bot cpplint

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot cloudtest

cvmfs/server/cvmfs_server_coda.sh
```diff
@@ -23,6 +23,10 @@ DEFAULT_LOCAL_STORAGE="/srv/cvmfs"
 
 LATEST_JSON_INFO_SCHEMA=1
 
+# Should we publish CVMFS and OS versions in meta.json?
+# Set to any non-true string to turn off this feature.
+PUBLISH_VERSIONS_IN_META_FILE=true
```
radupopescu: As this variable is supposed to be overridable in the repository configuration files, we should be consistent and rename it to start with `CVMFS_`.

radupopescu:
```suggestion
CVMFS_PUBLISH_VERSIONS_IN_META_FILE=true
```

cvmfs/server/cvmfs_server_json.sh
```diff
@@ -91,6 +91,11 @@ _render_info_file() {
     echo '  "last_geodb_update" : "'$modtime'",'
   fi
 
+  if [ "x${PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
```
radupopescu:
```suggestion
  if [ "x${CVMFS_PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
```

cvmfs/server/cvmfs_server_coda.sh
terjekv: Yup, good point!

terjekv:
Huh, https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1354/CVMFS_PLATFORM=el7,CVMFS_PLATFORM_CONFIG=x86_64,label=trampoline/testReport/(root)/ServerIntegrationTests/cvmfs_server_check__a/ is weird. 

This doesn't help me much. :( 

`````
Starting test.cern.ch.d5a293.replica-4 at Tue Mar 15 15:32:05 CET 2022
Verifying integrity of test.cern.ch.d5a293.replica-4...
failed to fetch repository manifest (8 - bad whitelist)
failed to load repository manifest
ERROR from cvmfs_server check!
Finished test.cern.ch.d5a293.replica-4 at Tue Mar 15 15:32:05 CET 2022
*** checking if check worked on test repo and replicas 1 and 3
  "check_status": "succeeded"
  "check_status": "succeeded"
  "check_status": "succeeded"
*** verifying check did not run on replica 2
*** verifying check failed on replica 4
  "check_status": "failed"
*** verify check was delayed on replica 1
`````

File: cvmfs/server/cvmfs_server_json.sh [hunk 3]
```diff
@@ -91,6 +91,11 @@ _render_info_file() {
     echo '  "last_geodb_update" : "'$modtime'",'
   fi
 
+  if [ "x${CVMFS_PUBLISH_VERSIONS_IN_META_FILE}" == "xtrue" ]; then
+    echo '  "cvmfs_version" : "$(cvmfs_version_string)",'
+    echo '  "os_version" : "' $(lsb_release -d | cut -f2) '", '
```
jblomer: For this we'd need to add the `lsb-release` dependency to the cvmfs-server package.

jblomer (changes requested):
Many thanks for the contribution!

cvmfs/server/cvmfs_server_json.sh
DrDaveD: And watch out, "there are no plans to ship lsb-release in RHEL9".  Should use /etc/os-release instead.

terjekv: At some point I will understand what the community decides on as a dependable, long term, consistent way of figuring out the running version of linux... I'm not sure we can depend on `/etc/os-release` as it is theoretically tied to systemd, and as much as Debians and RHEL clones have it, god knows for how long? I am happy to use it though if the CVMFS community prefers it they way, but it should probably be moved to a function somewhere ala `_get_os_description` or similar.

Other options is to drop the data, or doing `uname -r`, but `uname` leaks data we probably don't want to announce and doesn't really give me what I want anyway. Hm. I'm very open to ideas here.

terjekv: With regards to `/etc/os-release`, the default way to parse the file is to source it, ala:
````
test -e /etc/os-release && os_release='/etc/os-release' || os_release='/usr/lib/os-release'
. "${os_release}"
````
However, this pollutes the environment of the running script with environment variables like `NAME`, `ID`, `HOME_URL`, and more. The environment variables used in `cvmfs_server` are all sanely prefixed and should avoid being clobbered, but I'm still a bit nervous about doing a raw source late in the scripts run time. 

My thought would be to either parse the file, or if sourcing it, doing it very early in https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/server/cvmfs_server_coda.sh to ensure any clobbering is done by CMVFS code and not from `/etc/os-release`.

Comment (terjekv) [reply cvmfs/server/cvmfs_server_json.sh]: Implemented one approach in https://github.com/cvmfs/cvmfs/pull/2863/commits/03bbc163cf6e41cb8ecec8574dd4fb0263f244fd

File: cvmfs/server/cvmfs_server_util.sh [hunk 4]
```diff
@@ -733,6 +733,54 @@ has_selinux() {
...
+	fieldname=$1
+	if [ "x${fieldname}" == "x" ]; then
+		die "Internal error: _os_etc_release_get_field expects a field name to search for!"
+	fi
+
+	if [ "x${_CVMFS_OS_RELEASE_FILENAME}" == "x" ]; then
+		# If we are unable to find a proper /etc/os-release file return nothing.
+		:
+	else
+		grep "^${fieldname}=" ${_CVMFS_OS_RELEASE_FILENAME} | cut -f2 -d'"'
```
amadio: Suggestion: `eval $(sed -n "/^${fieldname}=/p" ${_CVMFS_OS_RELEASE_FILENAME})`

terjekv: Huh. Interesting approach. Saves a pipe but introduces an eval. Is there a specific reason for wanting this approach? I don't mind either way, I'm mostly asking to learn something. :)

cvmfs/server/cvmfs_server_util.sh
```diff
@@ -733,6 +733,54 @@ has_selinux() {
...
+#    but the question would be where to call this once and for all. In theory one could
+#    call it right after the declaration above?
+# 2. If a fieldname isn't provided, die due to wrongful usage of the (internal) function.
+# 3. If we couldn't find an os-release file, return nothing.
+# 4. If we grep fails to find anything, return nothing. We may want to check against known
+#    keys from https://www.freedesktop.org/software/systemd/man/os-release.html?
+_os_etc_release_get_field() {
+	_os_set_etc_release_filename
+
+	fieldname=$1
```
DrDaveD:
```suggestion
	local fieldname=$1
```

cvmfs/server/cvmfs_server_util.sh
amadio: Actually, nevermind, the eval will add the variable to the environment, and you said this is something you don't want, right? Otherwise you could just source the os-release file... I think the way you did is better, but a single sed can probably replace the grep + cut indeed.

DrDaveD: The grep..cut isn't going to work if the double-quotes are missing.  I see that is the case on debian for ID, where it is `ID=debian`.

On the other hand, the eval does not print to stdout as expected.  This one works for me:
```
eval echo $(sed -n "s/^${fieldname}=//p" /etc/os-release)
```

amadio: You don't need the eval in that case, just do `sed -n "s/^${fieldname}=//p" /etc/os-release` (you may get an extra newline like that, though).

DrDaveD: I tried that, but it doesn't remove the double quotes ... but you could do that with
```
sed -n "s/\"//g;s/^${fieldname}=//p" /etc/os-release
```

DrDaveD: Oh I thought that looked familiar.  I wrote that same line [before](https://github.com/cvmfs-contrib/config-repo/blob/osg/libexec/osgstorage-auth.conf#L20).

terjekv: Aha. :) https://github.com/cvmfs/cvmfs/pull/2863/commits/703dad1b9603d5fca9ee3d6359e3f5496f79c0ca now uses the line from @DrDaveD. Do note that supposedly the os-release-file may be located under `/usr/lib`.

cvmfs/server/cvmfs_server_util.sh
```diff
@@ -733,6 +733,54 @@ has_selinux() {
...
+	local fieldname=$1
+	if [ "x${fieldname}" == "x" ]; then
+		die "Internal error: _os_etc_release_get_field expects a field name to search for!"
+	fi
+
+	if [ "x${_CVMFS_OS_RELEASE_FILENAME}" == "x" ]; then
+		# If we are unable to find a proper /etc/os-release file return nothing.
+		:
+	else
+    sed -n "s/\"//g;s/^${fieldname}=//p" ${_CVMFS_OS_RELEASE_FILENAME}
```
jblomer, line 768: There seems to be a whitespacing problem here. Perhaps a mix of tabs and spaces? It should be spaces only.

jblomer (changes requested):
One tiny issue, otherwise it looks fine to me.  Many thanks for the PR!

cvmfs/server/cvmfs_server_util.sh
terjekv: Whops. Sorry. Fixed in https://github.com/cvmfs/cvmfs/pull/2863/commits/e5150a49ba45a423259a3397740734f0bf8bef03

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

terjekv:
> https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1358/

This again does stuff like this:

````
Starting test.cern.ch.21518a.replica-4 at Thu Mar 17 10:08:20 CET 2022
Verifying integrity of test.cern.ch.21518a.replica-4...
failed to fetch repository manifest (8 - bad whitelist)
failed to load repository manifest
ERROR from cvmfs_server check!
Finished test.cern.ch.21518a.replica-4 at Thu Mar 17 10:08:20 CET 2022
*** checking if check worked on test repo and replicas 1 and 3
  "check_status": "succeeded"
  "check_status": "succeeded"
  "check_status": "succeeded"
*** verifying check did not run on replica 2
*** verifying check failed on replica 4
  "check_status": "failed"
*** verify check was delayed on replica 1
````

Is this expected or something I've messed up?

jblomer:
> Is this expected or something I've messed up?

That may be a problem due to the patch. Possibly just an issue with the test itself. We'd need to look into it, it may take a few days until I get to it.

terjekv:
Ah, thanks! I'll wait to see how things go. :)

terjekv:
@jblomer Did anyone find out what caused the tests to fail for this PR?

jblomer:
Apologies for the delay! We need to fix the test, the problem shows with and without your patch.

I think it's good to go, we'll merge it tomorrow.

terjekv:
No worries, thank you for following up on this issue!

terjekv:
I noticed this didn't make it into 2.9.4. Are we looking at 2.10 for extra features such as this?

jblomer:
That's right, by default we merge into the devel branch, which becomes the next feature release (in this case: 2.10). Fixes for the patch releases need to be explicitly backported.  For 2.10, we plan to cut a release for the CernVM workshop in September.  We may be looking at a prelease, so ETA for 2.10 is 1-2 months.

terjekv:
Cool, thanks!

PR #2864, opened 2022-03-16
Drop chkconfig requirement for EL9

traylenator:
Drop chkconfig requirement for EL9 and newer.

Fairly sure it could be dropped for everything but for the future is
the main thing.

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -118,7 +118,9 @@ Requires: sysvinit
 Requires: sysvinit-tools
   %endif
 %else
+%if 0%{?el7}%{?el8}%{?suse_version}
 Requires: chkconfig
+%endif
 Requires: fuse-libs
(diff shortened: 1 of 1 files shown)


traylenator:
Curios , do you actually still build this for EL4, 5, ....  ? Looking at some of the %posts I think not.
I'd take an axe to this spec file and get rid of all the obsolete cases.

radupopescu:
Thanks for looking into this. We no longer build for anything older than RH7.

radupopescu:
@cernvm-bot clang-tidy

radupopescu:
@cernvm-bot cpplint

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot tidy

radupopescu:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot cloudtest

PR #2871, opened 2022-03-22
Fix build in Docker on macOS w/ virtio-fs

radupopescu:
Removes invalid symlinks from the source tarball which cause the RPM build process to fail inside Docker on macOS, when virtio-fs shared volumes are enabled.


radupopescu:
@cernvm-bot cloudtest

PR #2891, opened 2022-04-12
Track page cache state

### LINKED ISSUE #2879: Performance degradation at high core counts
As discussed with @elmsheus, ATLAS build machines using sources/dependencies from cvmfs see a performance degradation from around 32 cores to 64 cores.

@elmsheus @krasznaa If you could give us instructions on how to build Athena on a vanilla CentOS 7 node, we can try to reproduce the issue on our 64 core machine.

jblomer:
Adds a new page cache tracker to the data structures that track the state of the kernel caches. The page cache tracker is only used if the mount point is not in NFS mode and if it is not hotpatched from a previous version without one.

The page cache tracker remembers the content hash of the data that are in the page cache for a given inode. This allows us to make an informed decision on whether or not to flush the page cache and when to use direct I/O (if the inode has open files with a different content hash). The page cache tracker evicts entries if the corresponding inode receives a fuse FORGET message.  Still to be verified if the FORGET message is coupled to the page cache (and not just to the dentry cache).

- [x] Add counters for total number of files opened with cache (1), without cache (2), with direct i/o (3)
- [x] Add integration test
- [x] Confirm that removing from the tracker on FORGET is [correct](https://sourceforge.net/p/fuse/mailman/message/37642687/)
- [x] Respect "direct i/o" flag in the file catalogs
- [x] Possibly: take only single lock on inode tracker and page cache tracker in FORGET_MULTI

Fixes #2879

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -107,6 +107,7 @@
 #include "statistics.h"
 #include "talk.h"
 #include "tracer.h"
+#include "util/algorithm.h"
 #include "util/exception.h"
 #include "util_concurrency.h"
 #include "uuid.h"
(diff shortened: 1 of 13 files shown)


jblomer:
@cernvm-bot cloudtest

radupopescu:
Mac OS builds are broken right now, I'm aware and looking into it. Unfortunately, this also prevents the cloud tests from being run.

PR #2904, opened 2022-04-25
commitprocessor performance: truncate instead of re-initializing PathString in DiffRec

mharvey-jt:
The DiffRec function is the main function to compare the new and old catalogs. Inside this function, it compares the old and new entry and create PathStrings in each iteration of the loop. This is quite inefficient because each time it does a memcpy (ShortString::Assign) to initialize old_path and new_path from path. This gets worse as you traverse down the directory tree because the path gets longer.

This PR adds a Truncate function to ShortString which allows us to simply reduce the length of old_path and new_path and append the entry name, instead of re-initializing it from path.

### DIFF
--- a/cvmfs/catalog_diff_tool_impl.h
+++ b/cvmfs/catalog_diff_tool_impl.h
@@ -106,6 +106,15 @@ void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
   sort(old_listing.begin(), old_listing.end(), IsSmaller);
   AppendLastEntry(&old_listing);
 
+  // create these paths here so it can be "reused" in the loop without
+  // re-initializing every time, this should save time in doing memcpy
+  // especially when the path gets longer
+  PathString old_path(path);
+  old_path.Append("/", 1);
+  PathString new_path(path);
+  new_path.Append("/", 1);
+  unsigned char length_after_truncate = old_path.GetLength();
+
   catalog::DirectoryEntryList new_listing;
   AppendFirstEntry(&new_listing);
   new_catalog_mgr_->Listing(path, &new_listing);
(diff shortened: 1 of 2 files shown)


radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot cpplint

radupopescu:
@cernvm-bot tidy

radupopescu:
Thanks for the patch!

cvmfs/shortstring.h
```diff
@@ -87,6 +87,15 @@ class ShortString {
     this->length_ = new_length;
   }
 
+  void Truncate(unsigned char new_length) {
```
jblomer, line 90:
```suggestion
  void Truncate(unsigned new_length) {
```

It seems unnecessary to restrict `new_length` to 255 characters.

cvmfs/shortstring.h
```diff
@@ -87,6 +87,15 @@ class ShortString {
...
+    assert(new_length <= this->GetLength());
+    if (long_string_) {
```
jblomer, line 92: We should add a method documentation that the use of `Truncate()` can lead to a heap allocated string that is shorter than the reserved stack space.

cvmfs/catalog_diff_tool_impl.h
```diff
@@ -106,6 +106,15 @@ void CatalogDiffTool<RoCatalogMgr>::DiffRec(const PathString& path) {
...
   AppendLastEntry(&old_listing);
 
+  // create these paths here so it can be "reused" in the loop without
+  // re-initializing every time, this should save time in doing memcpy
+  // especially when the path gets longer
+  PathString old_path(path);
+  old_path.Append("/", 1);
+  PathString new_path(path);
+  new_path.Append("/", 1);
+  unsigned char length_after_truncate = old_path.GetLength();
```
jblomer, line 116:
```suggestion
  unsigned length_after_truncate = old_path.GetLength();
```

jblomer (changes requested):
Essentially good to go, just some minor issues.

HereThereBeDragons:
#3116 has the requested changes

PR #2905, opened 2022-04-25
commitprocessor performance: Increase kDefaultMaxName,kDefaultMaxLink

mharvey-jt:
In our environment, filesnames are consistently longer than 25 characters. Perf shows a fair bit of time spent in malloc because the ShortString on the stack is not big enough.  This PR increases the default lengths.

### DIFF
--- a/cvmfs/shortstring.h
+++ b/cvmfs/shortstring.h
@@ -19,8 +19,8 @@
 namespace CVMFS_NAMESPACE_GUARD {
 #endif
 
-const unsigned char kDefaultMaxName = 25;
-const unsigned char kDefaultMaxLink = 25;
+const unsigned char kDefaultMaxName = 80;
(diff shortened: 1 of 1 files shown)


radupopescu:
Thanks for the patch!

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot tidy

radupopescu:
@cernvm-bot cpplint

jblomer (changes requested):
Perhaps in a first step we can change that for the server side only.  On the client side, `ShortString` objects are part of the data structures that are preserved during hotpatch.  So if we change it, we need to migrate the data structures, too.

Ideally, I think, we make the stack size configurable but only on mount, i.e. a change of the configuration would not apply during hot patch.

PR #2908, opened 2022-04-25
commitprocessor performance: Avoid StatisticsDatabase::OpenStandardDB 

mharvey-jt:
Only call OpenStanardDB() if `params.upload_stats_db==true`

### DIFF
--- a/cvmfs/receiver/commit_processor.cc
+++ b/cvmfs/receiver/commit_processor.cc
@@ -287,13 +287,13 @@ CommitProcessor::Result CommitProcessor::Process(
     return kError;
   }
 
-  StatisticsDatabase *stats_db = StatisticsDatabase::OpenStandardDB(repo_name);
-  if (stats_db != NULL) {
-    if (!stats_db->StorePublishStatistics(statistics_, start_time_, true)) {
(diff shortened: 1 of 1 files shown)


jblomer (changes requested):
At the moment, keeping statistics (locally) and pushing them to the storage backend are two separate operations.  I think we should keep the logic consistent between gateway and simple publisher. We could

1. only keep the statistics if `CVMFS_UPLOAD_STATS_DB=true` consistently, or
2. introduce another parameter, `CVMFS_KEEP_STATS_DB` or similar (makes the whole statistics handling more complicated probably without much benefit.)

We could also combine 1) and 2) and deprecate `CVMFS_UPLOAD_STATS_DB` and replace it by `CVMFS_KEEP_STATS_DB`, which would turn on and off both keeping the statistics and uploading them. I have a slight preference for this last option but I'd be interested if there are other suggestions.

mharvey-jt:
Hi @jblomer I think your final suggestion of deprecating CVMFS_UPLOAD_STATS_DB and replacing it by CVMFS_KEEP_STATS_DB makes sense. Not sure there's value in keeping creation and uploading as separately-configurable activities.

PR #2915, opened 2022-04-25
Make open fd checks on publisher mount optional

mharvey-jt:
In some scenarios, it can be preferable to have the publisher continue even if there are open fds on the publisher mount . This PR adds  the configuration option `CVMFS_SKIP_FD_CHECK`. When set to 'true`, all open fd checks are elided.


cvmfs/server/cvmfs_server_ingest.sh
```diff
@@ -356,7 +356,7 @@ cvmfs_server_ingest() {
   fi
 
   # check again for open file descriptors (potential race condition)
-  if has_file_descriptors_on_mount_point $name && \
+  if  [ "$CVMFS_SKIP_FD_CHECK" != "true" ] &&  has_file_descriptors_on_mount_point $name && \
```
HereThereBeDragons, line 359:
```suggestion
  if [ "$CVMFS_SKIP_FD_CHECK" != "true" ] &&  has_file_descriptors_on_mount_point $name && \
```

and i would add above some output to say that the fd check is skipped:
```
# check again for open file descriptors (potential race condition)
 if [ "$CVMFS_SKIP_FD_CHECK" = "true" ]; then
    echo "Forced skipping check for file descriptors (potential race condition)"
  fi
  ```

File: cvmfs/server/cvmfs_server_publish.sh [hunk 2]
```diff
@@ -451,7 +455,7 @@ cvmfs_server_publish() {
     fi
 
     # check again for open file descriptors (potential race condition)
-    if has_file_descriptors_on_mount_point $name && \
+    if [ "$CVMFS_SKIP_FD_CHECK" != "true" ] && has_file_descriptors_on_mount_point $name && \
```
HereThereBeDragons, line 458: also add here above the notice about forced skipping of fds

```
# check again for open file descriptors (potential race condition)
 if [ "$CVMFS_SKIP_FD_CHECK" = "true" ]; then
    echo "Forced skipping check for file descriptors (potential race condition)"
  fi
```

cvmfs/server/cvmfs_server_publish.sh
```diff
@@ -350,7 +352,9 @@ cvmfs_server_publish() {
 
     # check if we have open file descriptors on /cvmfs/<name>
     local use_fd_fallback=0
-    handle_read_only_file_descriptors_on_mount_point $name $open_fd_dialog || use_fd_fallback=1
+    if [ "$CVMFS_SKIP_FD_CHECK" != "true" ]; then
```
HereThereBeDragons, line 355: @jblomer can you please verify that this fd check can be theoretically skipped?

HereThereBeDragons (changes requested):

Issue #2901, opened 2022-04-24
Make `x-amz-acl` header value configurable  (publisher)


mharvey-jt:
Presently the value of the `x-amz-acl` header is hardcoded in `cvmfs/s3fanout.cc` to `x-amz-acl:public-read`. 
It can useful to set this ACL to an alternate value, applicable globally to the repo

Desired behaviour:

* Default behavior remains to set `x-amz-acl: public-read`
* Allow `x-amz-acl` to be globllay set to one  of the values permitted by the S3 spec (https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl) , by means of an configuration environment variable `CVMFS_S3_X_AMZ_ACL`

mharvey-jt:
PR https://github.com/cvmfs/cvmfs/pull/2916/files

HereThereBeDragons:
#3102 got merged

PR #2916, opened 2022-04-25
Make x-amz-acl header value configurable (publisher)

mharvey-jt:
Ref https://github.com/cvmfs/cvmfs/issues/2901


HereThereBeDragons:
Close due to old version of #3102?

PR #2919, opened 2022-04-25
unconditionally compress catalogs irrespective of `CVMFS_COMPRESSION_ALGORITHM` setting

mharvey-jt:
If the publisher has `CVMFS_COMPRESSION_ALGORITHM=none`  then it will upload catalog objects in uncompressed form. But the cvmfs client always expects these to be compressed, irrespective of any `CVMFS_COMPRESSION_ALGORITHM` setting it may have. 
The PR causes the publisher to unconditionally compress catalog objects irrespective of `CVMFS_COMPRESSION_ALGORITHM`

Context: We are grafting external data which is held in uncompressed form. We have to set `CVMFS_COMPRESSION_ALGORITHM=none` so that when grafting these objects they are correctly marked as uncompressed in the catalog. This has the side-effect of then also not compressing the catalog objects.

### DIFF
--- a/cvmfs/ingestion/pipeline.cc
+++ b/cvmfs/ingestion/pipeline.cc
@@ -121,12 +121,19 @@ void IngestionPipeline::Process(
   bool allow_chunking,
   shash::Suffix hash_suffix)
 {
+  // always enable compression for anything that has a suffix
+  // (i.e. catalogs, certificates, histories)
+  zlib::Algorithms compression_algorithm = compression_algorithm_;
+  if (hash_suffix != shash::kSuffixNone) {
+      compression_algorithm = zlib::kZlibDefault;
+  }
+
   FileItem *file_item = new FileItem(
     source,
     minimal_chunk_size_,
     average_chunk_size_,
     maximal_chunk_size_,
-    compression_algorithm_,
+    compression_algorithm,
     hash_algorithm_,
     hash_suffix,
     allow_chunking && chunking_enabled_,


HereThereBeDragons:
it is true that catalogs are expected to be compressed with `kZlibDefault` as shown here in the fetch command of the client:
https://github.com/cvmfs/cvmfs/blob/b79d197b2fbeda5e60e67451d0cf7fdeba3fdcf7/cvmfs/catalog_mgr_client.cc#L258-L259

the fix you have is on the creation part of a catalog on the publisher. normally the `spooler` has the correct compression algorithm and then creates a new item based on this. while there was a patch https://github.com/cvmfs/cvmfs/pull/1584/commits/d8ea3088caea7622f198a65d1c251c1bcbf949e9 for uncompressed repos to always compress catalogs, it seems to be incomplete.

rather than putting it in https://github.com/cvmfs/cvmfs/blob/b79d197b2fbeda5e60e67451d0cf7fdeba3fdcf7/cvmfs/ingestion/pipeline.cc#L119-L124

i would put it in https://github.com/cvmfs/cvmfs/blob/b79d197b2fbeda5e60e67451d0cf7fdeba3fdcf7/cvmfs/upload.cc#L69-L72

with adding a new `Process()` function where you set the compression algorithm a la
```
ingestion_pipeline_->Process(new FileIngestionSource(local_path), false,
                               shash::kSuffixCatalog, zlib::kZlibDefault);
```


files that might have the wrong spoolers that could create this problem are: `swissknife_migrate.cc` and `swissknife_history.cc`


what do you think @jblomer?

HereThereBeDragons:
Superseeded by #3338

PR #2923, opened 2022-04-25
Restrict access to xattrs containing secrets; configure acl with CVMFS_PRIVILEGED_XATTR_GID

mharvey-jt:
Some extended attributes may contain secrets (eg URLs containing basic auth credentials). 
This PR:
* restricts Proxy and Host extended attributes to root and other authorised groups
* Adds the configuration option CVMFS_PRIVILEGED_XATTR_GID for specifying a comma-separated list of groupnames to be allowed access to sensitive xattrs
* Extends  Xattr class's PrepareValueFenced() to allow gating of access based on uid, gid, and pid.


HereThereBeDragons:
Replaced by PR #3104

PR #2924, opened 2022-04-25
Add user.external_url extended attribute

mharvey-jt:
Add a new extended attribute that gives the URL of an external data object.
This is useful if there is also a direct object store access route available.

### DIFF
--- a/cvmfs/magic_xattr.cc
+++ b/cvmfs/magic_xattr.cc
@@ -64,6 +64,7 @@ MagicXattrManager::MagicXattrManager(MountPoint *mountpoint,
   Register("xfsroot.rawlink", new RawlinkMagicXattr());
 
   Register("user.authz", new AuthzMagicXattr());
+  Register("user.external_url", new ExternalURLMagicXattr());
 }
 
 std::string MagicXattrManager::GetListString(catalog::DirectoryEntry *dirent) {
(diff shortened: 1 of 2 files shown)


cvmfs/magic_xattr.cc
```diff
@@ -541,3 +542,22 @@ std::string UsedDirPMagicXattr::GetValue() {
 std::string VersionMagicXattr::GetValue() {
   return std::string(VERSION) + "." + std::string(CVMFS_PATCH_LEVEL);
 }
+
+std::string ExternalURLMagicXattr::GetValue() {
+  std::vector<std::string> host_chain;
+  std::vector<int> rtt;
+  unsigned current_host;
+  mount_point_->external_download_mgr()->GetHostInfo(
```
jblomer, line 550: We should still catch the case when the external download manager is `NULL`

jblomer (changes requested):
Many thanks!  Besides the comment, let's also add a test for this attribute in integration test 615.  This test should at the same time become part of the quick test suite (`cvmfs_test_suites="quick"` in the preamble)

cvmfs/magic_xattr.cc
HereThereBeDragons: can you have a file marked as externalfile without an external downloadmanager? if not, then the fence should be good enough?

jblomer: Yes, the external file attribute is set in the catalog.  For it to work correctly, most likely the client needs to set an external host URL but that is nowhere enforced.

HereThereBeDragons:
@jblomer Change request and test can be found in #3101 
There are some unrelated failures for test 615

jblomer:
Replaced by #3101

Issue #2929, opened 2022-04-25
double-free from LoadCertificateMem caused by repo update race


mharvey-jt:
There is a race condition between the TryFinish() path in the fuse handlers and the Check(dryrun=true) in the CheckSynchronously() of the notification thread. Both try to reload the certificate, and can cause a double-free.

Steps to reproduce:

* set up a repo with a short catalog update polling time
* set up a transaction/publish in a continuous loop 
* mount the repo, and put some cvmfs load on it
* run `cvmfs_talk remount sync` in a continuous loop  
* wait (perhaps along time)

Observe that eventually the client will crash. The stack trace will show something like:
```
Thread 21 (Thread 0x7fff6e14d700 (LWP 48499)):
#0  0x00007ffff6ec3993 in select () from /lib64/libc.so.6
#1  0x00007ffff45ab696 in SafeSleepMs (ms=ms@entry=100)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/util/posix.cc:1730
#2  0x00007ffff4583241 in Watchdog::SendTrace (sig=6, 
    siginfo=<optimized out>, context=<optimized out>)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/monitor.cc:317
#3  <signal handler called>
#4  0x00007ffff6e04377 in raise () from /lib64/libc.so.6
#5  0x00007ffff6e05a68 in abort () from /lib64/libc.so.6
#6  0x00007ffff6e46ec7 in __libc_message () from /lib64/libc.so.6
#7  0x00007ffff6e4f6b9 in _int_free () from /lib64/libc.so.6
#8  0x00007ffff651f80d in CRYPTO_free () from /lib64/libcrypto.so.10
#9  0x00007ffff660951d in ASN1_STRING_free () from /lib64/libcrypto.so.10
#10 0x00007ffff65fd5fd in ASN1_primitive_free () from /lib64/libcrypto.so.10
#11 0x00007ffff65fd97f in ASN1_template_free () from /lib64/libcrypto.so.10
#12 0x00007ffff65fd80a in asn1_item_combine_free ()
   from /lib64/libcrypto.so.10
#13 0x00007ffff65fd8c5 in ASN1_item_free () from /lib64/libcrypto.so.10
#14 0x00007ffff459f345 in signature::SignatureManager::LoadCertificateMem (
    this=this@entry=0x7fff306f6930, 
[...]

PR #2930, opened 2022-04-25
add mutex lock to certificate operations

mharvey-jt:
https://github.com/cvmfs/cvmfs/issues/2929


mharvey-jt:
@jblomer this one's a medium-priority bug I overlooked, and I'm reasonably confident this is a sound fix for it, but would welcome your opinion

PR #2931, opened 2022-04-25
guard LoadCatalog() with a writelock

mharvey-jt:
https://github.com/cvmfs/cvmfs/issues/2932


mharvey-jt:
Not needed since the catalog reload refactoring

Issue #2897, opened 2022-04-24
hotpatch reload does not change debug mode


mharvey-jt:
A `cvmfs_config reload` does not honour any change to the CVMFS_DEBUGLOG configuration - eg a client normally running the non-debug build cannot be reloaded to  temporarily enable the debug build. 

This is an operational inconvenience as it means it is impossible to enable debugging without a complete remount.

Steps to reproduce:
1) Mount a repo using cvmfs-2.9.2, with a config file that does not have CVMFS_DEBUGLOG present
2) grep /proc/*/maps and observe libcvmfs_fuse3.so.2.9.2 in use
3) edit config file, set CVMFS_DEBUGLOG=
4) `cvmfs_config reload`
5) grep /proc/*/maps and observe libcvmfs_fuse3.so.2.9.2 is still in use, not libcvmfs_fuse3debug.so.2.9.2

Desired behaviour:
`cvmfs_config reload` loads the build of libcvmfs_fuse indicated by the current setting of CVMFS_DEBUGLOG, and not the version selected at initial mount time.

mharvey-jt:
PR https://github.com/cvmfs/cvmfs/pull/2933

mharvey-jt:
ref https://sft.its.cern.ch/jira/projects/CVM/issues/CVM-1745

vvolkl:
#3227 breaks our hotpatching, we'll  need to postpone this to the next patch release.

HereThereBeDragons:
With #3359 we should have now a correctly working version

PR #2933, opened 2022-04-25
allow switching of debug mode over hotpatch

mharvey-jt:
Ref #2897 
this patch has two short-comings:
1) it will break on hotpatch update of a client without this PR depending on the debug mode 
2) It only honours the CVMFS_DEBUGLOG from `default.local`, not the specfic repo's `.config`.


mharvey-jt:
This might still be generally desirable, but we haven't had operational recourse to this for some time, so we could close it.

Issue #2917, opened 2022-04-25
optionally use vendored zlib-ng


mharvey-jt:
For performance reasons, we have replaced the vendored zlib-1.2.8 with zlib-ng, https://github.com/zlib-ng/zlib-ng
I understand there's compatibility issues when comparing compressed objects produced by different versions of zlib, which may make switching from the default undesirable under some scenarios. However, where those concerns don't apply, it would be useful to have the option to select zlib-ng at compile time.

desired behaviour:

`cmake USE_ZLIB_NG=yes` at compile time to select a vendored zlib-ng over the vendored zlib 1.2.8

mharvey-jt:
https://github.com/cvmfs/cvmfs/pull/2934

HereThereBeDragons:
closing as won't fix. replacement will be to support zstd

PR #2934, opened 2022-04-25
zlib-ng as compile-time option with -DBUILD_ZLIB_NG=ON

mharvey-jt:
Adds zlib-ng-2.0.6, optionally build and used in place of zlib-1.2.8 if `-DBUILD_ZLIB_NG=ON` is specified 

Ref: https://github.com/cvmfs/cvmfs/issues/2917


DrDaveD:
Is it possible for zlib-ng to be used in a compatible way?  That would be preferable to a compile time option.  Can it be configured to generate compressions that standard zlib can decode?  That would be the best, and then make that the default.  The next best thing would be to link with both libraries, have the client automatically detect which one is used, and have the publisher have a runtime non-default configuration option to choose the new library.

DrDaveD:
https://github.com/zlib-ng/zlib-ng/blob/develop/PORTING.md says there's a zlib-compat mode; have you tried to see if that gives you sufficient speedup?  If not, it looks like zlib-ng native mode allows coexisting with zlib in the same application.

mharvey-jt:
We use the zlib-compat mode, that enables wrapper functions that make zlib-ng a drop-in replacement for zlib, at the API level. 
zlib-ng compressed data will always be decompressible by a standard zlib. 
I don't know if there's any guarantee that zlib-ng will ever produce compressed results that are bitwise identical to zlib's.

We use zlib-ng on the client side as well as publisher, as it helps reduce latency on catalog load.

DrDaveD:
That sounds good.  It sounds like it should be the default then, not an option.

mharvey-jt:
Closing this, we'll retain it as an internal patch, until zstd support is done.

PR #2935, opened 2022-04-25
make zlib compression level a compile-time option

mharvey-jt:
set `-DENABLE_ZLIB_BEST_SPEED=ON` to set compression level to `Z_BEST_SPEED`, otherwise retain default of `Z_DEFAULT_COMPRESSION`


mharvey-jt:
Closing, as we agreed not to proceed with this

Issue #2899, opened 2022-04-24
CVMFS_IGNORE_SIGNATURE not implemented


mharvey-jt:
Per https://cvmfs.readthedocs.io/en/stable/apx-parameters.html , there is a documented configuration option `CVMFS_IGNORE_SIGNATURE`  _When set to yes, don’t verify CernVM-FS file catalog signatures_.

As of at least CVMFS 2.8 and later, this option is not implemented. Its only occurrence in the source  is in: 
```
$ grep -R IGNORE_SIG *
cvmfs/cvmfs_config:switch_list="CVMFS_IGNORE_SIGNATURE CVMFS_STRICT_MOUNT CVMFS_SHARED_CACHE \
cvmfs/server/cvmfs_server_common.sh:CVMFS_IGNORE_SIGNATURE=no
cvmfs/cvmfs_server:CVMFS_IGNORE_SIGNATURE=no
```

This is actually quite a useful option operationally, if a catalog file has to be manually manipulated outside of cvmfs.

Desired behaviour:

*) This switch is implemented as documented

mharvey-jt:
Our quick-and-dirty implementation: https://github.com/cvmfs/cvmfs/pull/2936

HereThereBeDragons:
PR #3317 reintroduces `_CVMFS_DEVEL_IGNORE_SIGNATURE_FAILURES`. 

This is an option that is NOT designed TO BE USED IN PRODUCTION 
(rare exceptions might exist)

PR #2936, opened 2022-04-25
quick and dirty check of CVMFS_IGNORE_SIGNATURE

mharvey-jt:
Ref: https://github.com/cvmfs/cvmfs/issues/2899 
This is a quick-and-dirty implementation, checking the envvar at time of use, rather than storing the setting in a configuration variable. It serves only to indicate the check that should be made.


HereThereBeDragons:
Superseded by #3317

Issue #2834, opened 2021-12-23
EIO when entry changes file type


mharvey-jt:
Hi,

If an entry on a cvmfs file system changes file type (eg from a symlink to a directory), this can cause kernel to emit an EIO on a stat() / getattr() operation. This arises in 
 `fs/fuse/dir.c` at https://github.com/torvalds/linux/blob/bc491fb12513e79702c6f936c838f792b5389129/fs/fuse/dir.c#L1003 .
The cause of this is that CVMFS preserves inode numbers across a catalog reload, but doesn't account for the possibility of changing type. This means that the kernel can see an inode file type changing between successive lookup/getattr operations. This, apparently, breaks a fundamental assumption in the kernel https://www.spinics.net/lists/linux-fsdevel/msg190819.html  .
This can be mitigated by globally disabling kernel caching with `CVMFS_KCACHE_TIMEOUT=0` or by setting the specific inode's entry timeout to zero.

The `EIO` only happens once. That causes the kernel to invalide the inode, and a fuse `forget` is sent to CVMFS. This drops the cached inode entry and the next operation on the entry causes a new path->inode lookup resulting in a new inode number.

I'm not actually sure if there's a general solution to this that won't push a race elsewhere. Our specific scenario is the conversion of a symlink (that points to an off-cvmfs directory) into a directory of the same name.

Steps to repro:
1) Set up a repro publishing loop that changes an entry from a directory to a symlink on alternate publications. Make the repo TTL 1sec.
2) Mount the repo on a client and do a tight loop lstat()ing that entry, noting `errno`. - EIOs should quickly appear.

jblomer:
Many thanks for the report and the detailed analysis!

That's an interesting problem.  In any case, it seems cvmfs needs to remember the type of the active inodes. That should be relatively easy to do.  The question is what to do with this information when a type conflict arises.  What we cannot do is invalidating the inode within a fuse callback because that can lead to a deadlock. So most likely, in such a case we'd need to issue a new inode for the entry.  The tricky bit is how this will be handled in the cvmfs inode tables. Because at this point, we'd have two inodes for the same path, which is currently not foreseen.

Let me think a little about it.

jblomer:
Just for information, the corresponding [JIRA ticket](https://sft.its.cern.ch/jira/browse/CVM-2062)

mharvey-jt:
Our mitigation for this is to set the kernel cache timeout for off-cvmfs symlinks to zero. 
[...]

PR #2950, opened 2022-04-26
set entry timeout to 0 for off-cvmfs symlinks

mharvey-jt:
Ref https://github.com/cvmfs/cvmfs/issues/2834


jblomer:
I'll close this one for the fix in #2988

Issue #2872, opened 2022-03-23
New pacparser version 1.3.9 available


DrDaveD:
[pacparser 1.3.8](https://github.com/manugarg/pacparser/releases/tag/v1.3.8) has been released.  Update to that version when convenient.

DrDaveD:
There's another new [pacparser release 1.3.9](https://github.com/manugarg/pacparser/releases/tag/v1.3.9) that among other things did a lot of python fixes so hopefully it will pass the tests now.

radupopescu:
#2963 updates Pacparser to version 1.3.8 - this version compiles on newer macOS targets while maintaining compatibility with Python < 3.7 which is used on some of our older targets.

Once we do a cleanup of target platforms, we can revisit Pacparser and try to update to 1.4.0.

I suggest we close this ticket for now.

PR #2963, opened 2022-04-29
Fix x86_64 build on macOS

radupopescu:
* Fix includes in prng.h
* Updating pacparser to version 1.4.0


jblomer (approved):

jblomer:
Fixes #2872

jblomer (dismissed):
That may be a problem: unless sparsehash kept the in-memory layout the same, an update here will break the hotpatch because the information about currently open directories is kept in a sparsehash data structure.

radupopescu:
> That may be a problem: unless sparsehash kept the in-memory layout the same, an update here will break the hotpatch because the information about currently open directories is kept in a sparsehash data structure.

I reverted to the original version of Sparsehash.

radupopescu:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot cloudtest

PR #2969, opened 2022-05-11
Create SElinux label for /var/log/cvmfs

jblomer:


jblomer:
@cernvm-bot cloudtest

Issue #2895, opened 2022-04-23
Failure of the crash monitor process causes mount process to abort


mharvey-jt:
Under certain OOM conditions, we have observed the crash monitor process terminating. Since the monitor process is not essential to correct operation of the mount this should ideally be a survivable event. However, its failure causes the mount process itself to abort, leaving the mount broken, and giving ENOTCONN.

The exact reason for the abort of the monitor process hasn't been determined -- it  occurs when the OS is OOMing, but it is not a result of the OOM-killer reaping any cvmfs process.

Steps to reproduce:
1) mount a repo using cvmfs 2.9.2
2) `kill` the cvmfs2 monitor process 
3) observe that the main cvmfs2 process aborts, leaving accesses to the mount returnin ENOTCONN

Desired behaviour:
1) minimally, the mount process continues functioning in the absence of the monitor process
2) optimally:
  a) the monitor process does not abort in OOM conditions 
  b) the mount process attempts to restart any failed monitor process

mharvey-jt:
https://github.com/cvmfs/cvmfs/pull/2927

jblomer:
@traylenator This looks suspiciously like the issue we see on lxplus, doesn't it? Also here, the watchdog process mysteriously disappears.

Of course we can treat the broken pipe between the fuse process and the watchdog as a non-fatal error. But I feel a bit uneasy not having understood the root cause. It would be good to know what are these OOM conditions that trigger the failure but don't leave any trace in the logs. It somehow must be related to cvmfs, I find it hard to believe that we are the first ones seeing random processes disappearing.

@mharvey-jt Any chance you have more details about the conditions in which the failure occurred? Or how frequently?

mharvey-jt:
The only two things I can say for sure are 
1) The machine OOMed at the same time that CVMFS terminated
2) CVMFS processed were not OOM-killed (and were protected with `oom_score_adj`)
which suggests that the OOM condition caused an error on the pipe, for some reason.
I'll try and repro.

traylenator:
I agree the symptoms match.

Really feel have exhausted looking for OOM logs related or not.

We also already score those processes as don't kill for the OOM .. can not check exact details now but will next week.

jblomer:
We see the issue more frequently on clients backing an overlayfs mount (publishers). In this case, the process starts in the context of a systemd unit.  That may provide a hint to follow up on.

PR #2971, opened 2022-05-16
Fix link between watchdog and fuse module

jblomer:
Possibly fix the occasional unexplained termination of the watchdog (#2895). The watchdog blocks on `read()` on a control pipe. This `read()` call may be interrupted by a signal, which is currently not handled.

The patch is certainly an improvement. Unclear if it is the actual cause of the disappearing watchdog though. If it was, I would have expected to see a [syslog entry](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/monitor.cc#L564). Logging to syslog from the watchdog should have been fixed with [cvmfs-2.9.2](https://github.com/cvmfs/cvmfs/pull/2859) but I didn't see the entry in the CERN logs.

@traylenator Could I forcefully kill a cvmfs mount on an lxplus node, just to see if the reporting works as expected?

### DIFF
--- a/cvmfs/monitor.cc
+++ b/cvmfs/monitor.cc
@@ -250,7 +250,7 @@ string Watchdog::ReportStacktrace() {
   SetLogMicroSyslog(GetLogMicroSyslog());
 
   CrashData crash_data;
-  if (!pipe_watchdog_->Read(&crash_data)) {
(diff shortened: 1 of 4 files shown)


jblomer:
Seems to be an improvement but not necessarily fixing the issue.

PR #2975, opened 2022-05-19
Add support for https-only stratum0

### LINKED ISSUE #2974: Unable to create Stratum 1 with https
Our Stratum 0 stores data on S3 storage that is only available via HTTPS. When trying to create a Stratum1 the following line causes the creation to fail

https://github.com/cvmfs/cvmfs/blob/2d17aad8de3774437c0d3c5fe887f30c69a56005/cvmfs/server/cvmfs_server_util.sh#L876

cfenoy:
Fixes #2974

### DIFF
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -873,7 +873,7 @@ is_master_replica() {
   local name=$1
   local is_master_replica
 
-  if [ $(echo $name | cut --bytes=1-7) = "http://" ]; then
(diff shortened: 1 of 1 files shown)


radupopescu:
Thanks for the patch!

With this change and setting `CVMFS_S3_USE_HTTPS=true` in the S3 conf, are you able to use your stratum?

cfenoy:
Just with this change I'm able to create a Stratum1 with local storage. The Stratum0 was already using the `CVMFS_S3_USE_HTTPS=true`

radupopescu:
Thanks for clarifying, I missed that this was about a Stratum 1.

PR #2976, opened 2022-05-23
Fix memory leak in NentryTracker during reload

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #2850, opened 2022-02-14
Add lock between gc and check

DrDaveD:
See [CVM-2043](https://sft.its.cern.ch/jira/browse/CVM-2043).

This also fixes a bug introduced in #2803 where the is_checking_all lock was not released in the right place.

PR #2982, opened 2022-06-08
Fix releasing of gc lock

DrDaveD:
The new gc lock added in #2850 introduced in 2.9.1 is never getting released.  This PR fixes that.

I'm afraid this bug is bad enough to require a 2.9.3 release.  The tests added fail without the fix.

### DIFF
--- a/cvmfs/server/cvmfs_server_gc.sh
+++ b/cvmfs/server/cvmfs_server_gc.sh
@@ -252,20 +252,23 @@ __do_gc_cmd()
     [ $preserve_revisions   -ge 0 ] && additional_switches="$additional_switches -h $preserve_revisions"
     [ $preserve_timestamp   -gt 0 ] && additional_switches="$additional_switches -z $preserve_timestamp"
 
+    local trapcmd
     if [ $dry_run -eq 0 ]; then
       # if a check or other gc is in progress on this repo, abort
       acquire_gc_lock $name gc 1 || die "Failed to acquire gc lock for $name"
-      local trapcmd="release_gc_lock $name"
+      trapcmd="release_gc_lock $name"
       trap "$trapcmd" EXIT HUP INT TERM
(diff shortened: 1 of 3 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
Thanks for the patch! Just curious, how did you find out?

After the merge, I can cherry-pick the commit in the 2.9 branch and we can prepare a 2.9.3 release.

DrDaveD:
On the Nebraska stratum one where I am running cvmfs_server check -a I wanted to redo a check on one of the repositories after restoring some missing catalogs.  When I did that it said it was waiting for garbage collection to complete but I checked and there was no garbage collection running.  While investigating that I found that many repositories still had that lock and I looked at the code to find out how that could happen.

I think that systems that only run cvmfs_server gc -a will not be likely to notice this because each run is a different process and the lock will get overridden because the previous process is no longer running.  The process ID that goes into the lock file turns out to be the gc -a parent process, not each subprocess (because $$ stays at the parent process even inside a parentheses-created subshell) so there was a long running gc -a that was making all the locks be seen as valid, even when gc wasn't running on that particular repository.

radupopescu:
I don't think we can release this as 2.9.3, as we recently pushed a nightly build tagged 2.10.0-0.384 to the cvmfs-testing repositories. We can, however, also publish this to the testing repo as a new nightly build.

DrDaveD:
You're saying you can't also release a 2.9.3 to the testing repository?  Even if there's a newer one such that people can't install it with just `yum install cvmfs-server`, they could still install it when asking for a particular version number.

jblomer:
It's also backported to the 2.9 branch

radupopescu:
@DrDaveD Ok, in this case, if manually installing version 2.9.3 is useful, we can release it.

PR #2983, opened 2022-06-11
Cancel network fail-over cycle when fuse request is canceled

jblomer:


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

edquist:
Hi @jblomer, I'm seeing the following build error for el8:

```
--> Building pacparser
make clean && make for libpacparser (omitting test execution)...
make: Entering directory '/builddir/build/BUILD/cvmfs-2.9.3/externals_build/build_pacparser/src'
rm -f libpacparser.so libpacparser.so.1 libjs.a pacparser.o pactester pymod/pacparser_o_buildstamp jsapi_buildstamp
cd pymod && python setup.py clean --all
/bin/sh: python: command not found
```

Looks like `python` is required, but on el8, only `python2` and `python3` are available.

Do you recommend a fix?

Thanks!
Carl

DrDaveD:
@edquist Let's make a separate issue for that.  I don't understand why this is only affecting OSG builds, because the CERN CI tests passed.

jblomer:
@edquist If you can use pre-built packages, here's a build with the patch included: http://ecsft.cern.ch/dist/cvmfs/nightlies/cvmfs-git-3397/

edquist:
> edquist Let's make a separate issue for that. I don't understand why this is only affecting OSG builds, because the CERN CI tests passed.

It might be that there is something installed or configured in the PATH that provides `python` then ...

> If you can use pre-built packages, here's a build with the patch included

Thanks @jblomer - we need to build from source.  Does the source.tar.gz from your link contain a new fix for el8 python, or is it just that it builds successfully in the cern environment?

DrDaveD:
> It might be that there is something installed or configured in the PATH that provides `python` then ...

It has to be python3, because the pacparser build requires python3.

> Thanks @jblomer - we need to build from source. Does the source.tar.gz from your link contain a new fix for el8 python, or is it just that it builds successfully in the cern environment?

I'm sure it doesn't have that fix yet because #2985 hasn't yet been merged.  It's not clear to me that you couldn't just instruct the customer to use that daily build, although it would cause a conflict the released osg-oasis which would be a hassle.  So if you want to build from source, just add on to my 2.9.3-1.2.osg36 version.

jblomer:
> > Thanks @jblomer - we need to build from source. Does the source.tar.gz from your link contain a new fix for el8 python, or is it just that it builds successfully in the cern environment?

The reason it builds here is the different build environment. We use this build container for EL8:
https://gitlab.cern.ch/cernvm/build-images/-/blob/master/docker/centos_x86_64%3A8.Dockerfile

jblomer:
@cernvm-bot cloudtest

PR #2986, opened 2022-06-13
Require python3 on rhel>=8 and fedora

### LINKED ISSUE #2985: cvmfs-2.9.3 el8 compilation failure
As mentioned in [#2983](https://github.com/cvmfs/cvmfs/pull/2983#issuecomment-1154191593), the el8 build of cvmfs-2.9.3 fails on the OSG build system because pacparser tries to compile with "python" instead of "python2", and el8 has no python command.

I don't understand why the CERN compilation infrastructure didn't report a problem, but I'll propose a fix.

DrDaveD:
This fixes el8 builds of pacparser on OSG koji by requiring python3.  It also removes the %cvmfs_python macro which was no longer in use.

- Fixes #2985

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -43,19 +43,12 @@
 %define build_fuse3 1
 %endif
 
-%define cvmfs_python python
-%define cvmfs_python_devel python-devel
-%define cvmfs_python_setuptools python-setuptools
-%if 0%{?el8} || 0%{?fedora} >= 31
-%define cvmfs_python python2
-%define cvmfs_python_devel python2-devel
-%define cvmfs_python_setuptools python2-setuptools
-%endif
-# On SLES15, we need the python2 interpreter but python3 devel and setuptools
-# TODO(jblomer): upgrade all python components to Python3
-%if 0%{?sle15}
+%if 0%{?sle15} || 0%{?rhel} >= 8 || 0%{?fedora} >= 31
 %define cvmfs_python_devel python3-devel
 %define cvmfs_python_setuptools python3-setuptools
+%else
+%define cvmfs_python_devel python-devel
+%define cvmfs_python_setuptools python-setuptools
 %endif
 
 %define cvmfs_go golang


jblomer:
Thanks!  I'm afraid our release scripts are not prepared for RPM releases > 1.  So we'd need to resort to creating version 2.9.4. For this PR, please just revert the change the RPM release number, we'll bump the version number after the merge.

jblomer:
Hm, I still don't understand why it builds for us on EL8. Has koji no python2 packages on EL8?

radupopescu:
I fought with pacparser for a long time to get it to compile on all targets. The fact that we do not support optional external dependencies is very limiting here.

I would run a full build on this PR to see how it does.

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot cloudtest

DrDaveD:
Ok I removed the Release: 2 and changelog entry (since I didn't know what version number to put with it).

The OSG koji starts out with a base el8 image which has no python, python2, and python3.  I wonder if your base image is different.

In any case, I have patched the OSG build so our version 2.9.3-1.2 builds and we don't need another immediate release.

DrDaveD:
As a matter of fact python3 should be required on el7 too, if pacparser always needs it.  I am building cvmfs on [copr](https://copr.fedorainfracloud.org/coprs/dwd/cvmfs/build/4543196/) based on the OSG modified source rpm that includes this patch, and it failed building pacparser on el7.  Scroll down to the bottom of the [build log](https://download.copr.fedorainfracloud.org/results/dwd/cvmfs/epel-7-x86_64/04543196-cvmfs/builder-live.log.gz).  I don't know why it succeeded on OSG koji.

DrDaveD:
Well, no, I tried doing that with an el7 scratch build on OSG koji and [that failed too](https://kojihub0000.chtc.wisc.edu/kojifiles/work/tasks/2731/362731/build.log).  I don't know what the right thing to do is.  

Copr said it was missing python mock, and I see there's a python2-mock in EPEL.  Maybe that needs a BuildRequires.  I'll try that.

DrDaveD:
No, that's not it either.  It was trying to do `from unittest.mock import patch` and the mock module does not exist in the python2 on el7.  It is only in python3.  I found a rpm `python-unittest` with yum search that was described as `The new features in unittest backported to Python 2.4+` which looked promising, but it also does not have a mock module.

jblomer:
Let me try to create a build container with (only) the dependencies from the spec file. Should be possible with multi-stage builds.

jblomer:
@cernvm-bot cloudtest

jblomer:
Hm, I don't think we can fix that for the 2.9 branch, but we should do it for the 2.10 release. My favorite solution would be python3 only.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

DrDaveD:
Just because it was easy I also built 2.9.4 on copr, unpatched [here](https://copr.fedorainfracloud.org/coprs/dwd/cvmfs/build/4598250/).   It failed on el7 and el8 as expected, but built on el9.

DrDaveD:
I thought this problem was fixed in 2.9.4, but it still exists in 2.10.0.

jblomer:
@cernvm-bot cloudtest

jblomer:
Pushed to cvmfs-2.10 and devel, too

PR #2987, opened 2022-06-14
Fixing deb packaging on Ubuntu 22.04

radupopescu:


jblomer:
Should we make the fixups in the debian packaging explicit, with a `debian/control.in` file that has the variable bits clearly marked, e.g. `CVMFS_PYTHON_DEV` etc?

radupopescu:
Makes sense.

radupopescu:
An update here. I'm not able to reproduce and understand the CI build failures we see for Ubuntu 22.04. I'd like to have a working build for Ubuntu 22.04 before releasing CVMFS 2.9.3

radupopescu:
@cernvm-bot tidy

PR #2988, opened 2022-06-16
Keep file type in inode tracker

### LINKED ISSUE #2834: EIO when entry changes file type

jblomer:
Detects the issue described in #2834


jblomer:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot cloudtest

jblomer:
... looking into the 076 test failure

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2992, opened 2022-07-01
Fix use of spooler::Peek()

### LINKED ISSUE #2989: Avoid slash prefix in spooler Peek calls
The `Peek` function will add a `/` prefix so it is unnecessary to include it in the argument to `Peek`.

https://github.com/cvmfs/cvmfs/blob/94bba0b5d2ca7e475e5952d44054cdbe0bfd1753/cvmfs/upload_s3.cc#L467-L468

This double slash may cause unexpected HTTP response depending on what upstream CVMFS is talking to. So far I've only found this problem in below two places

https://github.com/cvmfs/cvmfs/blob/94bba0b5d2ca7e475e5952d44054cdbe0bfd1753/cvmfs/swissknife_pull.cc#L714

https://github.com/cvmfs/cvmfs/blob/94bba0b5d2ca7e475e5952d44054cdbe0bfd1753/cvmfs/signing_tool.cc#L96

jblomer:
Fixes #2989

### DIFF
--- a/cvmfs/signing_tool.cc
+++ b/cvmfs/signing_tool.cc
@@ -93,7 +93,7 @@ SigningTool::Result SigningTool::Run(
     }
   } else {
     LogCvmfs(kLogCvmfs, kLogVerboseMsg, "no reflog (ignoring)");
-    if (spooler->Peek("/.cvmfsreflog")) {
+    if (spooler->Peek(".cvmfsreflog")) {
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #2997, opened 2022-07-08
Fix SSL signature issues on CentOS Stream 9

### LINKED ISSUE #2990: CVMFS server failure on CentOS 9
### Description
Creating a repo on CentOS9 prints an error message because cvmfs is not able to sign the repo. The repository is created, but unusable, because a SEG fault is generated at the first transaction.

### Steps to reproduce
On CentOS9, run:
`sudo cvmfs_server mkfs <your_repo_name>`
and then
`sudo cvmfs_server transaction <your_repo_name>`

### Relevant output#

**Creating the repo**
```
$ sudo cvmfs_server mkfs razvan.test.repo
[sudo] password for razvan: 
Owner of razvan.test.repo [root]: razvan
Creating Configuration Files... done
Creating CernVM-FS Master Key and Self-Signed Certificate... done
Creating CernVM-FS Server Infrastructure... done
Creating Backend Storage... done
Signing 30 day whitelist with master key... The command rsautl was deprecated in version 3.0. Use 'pkeyutl' instead.
done
Creating Initial Repository... the private key '/etc/cvmfs/keys/razvan.test.repo.key' doesn't seem to match certificate '/etc/cvmfs/keys/razvan.test.repo.crt' (error:03000098:digital envelope routines::invalid digest)
failed to init repo connection
fail! (cannot sign repo)
```

**Using the repo**
```
$ sudo cvmfs_server transaction razvan.test.repo
/bin/cvmfs_server: line 4006: 173386 Segmentation fault      (core dumped) $(__publish_cmd dbg) transaction $@
```

radupopescu:
Fixes #2990

### DIFF
--- a/cvmfs/server/cvmfs_server_ssl.sh
+++ b/cvmfs/server/cvmfs_server_ssl.sh
@@ -64,10 +64,10 @@ create_whitelist() {
   local usemasterkeycard=0
   local hash_algorithm
 
-  local openssl_keyutil_cmd="rsautl"
+  local openssl_keyutil_cmd="openssl rsautl"
   local openssl_version=`openssl version | cut -d' ' -f2`
   if compare_versions "${openssl_version}" -ge "3.0.0" ; then
-    openssl_keyutil_cmd="pkeyutl"
+    openssl_keyutil_cmd="openssl pkeyutl"
   fi
 
   local whitelist
--- a/cvmfs/server/cvmfs_server_ssl.sh
+++ b/cvmfs/server/cvmfs_server_ssl.sh
@@ -64,6 +64,12 @@ create_whitelist() {
   local usemasterkeycard=0
   local hash_algorithm
 
+  local openssl_keyutil_cmd="rsautl"
+  local openssl_version=`openssl version | cut -d' ' -f2`
+  if compare_versions "${openssl_version}" -ge "3.0.0" ; then
+    openssl_keyutil_cmd="pkeyutl"
+  fi
+
   local whitelist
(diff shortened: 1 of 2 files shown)


jblomer:
This would use LibreSSL for curl. This is not what we need. Curl should continue to use the system SSL library, dynamically linked. We want LibreSSL's libcrypto for its cryptographic primitives.

jblomer (dismissed):
(see comment)

radupopescu:
So in that case, LibreSSL would only be used directly by CVMFS on Linux, except on macOS where it's used by CVMFS directly and also by CURL.

What is the benefit of using the system SSL on Linux if we also start using LibreSSL?

jblomer:
That's right, ideally I'd even like curl to use a native API such as Network.framework on macOS but that's unfortunately not supported.

TLS implementations are frequently subject to security patches and protocol updates. If we used a built-in LibreSSL with curl, we'd need to distribute such patches through updated cvmfs clients. Just the cryptographic primitives, however, are much more stable. Btw, I looked into libsodium, which unfortunately doesn't implement several of the algorithms we need.

radupopescu:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot cloudtest

radupopescu:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

PR #2998, opened 2022-07-12
Fix and simplify unit test runner

jblomer:
Fixes a faulty evaluation of `can_build_gateway`. Along the way, simplifies `can_build_ducc` to just use return codes instead of printed integers.


jblomer:
@cernvm-bot unittest

radupopescu:
Is it ok to merge this now?

PR #2999, opened 2022-07-12
Update deps for snapshotter

radupopescu:
Fix build issue with snapshotter


radupopescu:
@cernvm-bot cloudtest

PR #3001, opened 2022-07-15
Fix unit tests on EL9

jblomer:


jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot unittest

PR #3004, opened 2022-07-19
Catch all errors of publisher construction in txn cmd

jblomer:
Serendipity while looking int #2990


jblomer:
@cernvm-bot cloudtest

PR #3006, opened 2022-07-21
Workaround for OpenSSL issues on EL9

jblomer:
Adds a new binary `cvmfs_signing_helper`. This helper does not have any elevated privileges so that the `OPENSSL_ENABLE_SHA1_SIGNATURES` environment variable is effective. It is called from `cvmfs_swissknife` and `cvmfs_publish` so that the manifest and the whitelist can be signed on EL9 publishers.

That's not a great workaround.  It should buy us some time for a better fix.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -481,6 +481,17 @@ set (CVMFS_SWISSKNIFE_SOURCES
   xattr.cc
 )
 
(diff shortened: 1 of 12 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #3013, opened 2022-08-04
Fix: bash string comparison for cvmfs server scripts

HereThereBeDragons:
Problem could be see with e.g. command: sudo cvmfs_server update-info


radupopescu:
@cernvm-bot cloudtest

PR #3018, opened 2022-08-10
New timer support

razvanvirtan:
@radupopescu  One of the possible ways of profiling CernVM-FS is analyzing the output from some timers placed in the key points of the code.
This PR adds an enhanced timer support to the cvmfs code. Also, this draft adds some timers to the cvmfs code. Some of them should be merged (the callbacks timers), while others are not mandatory, but illustrate different usage scenarios.

The new timer implementation provides automatically creation of parent - children relations between a timer and a timer that have been created during it's run. The timers output is printed therefore in a tree-like form and added to the `cvmfs_talk` output. Basically, a user can create a partial flamegraph for the interesting code paths.

A timer is started by calling  the `HighPrecisionTimer()` constructor and automatically stopped when the destructor is called. The result is then added to the total time spent on that part of the code (like in a flamegraph).
- For simply timing a function, it is enough to call the constructor at the beginning of the function, as we do in https://github.com/cvmfs/cvmfs/pull/3018/commits/dba5a42f067ccfc1ab68ff8cb265b400b29615a0 for the callbacks.
- For timing only a few lines of code, those line should be enclosed in a scope and the constructor should be called at the beginning of the scope, as in https://github.com/cvmfs/cvmfs/pull/3018/commits/20aed1f0fdde9528d36d6040f542ffa2abb775a7#diff-7be7dd92cd800be25db196ca2a4c09df0af7143ba53be2133f21d45d0d33fe40R1785.
- There are situations when the timer value should be added to the total time only in a specific situation. Therefore, we can set `HighPrecisionTimer.ignore_result`. Also, the destructor can be called explicitly if we need a fine control, like in https://github.com/cvmfs/cvmfs/pull/3018/commits/20aed1f0fdde9528d36d6040f542ffa2abb775a7#diff-7be7dd92cd800be25db196ca2a4c09df0af7143ba53be2133f21d45d0d33fe40R543.

Please note that the second and third commits are here for showing different ways of using the timer interface, we may not want to keep the timers added there.

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -102,6 +102,7 @@
 #include "sqlitevfs.h"
 #include "statistics.h"
 #include "talk.h"
+#include "timers.h"
 #include "tracer.h"
 #include "util/algorithm.h"
 #include "util/atomic.h"
@@ -117,7 +118,6 @@
 using namespace std;  // NOLINT
 
 namespace cvmfs {
-
 FileSystem *file_system_ = NULL;
 MountPoint *mount_point_ = NULL;
 TalkManager *talk_mgr_ = NULL;
@@ -389,7 +389,8 @@ static void inline TraceInode(const int event,
  * We do check catalog TTL here (and reload, if necessary).
  */
 static void cvmfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name) {
-  HighPrecisionTimer guard_timer(file_system_->hist_fs_lookup());
+  TimerGuard timer_guard("cvmfs_lookup()", CVMFS_LOOKUP_TIMER,
+    file_system_->hist_fs_lookup());
 
   perf::Inc(file_system_->n_fs_lookup());
   const struct fuse_ctx *fuse_ctx = fuse_req_ctx(req);
@@ -498,7 +499,8 @@ static void cvmfs_forget(
   uint64_t nlookup
 #endif
 ) {
-  HighPrecisionTimer guard_timer(file_system_->hist_fs_forget());
+  TimerGuard timer_guard("cvmfs_forget()", CVMFS_FORGET_TIMER,
+    file_system_->hist_fs_forget());
 
   perf::Inc(file_system_->n_fs_forget());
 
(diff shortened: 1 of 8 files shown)


radupopescu:
@cernvm-bot tidy

radupopescu:
@cernvm-bot cpplint

radupopescu:
@cernvm-bot unittest

radupopescu:
Great job, this is really useful!

razvanvirtan:
@jblomer I have rebased and updated the PR, replacing the string key with the solution we've talked about this morning,
so now it is ready for a round of testing / reviewing.

I thought some more about the best final structure for the timers interface. I am not sure that adding
the timers interface (e.g. the timers map) to the mountpoint is the right choice. The timers output is
part of the `cvmfs_talk` command, and all the statistics (e.g. the histograms connected to our timers)
are currently part of the file system. Therefore, I think the timers output should also be "per file system",
at least for the moment.

Making the map (or a wrapper object that also includes the lock and the root vector)
a member of the file system class would complicate the current interface, adding one more parameter to the
`HighPrecisionTimer()` constructor. However, since we know that we only have one file system (and possibly
multiple mountpoints), we will also have only one map, so we will always send the same parameter to the
`HighPrecisionTimer()` constructor.

What do you think about this, is keeping the current implementation a good choice, maybe renaming the classes,
or shall we think about another structure?

razvanvirtan:
@jblomer I've rebased and updated the PR, replacing the string key in the timers map, as we've talked this morning.
Please note that the second and third commits are here for showing different ways of using the timer interface, we may not want to keep the timers added there.

I thought more about the best structure for the timer interface. I don't know if placing the timers map (or a wrapper that also includes the lock and the root timers vector) in the mountpoint is the right choice. The timers output is part of the `cvmfs_talk` output. Currently, all the statistics printed by this command, including the histograms connected to our timer, are bound to the file system, not to the mountpoint. Therefore, I think we should only have one timers map, "per file system".

If we use this approach, making the timers map part of the file system would complicate the timers interface, adding one more parameter to the `HighPrecisionTimer()` constructor (we will need to send the timers map / the wrapper object every time we create a timer). This approach makes sense for the histograms, because each callback timer is connected to a different histogram, but we only expect to have one timers map in a cvmfs process.

What do you think, shall we keep the current implementation (maybe renaming the classes) or shall we look more into this?

razvanvirtan:
Last changes: remove `PTHREADS_MUTEX_INITIALIZER` for the timers lock and replace it with explicit `pthread_mutex_init()` initialization in the filesystem.

razvanvirtan:
Last update
- Created new `TimerTree` singleton class and moved the static fields from `HighPrecisionTimer_` here. Now, the only static field is the `instance` member of `TimerTree`. Also, moved here `startTimer()`, `stopTimer()`, `printTimers()`, `printTimerTree()`
- Renamed `HighPrecisionTimer_` and `HighPrecisionTimer` to `Timer` and `GuardTimer`

razvanvirtan:
Last update: Solved synchronization issue for timers map insertion.

PR #3021, opened 2022-08-11
Safe garbage collection triggering on gateway

radupopescu:
This makes GC safe to run on the gateway machine.

* The `cvmfs_gateway` executable has been reorganized to provide a series of maintenance commands: list-leases, flush-leases, cancel-lease, set-repo-state
* Triggering the GC on the gateway will now lock the repos and flush any existing leases before starting the collection. Repos are unlocked at the end of the GC run


radupopescu:
@cernvm-bot unittest

radupopescu:
@cernvm-bot cloudtest

radupopescu:
The newly added calls to `cvmfs_gateway` in `cvmfs_server_gc.sh` are made even if the `cvmfs-gateway` service isn't running, breaking the integration tests.

radupopescu:
@cernvm-bot cloudtest

radupopescu:
The remaining integration tests which fail seem unrelated, so I think this can safely be merged now.

PR #3039, opened 2022-08-30
Add average times profiling tool

razvanvirtan:
This commit adds a profiling tool that provides user,
system, and total time statistics for a cvmfs benchmark.
Provided with a given benchmark script, this tool runs it for
a given number of times in all three cache scenarios and prints
the average times values, together with the ratios between
different cache scenarios.


razvanvirtan:
@jblomer I updated the PR according to this morning discussion.
- by providing a third parameter, the user can now combine the CVMFS benchmarks statistics with the time values measured for a benchmark that uses a local filesystem
- the temporary directory is now created using `mktemp`
- since `cvmfs_config umount` works for `autofs` cases too, I didn't change anything about that

jblomer (approved):

PR #3040, opened 2022-08-31
Skip foreign layers in container image conversion

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3046, opened 2022-09-02
Fix compile error "'time' was not declared in this scope"

fsiegert:
In Archlinux with gcc 12.2.0 we see a compilation error that seems to come from a missing `<ctime>` include. More details can be found in https://aur.archlinux.org/packages/cvmfs#comment-879961.

This PR fixes the issue.

### DIFF
--- a/cvmfs/authz/helper_log.cc
+++ b/cvmfs/authz/helper_log.cc
@@ -14,6 +14,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <ctime>
 
 using namespace std;  // NOLINT
 


jblomer (changes requested):
Many thanks for the patch!

Please just move the include down below `#include <cstring>`

fsiegert:
Thanks @jblomer, this is indeed more tidy and I have amended the commit. Hope it's fine now!

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cpplint

jblomer (approved):
Many thanks!

PR #2959, opened 2022-04-27
Add switch `CVMFS_CACHE_SYMLINKS` 

mharvey-jt:
Lets the kernel cache symlink references, if there's kernel fuse support (rhel8+) and libfuse support (3.10.0+)

Ref issue #2949

PR #3048, opened 2022-09-02
Add support for symlink kernel caching

### LINKED ISSUE #2949: enable fuse caching of symlinks, if supported by kernel and libfuse
Improve performance in symlink-heavy repos by allowing the kernel to cache symlink dereferences.

Requires rhel8 + fuse 3.10.

This has the potential to cause an issue over catalog update if the symlink changes  - the kernel will return the old value until the cache expires.

HereThereBeDragons:
Extended @mharvey-jt PR #2959 :
- fixes the wrongly truncated new symlinks when short -> long
- added fallback in case symlink caching is not possible 
- and fixed some small things.

Note: If you use this PR, enabling symlink kernel caching, you SHOULD NOT use symlinks as mountpoint. 
Currently it is not possible to just invalidate dentries in away that they are just considered stale in cache and need revalidation. Instead, the dentry of the symlink is evicted and the mountpoint on the symlink will be just gone.

The current libfuse implementation does not offer any solution to this problem.

Missing: Tests..

Fixes #2949

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -466,7 +466,10 @@ static void cvmfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name) {
   // 076 fails with the following line uncommented
   //
   // WARNING! ENABLING THIS BREAKS ANY TYPE OF MOUNTPOINT POINTING TO THIS INODE
-  if (mount_point_->cache_symlinks()) {  //  && dirent.IsLink()) {
+  // 
+  // only safe if fuse_expire_entry is available
+  if (mount_point_->fuse_expire_entry()
+      || (mount_point_->cache_symlinks() && dirent.IsLink())) {
     LogCvmfs(kLogCache, kLogDebug, "Dentry to evict: %s", name);
     mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
   }
@@ -1641,22 +1644,32 @@ static void cvmfs_init(void *userdata, struct fuse_conn_info *conn) {
   }
 
   if ( mount_point_->cache_symlinks() ) {
-#if FUSE_VERSION >= 310
-    if ((conn->capable & FUSE_CAP_CACHE_SYMLINKS) == 0) {
+#ifdef FUSE_CAP_CACHE_SYMLINKS
+    if ((conn->capable & FUSE_CAP_CACHE_SYMLINKS) == FUSE_CAP_CACHE_SYMLINKS) {
+      conn->want |= FUSE_CAP_CACHE_SYMLINKS;
+      LogCvmfs(kLogCvmfs, kLogDebug, "FUSE: "
+                                    "Enable symlink caching");
+    } else {
       mount_point_->DisableCacheSymlinks();
-      LogCvmfs(kLogCvmfs, kLogDebug,  // kLogDebug | kLogSyslogErr,
-           "Symlink caching requested but missing fuse kernel support, "
+      LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogErr,
+           "FUSE: Symlink caching requested but missing fuse kernel support, "
            "falling back to no caching");
     }
-    conn->want |= FUSE_CAP_CACHE_SYMLINKS;
-    LogCvmfs(kLogCvmfs, kLogDebug, "enabling symlink caching");
 #else
     mount_point_->DisableCacheSymlinks();
-    LogCvmfs(kLogCvmfs, kLogDebug,
-           "Symlink caching requested but fuse version not >=3.10, "
-           "falling back to no caching");
-#endif
+    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogErr,
+          "FUSE: Symlink caching requested but missing libfuse support, "
+          "falling back to no caching");
+#endif 
   }
+
+#ifdef FUSE_CAP_EXPIRE_ONLY
+  if ((conn->capable & FUSE_CAP_EXPIRE_ONLY) == FUSE_CAP_EXPIRE_ONLY) {
+    mount_point_->EnableFuseExpireEntry();
+    LogCvmfs(kLogCvmfs, kLogDebug, "FUSE: "
+                                   "Enable fuse_expire_entry");
+  }
+#endif
 }
 
 static void cvmfs_destroy(void *unused __attribute__((unused))) {
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1649,15 +1649,20 @@ static void cvmfs_init(void *userdata, struct fuse_conn_info *conn) {
       conn->want |= FUSE_CAP_CACHE_SYMLINKS;
       LogCvmfs(kLogCvmfs, kLogDebug, "FUSE: "
                                     "Enable symlink caching");
+      #ifndef FUSE_CAP_EXPIRE_ONLY
+        LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
+          "FUSE: Symlink caching enabled but no support for fuse_expire_entry, "
+          "mountpoints on top of symlinks will break!");
+      #endif
     } else {
       mount_point_->DisableCacheSymlinks();
-      LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogErr,
+      LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
            "FUSE: Symlink caching requested but missing fuse kernel support, "
            "falling back to no caching");
     }
 #else
     mount_point_->DisableCacheSymlinks();
-    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogErr,
+    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
           "FUSE: Symlink caching requested but missing libfuse support, "
           "falling back to no caching");
 #endif
@@ -1668,6 +1673,10 @@ static void cvmfs_init(void *userdata, struct fuse_conn_info *conn) {
     mount_point_->EnableFuseExpireEntry();
     LogCvmfs(kLogCvmfs, kLogDebug, "FUSE: "
                                    "Enable fuse_expire_entry");
+  } else if (mount_point_->cache_symlinks()) {
+    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
+      "FUSE: Symlink caching enabled but no support for fuse_expire_entry, "
+      "mountpoints on top of symlinks will break!");
   }
 #endif
 }
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -466,7 +466,7 @@ static void cvmfs_lookup(fuse_req_t req, fuse_ino_t parent, const char *name) {
   // 076 fails with the following line uncommented
   //
   // WARNING! ENABLING THIS BREAKS ANY TYPE OF MOUNTPOINT POINTING TO THIS INODE
-  if (mount_point_->cache_symlinks() && dirent.IsLink()) { 
+  if (mount_point_->cache_symlinks() && dirent.IsLink()) {
     mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
   }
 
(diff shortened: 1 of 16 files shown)


jblomer:
Related discussion on the fuse mailing list: https://sourceforge.net/p/fuse/mailman/message/37702061/

HereThereBeDragons:
Working tests for kernel caching including symlinks for fuse2 and >= fuse3.10. Tested on fuse3.10 with kernel patch and without
- [patched_kernel_702-symlink_caching.log](https://github.com/cvmfs/cvmfs/files/9617655/patched_kernel_702-symlink_caching.log)
- [unpatched_kernel_702-symlink_caching.log](https://github.com/cvmfs/cvmfs/files/9617656/unpatched_kernel_702-symlink_caching.log)

**Thinks that need further decision**

0) Ask libfuse to add some CAP_ define to check with the kernel supports notify_expire_entry
  - libfuse patch provides fuse_lowlevel_notify_expire_entry()
  - BUT in addition, i introduced: FUSE_CAP_DENTRY_EXPIRE_ONLY in libfuse/include/fuse_common.h to be able to see if the kernel supports it

1) One problem of the test is the clean up afterwards. It needs sudo rights
  ```
  -- Testing Symlink caching: Proper eviction policy for kernel cache (Wed Sep 21 16:23:48 CEST 2022 / test number 702)
  rm: cannot remove '/tmp/cvmfs-test/workdir/702-symlink_caching/private_mntc/just.test.repo': Permission denied
  ```

2) We should think about these lines 469-472 in `cvmfs.cc`:
```
  if (mount_point_->cache_symlinks()) {  //  && dirent.IsLink()) {
    LogCvmfs(kLogCache, kLogDebug, "Dentry to evict: %s", name);
    mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
  }
```

we can do something like
```c++
#ifdef FUSE_CAP_DENTRY_EXPIRE_ONLY //we can always deleten dentry tracker
    mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
#else // we only delete if symlink caching is requested and only symlink dentries
  if (mount_point_->cache_symlinks())  && dirent.IsLink()) {
    LogCvmfs(kLogCache, kLogDebug, "Dentry to evict: %s", name);
    mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
  }
#endif
```

Questions:
- Keep log message? (I just had it for testing) `LogCvmfs(kLogCache, kLogDebug, "Dentry to evict: %s", name);`

@jblomer (just to notify you)

HereThereBeDragons:
Ah and another thing: the linter here on github currently returns the wrong line numbers? 
If you call it locally you get the correct line numbers back

jblomer:
> Ah and another thing: the linter here on github currently returns the wrong line numbers? If you call it locally you get the correct line numbers back

That should eventually be fixed by [moving to clang-format](https://sft.its.cern.ch/jira/browse/CVM-2030).

jblomer:
>     0. Ask libfuse to add some CAP_ define to check with the kernel supports notify_expire_entry
> 
> 
>     * libfuse patch provides fuse_lowlevel_notify_expire_entry()
> 
>     * BUT in addition, i introduced: FUSE_CAP_DENTRY_EXPIRE_ONLY in libfuse/include/fuse_common.h to be able to see if the kernel supports it

Yep, that should be part of our feedback to the fuse mailing list.


>     1. One problem of the test is the clean up afterwards. It needs sudo rights
> 
> 
> ```
> -- Testing Symlink caching: Proper eviction policy for kernel cache (Wed Sep 21 16:23:48 CEST 2022 / test number 702)
> rm: cannot remove '/tmp/cvmfs-test/workdir/702-symlink_caching/private_mntc/just.test.repo': Permission denied
> ```

That's fine, we can just use `sudo` where needed. Integration tests are supposed to run on an account that has password-less sudo access.

>     2. We should think about these lines 469-472 in `cvmfs.cc`:
> 
> 
> ```
>   if (mount_point_->cache_symlinks()) {  //  && dirent.IsLink()) {
>     LogCvmfs(kLogCache, kLogDebug, "Dentry to evict: %s", name);
>     mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
>   }
> ```
> 
> we can do something like
> 
> ```c++
> #ifdef FUSE_CAP_DENTRY_EXPIRE_ONLY //we can always deleten dentry tracker
>     mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
> #else // we only delete if symlink caching is requested and only symlink dentries
>   if (mount_point_->cache_symlinks())  && dirent.IsLink()) {
>     LogCvmfs(kLogCache, kLogDebug, "Dentry to evict: %s", name);
>     mount_point_->dentry_tracker()->Add(parent_fuse, name, uint64_t(timeout));
>   }
> #endif
> ```

Hm, `ifdef` may not be enough because both kernel and library need to support it.  We need to check dynamically if the capability is set (see `fuse_conn_info->capable` and `fuse_conn_info->want`).

I think we should by default disable symlink caching if expiring dentries is not available, even if enable by cvmfs options. In this case, we should add a warning to syslog ("symlink caching enabled but not available..."). We may consider an additional option that enforces symlink caching even if the fuse requirements are missing.  In this case, we would resort to only add symlinks to the dentry tracker (and warn about breaking mountpoints on symlinks).

> Questions:
> 
>     * Keep log message? (I just had it for testing) `LogCvmfs(kLogCache, kLogDebug, "Dentry to evict: %s", name);`

This can stay, I think.

jblomer:
On second thought, regarding cvmfs options: we actually don't need an option `CVMFS_SYMLINK_CACHING`, I think. If fuse supports symlink caching _and_ dentry expiry, we just enable it.  Otherwise we don't. We may have an option `CVMFS_ENFORCE_SYMLINK_CACHING` for a transition period, where we enable symlink caching (with caveats) even if we cannot expire dentries.  Let's merge a first PR without that option though.

HereThereBeDragons:
ok then we first need a new version of fuse where we can check if the capability is available.
(thought it might be easier to commit immediately the `CVMFS_ENFORCE_SYMLINK_CACHING` because we already have all the functionality of it in the previous PR)

and i guess you are right that a "define" is not good enough as its defined in libfuse and not in the fuse kernel module

HereThereBeDragons:
@jblomer have a look. 
not sure how we can test it automatically unless we patch our own kernel for some of the cloudtest machines..

test/test_functions
```diff
@@ -2372,10 +2374,14 @@ EOF
   echo "  *** local mount config:"
   cat "$config"
 
+  if [[ ! -z $mount_options ]]; then
```
jblomer: `[[ ]]` is a bashism (doesn't work on Ubuntu's dash)
```suggestion
  if [ ! -z $mount_options ]; then
```

test/test_functions
```diff
@@ -2372,10 +2374,14 @@ EOF
...
+    mount_options=$(echo $mount_options",")
```
jblomer:
```suggestion
    mount_options="${mount_options},"
```

cvmfs/fuse_evict.cc
```diff
@@ -190,14 +199,26 @@ void *FuseInvalidator::MainInvalidator(void *data) {
...
+                                  *reinterpret_cast<struct fuse_session**>(
+                                  invalidator->fuse_channel_or_session_);
+#endif
+
+// we do not care if fuse kernel supports expire_entry as if it is
+// not support it will just be handled like a fuse_inval
+#ifdef FUSE_CAP_EXPIRE_ONLY
+      fuse_lowlevel_notify_expire_entry(channel_or_session,
+        entry_parent, entry_name.GetChars(), entry_name.GetLength(),
+        fuse_expire_flags::FUSE_LL_EXPIRE_ONLY);
```
jblomer: I guess this doesn't work with C++03
```suggestion
        FUSE_LL_EXPIRE_ONLY);
```

cvmfs/cvmfs.cc
```diff
@@ -1633,6 +1642,34 @@ static void cvmfs_init(void *userdata, struct fuse_conn_info *conn) {
...
+
+  if ( mount_point_->cache_symlinks() ) {
+#ifdef FUSE_CAP_CACHE_SYMLINKS
+    if ((conn->capable & FUSE_CAP_CACHE_SYMLINKS) == FUSE_CAP_CACHE_SYMLINKS) {
+      conn->want |= FUSE_CAP_CACHE_SYMLINKS;
+      LogCvmfs(kLogCvmfs, kLogDebug, "FUSE: "
+                                    "Enable symlink caching");
+    } else {
+      mount_point_->DisableCacheSymlinks();
+      LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogErr,
```
jblomer:
```suggestion
      LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
```

cvmfs/cvmfs.cc
```diff
@@ -1633,6 +1642,34 @@ static void cvmfs_init(void *userdata, struct fuse_conn_info *conn) {
...
+           "FUSE: Symlink caching requested but missing fuse kernel support, "
+           "falling back to no caching");
+    }
+#else
+    mount_point_->DisableCacheSymlinks();
+    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogErr,
```
jblomer:
```suggestion
    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
```

cvmfs/cvmfs.cc
```diff
@@ -1633,6 +1642,34 @@ static void cvmfs_init(void *userdata, struct fuse_conn_info *conn) {
...
+#endif
+  }
+
+#ifdef FUSE_CAP_EXPIRE_ONLY
+  if ((conn->capable & FUSE_CAP_EXPIRE_ONLY) == FUSE_CAP_EXPIRE_ONLY) {
+    mount_point_->EnableFuseExpireEntry();
+    LogCvmfs(kLogCvmfs, kLogDebug, "FUSE: "
+                                   "Enable fuse_expire_entry");
+  }
+#endif
```
jblomer, line 1682: I think we want an `#else` branch and print one more warning: if we have symlink caching turned on but expire entry turned off, we want to issue a warning about the fact that mountpoints on top of symlinks will break.

test/src/702-symlink_caching/main
```diff
@@ -0,0 +1,196 @@
...
+    sudo newfs_hfs $CVMFS_TEST_702_OSXMOUNTPOINT || return 30
+    sudo mount -t hfs $CVMFS_TEST_702_OSXMOUNTPOINT $mntpnt/testdir_mountpoint || return 31
+    sudo mkdir $mntpnt/testdir_mountpoint/self || return 32
+  else
+    sudo mount -t proc none $mntpnt/testdir_mountpoint || return 33
+  fi
+  trap cleanup_mount_on_top EXIT HUP INT TERM
+
+  ls $mntpnt/testdir_mountpoint/self > /dev/null || return 34
+  diff $mntpnt/testdir_mountpoint/self/maps /proc/self/maps || return 35
```
jblomer, line 78: I guess this fails on macOS, so perhaps restrict this check to linux only; or remove.

test/src/702-symlink_caching/main
```diff
@@ -0,0 +1,196 @@
...
+  # res_cached must be 0
+  local res_cached=$(check_readlink_is_cached $mntpnt $symlink_name)
+
+  echo ""
+  echo "  *** Test symlink short -> long replacement"
+  local old_symlink_pointer=$(readlink ${mntpnt}/${symlink_name})
+  replace_short_symlink_with_long $short_name $long_name $symlink_name
+  local new_symlink_pointer=$(readlink ${mntpnt}/${symlink_name})
+  # res_switch must be 0
+  if [ "$new_symlink_pointer" = "$long_name" ] && [ "$new_symlink_pointer" != "$short_name" ]; then
```
jblomer:
```suggestion
  if [ "$new_symlink_pointer" = "$long_name" ] && [ "$old_symlink_pointer" = "$short_name" ]; then
```

jblomer (changes requested):
Very nice!  Almost there, I think.  We'll need to exclude the 702 integration test on a number (all?) of platforms.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

test/src/702-symlink_caching/main
HereThereBeDragons: why? other directory possible to mount on top that exists on all platforms?

HereThereBeDragons:
> Very nice! Almost there, I think. We'll need to exclude the 702 integration test on a number (all?) of platforms.

i removed it now from all platforms. can be added back if better support for symlink caching and expire_entries is given. 

if backporting to libfuse2 becomes a common thing the integration test needs to be modified

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@jblomer  have a look. seems like the test fails do to unrelated stuff

jblomer (approved):

PR #3055, opened 2022-09-08
Fix ingestion with catalog of empty tarfile

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3056, opened 2022-09-09
[ducc] Add support for OCI manifests

### LINKED ISSUE #2851: cvmfs_ducc: support `application/vnd.oci.image.manifest.v1+json` manifest
Hello,

If we build and push an image with podman or buildah to a local registry, executing `cvmfs_ducc convert` would cause the following error :

```sh
{ "errors":[{ "code": "MANIFEST_UNKNOWN", "message": "OCI manifest found, but accepted header does not support OCI manifests"}]}

ERRO[0001] Error in getting manifest to determine singleton path error="Got empty manifest"
WARN[0001] Error in obtaining the path to save the Singularity file system: Got empty manifest error="Got empty manifest"
ERRO[0001] Error in converting wish (singularity), going on error="Error in getting the path where to save Singularity filesystem : Got empty manifest" input image="http://localhost:5000/library/symbolic-link:latest" output image="http://registry.fake/thin/library/symbolic-link:latest" repository=unpacked.fake
```

This is because podman and buildah use an OCI manifest by default.

On the client side, this can be fixed by using :

```sh
podman build -t <tag> --format docker .
# For buildah
buildah commit --format docker "$ctr" "<tag>"
```

On the cvmfs_ducc side, it would be great if DUCC could support `application/vnd.oci.image.manifest.v1+json`.

https://github.com/cvmfs/cvmfs/blob/e54432cbd2deb065947b6f7f1f28ca0765019461/ducc/lib/image.go#L178

The documentation of the OCI manifest is written here: https://github.com/opencontainers/image-spec/blob/main/manifest.md

jblomer:
Fixes #2851

### DIFF
--- a/ducc/lib/image.go
+++ b/ducc/lib/image.go
@@ -283,6 +283,7 @@ func (img *Image) GetOCIImage() (config image.Image, err error) {
 	}
 	req.Header.Set("Authorization", token)
 	req.Header.Set("Accept", "application/vnd.docker.distribution.manifest.v2+json")
+	req.Header.Set("Accept", "application/vnd.oci.image.manifest.v1+json")
 
 	resp, err := client.Do(req)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #3057, opened 2022-09-09
[ducc] fix registry use for config blob

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3059, opened 2022-09-11
Fix libarchive build on glibc >= 2.36

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3072, opened 2022-09-27
Set CPU affinity for CVMFS processes

mharvey-jt:
It's occasionally useful to set the CPU affinity of the cvmfs2 processes. This patch add the configuration option `CVMFS_CPU_AFFINITY` which takes a comma-delimited list of cpu cores to bind to. Eg

CVMFS_CPU_AFFINITY=0,2


cvmfs/loader.cc
```diff
@@ -736,6 +749,21 @@ int FuseMain(int argc, char *argv[]) {
...
+ if( options_manager->GetValue("CVMFS_CPU_AFFINITY", &parameter) ) {
+     cpu_set_t mask;
+     vector<int> cpus = splitToInts( parameter, ',' );
+     CPU_ZERO(&mask);
+     for( int i : cpus ) {
+        CPU_SET( i, &mask );
+     }
+
+     LogCvmfs(kLogCvmfs, kLogStdout, "Setting CPU Affinity to %s", parameter.c_str() );
+     int err = sched_setaffinity( 0, sizeof(mask), &mask );
```
jblomer: We'll need to ignore the request on macOS or use Mac's native API; I don't think `sched_setaffinity` exists there.

cvmfs/loader.cc
```diff
@@ -736,6 +749,21 @@ int FuseMain(int argc, char *argv[]) {
     fuse_opt_add_arg(mount_options, "-osuid");
     LogCvmfs(kLogCvmfs, kLogStdout, "CernVM-FS: running with suid support");
   }
+
+ if( options_manager->GetValue("CVMFS_CPU_AFFINITY", &parameter) ) {
+     cpu_set_t mask;
+     vector<int> cpus = splitToInts( parameter, ',' );
```
jblomer: I think we should use `SplitString` and `String2UInt64` from the cvmfs string utils instead.

jblomer:
Many thanks for the patch!  Do you have a recommendation how cvmfs pinning should be used?

mharvey-jt:
This is useful in environments where we want to confine cvmfs to dedicated service cores, and not cause variable latency in other processes colocated on the host. This could also be done with a cgroup, but would be a more involved change.

cvmfs/loader.cc
mharvey-jt: Made it a no-op on macos, as it's very unlikely to be a useful feature in that environment

cvmfs/loader.cc
mharvey-jt: converted to use those

jblomer:
Replaced by #3330

Issue #3028, opened 2022-08-17
Remove LookupOption kLookupSole


HereThereBeDragons:
Looks like that kLookupSole is not used anymore.

HereThereBeDragons:
`kLookupSole` seems to be only used in cvmfs/catalog_mrg_impl.h +246 
```c++
template <class CatalogT>
bool AbstractCatalogManager<CatalogT>::LookupPath(const PathString &path,
                                        const LookupOptions options,
                                        DirectoryEntry *dirent)
```

the only usage of the option is in line 325
```c++
if ((options & kLookupRawSymlink) == kLookupRawSymlink)
```

As `LookupOptions` is an enum and has already only 2 option I think we still need `kLookupSole`
```
enum LookupOptions {
  kLookupSole        = 0x01,
  // kLookupFull        = 0x02  not used anymore
  kLookupRawSymlink  = 0x10,
};
```

----------------------
Occurences of `kLookupSole` (tests and binaries excluded)
```c++
./cvmfs/catalog_balancer_impl.h:62:  retval = catalog_mgr_->LookupPath(PathString(path), kLookupSole, &parent);
./cvmfs/cvmfs.cc:269:      path, catalog::kLookupSole, dirent);
./cvmfs/cvmfs.cc:299:    if (catalog_mgr->LookupPath(path, catalog::kLookupSole, dirent)) {
./cvmfs/cvmfs.cc:312:      catalog_mgr->LookupPath(PathString(), catalog::kLookupSole, dirent);
./cvmfs/cvmfs.cc:332:  if (catalog_mgr->LookupPath(path, catalog::kLookupSole, dirent)) {
./cvmfs/cvmfs.cc:382:  retval = catalog_mgr->LookupPath(path, catalog::kLookupSole, dirent);
./cvmfs/cvmfs.cc:1115:    if (!catalog_mgr->LookupPath(path, catalog::kLookupSole, &dirent_origin)) {
./cvmfs/cvmfs.cc:1547:      catalog::kLookupSole | catalog::kLookupRawSymlink);
./cvmfs/libcvmfs_int.cc:198:  if (mount_point_->catalog_mgr()->LookupPath(path, catalog::kLookupSole,
./cvmfs/catalog_mgr_rw.cc:302:  if (!LookupPath(relative_source, kLookupSole, &source_dirent)) {
[...]

PR #3074, opened 2022-10-04
Change LookupOption

HereThereBeDragons:
Related to #3028 
- LookupOption type changes from enum to unsigned to be able to be a bit mask
- `kLookupSole` is renamed to `kLookupDefault` as this is the default treatment of a directory entry
- added some comments about the functionality

### DIFF
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -35,7 +35,7 @@ const unsigned kSqliteMemPerThread = 1*1024*1024;
 /**
  * LookupOption for a directory entry (bitmask).
  * kLookupDefault = Look solely at the given directory entry (parent is ignored)
- * kLookupRawSymlink = Mangle symlink to which the directory entry points to 
+ * kLookupRawSymlink = Don't resolve environment variables in symlink targets
  */
 typedef unsigned LookupOptions;
 const unsigned kLookupDefault = 0b1;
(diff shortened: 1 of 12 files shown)


cvmfs/catalog_mgr.h
```diff
@@ -31,15 +31,15 @@ namespace catalog {
 
 const unsigned kSqliteMemPerThread = 1*1024*1024;
 
+
 /**
- * Lookup a directory entry including its parent entry or not.
+ * LookupOption for a directory entry (bitmask).
+ * kLookupDefault = Look solely at the given directory entry (parent is ignored)
+ * kLookupRawSymlink = Mangle symlink to which the directory entry points to
```
```suggestion
 * kLookupRawSymlink = Don't resolve environment variables in symlink targets
```

jblomer (changes requested):
Very nice!  One nit, otherwise I'll put this on hold until the 2.10 branch is created, so this will go into 2.11

HereThereBeDragons:
Added your change request

jblomer:
While this is on hold, could you check the [clang-tidy report](https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsTidy/label=docker-x86_64/448/console) if it is a false positive or not? It's certainly not related to the patch but it may have still have found something.

HereThereBeDragons:
Just for the logs: clang-tidy problems will be handled by @jblomer

PR #3076, opened 2022-10-06
 add support for AWS-SIGv4

mharvey-jt:
This enables CVMFS to talk directly to an S3 endpoint requiring auth.
Add the S3 credentials as
```
CVMFS_SERVER_URL_AWS_CREDENTIALS=access_key:secret_key
CVMFS_EXTERNAL_URL_AWS_CREDENTIALS=access_key:secret_key

CVMFS_SERVER_URL_AWS_REGION=region
CVMFS_EXTERNAL_URL_AWS_REGION=region

```
Naturally, this requires `CVMFS_*_PROXY=DIRECT`
region defaults to `us-east-1`, and is only required for AWS.


jblomer:
Nice!  That update was anyway in the plan for the 2.11 release.  We should check if we can replace the AWSv4 signature code in the S3 fanout manager as well. In the AWSv4 signature scheme, the signing key can be reused for one day.  We should check if this optimization is part of the curl implementation. We should also read once through the libcurl changelog before merging.

mharvey-jt:
Updated `s3fanout.cc`
Looking at that now, it may be that we need to make the could region a configurable, rather than hard-coding `auto`

mharvey-jt:
Added `CVMFS_*_SERVER_AWS_REGION`

jblomer:
In #3093 libcurl has been upgraded to version 7.86.  Could you rebase this patch so that it just includes the AWS-SIGv4 parts?

mharvey-jt:
@cernvm-bot cloudtest

mharvey-jt:
@jblomer rebased as requested

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot tidy

HereThereBeDragons:
@cernvm-bot cpplint

HereThereBeDragons:
@cernvm-bot cpplint

HereThereBeDragons:
@cernvm-bot cloudtest

Issue #2907, opened 2022-04-25
commit processor critical path optimisations


mharvey-jt:
A set of optimisations informed by perf profiling of the commit processor critical path:

https://github.com/cvmfs/cvmfs/pull/2903 : remove ShortString counters
https://github.com/cvmfs/cvmfs/pull/2904 : truncate instead of re-initializing PathString in DiffRec 
https://github.com/cvmfs/cvmfs/pull/2905 : Increase kDefaultMaxName,kDefaultMaxLink
https://github.com/cvmfs/cvmfs/pull/2906 : avoid unnecessary construction of ShortString in MountSubtree
https://github.com/cvmfs/cvmfs/pull/2908 : avoid unnecessary call to StatisticsDatabase::OpenStandardDB()

HereThereBeDragons:
PR #3081 is a replacement for #2903

HereThereBeDragons:
I will edit on the run.

Following parts are done:
- [x] remove ShortString counters --> counters only available in debug
- [x] truncate instead of re-initializing PathString in DiffRec
- [ ] Increase kDefaultMaxName,kDefaultMaxLink
- [x] avoid unnecessary construction of ShortString in MountSubtree
- [ ] avoid unnecessary call to StatisticsDatabase::OpenStandardDB()

PR #2903, opened 2022-04-25
commitprocessor performance: remove ShortString counters

mharvey-jt:
In profiling the commit processor critical path, we observed that a substantial amount of time in `AbstractCatalogManager::MountSubtree`->`ShortString::ShortString`.  Approx half of the time spent in this constructor is in the atomic_inc64() of `num_instances_`.

In our environment, we don't use the `num_instances` or `num_overflows` statistics for `ShortString` derived objects, so this PR removes them.

It might, perhaps, be more elegant to make these counters configurable, so they can be disabled only where required.

```
12.68% 0.76% cvmfs_receiver cvmfs_receiver
11.93% catalog::AbstractCatalogManager<catalog: : Catalog>: :MountSubtree
- 6.31% ShortString<(unsigned char)200, (char)0>::ShortString 2.53% ShortString<(unsigned char) 200, (char)0>:: Assign 2.44% atomic_inc64
0.59% ShortString<(unsigned char)200, (char)0>::GetLength
2.64% ShortString<(unsigned char) 200, (char)0>:: Append
1.11% ShortString<(unsigned char)200, (char)0>::GetLength
1.01% ShortString<(unsigned char)200, (char)0>:: StartsWith
+0.76% libc_start_main
[.] catalog::AbstractCatalogManager<catalog::Catalog>:: MountSubtree
```

PR #3081, opened 2022-10-12
make shortstring counters only available for debug

HereThereBeDragons:
Replacement for #2903; part of issue #2907 

I think `DEBUGMSG` should be the correct variable

Can be checked by removing or adding `CVMFS_DEBUGLOG` to the client config.

With the same actions performed on the repo...
If debuglog is given:
```
sudo cvmfs_talk -i lhcb.cern.ch internal affairs | grep pathstring
pathstring.n_instances|18028|Number of instances
pathstring.n_overflows|0|Number of overflows
```
Otherwise:
```
sudo cvmfs_talk -i lhcb.cern.ch internal affairs | grep pathstring
pathstring.n_instances|0|Number of instances
pathstring.n_overflows|0|Number of overflow
```

Not sure if it is worth to have an explicit test for it?

### DIFF
--- a/cvmfs/shortstring.h
+++ b/cvmfs/shortstring.h
@@ -27,18 +27,26 @@ template<unsigned char StackSize, char Type>
 class ShortString {
  public:
   ShortString() : long_string_(NULL), length_(0) {
+#ifdef DEBUGMSG
     atomic_inc64(&num_instances_);
+#endif
   }
   ShortString(const ShortString &other) : long_string_(NULL) {
+#ifdef DEBUGMSG
     atomic_inc64(&num_instances_);
+#endif
     Assign(other);
   }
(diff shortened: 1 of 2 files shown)


jblomer (commented):
Very nice! We don't need an explicit test for this, I think. Let's just trigger the normal integration tests.

We can think about leaving the static member always in, and also the getters. Just not increment them in non-debug builds. That would save a couple of `#ifdef DEBUGMSG`.

HereThereBeDragons:
`#ifdef DEBUGMSG` removed and tested again.
small side note: `wipecache` does not switch between logging and not logging even though it is activated by `CVMFS_DEBUGLOG` in the client config. instead a full remount is needed.

jblomer:
@cernvm-bot cloudtest

Issue #3015, opened 2022-08-08
cvmfs_statfs() slow


mharvey-jt:
The Pycharm IDE likes to index filesystems. Its behaviour when parallel walking the filesystem is --  for no obvious reason -- to call statfs() + stat() on every entity. It turns out that cvmfs_statfs() is a rather slow operation, and as a result we see cvmfs2 consuming very large amounts of cpu. This causes the latency for other cvmfs filesystem operations to spike

This is arguably incorrect behaviour by Pycharm, but it has revealed a slow path in cvmfs. It seems that cvmfs_statfs() should be quite a quick operation, so perhaps this path could be optimised.

for reference, this is logged with Pycharm as bug https://intellij-support.jetbrains.com/hc/en-us/requests/4214503 and affects at least Pycharm 2022.1 Professional Edition

jblomer:
The time is spent in returning the space and number of files in the cache.  This can of course be cached for a while at the cost of potentially slightly outdated information. It may be a useful protection against rogue applications and monitoring scripts.

PR #3082, opened 2022-10-13
Cache statfs info object 

HereThereBeDragons:
Issue #3015 

Introduction of new client parameter `CVMFS_STATFS_CACHE_TIMEOUT` (using `platform_monotonic_time()`; should therefore be in seconds )

Not sure what is the best way to test it. As far as I could figure out there is no command line tool that calls `statfs()` to quickly test it?

If it works parameter needs to be added to the documentation

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1475,17 +1475,15 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
        QuotaManager::kCapIntrospectSize))
   {
     LogCvmfs(kLogCvmfs, kLogDebug, "QuotaManager does not support statfs");
-    fuse_reply_statfs(req, (file_system_->cache_mgr()->
-                                           quota_mgr()->statfsInfo()));
+    fuse_reply_statfs(req, (mount_point_->statfs_cache()->info()));
     return;
   }
 
   {
-  MutexLockGuard m(file_system_->cache_mgr()->quota_mgr()->statfsLock());
+  MutexLockGuard m(mount_point_->statfs_cache()->lock());
 
-  const uint64_t deadline = *file_system_->cache_mgr()->
-                                          quota_mgr()->statfsCachingDeadline();
-  struct statvfs *info = file_system_->cache_mgr()->quota_mgr()->statfsInfo();
+  const uint64_t deadline = *mount_point_->statfs_cache()->cachingDeadline();
+  struct statvfs *info = mount_point_->statfs_cache()->info();
 
   // cached version still valid
   if ( platform_monotonic_time() < deadline ) {
@@ -1518,8 +1516,8 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   info->f_ffree = info->f_favail = all_inodes - loaded_inode;
   fuse_remounter_->fence()->Leave();
 
-  *file_system_->cache_mgr()->quota_mgr()->statfsCachingDeadline() =
-    platform_monotonic_time() + mount_point_->statfs_time_cache_valid();
+  *mount_point_->statfs_cache()->cachingDeadline() =
+    platform_monotonic_time() + mount_point_->statfs_cache()->timeCacheValid();
 
   fuse_reply_statfs(req, info);
   }
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1470,6 +1470,8 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   // If we return 0 it will cause the fs to be ignored in "df"
   TraceInode(Tracer::kEventStatFs, ino, "statfs()");
 
+  perf::Inc(file_system_->n_fs_statfs());
+
   // Unmanaged cache (no lock needed - statfs is never modified)
   if (!file_system_->cache_mgr()->quota_mgr()->HasCapability(
        QuotaManager::kCapIntrospectSize))
@@ -1482,11 +1484,12 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   {
   MutexLockGuard m(mount_point_->statfs_cache()->lock());
 
-  const uint64_t deadline = *mount_point_->statfs_cache()->cachingDeadline();
+  const uint64_t deadline = *mount_point_->statfs_cache()->caching_deadline();
   struct statvfs *info = mount_point_->statfs_cache()->info();
 
   // cached version still valid
   if ( platform_monotonic_time() < deadline ) {
+    perf::Inc(file_system_->n_fs_statfs_cached());
     fuse_reply_statfs(req, info);
     return;
   }
@@ -1516,8 +1519,9 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   info->f_ffree = info->f_favail = all_inodes - loaded_inode;
   fuse_remounter_->fence()->Leave();
 
-  *mount_point_->statfs_cache()->cachingDeadline() =
-    platform_monotonic_time() + mount_point_->statfs_cache()->timeCacheValid();
+  *mount_point_->statfs_cache()->caching_deadline() =
+      platform_monotonic_time()
+    + mount_point_->statfs_cache()->time_cache_valid();
 
   fuse_reply_statfs(req, info);
   }
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1461,13 +1461,18 @@ static void cvmfs_release(fuse_req_t req, fuse_ino_t ino,
   fuse_reply_err(req, 0);
 }
 
-
+/**
+ * Returns information about a mounted filesystem. In this case it returns
+ * information about the local cache occupancy of cvmfs.
+ * 
+ * Note: If the elements of the struct statvfs *info are set to 0, it will cause
+ *       it to be ignored in commandline tool "df".
+ */
 static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   ino = mount_point_->catalog_mgr()->MangleInode(ino);
   LogCvmfs(kLogCvmfs, kLogDebug, "cvmfs_statfs on inode: %" PRIu64,
            uint64_t(ino));
 
-  // If we return 0 it will cause the fs to be ignored in "df"
   TraceInode(Tracer::kEventStatFs, ino, "statfs()");
 
   perf::Inc(file_system_->n_fs_statfs());
@@ -1481,10 +1486,9 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
     return;
   }
 
-  {
   MutexLockGuard m(mount_point_->statfs_cache()->lock());
 
-  const uint64_t deadline = *mount_point_->statfs_cache()->caching_deadline();
+  const uint64_t deadline = *mount_point_->statfs_cache()->expiry_deadline();
   struct statvfs *info = mount_point_->statfs_cache()->info();
 
   // cached version still valid
@@ -1519,12 +1523,11 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
   info->f_ffree = info->f_favail = all_inodes - loaded_inode;
   fuse_remounter_->fence()->Leave();
 
-  *mount_point_->statfs_cache()->caching_deadline() =
+  *mount_point_->statfs_cache()->expiry_deadline() =
       platform_monotonic_time()
-    + mount_point_->statfs_cache()->time_cache_valid();
+    + mount_point_->statfs_cache()->cache_timeout();
 
   fuse_reply_statfs(req, info);
-  }
 }
 
 #ifdef __APPLE__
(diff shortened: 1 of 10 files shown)


mharvey-jt:
Use `df`: 

```
$ strace -e statfs df /physical/cvmfs/REPO
statfs("/physical/cvmfs/REPO", {f_type=FUSE_SUPER_MAGIC, f_bsize=512, f_blocks=0, f_bfree=0, f_bavail=0, f_files=171282159, f_ffree=171226127, f_fsid={val=[0, 0]}, f_namelen=0, f_frsize=512, f_flags=ST_VALID|ST_RDONLY|ST_NOSUID|ST_NODEV|ST_RELATIME}) = 0
Filesystem     1K-blocks  Used Available Use% Mounted on
cvmfs2                 0     0         0    - /physical/cvmfs/REPO
+++ exited with 0 +++
```

HereThereBeDragons:
thanks matt! that helped a lot to find the problems

this is now a working version!
Verified by hand based on when cache updates would be reflected in the `strace -e statfs df /physical/cvmfs/REPO` output. Tested with 20 sec and 0 sec. (timing does not work in small examples)

new commit:
- fixes deadlock of mutex due to missing init/deinit
- add a log message in line 1477 in `cvmfs.cc` (@jblomer decide if you want it)
- i moved the new variables in the `quota.h` to `private` and exposed them in functions. hope that is the right style.

do we need some test for it or fine without it?

jblomer (changes requested):
Cool!  I think the general logic is fine!

Instead of adding new members to the quota manager, I think we should define a class, say `StatfsCache` in mountpoint.h, which has the members for the `struct statvfs`, the deadline, and the lock.  The `Mountpoint` class would the hold one `StatfsCache` object.

Let's also add an integration test using `df`.

HereThereBeDragons:
@jblomer have a look

jblomer:
@cernvm-bot cloudtest

jblomer:
The statfs integration test seems to fail

cvmfs/mountpoint.h
```diff
@@ -399,6 +404,39 @@ class FileSystem : SingleCopy, public BootFactory {
...
+  }
+  uint64_t *caching_deadline() { return &caching_deadline_; }
+  const uint64_t time_cache_valid() { return time_cache_valid_; }
+  struct statvfs *info() { return &info_; }
+  pthread_mutex_t *lock() { return lock_; }
+
+ private:
+  pthread_mutex_t *lock_;
+  // Timestamp/deadline when the currently cached statvfs info_ becomes invalid
+  uint64_t caching_deadline_;
```
jblomer:
```suggestion
  uint64_t expiry_deadline_;
```

cvmfs/mountpoint.h
```diff
@@ -399,6 +404,39 @@ class FileSystem : SingleCopy, public BootFactory {
...
+  // Time in seconds how long statvfs info_ should be cached
+  uint64_t time_cache_valid_;
```
jblomer:
```suggestion
  uint64_t cache_timeout_;
```

cvmfs/mountpoint.h
```diff
@@ -399,6 +404,39 @@ class FileSystem : SingleCopy, public BootFactory {
   bool has_custom_sqlitevfs_;
 };
 
+/**
+ * The StatfsCache class is a class purely designed as "struct" (= holding
+ * object for all its parameters). 
+ * All its logic, including the locking mechanism, is implemented in the 
+ * function cvmfs_statfs in cvmfs.cc
+ */
+class StatfsCache {
```
jblomer: Let's make sure we cannot copy the class (since it maintains the mutex resource):
```suggestion
class StatfsCache : SingleCopy {
```

cvmfs/mountpoint.cc
```diff
@@ -1783,6 +1792,19 @@ bool MountPoint::SetupBehavior() {
   LogCvmfs(kLogCvmfs, kLogDebug, "kernel caches expire after %d seconds",
            static_cast<int>(kcache_timeout_sec_));
 
+  uint64_t statfs_time_cache_valid = 0;
+  if (options_mgr_->GetValue("CVMFS_STATFS_CACHE_TIMEOUT", &optarg)) {
+    // 0ul does not work because of ubuntu 18
+    if ( optarg.size() > 0 ) {
+    statfs_time_cache_valid =
+      std::max(static_cast<uint64_t>(0),
```
jblomer: I think we don't need it: the second argument is already unsigned (`String2Uint64`), so it must be >=0.

cvmfs/cvmfs.cc
```diff
@@ -1468,45 +1468,63 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
            uint64_t(ino));
 
   // If we return 0 it will cause the fs to be ignored in "df"
```
jblomer: Let's move this comment up, before the function signature. Otherwise one might confuse it with a comment related to `TraceInode`.

cvmfs/cvmfs.cc
```diff
@@ -1468,45 +1468,63 @@ static void cvmfs_statfs(fuse_req_t req, fuse_ino_t ino) {
...
   if (!file_system_->cache_mgr()->quota_mgr()->HasCapability(
        QuotaManager::kCapIntrospectSize))
   {
-    fuse_reply_statfs(req, &info);
+    LogCvmfs(kLogCvmfs, kLogDebug, "QuotaManager does not support statfs");
+    fuse_reply_statfs(req, (mount_point_->statfs_cache()->info()));
+    return;
+  }
+
+  {
```
jblomer: The block can be removed (opening and closing bracelets)

cvmfs/mountpoint.cc
```diff
@@ -1783,6 +1792,19 @@ bool MountPoint::SetupBehavior() {
   LogCvmfs(kLogCvmfs, kLogDebug, "kernel caches expire after %d seconds",
            static_cast<int>(kcache_timeout_sec_));
 
+  uint64_t statfs_time_cache_valid = 0;
+  if (options_mgr_->GetValue("CVMFS_STATFS_CACHE_TIMEOUT", &optarg)) {
+    // 0ul does not work because of ubuntu 18
+    if ( optarg.size() > 0 ) {
```
jblomer: That's a good point! But I'd rather like handle this in `String2Uint64`, which should return 0 for the empty input string. This would leave us in a well-defined situation for the other numeric parameters, too.

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+
+  [ $origSize -ne $newSize2 ] || return 23
+
+  echo "   ... Success"
+}
+
+cvmfs_run_test() {
+  logfile=$1
+
+  # local scratch_dir=$(pwd)
```
jblomer: Let's remove the commented parts

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+  # mkdir $TEST097_PRIVATE_MOUNT
+  # TEST097_CONFIG_FILE="$(mktemp ${TEST097_PRIVATE_MOUNT}.conf.XXXXX)"
+
+  # echo "*** Set a trap for system directory cleanup"
+  # trap cleanup EXIT HUP INT TERM
+
+  check_no_parameter_set || return $?
+  check_parameter_10sec || return $?
+  check_parameter_0sec || return $?
+
```
jblomer, line 108: Let's remove one empty line

test/src/097-statfs/setup_teardown
```diff
@@ -0,0 +1,21 @@
+test097_mount() {
+  local test097_mount=$1
```
jblomer: This parameter is not used.

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
+cvmfs_test_name="Statfs"
+cvmfs_test_autofs_on_startup=false
+cvmfs_test_suites="quick"
+
+source ./src/097-statfs/setup_teardown
+
+TEST097_REPO=lhcb.cern.ch
+TEST097_PRIVATE_MOUNT=/cvmfs/lhcb.cern.ch
```
jblomer: We don't use a "private mountpoint" here
```suggestion
TEST097_MOUNTPOINT=/cvmfs/lhcb.cern.ch
```

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+  echo `strace -e statfs df $TEST097_PRIVATE_MOUNT 2>&1 | grep statfs | awk -F", " '{print $5}' | awk -F"=" '{print $2}'`
+}
+
+check_no_parameter_set() {
+  echo "*** Testing: Parameter CVMFS_STATFS_CACHE_TIMEOUT not set"
+  echo "    wipe cache"
+  test097_mount $TEST097_PRIVATE_MOUNT $TEST097_REPO ""
+  sudo cvmfs_config wipecache > /dev/null
+  echo "   found in config file: `sudo cvmfs_talk parameters | grep "STATFS"`"
+  sleep 1
```
jblomer: We don't need this sleep

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+  local origSize=$(get_cachesize)
+  fillCache
+  sleep 1
```
jblomer: We can remove the `sleep`

test/src/097-statfs/setup_teardown
```diff
@@ -0,0 +1,21 @@
...
+  local repo=$2
+  local params=$3
+
+  cvmfs_mount $repo $params
+}
+
+test097_unmount() {
+  cvmfs_umount $repo $params
```
jblomer: I think `$repo` is unknown here (may need `local repo=$1`) and we probably don't need `$params`

test/src/097-statfs/setup_teardown
```diff
@@ -0,0 +1,21 @@
...
+}
+
+cleanup() {
+  echo "running cleanup()..."
+  if [ "x$TEST097_PIDS" != "x" ]; then
```
jblomer: This block can be removed

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+  echo "   ... Success"
+}
+
+check_parameter_0sec() {
+  echo "*** Testing: Parameter CVMFS_STATFS_CACHE_TIMEOUT=0"
+  echo "    wipe cache"
+  test097_mount $TEST097_PRIVATE_MOUNT $TEST097_REPO "CVMFS_STATFS_CACHE_TIMEOUT=0"
+  sudo cvmfs_config wipecache > /dev/null
+  echo "   found in config file: `sudo cvmfs_talk parameters | grep "STATFS"`"
+  sleep 10
```
jblomer: A zero second timeout should work without any sleeps

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+  echo "   ... Success"
+}
+
+check_parameter_10sec() {
+  echo "*** Testing: Parameter CVMFS_STATFS_CACHE_TIMEOUT=10"
+  echo "    wipe cache"
+  test097_mount $TEST097_PRIVATE_MOUNT $TEST097_REPO "CVMFS_STATFS_CACHE_TIMEOUT=10"
+  sudo cvmfs_config wipecache > /dev/null
+  echo "   found in config file: `sudo cvmfs_talk parameters | grep "STATFS"`"
+  sleep 1
```
jblomer: Let's remove

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+  local origSize=$(get_cachesize)
+  fillCache
+  sleep 1
```
jblomer: Let's remove

test/src/097-statfs/main
```diff
@@ -0,0 +1,110 @@
...
+  [ $origSize -eq $newSize1 ] || return 22
+
+  sleep 8
+  local newSize2=$(get_cachesize)
+  test097_unmount
+
+
+  echo "   should not be equal: $origSize and $newSize2"
+
+  [ $origSize -ne $newSize2 ] || return 23
```
jblomer: At this point, let's also check the counter from internal affairs: `n_statfs_cached` should (only) now be > 0.

jblomer (changes requested):
I think we are coming close!  Some further comments inline.

cvmfs/mountpoint.cc
HereThereBeDragons: have a look in the new commit. if this kinda fits your idea this could also be implemented for `String2Int64` and `String2Uint64Pair` 

Short info about the return value of `sscanf`:
*On success, the function returns the number of items in the argument list successfully filled. This count can match the expected number of items or be less (even zero) in the case of a matching failure.
In the case of an input failure before any data could be successfully interpreted, [EOF](https://cplusplus.com/EOF) is returned.*

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot  cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

test/src/097-statfs/main
```diff
@@ -0,0 +1,113 @@
...
+
+fillCache() {
+  echo "   fill cache for $TEST097_MOUNTPOINT"
+  cat $TEST097_MOUNTPOINT/lib/*.sh > /dev/null
+}
+
+get_cachesize() {
+  #echo `strace -e statfs df $TEST097_MOUNTPOINT 2>&1 | grep statfs | awk -F", " '{print $5}' | awk -F"=" '{print $2}'`
+
+  # no statfs on integration test machine
```
jblomer:
```suggestion
```

jblomer (approved):

Issue #3089, opened 2022-10-25
watchdog process RSS a function of CVMFS_MEMCACHE_SIZE


mharvey-jt:
The watchdog process is forked after the memory cache is allocated, so it ends up with an RSS that's a function of `CVMFS_MEMCACHE_SIZE`, much larger than it needs to be. This should be deallocated and released to the OS on watchdog start.

https://github.com/cvmfs/cvmfs/pull/3092  for example implementation

PR #3092, opened 2022-11-01
free mountpoint object in monitor process

mharvey-jt:
ref https://github.com/cvmfs/cvmfs/issues/3089 (untested on `devel`)


jblomer:
I think we should go differently about this. Instead of destructing in the watchdog (which may have unforeseen consequences, e.g. if destruction of cache manager objects remove files etc.), we should fork the watchdog earlier. We can fork it off before the file system and the mountpoint objects are created in the `Init()` function. I'll submit a corresponding PR.

jblomer:
Replaced by #3219

PR #3093, opened 2022-11-05
Bump built-in libcurl to version 7.86

jblomer:
Partially replaces #3076 

Replacement of the custom AWS-SIGv4 code will be subject to a follow-up PR.


jblomer:
@cernvm-bot cloudtest

Issue #3096, opened 2022-11-08
Expose internal affairs telemetry


HereThereBeDragons:
Inspired by #2939  ; have a uniform interface to allow for different source code plugins to expose internal affairs telemetry to the outside

Idea:
- New `snapshot()` function for statistics.h that returns a (frozen) copy of `std::map<std::string, CounterInfo *> counters_`; -- similar to `string Statistics::PrintList(const PrintOptions print_options)`

- New Class/Thread called `TelemetryAggregator`
    - Thread (loop):
        - sleeps a configurable time, e.g. `CVMFS_TELEMETRY_RATE` (in seconds)
        - calls `statistics::snapshot()`
        - calls (virtual?) function `TelemetryAggregator::pushMetrics()`
    - `TelemetryAggregator::pushMetrics()`
        - virtual? function which is implemented by the specific `TelemetryAggregator`-Flavor (e.g. `TelemetryAggregatorInflux`)
        - prepares data in the right data format
        - sends the data off
   
- "MetricsAggregator" is only started if `CVMFS_TELEMTRY_CAPTURE=ON` is set in the client. `ON` and `OFF` (default) is decided during mountpoint creation

- A default `MetricsAggregator` provided could be `TelemetryAggregatorInflux` (see #2943 )

Questions:
- `pushMetrics(currentCounters)` or `pushMetrics(previousCounters, currentCounters)`
- --> probably the 2nd should be preferred to allow for classes to also send just deltas of the  counters

PR #3098, opened 2022-11-09
Telemetry framework + aggregator for influx

HereThereBeDragons:
Issue #3096 

Provides:
- Framework for source code replacement of telemetry aggregator
- telemetry aggregator sending influx data format over socket/udp

Missing
- [x] Verification that influx data format is correct
- [x] Comments in code
- [x] I guess this should have tests?
- [x] Documentation how to create your own telemetry aggregator ( https://github.com/cvmfs/doc-cvmfs/pull/195 )

Client options to enable:
```
CVMFS_TELEMETRY_SEND=ON  //could be later changed to name of aggregator if multiple are available
CVMFS_TELEMETRY_RATE=5  //send rate in seconds.. is the minimum time interval in between to sends
```

TelemetryAggregatorInflux:
- sends data in influx format
- sends additional to statistics counter also: timestamp and reponame
- sends data as absolute values and delta (seems like it is a bit harder to create the delta on the consuming site)

Parrameters required:
```
CVMFS_INFLUX_HOST=localhost
CVMFS_INFLUX_PORT=8092
CVMFS_INFLUX_METRIC_NAME=cvmfs_client_counters
CVMFS_INFLUX_EXTRA_TAGS=my_custom_tag=24,my_custom_tag2=12  //optional
CVMFS_INFLUX_EXTRA_FIELDS=os_version=8  //optional
```


HereThereBeDragons:
@jblomer maybe you can have a first look if that is more or less what you thought of

cvmfs/catalog_mgr.h
```diff
@@ -89,6 +90,8 @@ struct Statistics {
         "Number of listings of nested catalogs");
     n_detach_siblings = statistics->Register("catalog_mgr.n_detach_siblings",
         "Number of times the CVMFS_CATALOG_WATERMARK was hit");
+    catalog_revision = statistics->Register("catalog_revision",
+                                            "Catalog revision");
```
```suggestion
                                            "Revision number of the root file catalog");
```

cvmfs/catalog_mgr_impl.h
```diff
@@ -669,6 +669,8 @@ uint64_t AbstractCatalogManager<CatalogT>::GetRevision() const {
   ReadLock();
   const uint64_t revision = revision_cache_;
   Unlock();
+  statistics_.catalog_revision->Set(revision);
```
jblomer: I don't think that's the right place to update the counter. We should update the revision when it changes, not when we query it.

cvmfs/cvmfs.cc
```diff
@@ -2102,13 +2102,14 @@ static void Spawn() {
   cvmfs::fuse_remounter_->Spawn();
   if (cvmfs::mount_point_->dentry_tracker()->is_active()) {
     cvmfs::mount_point_->dentry_tracker()->SpawnCleaner(
-      cvmfs::mount_point_->kcache_timeout_sec());  // Usually every minute
+    cvmfs::mount_point_->kcache_timeout_sec());  // Usually every minute
```
jblomer: Not sure that this whitespace change was intended.

cvmfs/mountpoint.h
```diff
@@ -22,6 +22,7 @@
 #include "gtest/gtest_prod.h"
 #include "loader.h"
 #include "magic_xattr.h"
+#include "telemetry_aggregator.h"
```
jblomer: Forward declare the `TelemetryAggregator` instead of the include.

cvmfs/mountpoint.cc
```diff
@@ -1842,6 +1846,25 @@ bool MountPoint::SetupBehavior() {
     }
   }
 
+  // this can be later be changed to switch through different
+  // telemetryAggregators
+  if (options_mgr_->GetValue("CVMFS_TELEMETRY_SEND", &optarg)
+      && options_mgr_->IsOn(optarg)) {
+    uint64_t telemetry_send_rate = 0;
+    if (options_mgr_->GetValue("CVMFS_TELEMETRY_RATE", &optarg)) {
```
jblomer, line 1959: We should either fail if `CVMFS_TELEMETRY_SEND` is set but not `CVMFS_TELEMETRY_RATE`, or we use a built-in default value for `CVMFS_TELEMETRY_RATE`. I think I'd prefer the latter as it makes the code simpler. How about a 5 minutes default?

cvmfs/mountpoint.cc
```diff
@@ -1842,6 +1846,25 @@ bool MountPoint::SetupBehavior() {
...
+      telemetry_send_rate = static_cast<uint64_t>(String2Uint64(optarg));
```
jblomer: We should require a minimum for the rate. E.g. 5 seconds.

cvmfs/mountpoint.cc
```diff
@@ -1842,6 +1846,25 @@ bool MountPoint::SetupBehavior() {
...
+
+      telemetry_aggr_ = (perf::TelemetryAggregator*)
```
jblomer: No casting required here.

cvmfs/mountpoint.cc
```diff
@@ -1842,6 +1846,25 @@ bool MountPoint::SetupBehavior() {
...
+                         perf::TelemetryAggregatorInflux::Create(statistics_,
+                                                    telemetry_send_rate,
+                                                    options_mgr_,
+                                                    fqrn_);
+      LogCvmfs(kLogCvmfs, kLogSyslogWarn | kLogDebug,
```
```suggestion
      LogCvmfs(kLogCvmfs, kLogSyslog | kLogDebug,
```

cvmfs/mountpoint.cc
```diff
@@ -1842,6 +1846,25 @@ bool MountPoint::SetupBehavior() {
...
+               "TELEMETRY: Enable telemetry to report every %d seconds",
```
```suggestion
               "enable telemetry to report every %d seconds",
```

cvmfs/statistics.cc
```diff
@@ -136,10 +136,30 @@ string Statistics::PrintJSON() {
...
+ * CounterInfo struct.
+ * Elements will either be updated or inserted into the map.
+ * 
+ * Note: This function does NOT clear previous elements part of the map.
+*/
+void Statistics::SnapshotCounters(
+                        std::map<std::string, CounterInfo> *counters,
+                        uint64_t *monotonic_clock) {
+  MutexLockGuard lock_guard(lock_);
+  *monotonic_clock = platform_realtime_ns();
```
jblomer: Should it be monotonic time or real time? If monotonic time, please change to. Also, if ns precision is not needed, we can just use `platform_monotonic_time()`, or `time()` for realtime.
```suggestion
  *monotonic_clock = platform_monotonic_time_ns();
```

cvmfs/statistics.h
```diff
@@ -72,20 +99,13 @@ class Statistics {
   std::string LookupDesc(const std::string &name);
   std::string PrintList(const PrintOptions print_options);
   std::string PrintJSON();
+  void SnapshotCounters(std::map<std::string, CounterInfo> *counters,
+                        uint64_t *monotonic_clock);
```
jblomer:
```suggestion
                        uint64_t *timestamp);
```

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,70 @@
...
+#include <cstring>
+#include <map>
+#include <string>
+#include <vector>
+
+#include "options.h"
+#include "statistics.h"
+#include "util/single_copy.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
```
jblomer: We don't need the namespace guard here because telemetry is not part of the loader.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+#include <cstring>
+#include <map>
+#include <string>
+#include <vector>
+
+#include "telemetry_aggregator_influx.h"
+#include "util/logging.h"
+#include "util/posix.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
```
jblomer: We don't need the namespace guard here because telemetry is not part of the loader.

cvmfs/telemetry_aggregator.cc
```diff
@@ -0,0 +1,110 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include <errno.h>
+#include <poll.h>
+#include <unistd.h>
+
+#include "telemetry_aggregator.h"
```
jblomer: Include this first

cvmfs/telemetry_aggregator.cc
```diff
@@ -0,0 +1,110 @@
...
+#include "util/logging.h"
+#include "util/platform.h"
+#include "util/pointer.h"
+#include "util/posix.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
+namespace CVMFS_NAMESPACE_GUARD {
+#endif
+
+namespace perf {
```
jblomer, line 18: In the compilation unit: prefix with `perf::` as needed instead of the namespace scope

cvmfs/statistics.h
jblomer: Actually, we don't need to get the time in `SnapshotCounters` at all. The telemetry class can get the timestamp if it needs to.

cvmfs/statistics.h
```diff
@@ -58,11 +58,38 @@ inline int64_t Xadd(class Counter *counter, const int64_t delta) {
  * a Statistics class have a name and a description.  Thread-safe.
  */
 class Statistics {
+ private:
+  // must be on top here because of CounterInfo
+  struct AtomicCounterInfo {
```
jblomer: For reporting I think we don't need the `refcnt` neither the full functionality of `Counter`. The snapshotted `int64_t` should be sufficient. We also don't need to snapshot the description every time, this can be looked up once by the telemetry class. I think we can simply snapshot into a `std::map<std::string, int64_t>`.

cvmfs/statistics.h
```diff
@@ -72,20 +99,13 @@ class Statistics {
   std::string LookupDesc(const std::string &name);
   std::string PrintList(const PrintOptions print_options);
   std::string PrintJSON();
+  void SnapshotCounters(std::map<std::string, CounterInfo> *counters,
```
jblomer: See above.
```suggestion
  void SnapshotCounters(std::map<std::string, int64_t> *counters,
```

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,70 @@
...
+namespace CVMFS_NAMESPACE_GUARD {
+#endif
+
+namespace perf {
+
+class TelemetryAggregator : SingleCopy {
+ public:
+  static TelemetryAggregator *Create(Statistics* statistics, uint64_t send_rate,
+                                     OptionsManager *options_mgr,
+                                     std::string fqrn);
```
jblomer:
```suggestion
                                     const std::string &fqrn);
```

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,70 @@
...
+  void Spawn();
+
+ protected:
+  Statistics* statistics_;
+  const uint64_t maximum_send_rate_;
+  OptionsManager *options_mgr_;
+  int pipe_terminate_[2];
+  pthread_t thread_telemetry_;
+  std::string fqrn_;
+  bool allow_spawning_;
```
jblomer: I think it's simpler to let the `Create()` factory methods take care of missing options etc.  If something is wrong, return a `NULL` pointer, which has the same effect.

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,70 @@
...
+                      options_mgr_(options_mgr),
+                      fqrn_(fqrn),
+                      allow_spawning_(false),
+                      monotonic_clock_(0) {
+    pipe_terminate_[0] = pipe_terminate_[1] = -1;
+    counters_.clear();
+    memset(&thread_telemetry_, 0, sizeof(thread_telemetry_));
+  }
+
+  virtual void PushMetrics() {}
```
jblomer: Let's make it abstract, we don't have a use case for a no-op `PushMetrics()`. That also let us get rid of the `Create()` factory method for `TelemetryAggregator`. It's enough to have one for the concrete Influx aggregator.

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,70 @@
...
+  TelemetryAggregator(Statistics* statistics, uint64_t maximum_send_rate,
+                      OptionsManager *options_mgr, std::string fqrn) :
+                      statistics_(statistics),
+                      maximum_send_rate_(maximum_send_rate),
+                      options_mgr_(options_mgr),
+                      fqrn_(fqrn),
+                      allow_spawning_(false),
+                      monotonic_clock_(0) {
+    pipe_terminate_[0] = pipe_terminate_[1] = -1;
+    counters_.clear();
```
jblomer: Remove, `std::map` is empty on default construction.

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,70 @@
...
+    memset(&thread_telemetry_, 0, sizeof(thread_telemetry_));
```
jblomer, line 104: I think that can be removed, too.

cvmfs/telemetry_aggregator.cc
```diff
@@ -0,0 +1,110 @@
...
+
+void TelemetryAggregator::Spawn() {
+  if (allow_spawning_ == true) {
+    assert(pipe_terminate_[0] == -1);
+    assert(maximum_send_rate_ > 0);
+    MakePipe(pipe_terminate_);
+    int retval = pthread_create(&thread_telemetry_, NULL, MainTelemetry, this);
+    assert(retval == 0);
+    LogCvmfs(kLogCvmfs, kLogDebug,
+              "TELEMETRY: Spawning of telemetry thread.");
```
jblomer: Instead of prefixing the log messages manually with `TELEMETRY`, we should introduce a new enum constant `kLogTelemetry`, which we use instead of the generic `kLogCvmfs`.

cvmfs/telemetry_aggregator.cc
```diff
@@ -0,0 +1,110 @@
...
+    int retval = poll(&watch_term, 1, timeout_ms);
+    if (retval < 0) {
+      if (errno == EINTR) {  // external interrupt occured - no error for us
+        if (timeout_ms >= 0) {
+          uint64_t now = platform_monotonic_time();
+          timeout_ms = (now > deadline) ? 0 : (deadline - now) * 1000;
+        }
+        continue;
+      }
+      abort();
```
jblomer: Use `PANIC(...)`

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,70 @@
...
+  int pipe_terminate_[2];
+  pthread_t thread_telemetry_;
+  std::string fqrn_;
+  bool allow_spawning_;
+  uint64_t monotonic_clock_;
+  std::map<std::string, Statistics::CounterInfo> counters_;
+
+  static void *MainTelemetry(void *data);
+
+  TelemetryAggregator(Statistics* statistics, uint64_t maximum_send_rate,
```
jblomer: This constructor has a long parameter list.  I would suggest to reduce it to the minimum (statistics) and to use default values for the others, which you can then modify by setters. The options manager is only needed by the Influx derived class, so let's only use it there.

cvmfs/telemetry_aggregator_influx.h
```diff
@@ -0,0 +1,60 @@
...
+#include <map>
+#include <string>
+#include <vector>
+
+#include "options.h"
+#include "statistics.h"
+#include "telemetry_aggregator.h"
+#include "util/single_copy.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
```
jblomer: Namespace guard not needed.

cvmfs/telemetry_aggregator_influx.h
```diff
@@ -0,0 +1,60 @@
...
+  std::map<std::string, Statistics::CounterInfo> old_counters_;
+  std::string influx_host_;
+  int influx_port_;
+  std::string influx_metric_name_;
+  std::string influx_extra_fields_;
+  std::string influx_extra_tags_;
+
+  TelemetryAggregatorInflux(Statistics* statistics,
+                            uint64_t maximum_send_rate,
+                            OptionsManager *options_mgr, std::string fqrn);
```
jblomer:
```suggestion
                            OptionsManager *options_mgr, 
                            std::string fqrn);
```

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+#include <stdint.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+
+#include <cstring>
+#include <map>
+#include <string>
+#include <vector>
+
+#include "telemetry_aggregator_influx.h"
```
jblomer: Include first

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+#include "util/logging.h"
+#include "util/posix.h"
+
+#ifdef CVMFS_NAMESPACE_GUARD
+namespace CVMFS_NAMESPACE_GUARD {
+#endif
+
+namespace perf {
```
jblomer, line 22: Prefix with `perf` as needed

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+
+TelemetryAggregatorInflux *TelemetryAggregatorInflux::Create(
```
jblomer: You would use the factory method (instead of the constructor) to do the options parsing and, in case of errors, return a `NULL` pointer rather than an object.  The idea is that the factory method makes it easier to do complex error handling.  If you do it in the constructor, there is always the problem to destruct everything that has been already set up up to the error. With the factory method, you always have a valid object, although it is not usable until the factory method returns.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+TelemetryAggregatorInflux::~TelemetryAggregatorInflux() {
+  allow_spawning_ = false;
+}
+
+TelemetryAggregatorInflux::TelemetryAggregatorInflux(Statistics* statistics,
+                            uint64_t maximum_send_rate,
+                            OptionsManager *options_mgr, std::string fqrn) :
+                            TelemetryAggregator(statistics, maximum_send_rate,
+                                                options_mgr, fqrn) {
+  old_counters_.clear();
```
jblomer: Not needed.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+
+  int params = 0;
+
+  if (options_mgr_->GetValue("CVMFS_INFLUX_HOST", &influx_host_)) {
```
jblomer: Perhaps also check that the returned value is not empty (which is the case if users set `CVMFS_INFLUX_HOST=`)

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+      params++;
+  }
+
+  std::string opt;
+  if (options_mgr_->GetValue("CVMFS_INFLUX_PORT", &opt)) {
+      influx_port_ = atoi(opt.c_str());
```
jblomer:
```suggestion
      influx_port_ = String2UInt64(opt);
```

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+      if (influx_port_ > 0 && influx_port_ < 65536) {
+        params++;
+      } else {
+        LogCvmfs(kLogCvmfs, kLogDebug, "Invalid value for "
```
jblomer: Try to wrap the function call such that most of the error message is on a single line - makes it easier to grep the code for messages.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+    ret += tok + influx_extra_fields_;
+  }
+
+  snprintf(buf, sizeof(buf), "%lu", monotonic_clock_);
+  ret += " " + std::string(buf);
+
+  return ret;
+}
+
+int TelemetryAggregatorInflux::SendToInflux(std::string payload) {
```
jblomer: Use an enum for the different error cases

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+    if (sockfd < 0)  {
+       LogCvmfs(kLogCvmfs, kLogDebug, "Failed to open socket");
+       freeaddrinfo(res);
+       return 2;
+    }
+
+    s = reinterpret_cast<sockaddr_in*>(res->ai_addr);
+    s->sin_port = htons(port);
+
+    int n = sendto(sockfd, payload.c_str(), strlen(payload.c_str()), 0,
```
jblomer:
```suggestion
    ssize_t n = sendto(sockfd, payload.c_str(), strlen(payload.c_str()), 0,
```

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+                  (struct sockaddr*)s, sizeof(struct sockaddr_in));
+
+    if (n < 0)  {
```
jblomer: Check also for the total number of bytes to catch partial sends

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+      close(sockfd);
+      freeaddrinfo(res);
+      LogCvmfs(kLogCvmfs, kLogDebug, "Failed to send to influx. errno=%d",
+                                                                errno);
+      return 3;
+    }
+
+    close(sockfd);
```
jblomer: Move this one and the following line up before the `if` block.  Prevents the lines from being repeated in the `if` block.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+                                    influx_extra_fields_.c_str());
+  } else {
+    allow_spawning_ = false;
+    LogCvmfs(kLogCvmfs, kLogDebug,
+             "Not enabling influx metrics. Not all variables set");
+  }
+}
+
+std::string TelemetryAggregatorInflux::MakePayload() {
+  char buf[100];
```
jblomer: Use `std::string` to assemble the message.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  snprintf(buf, sizeof(buf), "%lu", monotonic_clock_);
+  ret += " " + std::string(buf);
+
+  return ret;
+}
+
+std::string TelemetryAggregatorInflux::MakeDeltaPayload() {
+  char buf[100];
```
jblomer: --> `std::string`

jblomer (changes requested):
Great work, that fits in very well!

Things can be simplified a bit here and there, see comments.

And, as you mention, tests, comments, docs should come with it.

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: ok moved it to `AttachCatalog()` --> cvmfs/catalog_mgr_impl.h +969 
```
  // The revision of the catalog tree is given by the root catalog revision
  if (catalogs_.empty()) {
    revision_cache_ = new_catalog->GetRevision();
    statistics_.catalog_revision->Set(revision_cache_);
    has_authz_cache_ = new_catalog->GetVOMSAuthz(&authz_cache_);
    volatile_flag_ = new_catalog->volatile_flag();
  }
```

cvmfs/statistics.cc
HereThereBeDragons: no we want realtime. renamed to timestamp to mark when the snapshot of the counters was taken

cvmfs/statistics.h
HereThereBeDragons: i do not recommend that. the processing in the telemetry class might be off. timestamp should present the time when the snapshot of counters was taken

cvmfs/telemetry_aggregator.h
HereThereBeDragons: in all the code samples i could find it was part of it... so not sure..

cvmfs/telemetry_aggregator.h
HereThereBeDragons: i believe it is most likely that any kind of telemetry class will need access to the optionmanager. as such i do not think removing it from the base telemetry class is a wise choice

cvmfs/statistics.h
HereThereBeDragons: "We also don't need to snapshot the description every time" -- we are not. existing elements are overwritten in the map given to `SnapshotCounters`

cvmfs/telemetry_aggregator.h
HereThereBeDragons: i tried to make it full virtual. compiler does not allow because it is called in `MainTelemetry` - but maybe i did something wrong.

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/mountpoint.cc
```diff
@@ -1842,6 +1845,30 @@ bool MountPoint::SetupBehavior() {
     }
   }
 
+  // this can be later be changed to switch through different
+  // telemetryAggregators
+  if (options_mgr_->GetValue("CVMFS_TELEMETRY_SEND", &optarg)
+      && options_mgr_->IsOn(optarg)) {
+    int telemetry_send_rate = 5 * 60;  // default send rate: 5min
```
jblomer: Let's make this a named constant in mountpoint.h, like `kDefaultKCacheTtlSec`.

cvmfs/mountpoint.cc
```diff
@@ -1842,6 +1845,30 @@ bool MountPoint::SetupBehavior() {
...
+    if (options_mgr_->GetValue("CVMFS_TELEMETRY_RATE", &optarg)) {
+      telemetry_send_rate = static_cast<int>(String2Uint64(optarg));
+
+      // minimum send rate: 5sec
+      if (telemetry_send_rate < 5) {
```
jblomer: Should also be a named constant.

cvmfs/statistics.h
```diff
@@ -58,6 +58,18 @@ inline int64_t Xadd(class Counter *counter, const int64_t delta) {
  * a Statistics class have a name and a description.  Thread-safe.
  */
 class Statistics {
+ private:
+  // must be on top here because of CounterInfo
```
jblomer: This can go back to its original place, I think.

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,113 @@
+/**
+ * This file is part of the CernVM File System.
+ *
+ * TelemetryAggregator class is a thread that snapshots the internal counters
```
jblomer:
```suggestion
 * TelemetryAggregator class manages a thread that snapshots the internal counters
```

cvmfs/telemetry_aggregator.h
```diff
@@ -0,0 +1,113 @@
...
+#include "options.h"
+#include "statistics.h"
+#include "util/single_copy.h"
+
+namespace perf {
+
+// Errors of telemetry classes (including custom classes)
+enum TelemetryError {
+  kTelemetrySuccess = 0,
+  kTelemetryHostAddress,
```
jblomer: Same for the ones below, let's indicate that these are error codes.

```suggestion
  kTelemetryFailHostAddress,
```

cvmfs/telemetry_aggregator.cc
```diff
@@ -0,0 +1,117 @@
...
+#include "telemetry_aggregator_influx.h"
+namespace perf {
+
+TelemetryAggregator* TelemetryAggregator::Create(Statistics* statistics,
+                                                 int send_rate,
+                                                 OptionsManager *options_mgr,
+                                                 const std::string &fqrn,
+                                                 const TelemetrySelector type) {
+  UniquePtr<TelemetryAggregatorInflux> telemetryInflux;
+  UniquePtr<TelemetryAggregator> *telemetry;
```
jblomer, line 26: I find having two unique pointers owning the same object confusing. As long as we only have a single concrete implementation, let's just use
```suggestion
  UniquePtr<TelemetryAggregatorInflux> *telemetry;
```

cvmfs/telemetry_aggregator.cc
```diff
@@ -0,0 +1,117 @@
...
+  uint64_t deadline = platform_monotonic_time() + telemetry->maximum_send_rate_;
+  while (true) {
+    // sleep and check if end - blocking wait for "maximum_send_rate_" seconds
+    watch_term.revents = 0;
+    int retval = poll(&watch_term, 1, timeout_ms);
+    if (retval < 0) {
+      if (errno == EINTR) {  // external interrupt occured - no error for us
+        if (timeout_ms >= 0) {
+          uint64_t now = platform_monotonic_time();
+          timeout_ms = (now > deadline) ? 0 :
```
jblomer: Once this happend, `timeout_ms` is never reset to `telemetry->maximum_send_rate_ * 1000`.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,245 @@
...
+#include <vector>
+
+#include "util/logging.h"
+#include "util/posix.h"
+#include "util/string.h"
+
+namespace perf {
+
+TelemetryAggregatorInflux::~TelemetryAggregatorInflux() {
+  is_zombie_ = true;
```
jblomer: That's not necessary, after calling the destructor `is_zombie_` is anyway not accessible anymore (ill-formed to still access it).

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,245 @@
...
+ * ( https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol/ )
+ * Syntax
+   <measurement>[,<tag_key>=<tag_value>[,<tag_key>=<tag_value>]] <field_key>=<field_value>[,<field_key>=<field_value>] [<timestamp>]
+ *
+ * Example
+   myMeasurement,tag1=value1,tag2=value2 fieldKey="fieldValue" 1556813561098000000
+*/
+std::string TelemetryAggregatorInflux::MakePayload() {
+  // measurement and tags
+  std::string ret = "" + influx_metric_name_ + "_absolute,repo=" + fqrn_;
```
jblomer:
```suggestion
  std::string ret = influx_metric_name_ + "_absolute,repo=" + fqrn_;
```

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,245 @@
...
+  std::string tok = ",";
+  for (std::map<std::string, int64_t>::iterator it
+      = counters_.begin(), iEnd = counters_.end(); it != iEnd; it++) {
+    if (it->second != 0) {
+      ret += tok + it->first + "=" + StringifyInt(it->second);
+
+      tok = ",";
+    }
+  }
+  if (influx_extra_fields_ != "") {
```
jblomer, line 133: If we don't have extra fields, the field list ends with a trailing comma. Is this ok?

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,245 @@
...
+  ret += "catalog_revision=" + StringifyUint(revision);
+
+  std::string tok = ",";
+  for (std::map<std::string, int64_t>::iterator it
+      = counters_.begin(), iEnd = counters_.end(); it != iEnd; it++) {
+    int64_t value = it->second;
+    int64_t old_value = old_counters_.at(it->first);
+    if (value != 0) {
+      ret += tok + it->first + "=" + StringifyInt(value - old_value);
+    }
```
jblomer, line 187: Missing the comma field separator, I think.

cvmfs/telemetry_aggregator_influx.cc
```diff
@@ -0,0 +1,245 @@
...
+    ret += tok + influx_extra_fields_;
+  }
+
+  // timestamp
+  ret += " " + StringifyUint(timestamp_);
+
+  return ret;
+}
+
+TelemetryError TelemetryAggregatorInflux::SendToInflux(
```
jblomer: I think that we should determine the host address and open the socket only once.

jblomer (changes requested):
Cool!  Some comments on the implementation, regarding tests I'd suggest
- Unit test for `MakePayload`, `MakeDeltaPayload`
- Integration test with netcat as an influx listener
The tests need to ignore the timestamp value.

cvmfs/telemetry_aggregator_influx.cc
HereThereBeDragons: no it isnt. might have confused you with `tok = ",";` which is unneccesary.
it is `ret += tok + it->first + "=" + StringifyInt(it->second);` -- so comma is only added if there is a field following

cvmfs/telemetry_aggregator.cc
HereThereBeDragons: just looked: where i took the pattern from `glue_buffer.cc` it's correct. just forgot to copy that section

HereThereBeDragons:
@jblomer have a look

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
Documentation PR here: https://github.com/cvmfs/doc-cvmfs/pull/195

HereThereBeDragons:
@jblomer  have a look

test/unittests/t_telemetry_aggregator.cc
```diff
@@ -0,0 +1,266 @@
...
+    ASSERT_EQ(0, result);
+
+    c1_ = statistics_.Register("test.c1", "test counter 1");
+    ASSERT_TRUE(c1_ != NULL);
+    c2_ = statistics_.Register("test.c2", "test counter 2");
+    ASSERT_TRUE(c2_ != NULL);
+    c3_ = statistics_.Register("test.c3", "test counter 3");
+    ASSERT_TRUE(c3_ != NULL);
+
+    options_manager_.ParsePath(config_file_, false);
```
jblomer: For this PR the approach is fine.  In general, we could have used a series `options_manager_->SetValue()` calls and avoid writing a temporary file.

test/unittests/t_telemetry_aggregator.cc
```diff
@@ -0,0 +1,266 @@
...
+  // fields are empty because counters are 0
+
+  std::vector<std::string> payload_split = SplitString(payload, ' ');
+
+  // remove whitespace-only strings from the vector
+  for (std::vector<std::string>::iterator itr = payload_split.begin();
+                                         itr != payload_split.end();
+                                         itr++) {
+    if (Trim(*itr).size() < 1) {
+      itr = payload_split.erase(itr);
```
jblomer: I think you need to increase the iterator in the `else` branch instead of the `for` loop header

test/unittests/t_telemetry_aggregator.cc
```diff
@@ -0,0 +1,266 @@
...
```
jblomer: See above

test/unittests/t_telemetry_aggregator.cc
```diff
@@ -0,0 +1,266 @@
...
+
+
+  string payload_header = "influx_test_absolute,repo=" + fqrn_+ ",test_tag=1";
+  string payload_fields = "test_field=5";
+  // fields are empty because counters are 0
+
+  std::vector<std::string> payload_split = SplitString(payload, ' ');
+
+  // remove whitespace-only strings from the vector
+  for (std::vector<std::string>::iterator itr = payload_split.begin();
```
jblomer: Perhaps make this a method of the test fixture

jblomer (changes requested):
One small issue in the unit tests, then this is good to go.

HereThereBeDragons:
@jblomer have a look

jblomer (approved):

jblomer:
@cernvm-bot cloudtest

jblomer:
@HereThereBeDragons Some issues in the integration tests: netcat is missing on some platforms. You can install it in the `test/cloud_testing/platforms/*setup.sh` scripts. Also, I'd suggest to rebase the PR to the latest devel branch, which should give you a green baseline for the integration tests.

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

test/src/098-telemetry/udp_listener.py
```diff
@@ -0,0 +1,74 @@
...
+  # while time.time() < endtime:
+
+  received_packages = 0
+
+  while received_packages < max_packages:
+  #Now receive data
+    reply = s.recv(8192)
+    if must_include in reply.decode():
+      print(reply.decode())
+      received_packages += 1
```
jblomer: Missing newline at end of file

HereThereBeDragons:
@cernvm-bot cloudtest

PR #3107, opened 2022-11-16
Use cvmfs-libs in cvmfs_talk, cvmfs_fsck

jblomer:
Adds cvmfs-libs as a dependency for the cvmfs client package. This is a careful first PR to increase the use of cvmfs-libs, replacing multiple static compilations of the same utility classes.


jblomer:
@cernvm-bot cloudtest

PR #3108, opened 2022-11-17
Link more binaries against cvmfs-libs

jblomer:
Covers the simple cases: libcvmfs_fuse, shrinkwrap, preloader. Still missing for a follow-up PR

- Static libraries (libcvmfs, libcvmfs_cache): probably best to change to shared libraries
- Unit tests: require cleanup of the CMakeLists file


jblomer:
@cernvm-bot cloudtest

PR #3118, opened 2022-11-30
libcurl - (re)enable http basic auth

mharvey-jt:
Needed to allow basic authentication with HTTP servers


jblomer (changes requested):
Looks good! Let's just make it implicit by adding `--enable-http-proxy` to the list of enabled options at the top

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

PR #3119, opened 2022-11-30
custom sharding hook

mharvey-jt:
This puts in place the ability to add sit-specific proxy selection by replacing `custom_sharding.cc` and setting `CVMFS_CUSTOM_SHARDING=1`


jblomer:
Is this replaced by #3121?

PR #3121, opened 2022-11-30
Custom sharding hook

mharvey-jt:
Add hooks in `download.cc` for site-specific customisation of proxy selection. Enabled with `CVMFS_CUSTOM_SHARDING=1`. Default implementation here simply `PANIC()`s at runtime if activated.


mharvey-jt:
@jblomer have reduced this to a minimal implementation, as discussed

HereThereBeDragons:
Superseeded by #3304

PR #3124, opened 2022-12-05
Remove openssl workaround

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3125, opened 2022-12-05
Fix undefined corner case behavior in external_url xattr

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3127, opened 2022-12-05
Set CVMFS_ARCH env variable in cvmfs client

jblomer:
Makes it possible to use CVMFS_ARCH (== uname -m) in variant symlinks


jblomer:
@cernvm-bot cloudtest

PR #3129, opened 2022-12-06
allow cvmfs_talk to evict chunked files

mharvey-jt:


jblomer:
Looks good to me! We should just add a minimal integration test for `cvmfs_talk evict`, unfortunately there is none, yet. @HereThereBeDragons can you take a look into it.

HereThereBeDragons:
superseded by #3168

PR #3130, opened 2022-12-06
[gw] Store publisher hostname as part of lease

jblomer:
The publisher transmits its own hostname when asking for a new lease, which gets stored in the lease db. When the lease db is queried, available hostnames are returned.

This requires a schema upgrade of the lease db from v2 to v3.  The gateway will do this automatically when opening a v2 db.  Old publishers won't transmit a hostname, in which case the remote IP address from the HTTP request is used as a fallback.  Old leases will not have a hostname set, in which case none is returned in the lease listing.  Old gateways speaking to new publishers will ignore the transmitted hostname.


jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

PR #3131, opened 2022-12-06
Skip AFS in cvmfs_config fuser

### LINKED ISSUE #3090: cvmfs_config fuser runs too long on host with afs
On a system in Wisconsin that uses AFS, the `lsof` command run by `cvmfs_config fuser` takes forever to run and strace showed it doing lots of things under `/afs`.   We found that adding the `-e /afs` option made the lsof instead complete in 30 seconds, on a system that had 159 users.  The option fails with an error if `/afs` is not mounted, so first check to see if it is there before adding the option.

jblomer:
Fixes #3090

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -1292,7 +1292,11 @@ cvmfs_config_fuser() {
   fi
   [ -n "$deviceid" ] || die "Could not obtain device id for $fqrn"
 
-  lsof | awk '{if ($6 == "'"${deviceid/:/,}"'") print $2}' | sort -un | paste -sd ' '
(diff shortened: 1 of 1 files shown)


cvuosalo:
I tested this change on a node where the original version of `cvmfs_config fuser` runs very long, and I confirmed that the change in this merge request allows `cvmfs_config` to complete quickly as it should.

jblomer:
Great, thanks for testing!  I'll merge it then and backport it to the 2.10 branch.

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #3142, opened 2022-12-15
Re-enable core-file generation after credentials drop

mharvey-jt:
We'd like the cvmfs2 process to dump core on assert() if the watchdog process is disabled. Since RHEL systems default to having the sysctl `fs.suid_dumpable==0`, we need to explicitly re-enable dumpability with an appropriate `prctl()` after dropping credentials.
(This is linux-only. I've not idea what the situation on macOS is)


jblomer (approved):

jblomer:
@cernvm-bot cloudtest

Issue #3143, opened 2022-12-15
fuse auto_unmount 


mharvey-jt:
As an alternative to the watchdog process, we could use libfuse3's `auto_unmount` option. 
Initial testing shows it's not working cleanly: 

Without the option:
```
[root@host ~]# /usr/bin/mount -t cvmfs -o libfuse=3,disable_watchdog my.cvmfs.repo /opt/cvmfs/my.cvmfs.repo
CernVM-FS: running with credentials 497:497
CernVM-FS: loading Fuse module... done
CernVM-FS: mounted cvmfs on /opt/cvmfs/my.cvmfs.repo
[root@host ~]# 
```
giving processes: 
```
[root@host ~]# ps -ef |grep cvmfs2
cvmfs     44479      1  0 10:08 ?        00:00:00 /usr/bin/cvmfs2 __cachemgr__ . 8 9 5085593600 2542796800 0 3 -1 :
cvmfs     44483      1  0 10:08 ?        00:00:00 /usr/bin/cvmfs2 __cachemgr__ . 8 9 5085593600 2542796800 0 3 -1 :
cvmfs     44490      1  0 10:08 ?        00:00:00 /usr/bin/cvmfs2 -o rw,libfuse=3,disable_watchdog,system_mount,fsname=cvmfs2,allow_other,grab_mountpoint,uid=497,gid=497 modules.firm.jump.gcp /opt/cvmfs/my.cvmfs.repo
```

But with the option:
```
[root@host~]# /usr/bin/mount -t cvmfs -o libfuse=3,disable_watchdog,auto_unmount my.cvmfs.repo /opt/cvmfs/my.cvmfs.repo
CernVM-FS: running with credentials 497:497
CernVM-FS: loading Fuse module... done
CernVM-FS: mounted cvmfs on /opt/cvmfs/my.cvmfs.repo
<DOES NOT RETURN>
```

with processes

```
 ps -ef |grep cvmfs
root      44932  30660  0 10:10 pts/43   00:00:00 /usr/bin/mount -t cvmfs -o libfuse=3,disable_watchdog,auto_unmount my.cvmfs.repo /opt/cvmfs/my.cvmfs.repo
root      44933  44932  0 10:10 pts/43   00:00:00 /sbin/mount.cvmfs my.cvmfs.repo /opt/cvmfs/my.cvmfs.repo -o rw,libfuse=3,disable_watchdog,auto_unmount
[...]

PR #3144, opened 2022-12-19
add waitpid check in mount.cvmfs.cc

mharvey-jt:
Ref https://github.com/cvmfs/cvmfs/issues/3143
Required when `auto_unmount` option is used. Without this PR, mount.cvmfs gets stuck on the select()

### DIFF
--- a/mount/mount.cvmfs.cc
+++ b/mount/mount.cvmfs.cc
@@ -684,7 +684,7 @@ int main(int argc, char **argv) {
   close(fd_stdout);
   close(fd_stderr);
 
-  if(!ended) {
+  if (!ended) {
     retval = waitpid(pid_cvmfs, &status, 0);
     if (retval == -1) {
       LogCvmfs(kLogCvmfs, kLogStderr, "Failed reading return code");
--- a/mount/mount.cvmfs.cc
+++ b/mount/mount.cvmfs.cc
@@ -680,18 +680,11 @@ int main(int argc, char **argv) {
           return 32;
       }
     }
-    ended = (waitpid(pid_cvmfs, &status, 0) == pid_cvmfs);
+    ended = (waitpid(pid_cvmfs, &status, WNOHANG) == pid_cvmfs);
   } while ((stdout_open || stderr_open) && !ended);
   close(fd_stdout);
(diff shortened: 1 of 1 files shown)


mharvey-jt:
@cernvm-bot unittest

mharvey-jt:
@cernvm-bot unittest linux

mount/mount.cvmfs.cc
```diff
@@ -676,15 +680,17 @@ int main(int argc, char **argv) {
...
+  } while ((stdout_open || stderr_open) && !ended);
   close(fd_stdout);
   close(fd_stderr);
 
-  int status;
-  retval = waitpid(pid_cvmfs, &status, 0);
-  if (retval == -1) {
-    LogCvmfs(kLogCvmfs, kLogStderr, "Failed reading return code");
-    return 32;
+  if (!ended) {
```
jblomer: At this point, we are sure that `ended == true`, aren't we? So we can remove the `if` block.

mount/mount.cvmfs.cc
```diff
@@ -676,15 +680,17 @@ int main(int argc, char **argv) {
           return 32;
       }
     }
-  } while (stdout_open || stderr_open);
+    ended = (waitpid(pid_cvmfs, &status, 0) == pid_cvmfs);
```
jblomer: Shouldn't this use the `WNOHANG` option? Otherwise, there may be a deadlock here: the child process fills the stdout/stderr pipe with content and blocks. But the parent (== mount helper) does not read the data out of the pipe because it's waiting for the child to end.

jblomer (changes requested):

mount/mount.cvmfs.cc
mharvey-jt: correct, that was left over from dev. Removed the block.

mount/mount.cvmfs.cc
mharvey-jt: changed to `WNOHANG`

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

PR #3146, opened 2022-12-21
Add CVMFS_CURL_DEBUGLOG to enable Curl transaction tracing

mharvey-jt:
Useful for logging Curl's transactions, can be used to test features that set particular headers.

### DIFF
--- a/cvmfs/download.cc
+++ b/cvmfs/download.cc
@@ -1663,8 +1663,8 @@ void DownloadManager::Init(const unsigned max_pool_handles,
   curl_debug_file_ = NULL;
 }
 
-void DownloadManager::SetCurlDebug(std::string f) {
-  curl_debug_file_ = fopen(f.c_str(), "w");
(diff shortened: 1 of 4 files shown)


mharvey-jt:
@cernvm-bot unittest

jblomer (changes requested):
In principle looks good to me.  We should just use different files for the different download managers. Otherwise they will double-close the file. Also, `DownloadManager::Clone()` should specifically handle this member (or have a comment that the debug log file won't be cloned).

jblomer:
@mharvey-jt looking once more into this: is it important to separate the curl debug log from the cvmfs debug log? Otherwise, we can use `CURLOPT_DEBUGFUNCTION` to emit cvmfs debug log output. The advantage of that is that log messages from different files don't need to be correlated, all is serialized in a single file. The disadvantage is that the one might need to grep for the relevant curl-related lines, and that the curl debugging is coupled to the cvmfs debug mode.

jblomer:
Replaced by #3329

PR #2926, opened 2022-04-25
replace asserts() with return error

mharvey-jt:
https://github.com/cvmfs/cvmfs/issues/2896

PR #3148, opened 2022-12-22
Add conditional_assert()

mharvey-jt:
An alternative assert function. aborts if test if false unless CVMFS_SUPPRESS_ASSERTS=on, in which case the abort is suppressed.

Useful for situations where there is an assert()ing bug that doesn't yet have a fix but which has a tactical workaround in place.

### DIFF
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1887,6 +1887,13 @@ bool MountPoint::SetupBehavior() {
       return false;
     }
   }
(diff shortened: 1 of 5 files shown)


mharvey-jt:
@cernvm-bot unittest

mharvey-jt:
@cernvm-bot unittest

HereThereBeDragons:
I guess this is a replacement/better solution for #2926 ?

HereThereBeDragons:
Superseded by https://github.com/cvmfs/cvmfs/pull/3169

PR #3153, opened 2023-01-13
Fix several issues found in integration tests

jblomer:


jblomer:
@cernvm-bot cloudtest

Issue #3031, opened 2022-08-22
Improve readability of ClientCatalogManager::LoadCatalog function


HereThereBeDragons:
Function ClientCatalogManager::LoadCatalog needs reworking to improve readability.

Steps could be:
- Create nice readable diagram of the different use cases that need to be covered from current implementation
- Restructure function into smaller subroutines based on the diagram

https://github.com/cvmfs/cvmfs/blob/186cbd51f563daa5f0ac574d000f666cf6468872/cvmfs/catalog_mgr_client.cc#L109

Issue #2932, opened 2022-04-25
double-free from Manifest from race on repo update


mharvey-jt:
Similar to #2929, a race between CheckSynchronously() and polled repo updates can cause this double-free the Manifest class.  Can be fixed with a writelock guard around LoadCatalog() https://github.com/cvmfs/cvmfs/pull/2931

```
Thread 19 (Thread 0x7fffd6ffd700 (LWP 56374)):
#0  0x00007ffff6ec3993 in select () from /lib64/libc.so.6
#1  0x00007ffff45a9b74 in SafeSleepMs (ms=ms@entry=100)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/util/posix.cc:1730
#2  0x00007ffff45813a9 in Watchdog::SendTrace (sig=6, 
    siginfo=<optimized out>, context=<optimized out>)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/monitor.cc:321
#3  <signal handler called>
#4  0x00007ffff6e04377 in raise () from /lib64/libc.so.6
#5  0x00007ffff6e05a68 in abort () from /lib64/libc.so.6
#6  0x00007ffff6e46ec7 in __libc_message () from /lib64/libc.so.6
#7  0x00007ffff6e4f6b9 in _int_free () from /lib64/libc.so.6
#8  0x00007ffff7771b63 in std::basic_string<char, std::char_traits<char>, std::allocator<char> >::~basic_string() () from /lib64/libstdc++.so.6
#9  0x00007ffff4546141 in ~Manifest (this=0x7fffc800b0c0, 
    __in_chrg=<optimized out>)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/manifest.h:41
#10 UniquePtr<manifest::Manifest>::Free (this=this@entry=0x13c32e8)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/util/pointer.h:68
#11 0x00007ffff4545070 in Free (this=0x13c32e8)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/util/pointer.h:50
#12 operator= (ref=0x7fffc800b1c0, this=0x13c32e8)
    at /builddir/build/BUILD/cvmfs-2.8.1/cvmfs/util/pointer.h:38
#13 catalog::ClientCatalogManager::LoadCatalog (this=this@entry=0x13c3150, 
[...]

PR #3156, opened 2023-01-16
Refactoring of file catalog reloads

HereThereBeDragons:
Issue #3031 

- `LoadCatalog` replaced by `GetNewRootCatalogInfo` and `LoadCatalogByHash`
- `Remount(bool)` replaced by `RemountDryrun()` and `Remount()`
- `LoadError` renamed to `LoadReturn`
- `LoadCatalogCas` renamed to `FetchCatalogByHash`
- Added revision to breadcrumb
- Renamed some more variables for better readability
- Added unittest
- Added server integration test

Comments of interest to read are marked with `//TODO(heretherebedragons)`

Note: because i removed unittest/integration test fixes following probably fail but have fixes:
- t_dns HostfileResolverBlankLines
- T_Util.ManagedExecClearEnv
- T_Util.ManagedExecExecuteBinaryDoubleFork
- slowtests of unittest t_catalog_mgr_rw.cc


HereThereBeDragons:
we might need to add some mutexes? see #2932

HereThereBeDragons:
and #2929

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot unittest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/manifest.h
```diff
@@ -16,21 +16,28 @@
...
+                                                          catalog_hash(h),
+                                                          timestamp(t),
+                                                          revision(r) {}
   explicit Breadcrumb(const std::string &from_string);
 
   bool Export(const std::string &fqrn,
               const std::string &directory, const int mode) const;
   std::string ToString() const;
-  bool IsValid() const { return !catalog_hash.IsNull() && (timestamp > 0); }
+  // TODO(heretherebedragons) at a later time also check for valid revision?
```
jblomer: Isn't the code checking already (not `-1ul`)?

cvmfs/manifest.cc
```diff
@@ -27,9 +27,26 @@ Breadcrumb::Breadcrumb(const std::string &from_string) {
     shash::MkFromHexPtr(shash::HexPtr(from_string.substr(0, separator_pos)),
                                       shash::kSuffixCatalog);
 
-  // Get local last modified time
-  if ((from_string[separator_pos] == 'T') && (len > (separator_pos + 1))) {
-    timestamp = String2Uint64(from_string.substr(separator_pos + 1));
+  // check if revision number is included
+  int separator_pos_revision_start = separator_pos;
```
jblomer: `SplitString` for the parsing?

cvmfs/manifest.cc
```diff
@@ -27,9 +27,26 @@ Breadcrumb::Breadcrumb(const std::string &from_string) {
...
+            && (from_string[separator_pos_revision_start] != 'R');
+        ++separator_pos_revision_start)
+  { }
+
+  // Get local last modified time and revision
+  if (separator_pos < len - 1) {
+    if (separator_pos_revision_start + 1 >= len) {
+      timestamp = String2Uint64(from_string.substr(separator_pos + 1));
+      // TODO(heretherebedragons) old format that does not include
+      // the revision number- IS THIS THE BEST SOLUTION???
```
jblomer: It's probably not bad. We should just document what happens if an old breadcrumb is picked up. I guess potentially one more root catalog download than necessary?

cvmfs/cache.proto
```diff
@@ -104,6 +104,7 @@ message MsgBreadcrumb {
   required string fqrn      = 1;
   required MsgHash hash     = 2;
   required uint64 timestamp = 3;
+  required uint64 revision  = 4;
```
jblomer: Perhaps this should be `optional`? We should think about compatibility between new client and old cache plugin.

cvmfs/cache_plugin/libcvmfs_cache.h
```diff
@@ -109,6 +109,7 @@ struct cvmcache_session {
 struct cvmcache_breadcrumb {
   struct cvmcache_hash catalog_hash;
   uint64_t timestamp;
+  uint64_t revision;
```
jblomer, line 114: I guess we should update the library revision, too

cvmfs/libcvmfs_int.cc
```diff
@@ -721,15 +721,16 @@ int LibContext::Close(int fd) {
 
 int LibContext::Remount() {
   LogCvmfs(kLogCvmfs, kLogDebug, "remounting root catalog");
-  catalog::LoadError retval =
-    mount_point_->catalog_mgr()->Remount(true /* dry_run */);
+  catalog::CatalogInfo ctlg_info;
```
jblomer: Seems unused.

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,29 @@ enum LoadError {
   kLoadNumEntries
 };
 
-inline const char *Code2Ascii(const LoadError error) {
+/**
+ * Location of the most recent root catalog.
+ * Used as part of loading the catalog.
+ */
+enum RootCatalogLocation {
```
jblomer, line 64: I think it should be clearer how the information is used / what the purpose of this enum is.

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,29 @@ enum LoadError {
...
+};
+
+/**
+ * Struct to contain all necessary parameters for nested and root catalogs
+ * to load them.
+ */
+struct CatalogInfo {
+  shash::Any hash;                              // mandatory
+  PathString mountpoint;                        // mandatory
+  std::string sql_catalog_handle;               // out parameter
```
```suggestion
  std::string sqlite_path;               // out parameter, path name of the sqlite catalog file
```

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,29 @@ enum LoadError {
...
+  uint64_t root_ctlg_revision;                  // root catalog revision
```
jblomer: Mention if this is unused for nested catalogs (same for the next member)

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,29 @@ enum LoadError {
...
+  kCtlgLocationMounted = 0,      // already loaded in mounted_catalogs_
+  kCtlgLocationServer,
+  kCtlgLocationBreadcrumb
+};
+
+/**
+ * Struct to contain all necessary parameters for nested and root catalogs
+ * to load them.
+ */
+struct CatalogInfo {
```
jblomer: Consider making this a "proper class". E.g., the fact that `hash` and `mountpoint` are mandatory could be enforced by a corresponding constructor.

cvmfs/catalog_mgr.h
```diff
@@ -210,10 +233,8 @@ class AbstractCatalogManager : public SingleCopy {
    * class can decide if it wants to use the hash or the path.
    * Both the input as well as the output hash can be 0.
    */
```
jblomer, line 329: Comment needs adjustment

cvmfs/catalog_mgr_client.h
```diff
@@ -88,6 +92,7 @@ class ClientCatalogManager : public AbstractCatalogManager<Catalog> {
    */
   std::map<PathString, shash::Any> loaded_catalogs_;
   std::map<PathString, shash::Any> mounted_catalogs_;
+  uint64_t mounted_root_ctlg_revision_; /**< invalid value: -1ul */
```
jblomer: Do we need to store this value or can we ask the currently attached root catalog directly when necessary?

cvmfs/catalog_mgr_client.cc
```diff
@@ -86,179 +88,246 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
 
   // Happens only on init/remount, i.e. quota won't delete a cached catalog
-  shash::Any cache_hash(shash::kSha1, shash::kSuffixCatalog);
-  uint64_t cache_last_modified = 0;
+  shash::Any breadcrumb_hash(shash::kSha1, shash::kSuffixCatalog);
+  uint64_t breadcrumb_timestamp = 0;
+  uint64_t breadcrumb_revision = -1ul;
 
+  // TODO(heretherebedragons) how to handle old breadcrumbs that are loaded by
+  // default with revision number 0?
```
jblomer: We may need to introduce a constant like `kUnknownRevision = -2ul`

jblomer: Alternatively: we use the revision if we have it, else we fall back to time stamp comparison.

cvmfs/catalog_mgr_client.cc
```diff
@@ -86,179 +88,246 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+  string alt_root_catalog_path = "";
+
+  // root catalog needs special handling because of alt_root_catalog_path
+  CachedManifestEnsemble ensemble(fetcher_->cache_mgr(), this);
+  if (ctlg_info->mountpoint.IsEmpty()) {
+    if (fixed_alt_root_catalog_) {
+      alt_root_catalog_path = ctlg_info->hash.MakeAlternativePath();
+    }
+
+    // TODO(heretherebedragons) is this test necessary??
```
jblomer: I don't see that this is necessary, I think we should drop it.

cvmfs/catalog_mgr_client.cc
```diff
@@ -86,179 +88,246 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+  // TODO(heretherebedragons) could help: fetch should return if fetch from
+  // cache or from remote would save us the if in L283
+  LoadReturn load_ret = FetchCatalogByHash(ctlg_info->hash, catalog_descr,
+                                           alt_root_catalog_path,
+                                           &ctlg_info->sql_catalog_handle);
+  // *catalog_hash = ctlg_info->hash;
+  if (load_ret == catalog::kLoadNew) {
+    loaded_catalogs_[ctlg_info->mountpoint] = ctlg_info->hash;
+
+    if (ctlg_info->mountpoint.IsEmpty()) {  // root catalog
```
jblomer: `CatalogInfo` should get a helper method `bool IsRootCatalog() const`

test/src/703-catalog_mgr_client/setup_teardown
```diff
@@ -0,0 +1,57 @@
...
+cleanup() {
+  echo "running cleanup()..."
+  if [ "x$TEST703_PIDS" != "x" ]; then
+    sudo kill -9 $TEST703_PIDS
+  fi
+  if [ "x$TEST703_PRIVATE_MOUNT" != "x" ]; then
+    private_unmount
+  fi
+  sudo rm -rf /tmp/cvmfs-test/workdir/703-catalog_mgr_client
+}
```
jblomer: Missing newline at the end of the file

test/unittests/t_catalog_mgr.cc
```diff
@@ -324,9 +324,12 @@ TEST_F(T_CatalogManager, Balance) {
 
 TEST_F(T_CatalogManager, Remount) {
   EXPECT_TRUE(catalog_mgr_.Init());
-  LoadError le;
-  EXPECT_EQ(kLoadNew, le = catalog_mgr_.Remount(true));
-  EXPECT_EQ(kLoadNew, catalog_mgr_.Remount(false));
+  LoadReturn le;
+  // EXPECT_EQ(kLoadNew, le = catalog_mgr_.Remount(true));
+  // EXPECT_EQ(kLoadNew, catalog_mgr_.Remount(false));
```
jblomer, line 329: Hm, why was it returning `kLoadNew` before?

jblomer (changes requested):
Very nice!  I took a first pass at this and this *much* better than what we currently have!

Some comments inline.  Let me know how you want to proceed to get this merged.

cvmfs/manifest.h
HereThereBeDragons: yes as in the end i decided that breadcrumbs that have no `revision` will get `revision = 0` - not sure if this is the best solution

cvmfs/manifest.cc
HereThereBeDragons: yes that would be most likely the result as the revision number is set to `0`

cvmfs/cache.proto
HereThereBeDragons: ok changed it to `optional`. not sure where the proto stuff is exactly used and if a remount/mount would automatically write a new breadcrumb ( in the new format)

cvmfs/catalog_mgr.h
HereThereBeDragons: it's called `root_ctlg` - is that not hint enough?

cvmfs/catalog_mgr_client.h
HereThereBeDragons: i tried `GetRevision()` but this results in a deadlock, see
```
uint64_t Catalog::GetRevision() const {
  MutexLockGuard m(lock_);
  return database().GetPropertyDefault<uint64_t>("revision", 0);
}
```
if there is any better solution than having `mounted_root_ctlg_revision_` i would totally go for it

cvmfs/cache_plugin/libcvmfs_cache.h
HereThereBeDragons: where?

test/unittests/t_catalog_mgr.cc
HereThereBeDragons: because the return value of `LoadCatalogCas` is propagated through mutliple functions and that can only have the values `kLoadNew`, `kLoadNoSpace` and `kLoadFail` (the underlying `Fetch()` does not know where exactly it fetched from)

cvmfs/catalog_mgr_client.h
HereThereBeDragons: make private version of `getrevision` without lock

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: i just realized if we do this then we `CatalogInfo` needs to keep a copy of the manifest when `GetNewRootCatalogInfo` is being called

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
missing: how to handle an old breadcrumb that does not contain yet a revision

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cpplint

HereThereBeDragons:
@cernvm-bot unittest

HereThereBeDragons:
@cernvm-bot cpplint

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot help

HereThereBeDragons:
@cernvm-bot cloudtest cc8

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cpplint

jblomer:
@cernvm-bot tidy

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot  cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/cache_plugin/libcvmfs_cache.h
jblomer: At the beginning of the file

cvmfs/cache.proto
```diff
@@ -104,6 +104,7 @@ message MsgBreadcrumb {
   required string fqrn      = 1;
   required MsgHash hash     = 2;
   required uint64 timestamp = 3;
+  optional uint64 revision  = 4;
```
jblomer, line 108: Add changelog entry at the beginning of the file

cvmfs/cache_plugin/channel.cc
```diff
@@ -215,6 +215,7 @@ void CachePlugin::HandleBreadcrumbStore(
     msg_reply.set_status(cvmfs::STATUS_MALFORMED);
   } else {
     breadcrumb.timestamp = msg_req->breadcrumb().timestamp();
+    breadcrumb.revision = msg_req->breadcrumb().revision();
```
jblomer: We don't necessarily have a revision in the breadcrumb message (it is declared as optional). Add a check and handle as in other cases of missing revision.

cvmfs/cache_extern.cc
```diff
@@ -701,6 +701,7 @@ manifest::Breadcrumb ExternalCacheManager::LoadBreadcrumb(
     assert(rv);
     breadcrumb.catalog_hash.suffix = shash::kSuffixCatalog;
     breadcrumb.timestamp = msg_reply->breadcrumb().timestamp();
+    breadcrumb.revision = msg_reply->breadcrumb().revision();
```
jblomer: We may not have a revision (optional member). Add a check.

cvmfs/manifest.h
```diff
@@ -16,21 +16,28 @@
...
 /**
- * The breadcrumb stores the catalog root hash and a time stamp.  It is used
- * to store the last known copy of the catalog in the cache.
+ * The breadcrumb stores the catalog root hash, its revision and a time stamp.
+ * It is used to store the last known copy of the catalog in the cache.
  */
 struct Breadcrumb {
-  Breadcrumb() : catalog_hash(), timestamp(0) {}
-  Breadcrumb(const shash::Any &h, uint64_t t) : catalog_hash(h), timestamp(t) {}
+  Breadcrumb() : catalog_hash(), timestamp(0), revision(-1ul) {}
```
jblomer: Make `-1ul` a named static constant, i.e.
```suggestion
  static const uint64_t kInvalidRevision = -1ul;
  Breadcrumb() : catalog_hash(), timestamp(0), revision(kInvalidRevision) {}
```

cvmfs/manifest.h
```diff
@@ -16,21 +16,28 @@
...
   explicit Breadcrumb(const std::string &from_string);
 
   bool Export(const std::string &fqrn,
               const std::string &directory, const int mode) const;
   std::string ToString() const;
-  bool IsValid() const { return !catalog_hash.IsNull() && (timestamp > 0); }
+  // TODO(heretherebedragons) at a later time also check for valid revision?
+  bool IsValid() const { return !catalog_hash.IsNull()
+                                && (timestamp > 0)
+                                && (revision != -1ul); }
```
jblomer:
```suggestion
                                && (revision != kInvalidRevision); }
```

cvmfs/manifest.cc
```diff
@@ -6,30 +6,44 @@
...
+    if (vec_split_timestamp.size() > 1) {
+      // check if revision number is included
+      std::vector<std::string> vec_split_revision =
+                                       SplitString(vec_split_timestamp[1], 'R');
+
+      // Get local last modified time and revision
+      if (vec_split_revision.size() == 1) {
+          timestamp = String2Uint64(vec_split_revision[0]);
+          // TODO(heretherebedragons) old format that does not include
+          // the revision number- IS THIS THE BEST SOLUTION???
```
jblomer:
```suggestion
          // No revision found, i.e. the breadcrumb was stored by an older cvmfs version.
          // We set the revision to 0 as a workaround.
```

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,113 @@ enum LoadError {
...
+  kCtlogLocationUnknown = 0,
+  kCtlgLocationMounted,      // already loaded in mounted_catalogs_
+  kCtlgLocationServer,
+  kCtlgLocationBreadcrumb
+};
+
+/**
+ * CatalogContext class contains all necessary information to load a catalog and
+ * also keeps track of the resulting output.
+ * It works the following:
```
```suggestion
 * It works as follows:
```

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,113 @@ enum LoadError {
...
+              mountpoint_(mountpoint),
+              sqlite_path_(""),
+              root_ctlg_revision_(-1ul),
+              root_ctlg_location_(kCtlogLocationUnknown),
+              manifest_ensemble_(NULL) { }
+
+  CatalogContext(shash::Any hash, const PathString &mountpoint,
+              RootCatalogLocation location) :
+              hash_(hash),
+              mountpoint_(mountpoint),
```
jblomer: Wouldn't the mountpoint always be empty?
```suggestion
  CatalogContext(const shash::Any &hash, RootCatalogLocation location) :
              hash_(hash),
              mountpoint_(),
```

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,113 @@ enum LoadError {
...
+      - Nested catalog: CatalogContext(shash::Any hash, PathString mountpoint)
+      - Note: in this case root_ctlg_revision is not used
+ *  - Call LoadCatalogByHash()
+      - This will set: sqlite_path
+ */
+struct CatalogContext {
+ public:
+  CatalogContext() :
+              hash_(shash::Any()),
+              mountpoint_(PathString("invalid", 7)),
```
```suggestion
              mountpoint_(),
```

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,113 @@ enum LoadError {
...
+  void SetMountpoint(const PathString &mountpoint) { mountpoint_ = mountpoint; }
+  void SetSqlitePath(const std::string &sqlite_path)
+                                                 { sqlite_path_ = sqlite_path; }
+  void SetRootCtlgRevision(uint64_t root_ctlg_revision)
+                                   { root_ctlg_revision_ = root_ctlg_revision; }
+  void SetRootCtlgLocation(RootCatalogLocation root_ctlg_location)
+                                   { root_ctlg_location_ = root_ctlg_location; }
+  /**
+   * Gives ownership to CatalogContext
+  */
```
```suggestion
  /**
   * Gives ownership to CatalogContext
   */
```

cvmfs/catalog_mgr.h
```diff
@@ -53,7 +53,113 @@ enum LoadError {
...
+struct CatalogContext {
+ public:
+  CatalogContext() :
+              hash_(shash::Any()),
+              mountpoint_(PathString("invalid", 7)),
+              sqlite_path_(""),
+              root_ctlg_revision_(-1ul),
+              root_ctlg_location_(kCtlogLocationUnknown),
+              manifest_ensemble_(NULL) { }
+  CatalogContext(shash::Any hash, const PathString &mountpoint) :
```
```suggestion
  CatalogContext(const shash::Any &hash, const PathString &mountpoint) :
```

cvmfs/catalog_mgr.h
```diff
@@ -209,14 +317,18 @@ class AbstractCatalogManager : public SingleCopy {
...
   /**
-   * Load the catalog and return a file name and the catalog hash. Derived
-   * class can decide if it wants to use the hash or the path.
-   * Both the input as well as the output hash can be 0.
+   * Load the catalog and return a file name and the catalog hash.
+   *
+   * GetNewRootCatalogContext() populates CatalogContext object with the
+   * information needed to retrieve the most recent root catalog independent of
+   * its location.
+   * CatalogContext object must be populated with at least hash and mountpoint
```
```suggestion
   * The CatalogContext object must be populated with at least hash and mountpoint
```

cvmfs/catalog_mgr_impl.h
```diff
@@ -115,28 +115,41 @@ bool AbstractCatalogManager<CatalogT>::Init() {
...
+  CatalogContext ctlg_context;
+  return GetNewRootCatalogContext(&ctlg_context);
+}
+
+template <class CatalogT>
+LoadReturn AbstractCatalogManager<CatalogT>::Remount() {
+  LogCvmfs(kLogCatalog, kLogDebug, "remounting repositories");
+  CatalogContext ctlg_context;
+
+  // TODO(heretherebedragons) Is this necessary or can we move it outside?
```
jblomer: Let's leave it as is for the time being. We will come back here for background loading of catalogs.

cvmfs/catalog_mgr_impl.h
```diff
@@ -153,28 +166,32 @@ LoadError AbstractCatalogManager<CatalogT>::Remount(const bool dry_run) {
...
-                                           &catalog_path,
-                                           &catalog_hash);
-  if (load_error == kLoadNew) {
+  CatalogContext ctlg_context(root_hash, PathString("", 0),
+                                                          kCtlgLocationMounted);
+  // we do not need to set revision as LoadCatalogByHash
+  // needs only mountpoint, hash and root_ctlg_location
+
+  // TODO(heretherebedragons) HERE IS A PROBLEM: THE CATALOG REVISION IS NOT
+  // SET CORRECTLY. WOULD NEED SPECIAL TREATMENT
```
jblomer: Does it mean that this code path fails?

cvmfs/catalog_mgr.h
```diff
@@ -176,6 +283,7 @@ class AbstractCatalogManager : public SingleCopy {
   }
   bool volatile_flag() const { return volatile_flag_; }
   uint64_t GetRevision() const;
+  uint64_t GetRevisionNoLock() const;
```
jblomer: Should be protected

cvmfs/catalog_mgr_impl.h
```diff
@@ -673,6 +690,15 @@ uint64_t AbstractCatalogManager<CatalogT>::GetRevision() const {
   return revision;
 }
 
+/**
+ * Like GetRevision() only without any locking mechanism.
+ * As such should only be used in conditions where a lock was already taken
+ * and calling GetRevision() would otherwise result in a deadlock.
+ */
+template <class CatalogT>
+uint64_t AbstractCatalogManager<CatalogT>::GetRevisionNoLock() const {
```
jblomer, line 696: The `GetRevision()` method should reuse this one.

cvmfs/catalog_mgr_impl.h
```diff
@@ -879,23 +905,34 @@ CatalogT *AbstractCatalogManager<CatalogT>::MountCatalog(
                                               CatalogT *parent_catalog)
 {
   CatalogT *attached_catalog = NULL;
-  if (IsAttached(mountpoint, &attached_catalog))
+  if (IsAttached(mountpoint, &attached_catalog)) {
     return attached_catalog;
+  }
+
+  CatalogContext ctlg_context(hash, mountpoint, kCtlgLocationMounted);
```
jblomer, line 928: Here we seem to use a constructor that is documented as "for root catalog only" for both nested and root catalog.

My suggestion is to distinguish based on the MountCatalog input parameters whether it is a root catalog or not and then to use the CatalogContext object differently depending on the case.

jblomer: Second question: why do we set the location to "newest revision already mounted" and not to "I don't know where the newest revision is"?

cvmfs/catalog_mgr_impl.h
```diff
@@ -919,15 +956,20 @@ CatalogT *AbstractCatalogManager<CatalogT>::LoadFreeCatalog(
                                             const PathString     &mountpoint,
                                             const shash::Any     &hash)
 {
-  string new_path;
-  shash::Any check_hash;
-  const LoadError load_error = LoadCatalog(mountpoint, hash, &new_path,
-                                           &check_hash);
-  if (load_error != kLoadNew)
+  CatalogContext ctlg_context(hash, mountpoint, kCtlgLocationMounted);
```
jblomer: I would have naturally assumed that we set the last parameter to "I don't know where the newest revision is"

cvmfs/catalog_mgr_impl.h
```diff
@@ -919,15 +956,20 @@ CatalogT *AbstractCatalogManager<CatalogT>::LoadFreeCatalog(
...
+
+
+  const LoadReturn load_ret = LoadCatalogByHash(&ctlg_context);
+
+  // TODO(heretherebedragons) correct if statement?
+  if (load_ret != kLoadNew && load_ret != kLoadUp2Date) {
```
jblomer: Not sure. Does it work when you change back to the original if statement?

cvmfs/catalog_mgr_impl.h
```diff
@@ -919,15 +956,20 @@ CatalogT *AbstractCatalogManager<CatalogT>::LoadFreeCatalog(
...
     return NULL;
-  assert(hash == check_hash);
+  }
+
+  assert(hash == ctlg_context.hash());  // TODO(heretherebedragons) why?
```
jblomer: Why not? If you think it's obvious / not needed anymore, please remove.
```suggestion
  assert(hash == ctlg_context.hash());  // TODO(heretherebedragons) why?
```

cvmfs/catalog_mgr_client.h
```diff
@@ -68,18 +74,16 @@ class ClientCatalogManager : public AbstractCatalogManager<Catalog> {
   int root_fd() const { return root_fd_; }
 
  protected:
-  LoadError LoadCatalog(const PathString  &mountpoint,
-                        const shash::Any  &hash,
-                        std::string       *catalog_path,
-                        shash::Any        *catalog_hash);
+  LoadReturn GetNewRootCatalogContext(CatalogContext *result);
+  LoadReturn LoadCatalogByHash(CatalogContext* ctlg_context);
```
jblomer:
```suggestion
  virtual LoadReturn GetNewRootCatalogContext(CatalogContext *result);
  virtual LoadReturn LoadCatalogByHash(CatalogContext *ctlg_context);
```

cvmfs/libcvmfs_int.cc
```diff
@@ -721,17 +721,17 @@ int LibContext::Close(int fd) {
...
+
   switch (retval) {
     case catalog::kLoadUp2Date:
       LogCvmfs(kLogCvmfs, kLogDebug, "catalog up to date");
       return 0;
 
     case catalog::kLoadNew:
-      retval = mount_point_->catalog_mgr()->Remount(false /* dry_run */);
-      if (retval != catalog::kLoadNew)
-        return -1;
```
jblomer: Why can we ignore the return value here?

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
-    return load_error;
-  }
+  // 1) Get alien cache root catalog (local)
 
   // Happens only on init/remount, i.e. quota won't delete a cached catalog
-  shash::Any cache_hash(shash::kSha1, shash::kSuffixCatalog);
-  uint64_t cache_last_modified = 0;
+  shash::Any breadcrumb_hash(shash::kSha1, shash::kSuffixCatalog);
+  uint64_t breadcrumb_timestamp = 0;
+  uint64_t breadcrumb_revision = -1ul;
```
jblomer: The breadcrumb can be invalid and then it is strange to continue using these variables.

My suggestion is to rename them to `local_newest_...`, as you introduce later for hash and timestamp only. Let's have them for all three, hash, timestamp, revision. After step two, these three as set consistenty to either

- Information from the breadcrumb
- Information from the currently mounted catalog
- NULL, 0, -1, if we have neither

This should also make it in a comment for the three variables.

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+  shash::Any local_newest_hash = breadcrumb_hash;
+  uint64_t local_newest_revision = breadcrumb_revision;
+  result->SetRootCtlgLocation(kCtlgLocationBreadcrumb);
+  LoadReturn success_code = catalog::kLoadNew;
+
+  // We only look for currently loaded catalog if the revision is newer than
+  // the breadcrumb revision and both revision numbers are valid (!= -1ul).
+  if (breadcrumb_revision <= GetRevisionNoLock()
+     || breadcrumb_revision == -1ul) {
+    const std::map<PathString, shash::Any>::iterator curr_hash_itr =
```
jblomer: We may not find a currently mounted catalog when this is run for the first time on mounting and we have a breadcrumb without revision (i.e. both revisions return 0). We need to check this case.

Also add a test: e.g. mount normally, unmount, manipulate the breadcrumb and remove the revision, mount again.

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+  // 2) Select local newest catalog: mounted vs alien
+
+  shash::Any local_newest_hash = breadcrumb_hash;
+  uint64_t local_newest_revision = breadcrumb_revision;
+  result->SetRootCtlgLocation(kCtlgLocationBreadcrumb);
+  LoadReturn success_code = catalog::kLoadNew;
+
+  // We only look for currently loaded catalog if the revision is newer than
+  // the breadcrumb revision and both revision numbers are valid (!= -1ul).
+  if (breadcrumb_revision <= GetRevisionNoLock()
```
jblomer: If the breadcrumb revision is zero, we should compare timestamps as a fallback.

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
-  CachedManifestEnsemble ensemble(fetcher_->cache_mgr(), this);
-  manifest_failure = manifest::Fetch("", repo_name_, cache_last_modified,
-                                     &cache_hash, signature_mgr_,
+  UniquePtr<CachedManifestEnsemble> ensemble(
+                       new CachedManifestEnsemble(fetcher_->cache_mgr(), this));
+  // TODO(heretherebedragons)
+  // breadcrumb_timestamp: does breadcrumb timestamp and revision always go hand
+  // in hand?
+  // can it be that the timestamp is older on the server but the revision higher
+  // than the breadcrumb?
```
jblomer: No, that should not be possible.

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
-        LoadCatalogCas(cache_hash, cvmfs_path, "", catalog_path);
-      if (success_code != catalog::kLoadNew)
-        return success_code;
-      loaded_catalogs_[mountpoint] = cache_hash;
+    result->TakeManifestEnsemble(
+                  static_cast<manifest::ManifestEnsemble*>(ensemble.Release()));
+    offline_mode_ = false;
+
+    return catalog::kLoadNew;
     }
```
jblomer: Indentation broken

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
-    if (iter != mounted_catalogs_.end()) {
-      if (breadcrumb.IsValid() && (iter->second != cache_hash)) {
-        success_code = catalog::kLoadNew;
-      }
+                                     ensemble.weak_ref());
+
+  if (manifest_failure == manifest::kFailOk) {
+    // fixed catalog: just check if server is reachable but use local version
+    // TODO(heretherebedragons) OK that this could be in theory the breadcrumb?
+    if (fixed_root_catalog_ != shash::Any()) {
```
jblomer:
```suggestion
    if (!fixed_root_catalog_.IsNull()) {
```

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
-      mounted_catalogs_.find(mountpoint);
-    if (iter != mounted_catalogs_.end()) {
-      if (breadcrumb.IsValid() && (iter->second != cache_hash)) {
-        success_code = catalog::kLoadNew;
-      }
+                                     ensemble.weak_ref());
+
+  if (manifest_failure == manifest::kFailOk) {
+    // fixed catalog: just check if server is reachable but use local version
+    // TODO(heretherebedragons) OK that this could be in theory the breadcrumb?
```
jblomer: What happens if we mount a fixed catalog for the first time, i.e. if there is no local copy yet? Do we call this method at all if we have a fixed catalog?

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+  }
+  LogCvmfs(kLogCache, kLogDebug,
+            "failed fetch manifest from server: "
+            "manifest too old or server unreachable (%d - %s)",
+            manifest_failure, manifest::Code2Ascii(manifest_failure));
+
+  // corner case: server not reachable, local revision is 0. We can not
+  // distinguish if local revision number is valid or invalid. Most likely
+  // it is invalid - so fail
+  if (local_newest_revision == 0 && !manifest_failure == manifest::kFailOk) {
```
jblomer:
```suggestion
  if (local_newest_revision == 0 && (manifest_failure != manifest::kFailOk)) {
```

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
-      if (iter->second != cache_hash) {
-        LogCvmfs(kLogCache, kLogDebug, "updating from %s to alien cache copy",
-                 iter->second.ToString().c_str());
-        success_code = catalog::kLoadNew;
+/**
+ * Loads (and fetches) a catalog by hash for a given mountpoint.
+ *
+ * Special case for root catalog: ctlg_context->root_ctlg_location must be given.
+ * If it is kCtlgLocationServer than another check for the most recent
+ * manifest is done.
```
jblomer: This doesn't seem to be the case (additional check for the manifest). It's probably OK but then the comment should be adjusted.

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+
+  // TODO(heretherebedragons) could help: fetch should return if fetch from
+  // cache or from remote would save us the if in L283
+  const LoadReturn load_ret = FetchCatalogByHash(ctlg_context->hash(),
+                                           catalog_descr, alt_root_catalog_path,
+                                           ctlg_context->GetSqlitePathPtr());
+  if (load_ret == catalog::kLoadNew) {
+    loaded_catalogs_[ctlg_context->mountpoint()] = ctlg_context->hash();
+
+    if (ctlg_context->IsRootCatalog()) {  // root catalog
```
jblomer:
```suggestion
    if (ctlg_context->IsRootCatalog()) {
```

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+  // root catalog needs special handling because of alt_root_catalog_path
+  if (ctlg_context->IsRootCatalog() && fixed_alt_root_catalog_) {
+    alt_root_catalog_path = ctlg_context->hash().MakeAlternativePath();
+  }
+
+  // TODO(heretherebedragons) could help: fetch should return if fetch from
+  // cache or from remote would save us the if in L283
+  const LoadReturn load_ret = FetchCatalogByHash(ctlg_context->hash(),
+                                           catalog_descr, alt_root_catalog_path,
+                                           ctlg_context->GetSqlitePathPtr());
```
jblomer: I don't really understand the comment. I wonder though, if the context says "use the locally mounted catalog", shouldn't we avoid calling `FetchCatalogByHash`?

cvmfs/catalog_mgr_ro.cc
```diff
@@ -14,43 +14,56 @@ using namespace std;  // NOLINT
...
+ * 
+ * @return kLoadNew on success
  */
-LoadError SimpleCatalogManager::LoadCatalog(const PathString  &mountpoint,
-                                            const shash::Any  &hash,
-                                            std::string       *catalog_path,
-                                            shash::Any        *catalog_hash)
-{
-  shash::Any effective_hash = hash.IsNull() ? base_hash_ : hash;
+// TODO(herethebedragons) CORRECT?
```
jblomer:
```suggestion
```

cvmfs/catalog_mgr_ro.cc
```diff
@@ -14,43 +14,56 @@ using namespace std;  // NOLINT
...
+  const shash::Any effective_hash = ctlg_context->hash();
   assert(shash::kSuffixCatalog == effective_hash.suffix);
   const string url = stratum0_ + "/data/" + effective_hash.MakePath();
 
-  FILE *fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w",
-                                  catalog_path);
+  std::string tmp;
+
+  FILE *fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w", &tmp);
+  ctlg_context->GetSqlitePathPtr()->assign(tmp);
```
jblomer: Also: let's move to after the `if` block
```suggestion
  ctlg_context->SetSqlitePath(tmp);
```

cvmfs/catalog_mgr_ro.cc
```diff
@@ -14,43 +14,56 @@ using namespace std;  // NOLINT
...
-  download::Failures retval = download_manager_->Fetch(&download_catalog);
-  fclose(fcatalog);
+cvmfs::FileSink filesink(fcatalog);
+download::JobInfo download_catalog(&url, true, false,
+                                  &effective_hash, &filesink);
+const download::Failures retval = download_manager_->Fetch(&download_catalog);
+if (fclose(fcatalog) != 0) {
+    PANIC(kLogStderr, "could not close temporary file %s: error %d",
+                tmp.c_str(), retval);
+  }
```
jblomer: Indentation error.

jblomer: It's ok to not check -- we generally don't check for file close errors

test/src/703-catalog_mgr_client/setup_teardown
```diff
@@ -0,0 +1,57 @@
...
+
+  do_local_mount_as_root "$mntpnt"          \
+                 "$CVMFS_TEST_REPO" \
+                 "$(get_repo_url $CVMFS_TEST_REPO)" \
+                 "" \
+                 "CVMFS_SYSLOG_LEVEL=2
+CVMFS_HIDE_MAGIC_XATTRS=no
+CVMFS_QUOTA_LIMIT=15" \
+                 || return 1
+                 #"$mount_options" || return 1
```
jblomer: Remove unused commented code

test/src/703-catalog_mgr_client/setup_teardown
```diff
@@ -0,0 +1,57 @@
...
+
+cleanup() {
+  echo "running cleanup()..."
+  if [ "x$TEST703_PIDS" != "x" ]; then
+    sudo kill -9 $TEST703_PIDS
+  fi
+  if [ "x$TEST703_PRIVATE_MOUNT" != "x" ]; then
+    private_unmount
+  fi
+  sudo rm -rf /tmp/cvmfs-test/workdir/703-catalog_mgr_client
```
jblomer: Don't remove the workdir. That's handled by the test driver. It keeps it around for debugging if the test fails.

test/unittests/t_catalog_mgr_client.cc
```diff
@@ -0,0 +1,352 @@
...
+#include "upload_spooler_definition.h"
+#include "util/pointer.h"
+#include "util/posix.h"
+#include "util/uuid.h"
+
+#include "c_file_sandbox.h"
+
+using namespace std;  // NOLINT
+
+#define GTEST_COUT std::cerr << "[          ] [ INFO ]"
```
jblomer: That's confusing... `GTEST_COUT` prints to `cerr`?

It would be preferable to avoid additional output in the unit tests, except `LogCvmfs` debug output.

test/unittests/t_catalog_mgr_client.cc
```diff
@@ -0,0 +1,352 @@
...
+  EXPECT_EQ(catalog::kLoadNew,
+            mp->catalog_mgr()->LoadCatalogByHash(&ctlg_context));
+
+  // also chached should return the same answer
+  EXPECT_EQ(catalog::kLoadNew,
+            mp->catalog_mgr()->LoadCatalogByHash(&ctlg_context));
+
+  // break URL to repo
+  mp->download_mgr()->SetProxyChain("file://noValidURL", "",
+                       download::DownloadManager::kSetProxyBoth);
```
jblomer: No need to touch the proxies. You are setting an invalid host later.
```suggestion
```

test/unittests/t_catalog_mgr_client.cc
```diff
@@ -0,0 +1,352 @@
...
+  EXPECT_EQ(loader::kFailOk, mp->boot_status());
+  EXPECT_EQ(root_hash_str, mp->catalog_mgr()->GetRootHash().ToString());
+
+  // load new root catalog without providing the hash
+  // this will perform a check vs storage loaction which has the most recent one
+  CatalogContext root_info;
+
+  // TODO(heretherebedragons) is it true that MountPoint::Create does NOT
+  // load the root catalog???
+  // apparently it does.. loadcatalogbyhash had a small bug with offline_mode_
```
jblomer: What about this comment? Is it resolved?

jblomer (changes requested):

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,251 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
     }
+  }
+  LogCvmfs(kLogCache, kLogDebug,
+            "failed fetch manifest from server: "
+            "manifest too old or server unreachable (%d - %s)",
+            manifest_failure, manifest::Code2Ascii(manifest_failure));
+
+  // corner case: server not reachable, local revision is 0. We can not
+  // distinguish if local revision number is valid or invalid. Most likely
+  // it is invalid - so fail
```
jblomer: We may have the case that we update a cvmfs client that has no connection to the stratum 1 (and, because of upgrade, has no revision in the breadcrumb). We should still ensure that this mounts from cache (add as an explicit test).

cvmfs/catalog_mgr.h
HereThereBeDragons: no there are cases where we prefill the mountpoint that does not have to be the root catalog. e.g. in `MountCatalog` or in `LoadFreeCatalog`

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: not needed anymore. we do not change the hash. we just try to retrieve/download based on the hash given.

cvmfs/libcvmfs_int.cc
HereThereBeDragons: i added it back.
i dont know why I removed it. maybe by accident (took me quite a while to get the `remount` correctly working). or maybe i thought that because of the `remountDryrun` that has to be successful to go in this branch of the switch statement

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: no we dont call it as we set the hash in `initfixed()` directly called from `mountpoint.cc`
`initfixed()` -> `mountcatalog()` --> `GetNewRootCatalogContext()` is skipped as `!hash.Isnull()` -->  `LoadCatalogByHash()`

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: true. i think we talked about it a while ago and decided that we dont need another request to get the manifest

test/unittests/t_catalog_mgr_client.cc
HereThereBeDragons: yes resolved

test/unittests/t_catalog_mgr_client.cc
HereThereBeDragons: ah that was just for testing. i remove it

HereThereBeDragons: and yes. google test uses `std::cerr` as output

cvmfs/cache.proto
HereThereBeDragons: what do you mean with `libcvmfs_cache.h` "at the beginning of the file?

cvmfs/cache_plugin/channel.cc
HereThereBeDragons: i wonder this is true: the new client always has a revision. so storing with the revision should always succeed?
for loading you are right. there i definitely have to add it

HereThereBeDragons: i added it, but we should have another talk  about the breadcrumb deal. i am a bit confused now. the only way the breadcrumb revision can be missing is when an old client wrote the breadcrumb and it is loaded from cache with a new client. 

e.g. the posix_cache uses manifest::breadcrumb::breadcrumb(string) to set up the breadcrumb. here the constructor takes care and sets the revision automatically to 0 if it is not part of the (outdated) breadcrumb.

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: in the end i think not. the `loadCatalogByHash` does not work on the revision.

i think i was unsure where the local revision will be set (the call to `GetRevisionNoLock()`), but as far as i see it is set in `attachCatalog` and is retrieved directly from the `catalog` object.

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: i will answer here also for your change request below:

`LoadFreeCatalog` is only called for nested mountpoints. the catalog location is only important for the root catalog, as this means that you might want to cache the manifest locally. so here we dont care.

`MountCatalog` can be called both for nested and root catalogs. and the root catalog can either be a fixed hash or null. the location of the root catalog (fixed or not) can be local or remote.

there is now 1 border case i did not see before: if it is a fixed root catalog with the current setting `kCtlgLocationMounted` it will not safe the manifest.
problem is: we also do not have the manifest. and manifest::Fetch should return the newest manifest, not the manifest related to the hash?

in anyway i guess we should add the feature to verify the manifest for fixed root catalogs?

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: see comment above

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: if `LoadFreeCatalog` is never called for a root catalog then only `kLoadNew` can be a succeful return value

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: ok.. just a note: the currently loaded catalog has no timestamp. as such i will keep it at `breadcrumb_timestamp`

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/cache.proto
jblomer: Comments might have gotten scrambled. It was a reply to the question where to edit the changelog. The changelog for libcvmfs_cache is at the beginning of the file libcvmfs_cache.h.

cvmfs/cache_plugin/channel.cc
jblomer: That's correct. The only way is a breadcrumb written by a previous version. But this case is common when we update or when clients in different versions share the same alien cache.

cvmfs/catalog_mgr.h
jblomer: But if it is not a root catalog we are loading, why would we use the constructor that takes a RootCatalogLocation?

cvmfs/catalog_mgr_impl.h
jblomer: Great, so then we can remove the comment.

cvmfs/catalog_mgr_impl.h
jblomer: Hm, I'm afraid I don't understand the relation to the `CatalogContext `object, i.e. how this answers the questions above.

cvmfs/catalog_mgr_impl.h
jblomer: Ok, so then this needs to be checked and the if statement adjusted accordingly.

cvmfs/catalog_mgr_client.cc
jblomer: You can query the currently loaded catalog for its timestamp with `Catalog::GetLastModified()`

cvmfs/catalog_mgr_client.cc
jblomer: So... should the comment then be removed? And the `if` block?

HereThereBeDragons:
performance tests show that current devel and loatctlg have the same performance

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: added `kCtlgNoLocationNeeded`

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,274 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+            "failed fetch manifest from server: "
+            "manifest too old or server unreachable (%d - %s)",
+            manifest_failure, manifest::Code2Ascii(manifest_failure));
 
-    *catalog_hash = cache_hash;
-    offline_mode_ = true;
+  if (success_code == catalog::kLoadFail) {
+    LogCvmfs(kLogCache, kLogDebug, "No valid root catalog found!");
     return success_code;
   }
```
HereThereBeDragons: this if is just for debug output. we could also just not have it.

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot help

HereThereBeDragons:
@cernvm-bot unittest full

HereThereBeDragons:
@cernvm-bot unittest full

HereThereBeDragons:
failure of unittest due to new ip address for b.root-servers.net

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/manifest.cc
```diff
@@ -6,30 +6,44 @@
...
-  { }
-  catalog_hash =
-    shash::MkFromHexPtr(shash::HexPtr(from_string.substr(0, separator_pos)),
-                                      shash::kSuffixCatalog);
-
-  // Get local last modified time
-  if ((from_string[separator_pos] == 'T') && (len > (separator_pos + 1))) {
-    timestamp = String2Uint64(from_string.substr(separator_pos + 1));
+
+  if (from_string.size() > 0) {
```
jblomer: Ok for now but usually it's more readable to reverse such if statements, like

    if (from_string.empty())
      return;

    // Everything else unintended

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,281 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+                                     ensemble.weak_ref());
+
+  if (manifest_failure == manifest::kFailOk) {
+    // server has newest revision or no valid local revision
+    if (ensemble->manifest->revision() > local_newest_revision
+          || local_newest_revision == manifest::Breadcrumb::kInvalidRevision
+          // if revision is 0 both local and server, load catalog from server
+          // as local is most likely just "initialized" without valid value
+          || (ensemble->manifest->revision() == 0 &&
+              local_newest_revision == 0)) {
```
jblomer, line 228: Ok for now, but we should revisit this later. I think we can remove this last condition.

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,281 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+  }
+  LogCvmfs(kLogCache, kLogDebug,
+            "Failed fetch manifest from server: "
+            "manifest too old or server unreachable (%d - %s)",
+            manifest_failure, manifest::Code2Ascii(manifest_failure));
+
+  // corner case: server not reachable, local revision is 0. We can not
+  // distinguish if local revision number is valid or invalid. Most likely
+  // it is invalid - so fail
+  if (local_newest_revision == 0 && (manifest_failure != manifest::kFailOk)
```
jblomer: Does not need to be addressed here, but in general I think consistent parenthesis use would be preferable:
```suggestion
  if ((local_newest_revision == 0) && (manifest_failure != manifest::kFailOk)
```

Also: why do we need the revision check here? Isn't that strictly superseded by checking that the local hash is unset?

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,281 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+          || (ensemble->manifest->revision() == 0 &&
+              local_newest_revision == 0)) {
+      result->SetHash(ensemble->manifest->catalog_hash());
+      result->SetRootCtlgRevision(ensemble->manifest->revision());
+      result->SetRootCtlgLocation(kCtlgLocationServer);
+      if (ensemble->manifest->has_alt_catalog_path()) {
+        fixed_alt_root_catalog_ = true;
+      } else {
+        fixed_alt_root_catalog_ = false;
       }
```
jblomer: Can be written in fewer lines as 

    fixed_alt_root_catalog_ = ensemble->manifest->has_alt_catalog_path();

cvmfs/catalog_mgr_client.cc
```diff
@@ -87,188 +88,281 @@ shash::Any ClientCatalogManager::GetRootHash() {
...
+    local_newest_hash = mounted_hash;
+    local_newest_revision = GetRevisionNoLock();
+    // if needed for integration test 707: breadcrumb_timestamp_newer()
+    local_newest_timestamp = GetTimestampNoLock() > local_newest_timestamp ?
+                                  GetTimestampNoLock() : local_newest_timestamp;
+    result->SetRootCtlgLocation(kCtlgLocationMounted);
+    success_code = catalog::kLoadUp2Date;
+  } else if (local_newest_revision == 0 && mounted_catalogs_.size() > 0) {
+    // breadcrumb has no revision
+    // TODO(heretherebedragons) this branch can be removed in futur versions
```
jblomer:
```suggestion
    // TODO(heretherebedragons) this branch can be removed in future versions
```

jblomer (approved):
Great!  The remaining comments don't block merging, as far as I am concerned. They may be addressed in follow-up PRs.

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: oops  not sure what i thought there

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: could we have in theory the following condition: breadcrumb has a higher revision number with a timestamp that is < server timestamp?

cvmfs/manifest.cc
HereThereBeDragons: simplified a bit more of the entire constructor

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: ok lets try this and see what the cloudtests are saying.

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot unittest full

HereThereBeDragons:
@cernvm-bot unittest full

HereThereBeDragons:
@cernvm-bot unittest full

HereThereBeDragons:
@cernvm-bot unittest full

PR #3161, opened 2023-01-19
Improve mount transfer syslog message

### LINKED ISSUE #2956: Misleading "Transfer fuse connection to new mount...success" message
If a cvmfs repository is unmounted but the fuse process does not go away (because something under a namespace is holding it open) then as of version 2.9 there's a message in the log file like:
```
Transfer fuse connection to new mount (via /var/lib/cvmfs/shared/fusefd.UrOfEt/sock): success
```
This happens even if the kernel is not new enough, and I think the message in that case is misleading because the mount still fails.

jblomer:
Fixes #2956


DrDaveD (approved):
I think that's good enough, although maybe it would be a little better if it was "Attempt to send" instead of "Send".

Issue #2991, opened 2022-07-01
snapshot data file has doubled data


DrDaveD:
This is a new problem I haven't seen before.    The cvmfs-server version is 2.7.1.

The current secondary stratum 1 at IHEP is repeatedly seeing this error when attempting to snapshot from the current primary stratum 1 (they are using my cvmfs-hastratum1 scripts):
```
failed to download http://cvmfs-stratum-one-b.ihep.ac.cn:8081/cvmfs/atlas.cern.ch/stage/data/c9/ee7b96e5e472c6fab19c102449663c54f503ec (9 - host returned HTTP error)
```
That by itself isn't so unusual, but I looked into the file, and it exists on the -b machine, but it is twice the size it is supposed to be compared to the upstream stratum 1 that it read from (oasis-replica.opensciencegrid.org:8002).  In fact, the data file is 2752 bytes, exactly twice the size of 1376 bytes it is supposed to be, and I did a cmp on the data and found it was the original data repeated twice.

I found the corresponding log where it pulled the snapshot from the upstream stratum 1, and there were no anomalies printed.  It was toward the end of its run there:
```
Pulling to /srv/cvmfs/atlas.cern.ch started at Tue Jun 28 22:53:05 CST 2022 (lock 6)
...
Replicating from catalog at /repo/sw/software/0.0
  Processing chunks [46835 registered chunks]: .......................... fetched 1313 new chunks out of 26913 unique chunks
Replicating from historic catalog 6325c95bc4f44e8c9814cd97fb480de20f2eb66b
  Catalog up to date
Checking tagged snapshots...
Stopping 16 workers
Uploading manifest ensemble
Serving revision 106123
Fetched 1313 new chunks out of 32929 processed chunks

Pulling to /srv/cvmfs/lhcb-condb.cern.ch started at Tue Jun 28 22:53:31 CST 2022
```
where on the failing copy to the secondary machine it ends like this:
```
Pushing /srv/cvmfs/atlas.cern.ch started at Tue Jun 28 22:53:31 CST 2022 (lock 6)
...
Replicating from catalog at /repo/sw/software/0.0
  Processing chunks [46835 registered chunks]: .....................failed to download http://cvmfs-stratum-one-b.ihep.ac.cn:8081/cvmfs/atlas.cern.ch/stage/data/c9/ee7b96e5e472c6fab19c102449663c54f503ec (9 - host returned HTTP error)
unexpected HTTP error code 200 - please check the stratum 0 health
[...]

PR #3166, opened 2023-01-23
Improve file truncate logic on download retry

jblomer:
Further improvement on #2991


jblomer:
@cernvm-bot cloudtest

PR #3167, opened 2023-01-24
Avoid checking deduplicated files twice in cvmfs_server check

### LINKED ISSUE #3138: cvmfs_server check can be extremely slow with many references to deduplicated files
I have seen extremely long cvmfs_server check times, on the order of months, especially in repositories that have many references to deduplicated files, such as unpacked.cern.ch and singularity.opensciencegrid.org.  It would be good if they only got checked once, because otherwise each check turns into an http request.

vvolkl:
First implementation of the suggestion in  https://github.com/cvmfs/cvmfs/issues/3138. Adds a map that keeps track of checksums, and skips the `Exists()` check if the file has already been seen. In local tests this does speed up cvmfs_server check: Using `525-bigrepo` (modified to have half duplicated files) it reduces `cvmfs_server check` times from 2min to 1min20s.

Still to do:
- add formal testing
- restructure code (overload Exists())

Fixes #3138

### DIFF
--- a/cvmfs/swissknife_check.cc
+++ b/cvmfs/swissknife_check.cc
@@ -342,10 +342,12 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
   typedef map< uint32_t, vector<catalog::DirectoryEntry> > HardlinkMap;
   HardlinkMap hardlinks;
   bool found_nested_marker = false;
-  char duplicate_lookup_result = 0;
 
   for (unsigned i = 0; i < entries.size(); ++i) {
-    entry_needs_check_ = 1;
+    const bool entry_needs_check = 
+          !entries[i].checksum().IsNull() && duplicates_map_.Contains(entries[i].checksum());
+    if (entry_needs_check) 
+        duplicates_map_.Insert(entries[i].checksum(), 1);
     PathString full_path(path);
     full_path.Append("/", 1);
     full_path.Append(entries[i].name().GetChars(),
@@ -353,13 +355,6 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
     LogCvmfs(kLogCvmfs, kLogVerboseMsg, "[path] %s",
              full_path.c_str());
 
-     if (!duplicates_map_.Lookup(entries[i].checksum(),
-                                 &duplicate_lookup_result))
-     {
-         duplicates_map_.Insert(entries[i].checksum(), 1);
-     } else {
-         entry_needs_check_ = 0;
-     }
 
     // Name must not be empty
     if (entries[i].name().IsEmpty()) {
@@ -389,7 +384,7 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
     }
 
     // Check if the chunk is there
-    if (check_chunks_ && entry_needs_check_ &&
+    if (check_chunks_ && entry_needs_check &&
         !entries[i].checksum().IsNull() && !entries[i].IsExternalFile())
     {
       string chunk_path = "data/" + entries[i].checksum().MakePath();
--- a/cvmfs/swissknife_check.cc
+++ b/cvmfs/swissknife_check.cc
@@ -603,7 +603,7 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
           bool chunk_needs_check = true;
           if (!no_duplicates_map_ && !duplicates_map_.Contains(chunk_hash)) {
             duplicates_map_.Insert(chunk_hash, 1);
-          } else if (!no_duplicates_map) {
+          } else if (!no_duplicates_map_) {
             chunk_needs_check = false;
           }
           if (chunk_needs_check) {
--- a/cvmfs/swissknife_check.cc
+++ b/cvmfs/swissknife_check.cc
@@ -587,7 +587,8 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
           const shash::Any &chunk_hash = this_chunk.content_hash();
           // for performance reasons, only perform the check once
           // and skip if the hash has been checked before
-          if (duplicates_map_.Contains(chunk_hash)) {
+          if (!duplicates_map_.Contains(chunk_hash)) {
+            duplicates_map_.Insert(chunk_hash, 1);
             const string chunk_path = "data/" + chunk_hash.MakePath();
             if (!Exists(chunk_path)) {
               LogCvmfs(kLogCvmfs, kLogStderr, "partial data chunk %s (%s -> "
@@ -598,9 +599,6 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
                        this_chunk.size());
               retval = false;
             }
-          } else {
-            // now hash is checked - add it to map
-            duplicates_map_.Insert(chunk_hash, 1);
           }
         }
       }
--- a/cvmfs/swissknife_check.cc
+++ b/cvmfs/swissknife_check.cc
@@ -49,7 +49,7 @@ namespace swissknife {
 CommandCheck::CommandCheck()
     : check_chunks_(false)
     , is_remote_(false) {
-    const shash::Any hash_null(shash::kMd5);
+    const shash::Any hash_null;
     duplicates_map_.Init(16, hash_null, hasher_any);
 }
 
--- a/cvmfs/swissknife_check.cc
+++ b/cvmfs/swissknife_check.cc
@@ -1089,8 +1089,8 @@ int CommandCheck::Main(const swissknife::ArgumentList &args) {
     return 1;
   }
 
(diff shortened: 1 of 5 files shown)


cvmfs/swissknife_check.cc
```diff
@@ -1065,6 +1087,9 @@ int CommandCheck::Main(const swissknife::ArgumentList &args) {
     return 1;
   }
 
+  const shash::Any hash_zero(shash::kMd5);
```
jblomer:
```suggestion
  const shash::Any hash_null();
```
zero --> null for consistency with the rest of the code.

cvmfs/swissknife_check.cc
```diff
@@ -332,15 +342,25 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
...
   for (unsigned i = 0; i < entries.size(); ++i) {
+    entry_needs_check_ = 1;
     PathString full_path(path);
     full_path.Append("/", 1);
     full_path.Append(entries[i].name().GetChars(),
                      entries[i].name().GetLength());
     LogCvmfs(kLogCvmfs, kLogVerboseMsg, "[path] %s",
              full_path.c_str());
 
+     if (!duplicates_map_.Lookup(entries[i].checksum(),
```
jblomer: I think we should only look at regular files. Otherwise `checksum()` is NULL, and looking for the special key is not really supported. Also, we can use `duplicates_map_.Contains()`.

cvmfs/swissknife_check.cc
```diff
@@ -332,15 +342,25 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
   typedef map< uint32_t, vector<catalog::DirectoryEntry> > HardlinkMap;
   HardlinkMap hardlinks;
   bool found_nested_marker = false;
+  char duplicate_lookup_result = 0;
 
   for (unsigned i = 0; i < entries.size(); ++i) {
+    entry_needs_check_ = 1;
```
jblomer: I guess this can be local to the for loop, so like
```suggestion
    const bool entry_needs_check = 
      !entries[i].checksum().IsNull() && duplicates_map_.contains(entries[i].checksum());
```

Perhaps we can also fold in here the check for external file from further down, and than simplify that if further down to `check_chunks_ && entry_needs_check_`.

cvmfs/swissknife_check.cc
```diff
@@ -369,12 +389,14 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
     }
 
     // Check if the chunk is there
-    if (check_chunks_ &&
+    if (check_chunks_ && entry_needs_check_ &&
```
jblomer: We use the duplicates map also for the individual chunks of the entry (further down the method in `if (entries[i].IsChunkedFile())`).

jblomer (changes requested):
Very nice! Some comments inline.

vvolkl:
Thanks for the comments! Should be adressed. I'll still look into testing.

cvmfs/swissknife_check.cc
```diff
@@ -369,12 +385,13 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
...
 
     // Check if the chunk is there
-    if (check_chunks_ &&
-        !entries[i].checksum().IsNull() && !entries[i].IsExternalFile())
+    if (check_chunks_ && entry_needs_check)
     {
       string chunk_path = "data/" + entries[i].checksum().MakePath();
       if (entries[i].IsDirectory())
         chunk_path += shash::kSuffixMicroCatalog;
+    LogCvmfs(kLogCvmfs, kLogVerboseMsg, "[chunkpath] %s",
```
jblomer: - I guess here is a whitespace error
- The word "chunk" is now actually used for the chunks of a multi-part file. To avoid confusion, I' suggest `[object path]`

cvmfs/swissknife_check.cc
```diff
@@ -566,15 +583,22 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
...
-            LogCvmfs(kLogCvmfs, kLogStderr, "partial data chunk %s (%s -> "
-                                            "offset: %d | size: %d) missing",
-                     this_chunk.content_hash().ToStringWithSuffix().c_str(),
-                     full_path.c_str(),
-                     this_chunk.offset(),
-                     this_chunk.size());
-            retval = false;
+          // for performance reasons, only perform the check once
+          // and skip if the hash has been checked before
+          if (duplicates_map_.Contains(chunk_hash)) {
```
jblomer:
```suggestion
          if (!duplicates_map_.Contains(chunk_hash)) {
            duplicates_map_.Insert(chunk_hash, 1);
```
and then no `else` branch.

NB: This method has grown uncomfortably complicated. Perhaps let's add a TODO to split it into smaller methods.

jblomer (changes requested):
Great! I noticed just a couple more things.

cvmfs/swissknife_check.cc
vvolkl: Yes,  this was only  a debug printout for myself, I'm not sure it's needed for the logs, I'll remove it alltogether

cvmfs/swissknife_check.cc
vvolkl: Indeed that was the wrong way around, thanks!

cvmfs/swissknife_check.cc
```diff
@@ -35,8 +35,24 @@
...
+             reinterpret_cast<const uint32_t *>(key.digest) + 1);
+}
+
+
 namespace swissknife {
 
+CommandCheck::CommandCheck()
+    : check_chunks_(false)
+    , is_remote_(false) {
+    const shash::Any hash_null(shash::kMd5);
```
jblomer: In principle correct but unclear why we bother picking the MD5 algorithm. Let's just do
```suggestion
    const shash::Any hash_null;
```

jblomer (changes requested):
Very nice! One more nit and we can merge.

jblomer:
@cernvm-bot cloudtest

jblomer (changes requested):
As discussed, two more issues
  - A minimal integration test
  - Follow-up on the clang-tidy error

jblomer:
Fixes #3138

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@jblomer This should be ready now - I've checked it again on unpacked-test.cern.ch and the increase in memory consumption seems reasonable (1.6 vs 1.2 GB). The failing cloudtests look unrelated to me.

cvmfs/swissknife_check.cc
```diff
@@ -566,15 +598,24 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
...
-                                            "offset: %d | size: %d) missing",
-                     this_chunk.content_hash().ToStringWithSuffix().c_str(),
-                     full_path.c_str(),
-                     this_chunk.offset(),
-                     this_chunk.size());
-            retval = false;
+          // for performance reasons, only perform the check once
+          // and skip if the hash has been checked before
+         bool chunk_needs_check = true;
+          if (!no_duplicates_map_ && duplicates_map_.Contains(chunk_hash)) {
```
jblomer: Here seems to be two issues

- Whitespacing (also following lines)
- Logic: we should insert the chunk if the map does _not_ contain it already.

jblomer (changes requested):
I think there is one more logical error for chunked files. Otherwise good to go.

cvmfs/swissknife_check.cc
vvolkl: True, thanks!

cvmfs/swissknife_check.cc
```diff
@@ -600,14 +600,14 @@ bool CommandCheck::Find(const catalog::Catalog *catalog,
...
           // for performance reasons, only perform the check once
           // and skip if the hash has been checked before
-         bool chunk_needs_check = true;
-          if (!no_duplicates_map_ && duplicates_map_.Contains(chunk_hash)) {
+          bool chunk_needs_check = true;
+          if (!no_duplicates_map_ && !duplicates_map_.Contains(chunk_hash)) {
             duplicates_map_.Insert(chunk_hash, 1);
-           chunk_needs_check = false;
-         }
+            chunk_needs_check = false;
```
jblomer, line 607: Sorry, still one more: if the file was not yet in the duplicates map, it does need checking.

jblomer (changes requested):

cvmfs/swissknife_check.cc
vvolkl: Oh right - adding the fallback option didn't really make this method less complicated unfortunately.

jblomer (approved):

Issue #3157, opened 2023-01-17
Allow asserts to be suppressed


HereThereBeDragons:
In certain non-cirtical paths asserts are triggered in rare cases causing a fatal mount crash. Supressing asserts and performing graceful logging instead could be a better workaround.

Issue to track PRs related to this topic
#3148 - conditional_assert() function
#2926 - replace asserts() with return error

Other issues part of it
#3135 - assert from glue::PageCacheTracker::Close 
#2896 - assert()s in xattr handling functions can cause client to abort

PR #3169, opened 2023-01-25
build option: allow replacement of certain `asserts` by error logging

HereThereBeDragons:
Issue #3157 

Certain asserts are now replaced by `assert_or_log_error()`
If for the build `CVMFS_SUPPRESS_ASSERTS` is defined: replace assert by logging an error message. Needs then special handling for failure-path.

Unittest either checks for successful `assert` or error logging depending on the build option.

Integrates change requests of PR #2926 and #3148


HereThereBeDragons:
Functionality of unittests was tested with and without `CVMFS_SUPPRESS_ASSERTS`

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
maybe we can also do a cloudtest with `CVMFS_SUPPRESS_ASSERTS` on?

HereThereBeDragons:
still missing: changing asserts in `glue_buffer.cc`

HereThereBeDragons:
@jblomer have a look

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/util/exception.h
```diff
@@ -36,6 +37,27 @@ CVMFS_EXPORT
...
+    LogCvmfs(log_source, log_mask, log_error_msg);
+    return false;
+  }
+  return true;
+}
+#else
+static inline bool AssertOrLog(int t,
+                               const LogSource /*log_source*/,
+                               const int /*log_mask*/,
+                               const char */*log_error_msg*/) {
```
jblomer:
```suggestion
                               const char * /*log_error_msg*/) {
```

cvmfs/util/exception.h
```diff
@@ -5,6 +5,7 @@
 #ifndef CVMFS_UTIL_EXCEPTION_H_
 #define CVMFS_UTIL_EXCEPTION_H_
 
+#include <assert.h>
```
jblomer:
```suggestion
#include <cassert>
```

cvmfs/util/exception.h
```diff
@@ -36,6 +37,27 @@ CVMFS_EXPORT
 __attribute__((noreturn))
 void Panic(const char *coordinates, const LogSource source, const char *nul);
 
+#ifdef CVMFS_SUPPRESS_ASSERTS
```
jblomer, line 46:
```suggestion
// The AssertOrLog function is used for rare error cases that are not yet fully understood. By default,
// these error cases are considered as unreachable and thus an assert is thrown. Some users may prefer though
// to continue operation (even though the file system state may be scrambled).
// Ideally, we can at some point remove all AssertOrLog statements.
#ifdef CVMFS_SUPPRESS_ASSERTS
```

cvmfs/cvmfs.cc
```diff
@@ -311,7 +311,15 @@ static bool GetDirentForInode(const fuse_ino_t ino,
   if (ino == catalog_mgr->GetRootInode()) {
     bool retval =
       catalog_mgr->LookupPath(PathString(), catalog::kLookupDefault, dirent);
-    assert(retval);
+
+    std::string errormsg = "GetDirentForInode: Race condition? "
+                           "Not found dirent ";
+    errormsg.append(dirent->name().c_str());
```
jblomer: This triggers additional memory allocation(s) for the string concatenation. I'd prefer to do this only if `!retval`. Or, we make AssertOrLog a C variadic function like LogCvmfs, so that we can pass static C strings (mostly).

Actually, I think I prefer the second option because the code using it becomes nicer. See here: https://c-faq.com/varargs/handoff.html
We will need to split `LogCvmfs` into two functions, which is hopefully trivial.

cvmfs/cvmfs.cc
```diff
@@ -430,7 +438,17 @@ static bool GetPathForInode(const fuse_ino_t ino, PathString *path) {
   LogCvmfs(kLogCvmfs, kLogDebug, "MISS %d - looking in inode tracker", ino);
   glue::InodeEx inode_ex(ino, glue::InodeEx::kUnknownType);
   bool retval = mount_point_->inode_tracker()->FindPath(&inode_ex, path);
-  assert(retval);
+
+  std::string errormsg =
```
jblomer: See above

cvmfs/cvmfs.cc
```diff
@@ -1630,7 +1648,15 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   }
   if (d.HasXattrs()) {
     retval = catalog_mgr->LookupXattrs(path, &xattrs);
-    assert(retval);
+
+    std::string errormsg =
```
jblomer: See above.

jblomer (changes requested):
I think we should make the error message handling slightly more elegant/efficient.

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer (approved):

Issue #3117, opened 2022-11-29
cvmfs_swissknife gc has no memory of what it has deleted


mharvey-jt:
Debugging the garbage collection of a repo today (with 2.8.1) I observe that gc appears not to retain any memory of objects that it has already deleted, leading to (in our case very many) redundant `DELETE` calls to object storage, and much slower performance than expected.

jblomer:
Good point!  We keep the hash set of the preserved items but not of the already deleted items. That would be a very nice performance improvement!

PR #3175, opened 2023-01-30
[GC] delete objects only once

HereThereBeDragons:
Issue #3117 
new hashmap that remembers which hashes are already requested to be deleted and skips them if called again
new counter `duplicate_delete_requests_` to count how many of the requests were skipped

2 questions:
- `// TODO(jblomer): turn current counters into perf::Counters` - ??
- `generate_stats_plots.c` - extend it?

extended test by some super simple lines. not sure if needed?


HereThereBeDragons:
@cernvm-bot cloudtest

jblomer (changes requested):
In principle looks good to me but we also need to upgrade existing statistics databases with an older schema version. See `LiveSchemaUpgradeIfNecessary()` for details.

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer (approved):

PR #3178, opened 2023-01-31
Avoid segfault when calling dtor on uninitialized smallhash

### LINKED ISSUE #3177: Calling the destructor on an unitialized Smallhash Object segfaults
All uses of the Smallhash classes in CVMFS consistently call Init() before being destroyed, but a small check can avoid the possible segfault.

vvolkl:
Fixes #3177


jblomer:
@cernvm-bot cloudtest

jblomer (approved):

PR #3180, opened 2023-02-01
Add options to ignore the cache checks and make the memory threshold configurable

dvandok:
There are two additional configuration options to add to the check_cvmfs script:

1. An option to set the memory threshold instead of the hard-coded default
2. A flag to ignore the cache related checks, mostly for osgstorage.org which doesn't play well with the cache

According to DrDave, the osgstorage.org repos don't expect many cache hits. So the default (non-shared) cache size should be small. This is going to trip up the checks. For these repos we want to ignore the cache checks.

### DIFF
--- a/add-ons/check_cvmfs.sh
+++ b/add-ons/check_cvmfs.sh
@@ -6,6 +6,7 @@
 # ChangeLog
 # 1.12a - 31.01.2023
 #    - Add -c option to skip cache checks (for osgstorage.org)
+#    - Add -M option to Customize the memory check threshold
 # 1.12 - 05.11.2021
 #    - Add -p parameter for I/O error retention period
 # 1.11 - 16.03.2021
@@ -28,7 +29,7 @@
 #    - return immediately if transport endpoint is not connected
 #    - start of ChangeLog
 
-VERSION=1.12
+VERSION=1.12a
 
 STATUS_OK=0
 STATUS_WARNING=1     # Check timed out or CernVM-FS resource consumption high or
@@ -42,6 +43,7 @@ RETURN_STATUS=$STATUS_OK
 MAX_FILL_RATIO=95
 TIMEOUT_SECONDS=120
 IO_ERROR_PERIOD=180 # By default, ignore I/O errors older than 3 hours
+MEM_THRESHOLD=50
 
 usage() {
    /bin/echo "Usage:   $0 [-t <seconds>][-m] [-n] [-f fill_ratio] [-i] [-p minutes] <repository name> [expected cvmfs version]"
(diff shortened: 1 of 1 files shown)


add-ons/check_cvmfs.sh
```diff
@@ -1,9 +1,12 @@
 #!/bin/bash
 # CernVM-FS check for Nagios
-# Version 1.10, last modified: 19.06.2017
+# Version 1.12a, last modified: 31.01.2023
```
jblomer:
```suggestion
# Version 1.13, last modified: 31.01.2023
```

add-ons/check_cvmfs.sh
```diff
@@ -1,9 +1,12 @@
...
 # Bugs and comments to Jakob Blomer (jblomer@cern.ch)
 #
 # ChangeLog
+# 1.12a - 31.01.2023
```
jblomer:
```suggestion
# 1.13 - 31.01.2023
```

add-ons/check_cvmfs.sh
```diff
@@ -26,7 +29,7 @@
 #    - return immediately if transport endpoint is not connected
 #    - start of ChangeLog
 
-VERSION=1.12
+VERSION=1.12a
```
jblomer:
```suggestion
VERSION=1.13
```

add-ons/check_cvmfs.sh
```diff
@@ -309,17 +322,19 @@ do_check() {
...
-  if [ $? -ne 0 ]; then
-    append_info "failed to run /bin/df -P $CVMFS_CACHE_BASE"
-    RETURN_STATUS=$STATUS_CRITICAL
-  else
-    FILL_RATIO=`/bin/echo "$DF_CACHE" | /usr/bin/tail -n1 | \
-                /bin/awk '{print $5}' | /usr/bin/tr -Cd [:digit:]`
-    if [ $FILL_RATIO -gt $MAX_FILL_RATIO ]; then
-      append_info "space on cache partition low"
-      RETURN_STATUS=$STATUS_WARNING
+  if [ $OPT_IGNORE_CACHE = 0 ]; then
```
jblomer:
```suggestion
  if [ $OPT_IGNORE_CACHE -eq 0 ]; then
```

add-ons/check_cvmfs.sh
```diff
@@ -358,10 +373,12 @@ do_check() {
     fi
   fi
 
-  # Check for number of cache cleanups within the last 24 hours
-  if [ $NCLEANUP24 -gt 24 ]; then
-    append_info "frequent cache cleanups, cache might be undersized"
-    RETURN_STATUS=$STATUS_WARNING
+  if [ $OPT_IGNORE_CACHE = 0 ]; then
```
jblomer:
```suggestion
  if [ $OPT_IGNORE_CACHE -eq 0 ]; then
```

jblomer (changes requested):
Many thanks!  Looks in principle good to me. Just some nits.

dvandok:
Resolved the nits.

jblomer (approved):

PR #3188, opened 2023-02-08
Update individual cvmfs_server check status when successful

### LINKED ISSUE #3147: Succeeding cvmfs_server check should clear status in .cvmfs_status.json
When I implemented `cvmfs_server check -a` I deliberately had it only update the .cvmfs_status.json "last_check" and "check_status" fields, as opposed to somebody manually running `cvmfs_server check reponame`.  That was because I didn't want failed status from a manual check to get flagged by a remote monitor such as cvmfs-servermon.  However, I now think that it makes sense for a manual `cvmfs_server check reponame` that succeeds to clear out a failure from a previous run.  I think that would be a common use case, that someone would run the check manually after they have repaired whatever the background check detected.  It could update the "last_check" timestamp too, although I don't think it should add the timestamp if it wasn't already there from a previous "check -a".

DrDaveD:
- Fixes #3147

### DIFF
--- a/cvmfs/server/cvmfs_server_check.sh
+++ b/cvmfs/server/cvmfs_server_check.sh
@@ -317,6 +317,18 @@ cvmfs_server_check() {
       __do_check "$@"
     fi
     retcode=$?
+    if [ $retcode = 0 ]; then
+      # Intentionally do not store the status for a failure when a check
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer (approved):

PR #3189, opened 2023-02-09
Limit parallel cvmfs_server snapshot -a commands

### LINKED ISSUE #3064: Make a parallelization limit on `cvmfs_server snapshot -a`
The OSG stratum 1s at U of Nebraska-Lincoln starts `cvmfs_server snapshot -a` 4 times a minute, and we had a case once where it ran into trouble and had huge numbers of parallel processes started.  So for quite some time we have protected each cron entry with the following command
```
sem --id cvmfs_snapshot_a -j +0 --st -1 cvmfs_server snapshot -ais
```
That limits the number of parallel processes to the number of cores on the machine by default.  The limit can be set to a specific limit by replace `+0` with the number.

It would be good if the `cvmfs_server snapshot -a` command would do this automatically, and add a config variable to override the default limit.

DrDaveD:
This limits the number of `cvmfs_server snapshot -a` that run in parallel to $CVMFS_MAX_PARALLEL_SNAPSHOTS, default the number of cores.

- Fixes #3064

### DIFF
--- a/cvmfs/server/cvmfs_server_snapshot.sh
+++ b/cvmfs/server/cvmfs_server_snapshot.sh
@@ -234,13 +234,36 @@ EOF
     exit 1
   fi
 
+  local maxparallel="${CVMFS_MAX_PARALLEL_SNAPSHOTS:-`nproc`}"
+
+  # make locks in a tmpfs directory so they will go away after system crash
+  local tmpdir=/dev/shm/cvmfs_snapshot_all
+  if [ ! -d $tmpdir ]; then
+    mkdir -m 1777 $tmpdir
+  fi
+  local locknum=0
+  local lockfile=""
+  while [ "$locknum" -lt "$maxparallel" ]; do
+    lockfile=$tmpdir/$locknum
+    if acquire_lock $lockfile; then
+      break
+    fi
+    let locknum+=1
+  done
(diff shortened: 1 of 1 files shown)


cvmfs/server/cvmfs_server_snapshot.sh
```diff
@@ -234,13 +234,36 @@ EOF
...
+  local lockfile=""
+  while [ "$locknum" -lt "$maxparallel" ]; do
+    lockfile=$tmpdir/$locknum
+    if acquire_lock $lockfile; then
+      break
+    fi
+    let locknum+=1
+  done
+  if [ "$locknum" -ge "$maxparallel" ]; then
+    # reached the maximum, silently exit
```
jblomer: I think it would be useful to log something in this case.  We could report in `$fulllog` that the command was skipped.

jblomer (changes requested):

cvmfs/server/cvmfs_server_snapshot.sh
DrDaveD: Done, please re-review

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

Issue #3095, opened 2022-11-08
Refactoring/Extending DownloadManager


HereThereBeDragons:
Here a first overview of the changes we talked about:

**Refactoring**
- Move all the componentes in a new directory cvmfs/network/

- Make struct Counters in download.h a subclass of DownloadManager

- JobInfo
    - Becomes a class
    - Has 3 sections: Input parameters, Statusinfo, Output parameters

- Download sink
    - Independent of the exact download destination, all should inherit from sink.h
    - JobInfo should then only have a single parameter "destinationSink"

- Create a Pipe Class
    - Extremly light-weight (maybe just typedefs?, or some template that takes an enum?)
    - Goal: Better understand which pipe communicates with which classes

    
- Replace Init/Fini functions of DownloadManager with Create/Destructor to comfort latest coding style

- Create classes for 
    - HostInfo
    - HostList
    - ProxyList
    - ProxyGroups

    - Must support the following functionalities:
        - Add/Remove

**New features**
- Add health check for proxies and? hosts (see #2945 )
- Add CVMFS_DOWNLOAD_INFINITE_RETRY client parameter and allow for infinite retries for downloading, iterating through proxies/hosts even if they fail until one is back up online ( see #2944 )

HereThereBeDragons:
First couple of Tasks:

- [x] Move files into newly created `network` folder
- [x] JobInfo: should then only have a single parameter "destinationSink"
- [x] Template pipe class in `posix.h`
- [x] Document pipe class for developers
- [x] Replacement of `Init()/Fini()` by constructor + `Create()`

HereThereBeDragons:
health check: allow to define a health check URL either 1) one url for all proxy groups or 2) one url per proxy group

HereThereBeDragons:
next steps:

- [x] Replace `Fini()` with destructor
- [ ] replace placeholder for custome proxy health check and sharding policy with proper version (#2945 , #2944)
    - see PRs #3121
- [ ] replace `jobinfo` with `jobcontext` (#3217)
- [ ] add zstd compression algorithm (#3141)
- [ ] maybe also: Add CVMFS_DOWNLOAD_INFINITE_RETRY client parameter and allow for infinite retries for downloading, iterating through proxies/hosts even if they fail until one is back up online ( #2944 )

HereThereBeDragons:
- [ ] For debug messages: link jobinfo object and curl requests #3492

HereThereBeDragons:
[...]

PR #3193, opened 2023-02-10
Jobinfo: Introduce Sinks for everything

HereThereBeDragons:
Issue #3095

Needs some more decision taking when it comes to streamling functionality over all the different sinks. 

Could be to add for every sink: 
`data_` (either char * or FILE or ,...)
`pos_` (for FILE it would be always null unless we offer an option for appended writing or whatever)
`size_` (not sure what it would be for FILE? some large number? )

TransactionSink is a bit different. Should be extracted from fetch?

`Flush` (for file) and `Close` (for path) also dont have a equivalent -- maybe `Finalize`?
For memory there are in addition `AllocData`, `Release` and `Set`

New function for all sinks `IsValid`


HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@jblomer have a look
we will have to talk about how we can unify the differents sinks a bit more

1) extra functions per sink:
mem sink: AllocData, Release, Set 
file sink: Flush
path sink: Close, Release

2) about ownership:
mem sink: owns, can release ownership
file sink: does not own
path sink: owns
transaction sink: does not own?

3) maybe it is also possible to replace some file sinks with path sinks to make the owner ship more clear?

4) move transactionsink also to `network/`?

cvmfs/network/sink.h
```diff
@@ -29,6 +29,11 @@ class Sink {
    * Truncate all written data and start over at position zero.
    */
   virtual int Reset() = 0;
+  // virtual void Finalize() = 0;
+  /**
+    * Returns true if the object is correctly initialized.
+  */
+  virtual bool IsValid() = 0;
```
jblomer, line 53: The "zombie object" pattern should only be a last resort.  If possible to avoid, the code becomes usually easier to maintain.  Since writing can anyway always fail, I think we don't need to explicitly check for invalid sinks.  If the sink is invalid, its implementation has to deal with it.  E.g.: no valid file descriptor --> writing fails.

jblomer: The class documentation needs adjustment.

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,96 @@
...
+    is_owner_ = is_owner;
+    size_ = size;
+    pos_ = pos;
+    data_ = data;
+  }
+
+ public:
+  size_t size_;
+  size_t pos_;
+  char *data_;
```
jblomer:
```suggestion
  unsigned char *data_;
```

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,96 @@
...
+  explicit MemSink(size_t size) : size_(size), pos_(0) {
+    data_ = static_cast<char *>(smalloc(size));
+    is_owner_ = true;
+  }
+
+  virtual ~MemSink() { if (is_owner_ && data_) { free(data_); } }
+
+  virtual int64_t Write(const void *buf, uint64_t sz) {
+    if (pos_ + sz > size_) {
+      return static_cast<int64_t>(pos_ + sz - size_);
```
jblomer: Let's return a `write` errno here, I'd suggest
```suggestion
      return -ENOSPC;
```

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,96 @@
...
+    return ret;
+  }
+
+  virtual bool IsValid() {
+    return (size_ == 0 && pos_ == 0 && data_ == NULL) ||
+           (size_ > 0 && pos_ >= 0 && data_ != NULL);
+  }
+
+  void AllocData(size_t size) {
+    if (is_owner_ && data_) {
```
jblomer: Also used in the destructor. Let's make it a private function `FreeData()`

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,96 @@
...
+    if (pos_ + sz > size_) {
+      return static_cast<int64_t>(pos_ + sz - size_);
+    }
+
+    memcpy(data_ + pos_, buf, sz);
+    pos_ += sz;
+    return 0;
+  }
+
+  virtual int Reset() {
```
jblomer: Do we not just want to reset `pos_` to zero?

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,96 @@
...
+    } else {
+      data_ = static_cast<char *>(smalloc(size));
+    }
+  }
+
+  void Release() {
+    is_owner_ = false;
+  }
+
+  void Set(size_t size, size_t pos, char *data, bool is_owner = true) {
```
jblomer: I think here we always want to adopt memory, so I'd suggest

```suggestion
  void Adopt(size_t size, size_t pos, char *data) {
```

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,96 @@
...
+    assert(size >= pos);
```
jblomer: include `cassert`

cvmfs/network/sink_file.h
```diff
@@ -0,0 +1,45 @@
...
+#define CVMFS_NETWORK_SINK_FILE_H_
+
+#include "sink.h"
+#include "util/posix.h"
+
+namespace cvmfs {
+
+class FileSink : public Sink {
+ public:
+  explicit FileSink(FILE *destination_file) : file_(destination_file) {
```
jblomer: explicitly include `cstdio` for `FILE`

cvmfs/network/sink_file.h
```diff
@@ -0,0 +1,45 @@
...
+    assert(file_ != NULL);
+  }
+
+  virtual ~FileSink() { }
+
+  virtual int64_t Write(const void *buf, uint64_t sz) {
+    return fwrite(buf, 1ul, sz, file_);
```
jblomer: `fwrite` and `write` have different return value semantics.  We have should decide to use one of the for `Sink::Write`. I'd suggest `write`. In this case, we need to check for short writes here and then use `feof`, `ferror` to return an errno.

cvmfs/network/sink_file.h
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  virtual int Reset() {
+    return !((fflush(file_) != 0) ||
```
jblomer: Looks like this can be expressed positively, like `fflush` works and `ftruncate` works and `freopen` works.

cvmfs/network/sink_file.h
```diff
@@ -0,0 +1,45 @@
...
+           (ftruncate(fileno(file_), 0) != 0) ||
+           (freopen(NULL, "w", file_) != file_));
+  }
+
+  int Flush() {
+    return fflush(file_);
```
jblomer: return -errno on failure

cvmfs/network/sink_path.h
```diff
@@ -0,0 +1,63 @@
...
+#define CVMFS_NETWORK_SINK_PATH_H_
+
+#include <string>
+
+#include "sink.h"
+#include "util/posix.h"
+
+namespace cvmfs {
+
+class PathSink : public Sink {
```
jblomer, line 30: You may want to use a `PathFile` for the implementation.

cvmfs/network/sink_file.h
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  bool IsValid() {
+    return file_ != NULL;
+  }
+
+ public:
```
jblomer:
```suggestion
 private:
```

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,96 @@
...
+      free(data_);
+    }
+
+    is_owner_ = is_owner;
+    size_ = size;
+    pos_ = pos;
+    data_ = data;
+  }
+
+ public:
```
jblomer:
```suggestion
 private:
```

cvmfs/network/sink_path.h
```diff
@@ -0,0 +1,63 @@
...
+
+  void Release() {
+    is_owner_ = false;
+  }
+
+  bool IsValid() {
+    return file_ != NULL;
+  }
+
+ public:
```
jblomer:
```suggestion
 private:
```

jblomer (changes requested):
Cool!

I took a first pass through the sink implementation.  I think it makes sense to stop here before looking into the rest of the changes.

I think the abstract interface should be amended by
  - `int Release() // release ownership of the sink's resources (whatever it is)`
  - `int Flush() // commit data to the sink`
  - `int Reserve(uint64_t size) // allocate space in the sink`

The implementation of these functions can be empty (e.g. `Reserve` for files, `Flush` for mem).

The `AdoptMemory` method as well as methods to get the sink's resource after it has been released should remain custom to the concrete sink.

So the idea is that the download manager does not need to know about the sink type. But the user of the sink of course knows the type and can call the extra functions when needed.

cvmfs/network/sink.h
HereThereBeDragons: in `download.cc` is a check - you want the check of validity gone?:
```
static Failures PrepareDownloadDestination(JobInfo *info) {
  if (info->destination != kDestinationNone &&
      !info->destination_sink->IsValid()) {
```

cvmfs/network/sink_mem.h
HereThereBeDragons: `Take` vs `Adopt`?

cvmfs/network/sink_path.h
HereThereBeDragons: `grep` for `PathFile` only finds `.go` files?

cvmfs/network/sink_mem.h
HereThereBeDragons: in theory i would agree, but `Reset()` replaces the following sequence in `download.cc`

```c++
// for mem sink
free(info->destination_mem.data);
info->destination_mem.data = NULL;
info->destination_mem.size = 0;
info->destination_mem.pos = 0;


// for  file/pathsink
if ((fflush(info->destination_file) != 0) ||
    (ftruncate(fileno(info->destination_file), 0) != 0) ||
    (freopen(NULL, "w", info->destination_file) != info->destination_file)) {
    {
        info->error_code = kFailLocalIO;
        goto verify_and_finalize_stop;
    }
 ```

File: cvmfs/network/sink_path.h (hunk 12 above)
Comment (jblomer) [reply cvmfs/network/sink_path.h:30]: Sorry, I meant `FileSink`

File: cvmfs/network/sink_mem.h (hunk 5 above)
Comment (jblomer) [reply cvmfs/network/sink_mem.h]: You're right, so let's leave the current implementation.

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
current commit is mainly a cloudtest check

i've wrote a few more `TODOs` in the code we need to talk about and we have to have another look at `fwrite` vs `write` and replacing some of the `filesinks` with `pathsinks`

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot  cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot help

**Comment by HereThereBeDragons** [comment]
@cernvm-bot tidy

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by jblomer** [comment]
@cernvm-bot tidy

**Comment by jblomer** [comment]
@cernvm-bot cpplint

File: cvmfs/network/sink.h [hunk 16]
```diff
@@ -22,13 +23,60 @@ class Sink {
...
+   * 
+   * @returns true  - reservation is needed
+   *          false - no reservation is needed
+   */
+  virtual bool RequiresReserve() = 0;
+
+  /**
+   * Return a string representation of the sink
+  */
+  virtual std::string ToString() = 0;
```
jblomer: A `ToString()` method makes it look like the contents of the sink is being represented as a string.
```suggestion
  virtual std::string Describe() = 0;
```

cvmfs/network/sink_file.h
jblomer, line 1: Only put one-liners in the header, the rest in a proper .cc compilation unit

cvmfs/network/sink.h
```diff
@@ -22,13 +23,60 @@ class Sink {
...
+   * @returns on success: number of bytes written 
+   *          on failure: -errno.
    */
   virtual int64_t Write(const void *buf, uint64_t sz) = 0;
+
   /**
    * Truncate all written data and start over at position zero.
+   * 
+   * @returns Success = 0
+   *          Failure = -1
```
jblomer: For consistency, we also return `-errno` here in case of failure

cvmfs/network/sink.h
```diff
@@ -22,13 +23,60 @@ class Sink {
...
+
+  /**
+   * Release ownership of the sink's resource
+   */
+  void Release() { is_owner_ = false; }
+
+  /**
+   * Commit data to the sink
+   * @returns success 0
+   *          otherwise failure
```
jblomer: `-errno` on failure

cvmfs/network/sink.h
```diff
@@ -22,13 +23,60 @@ class Sink {
...
+   */
+  virtual int Flush() = 0;
+
+  /**
+   * Allocate space in the sink
+   * 
+   * @returns success 0
+   */
+  virtual int Reserve(size_t size) = 0;
```
jblomer:
```suggestion
  virtual bool Reserve(size_t size) = 0;
```

cvmfs/network/sink_file.h
jblomer: This is still public. It should be a `private` section.

cvmfs/network/sink_path.h
jblomer, line 1: Only put one-liners in the header, the rest in a proper .cc compilation unit

cvmfs/network/sink_path.h
```diff
@@ -0,0 +1,111 @@
...
+
+namespace cvmfs {
+
+class PathSink : public Sink {
+ public:
+  explicit PathSink(const std::string &destination_path) :
+                                                       path_(destination_path) {
+    is_owner_ = true;
+    file_ = fopen(destination_path.c_str(), "w");
+    assert(file_ != NULL);
```
jblomer: I don't think we want to crash in this case.  We want to return an I/O error to the user.  All the infrastructure to do so is already in place, I think we can simply remove the assert.

cvmfs/network/sink_path.h
```diff
@@ -0,0 +1,111 @@
...
+#include "util/posix.h"
+
+namespace cvmfs {
+
+class PathSink : public Sink {
+ public:
+  explicit PathSink(const std::string &destination_path) :
+                                                       path_(destination_path) {
+    is_owner_ = true;
+    file_ = fopen(destination_path.c_str(), "w");
```
jblomer: include `cstdio`

cvmfs/network/sink_path.h
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Commit data to the sink
+   * @returns success 0
+   *          otherwise failure
+   */
+  int Flush() {
+    // A zero value indicates success.
+    // For error: EOF is returned and the error indicator is set (see ferror)
+    return fflush(file_) * -1;
```
jblomer: That doesn't work.  The `fflush` call returns `EOF` on error. We want to return `-errno` in this case.

cvmfs/network/sink_path.h
jblomer: This is still `public`. It should be `private`.

cvmfs/network/sink_path.h
```diff
@@ -0,0 +1,111 @@
...
+  std::string ToString() {
+    std::string result = "Path sink for ";
+    result += "path " + path_ + " and ";
+    result += IsValid() ? " valid file pointer" : " invalid file pointer";
+    return result;
+  }
+
+ public:
+  FILE *file_;
+  const std::string &path_;
```
jblomer: In order to make sure not to end up with a dangling reference, let's use
```suggestion
  std::string path_;
```

cvmfs/network/sink_mem.h
jblomer, line 1: Only put one-liners in the header, the rest in a proper .cc compilation unit. I think the memory sink should have a safety limit for the size, e.g. 4MB max.

cvmfs/network/sink_mem.h
jblomer: This is still `public`. It should be `private`. You can add public getters for the buffer and the size (i.e., nbytes written aka `pos_`).

cvmfs/network/sink_mem.h
jblomer: Still needs fixing.

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,158 @@
...
+  size_t size_;
+  size_t pos_;
+  char *data_;
+
+ private:
+  void FreeData() {
+    if (is_owner_ && data_) {
```
jblomer: The `free()` call already handles the `NULL` pointer, hence
```suggestion
    if (is_owner_) {
```

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /**
+   * Truncate all written data and start over at position zero.
+   * 
+   * @returns Success = 0
+   *          Failure = -1
+   */
+  virtual int Reset() {
+    int ret = 0;
```
jblomer: We don't need `ret`, the call always succeeds

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,158 @@
...
+    if (is_owner_ && data_) {
```
jblomer: The `free()` call already handles the `NULL` pointer, hence
```suggestion
    if (is_owner_) {
```

cvmfs/network/sink_mem.h
jblomer: I don't think we ever call `Reset` when we don't own the buffer.  Still, to make this behave correctly, I think if we don't own the buffer we should only set `pos_` to 0.

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,158 @@
...
+  virtual int Flush() {
+    return 0;
+  }
+
+  /**
+   * Allocate space in the sink
+   * 
+   * @returns success 0
+   */
+  virtual int Reserve(size_t size) {
```
jblomer: We should first check if the current size is already big enough. Secondly, if we don't own the buffer and the current size is too small, return an error.

cvmfs/network/sink_mem.h
```diff
@@ -0,0 +1,158 @@
...
+    FreeData();
+
+    is_owner_ = true;
+    size_ = size;
+    pos_ = 0;
+    if (size == 0) {
+      data_ = NULL;
+      return 0;
```
jblomer: Move to after the if/else block

cvmfs/catalog_mgr_ro.cc
```diff
@@ -30,15 +30,19 @@ LoadError SimpleCatalogManager::LoadCatalog(const PathString  &mountpoint,
   shash::Any effective_hash = hash.IsNull() ? base_hash_ : hash;
   assert(shash::kSuffixCatalog == effective_hash.suffix);
   const string url = stratum0_ + "/data/" + effective_hash.MakePath();
+
+  // TODO(heretherebedragons) change this to a pathsink???
```
jblomer: What would be the advantage?
```suggestion
```

cvmfs/network/jobinfo.h
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Where to store downloaded data.
+ */
+enum Destination {
+  kDestinationMem = 1,
+  kDestinationFile,
+  kDestinationPath,
+  kDestinationTransaction,
+  kDestinationNone
+};  // Destination
```
jblomer: This is not needed anymore, let's remove. We can replace the check for `kDestinationNone` by looking if the sink is `NULL` or not. Where you need to distinguish the destination type for better error reporting, you can  use `dynamic_cast`.

cvmfs/network/jobinfo.h
```diff
@@ -0,0 +1,155 @@
...
+  bool probe_hosts;
+  bool head_request;
+  bool follow_redirects;
+  bool force_nocache;
+  pid_t pid;
+  uid_t uid;
+  gid_t gid;
+  void *cred_data;  // Per-transfer credential data
+  InterruptCue *interrupt_cue;
+  Destination destination;
```
jblomer: Let's remove

cvmfs/network/sink.h
```diff
@@ -22,13 +23,60 @@ class Sink {
...
+   */
+  virtual bool RequiresReserve() = 0;
+
+  /**
+   * Return a string representation of the sink
+  */
+  virtual std::string ToString() = 0;
+
+ protected:
+    bool is_owner_;
```
jblomer, line 99: To make sure this is always initialized, we should have an explicit `Sink` constructor where the value for `is_owner_` is passed.

cvmfs/fetch.h
```diff
@@ -35,14 +35,33 @@ class TransactionSink : public Sink {
   TransactionSink(CacheManager *cache_mgr, void *open_txn)
     : cache_mgr_(cache_mgr)
     , open_txn_(open_txn)
-  { }
-  virtual ~TransactionSink() { }
-  virtual int64_t Write(const void *buf, uint64_t sz) {
+  { is_owner_ = false; }
+  ~TransactionSink() { }
+  int64_t Write(const void *buf, uint64_t sz) {
```
jblomer: Keep the `virtual` keyword for overwritten virtual methods (also further down).

cvmfs/manifest_fetch.cc
```diff
@@ -217,9 +225,10 @@ Failures Verify(char *manifest_data, size_t manifest_size,
                 signature::SignatureManager *signature_manager,
                 download::DownloadManager *download_manager,
                 ManifestEnsemble *ensemble) {
-  char *manifest_copy = reinterpret_cast<char *>(smalloc(manifest_size));
-  memcpy(manifest_copy, manifest_data, manifest_size);
-  return DoVerify(manifest_copy, manifest_size, base_url, repository_name,
+  cvmfs::MemSink sink(manifest_size);
+  memcpy(sink.data_, manifest_data, manifest_size);
+  sink.pos_ = manifest_size;
+  return DoVerify(&sink, base_url, repository_name,
```
jblomer: The `DoVerify` method should not take a sink. It can continue taking a manifest buffer and size (nothing is downloaded here).

cvmfs/network/download.cc
```diff
@@ -300,7 +294,8 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
                   num_bytes, info->hash_context);
   }
 
-  if (info->destination == kDestinationSink) {
+  // TODO(heretherebedragons) do we need the check?? info->destination != kDestinationNone
```
jblomer: Because without a sink we can only do a `HEAD` request and therefore `CallbackCurlData` is never called?  If that's the assumption, replace the check by an assert.

cvmfs/network/download.cc
```diff
@@ -318,78 +313,17 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
...
-                 PRId64 ")", info->url->c_str(), written);
-        info->error_code = kFailLocalIO;
-        return 0;
-      }
-    }
-  } else if (info->destination == kDestinationMem) {
-    // Write to memory
-    if (info->destination_mem.pos + num_bytes > info->destination_mem.size) {
-      if (info->destination_mem.size == 0) {
+      if (static_cast<uint64_t>(written) != num_bytes) {
```
jblomer: That's in principle unsafe. The negative return value could by chance be equal to `num_bytes`.

cvmfs/network/download.cc
```diff
@@ -318,78 +313,17 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
...
-        return 0;
-      }
-    } else {
-      if (fwrite(ptr, 1, num_bytes, info->destination_file) != num_bytes) {
-       LogCvmfs(kLogDownload, kLogSyslogErr,
-                 "downloading %s, IO failure: %s (errno=%d)",
-                 info->url->c_str(), strerror(errno), errno);
-        info->error_code = kFailLocalIO;
-        return 0;
+          "Failed to perform write of %d bytes to sink %s",
```
jblomer: The `num_bytes` variable is of type `size_t`, so this should be `%zu`. It would be helpful to additionally report the error number returned by `written`.

cvmfs/network/jobinfo.h
```diff
@@ -0,0 +1,155 @@
...
+  cvmfs::Sink *destination_sink;
```
jblomer:
```suggestion
  cvmfs::Sink *sink;
```

cvmfs/network/download.cc
```diff
@@ -1048,12 +982,21 @@ void DownloadManager::SetUrlOptions(JobInfo *info) {
...
 
-  if ((info->destination == kDestinationMem) &&
-      (info->destination_mem.size == 0) &&
-      HasPrefix(url, "file://", false))
-  {
-    info->destination_mem.size = 64*1024;
-    info->destination_mem.data = static_cast<char *>(smalloc(64*1024));
+  // TODO(heretherebedragons) how do we make size_ == 0 generic?
+  // MAYBE have an empty non-functional sink, so that you do not have to check
+  // for kDestinationNone???
```
jblomer: Let's simply use `destination_sink == NULL` as the marker for "HEAD request / no sink"

cvmfs/network/download.cc
```diff
@@ -1048,12 +982,21 @@ void DownloadManager::SetUrlOptions(JobInfo *info) {
...
+  if ((info->destination != kDestinationNone) &&
+      info->destination_sink->RequiresReserve() &&
+      (static_cast<cvmfs::MemSink*>(info->destination_sink)->size_ == 0) &&
```
jblomer: This should not be needed anymore

cvmfs/network/download.cc
```diff
@@ -1048,12 +982,21 @@ void DownloadManager::SetUrlOptions(JobInfo *info) {
...
+      HasPrefix(url, "file://", false)) {
+    platform_stat64 stat_buf;
+    int retval = platform_stat(url.c_str(), &stat_buf);
+    if (retval != 0) {
+      // TODO(heretherebedragons) WHAT DO WE DO IN AN ERROR CASE?
+      // info->destination_sink->Reserve(64ul * 1024ul);
```
jblomer: Let's generally reserve 64kB. You already have the code to increase the buffer as needed.

cvmfs/network/download.cc
```diff
@@ -1413,38 +1335,25 @@ bool DownloadManager::VerifyAndFinalize(const int curl_error, JobInfo *info) {
...
-    if ((info->destination == kDestinationMem) && info->destination_mem.data) {
-      free(info->destination_mem.data);
-      info->destination_mem.data = NULL;
-      info->destination_mem.size = 0;
-      info->destination_mem.pos = 0;
+    if (info->destination != kDestinationNone &&
+        info->destination_sink->Reset() != 0) {
+      // TODO(heretherebedragons) in theory this is unnecessary to set
+      // the error code because in the goto the flush is checked again
+      // and error_code set
```
jblomer: Under `verify_and_finalize_stop`, the error code is only set if flushing fails. So I think we should remove the comment.

cvmfs/network/download.cc
```diff
@@ -1797,13 +1702,14 @@ Failures DownloadManager::Fetch(JobInfo *info) {
...
+                                                                 path_.c_str());
+    }
 
-    if (info->destination_mem.data) {
-      free(info->destination_mem.data);
-      info->destination_mem.data = NULL;
-      info->destination_mem.size = 0;
+    if (info->destination == kDestinationMem) {
+      // This IF could be removed? just always reset the sink?
+      info->destination_sink->Reset();
```
jblomer: Let's handle this consistently with a virtual `Sink::Purge()` method that releases all the resources of the sink.  I.e., memory sink frees the memory, path sink unlinks the file, file sink closes the fd. The `Purge()` method would only take action if the sink is owning and then leave the sink in an invalid state.

cvmfs/network/jobinfo.cc
```diff
@@ -0,0 +1,17 @@
...
+#include "util/string.h"
+
+namespace download {
+
+bool JobInfo::IsFileNotFound() {
+  if (HasPrefix(*url, "file://", true /* ignore_case */))
+    return error_code == kFailHostConnection;
+
+  return http_code == 404;
+}
```
jblomer, line 15: I think this is not used anywhere and hence should be removed.

cvmfs/swissknife_info.cc
```diff
@@ -127,18 +127,21 @@ int swissknife::CommandInfo::Main(const swissknife::ArgumentList &args) {
...
     }
-    char *buffer = download_manifest.destination_mem.data;
-    const unsigned length = download_manifest.destination_mem.pos;
+
     manifest = manifest::Manifest::LoadMem(
-        reinterpret_cast<const unsigned char *>(buffer), length);
-    free(download_manifest.destination_mem.data);
+                reinterpret_cast<const unsigned char *>(manifest_memsink.data_),
+                manifest_memsink.pos_);
+    manifest_memsink.Reset();
```
jblomer: Not needed, part of sink destructor

jblomer (changes requested):
Very nice work! Some comments inline, general comments
- Overwritten virtual methods should again be marked virtual
- A couple of `Reset()` calls on memory sinks can be removed, as the destructor of the sink takes care of releasing the memory
- We should have minimal unit tests for the sinks. E.g., one unit test for the new sinks (file, path, mem) that checks writes one byte, resets, writes again. It is then much easier to extend if we find bugs.

cvmfs/network/sink.h
HereThereBeDragons: is this the case for the `transactionSink`? otherwise i do not see it

cvmfs/catalog_mgr_ro.cc
HereThereBeDragons: less lines of code and not having to think about to manually call `fclose()`

cvmfs/network/jobinfo.h
HereThereBeDragons: you just want to remove `kDestinationNone`? because the other ones we for sure still need. or some way to know what kind of sink we are

cvmfs/manifest_fetch.cc
HereThereBeDragons: it uses the prefilled sink here:
```c++
  manifest_sink->Release();
  ensemble->raw_manifest_buf =
                        reinterpret_cast<unsigned char *>(manifest_sink->data_);
  ensemble->raw_manifest_size = manifest_sink->pos_;
  ensemble->manifest = manifest::Manifest::LoadMem(
    ensemble->raw_manifest_buf, ensemble->raw_manifest_size);
    ```

Comment (jblomer) [reply cvmfs/manifest_fetch.cc]: I understand that `DoVerify` is using the sink data. But I find it confusing that the sink is used here only as a memory buffer.  Nothing is actually downloaded into the sink. Therefore, let's just use a pointer & size.

File: cvmfs/network/jobinfo.h (hunk 30 above)
Comment (jblomer) [reply cvmfs/network/jobinfo.h]: No, all of them should be removed.  We can replace all but the `kDestinationNone` by `dynamic_cast` calls, and the `kDestinationNone` by a check for the `NULL` pointer.

File: cvmfs/catalog_mgr_ro.cc (hunk 29 above)
Comment (jblomer) [reply cvmfs/catalog_mgr_ro.cc]: I see, if it is something that should be addressed in a follow-up PR, then just remove the question marks from the comment. It is then clear that we should really do this.

File: cvmfs/network/sink.h (hunk 17 above)
Comment (jblomer) [reply cvmfs/network/sink.h]: What I mean is: the other methods in the abstract `Sink` interface return `-errno` for an error (not just -1). So `Truncate` should do the same.

File: cvmfs/network/jobinfo.cc (hunk 44 above)
Comment (HereThereBeDragons) [reply cvmfs/network/jobinfo.cc:15]: Short search says:
It is used here
https://github.com/cvmfs/cvmfs/blob/943eff2c581788030cb3f679d3e7cc64cc3d8a85/cvmfs/publish/repository.cc#L179
and here
https://github.com/cvmfs/cvmfs/blob/943eff2c581788030cb3f679d3e7cc64cc3d8a85/cvmfs/publish/repository.cc#L248

Comment (jblomer) [reply cvmfs/network/jobinfo.cc:15]: Ah, missed these two. Never mind then.

File: cvmfs/network/sink.h (hunk 17 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink.h]: changed `Reset()` now to (where applicable)
```
return ((fflush(file_) == 0) &&
           (ftruncate(fileno(file_), 0) == 0) &&
           (freopen(NULL, "w", file_) == file_)) ? 0 : -errno;
```

File: cvmfs/network/sink.h (hunk 19 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink.h]: for `memSink` i thought for a moment that you want to give different return types if you are the owner or not. but in the end i decided: if you call "reserve" you get new space - so you will become the owner of the new data eventhough of the old data you might have not been (frees old data if needeD)

File: cvmfs/network/sink_file.h (hunk 13 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink_file.h]: i guess then i need some more getters...

File: cvmfs/network/sink_path.h (hunk 20 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink_path.h]: so we want to add some error handling in the constructor or what is the idea?

File: cvmfs/network/sink_path.h (hunk 22 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink_path.h]: thanks.. and weird.. in `sink_file` it was correct

File: cvmfs/network/sink_mem.h (hunk 25 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink_mem.h]: some leftover code: at one point i wanted to distinguish between `Reset()` called on owning data and not owning data. (if data was free or not).. but in the end i decided not do it)

File: cvmfs/network/sink_mem.h (hunk 5 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink_mem.h]: i guess this needs then a bit more comment about that not wrong assumptions are being taken. as when you are the owner your entire buffer is being deleted.

but for now i moved `data_` and `size_` into the `if (is_owner_) `
```
  virtual int Reset() {
    if (is_owner_) {
      free(data_);
      data_ = NULL;
      size_ = 0;
    }

    pos_ = 0;

    return 0;
  }
```

File: cvmfs/network/sink_mem.h (hunk 27 above)
Comment (HereThereBeDragons) [reply cvmfs/network/sink_mem.h]: added a check.. but right now `Reserve` creates a new buffer anyway. Currently when you call `Reserve` it always succeeds and even if you were not before the owner you become the new owner for the new buffer..

i guess `smalloc` should be changed to `realloc`? but what would we do with `pos`? right now it is set to 0. but with realloc i guess it should stay at current pos and then we would need some `reset pos` function or `reset()` does this and `purge()` allows freeing the data.

File: cvmfs/catalog_mgr_ro.cc (hunk 29 above)
Comment (HereThereBeDragons) [reply cvmfs/catalog_mgr_ro.cc]: i first wanted to hear your opinion what you would prefer :)

File: cvmfs/network/download.cc (hunk 35 above)
Comment (HereThereBeDragons) [reply cvmfs/network/download.cc]: you tell me.
i have currently only focused on replacing stuff and not how all the different functions interact with each other in detail.

but the name `CallbackCurlData` suggest that this should not be called if only the header is requested?

File: cvmfs/network/download.cc (hunk 39 above)
Comment (HereThereBeDragons) [reply cvmfs/network/download.cc]: yes but that does not solve this here:
```c++
if ((info->destination != kDestinationNone) &&
      info->destination_sink->RequiresReserve() &&
      (static_cast<cvmfs::MemSink*>(info->destination_sink)->size_ == 0) && << this line
      HasPrefix(url, "file://", false)) {
```

HereThereBeDragons: ah didnt see the comment below.. 
ok what should this code snipped do?

if the sink is one where you need to reserve space and it's size is 0, either allocate 64 kB or size of the file if the url points to a local file?

as far as i understand we cannot remove the precondition of `size == 0` as `DownloadManager::SetUrlOptions(JobInfo *info)` seems to be called from multiple places multiple time on the same jobinfo obj that we do not want to reset

https://github.com/cvmfs/cvmfs/blob/943eff2c581788030cb3f679d3e7cc64cc3d8a85/cvmfs/network/download.cc#L988-L998

cvmfs/network/download.cc
HereThereBeDragons: see comment above

cvmfs/network/download.cc
HereThereBeDragons: so do  you want to remove the part again to allocate as much space as needed for a local file?

https://github.com/cvmfs/cvmfs/blob/943eff2c581788030cb3f679d3e7cc64cc3d8a85/cvmfs/network/download.cc#LL998

cvmfs/network/download.cc
HereThereBeDragons: i think we will run here into problems:
- mem sink expects the `reset` to `purge`
- file sink expects the `reset` to NOT do `unlink` 

the `unlink` i had once in the destructor of the file sink and it crashed due to the `unlink`

cvmfs/network/sink_mem.h
HereThereBeDragons: we then need to check the return type when we call `reserve` - what should be the behavior?

cvmfs/network/sink_mem.h
HereThereBeDragons: apparently `malloc` is supposed to be faster than `realloc` when assigning larger chunks and as we dont want to preserve the data no reason to use `realloc`

cvmfs/network/download.cc
HereThereBeDragons: @jblomer having a second look at it
i think we could do here some error handling? != 0 should only be returned when stat64 returns an error -- so i guess this is only the case when the file does not exist or out of memory https://linux.die.net/man/2/stat64

should we do something there or just wait for the error to be caught at a different point?

```c++
int retval = platform_stat(url.c_str(), &stat_buf); 
   if (retval != 0)  { // this should mean here is an error that the file(url) does not exist }
   ```

File: cvmfs/network/download.cc (hunk 35 above)
Comment (HereThereBeDragons) [reply cvmfs/network/download.cc]: ok seems like the `assert` is not being triggered.. but lets wait for the cloud test

File: cvmfs/network/download.cc [hunk 46]
```diff
@@ -1048,12 +979,22 @@ void DownloadManager::SetUrlOptions(JobInfo *info) {
...
-  {
-    info->destination_mem.size = 64*1024;
-    info->destination_mem.data = static_cast<char *>(smalloc(64*1024));
+  // TODO(heretherebedragons) before removing
+  // static_cast<cvmfs::MemSink*>(info->sink)->size() == 0
+  // and just always call info->sink->Reserve()
+  // we should do a speed check
+  if ((info->sink != NULL) && info->sink->RequiresReserve() &&
+      (static_cast<cvmfs::MemSink*>(info->sink)->size() == 0) &&
+      HasPrefix(url, "file://", false)) {
```
HereThereBeDragons, line 988: please comment on this

cvmfs/network/download.cc
```diff
@@ -1048,12 +979,22 @@ void DownloadManager::SetUrlOptions(JobInfo *info) {
...
+    platform_stat64 stat_buf;
+    int retval = platform_stat(url.c_str(), &stat_buf);
+    if (retval != 0) {
+      // this is an error: file does not exist or out of memory
+      // error is caught in other code section.
+      info->sink->Reserve(64ul * 1024ul);
```
HereThereBeDragons, line 994: please comment on this

cvmfs/fetch.h
```diff
@@ -33,17 +33,63 @@ namespace cvmfs {
...
+
+  // TODO(heretherebedragons) does this always return 0 if successful? (also for tiered cache)
+  /**
+   * Truncate all written data and start over at position zero.
+   * 
+   * @returns Success = 0
+   *          Failure = -errno
+   */
   virtual int Reset() {
     return cache_mgr_->Reset(open_txn_);
```
HereThereBeDragons: please comment on this

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
if the cloudtest turn out green:
we might want to consider what `Manifest::Verify()` does. If it has to create its on manifest copy or if we can tell `DoVerify()` that it does not on the manifest buffer from where it is called.

and we really need is some debug version for the server (also for the unittests where the server unittests are not build in debug, only the client unittests). we might also consider playing with [`mallopt(M_CHECK_ACTION, n)](https://man7.org/linux/man-pages/man3/mallopt.3.html) to get better debug when e.g. double free occurs.

cvmfs/network/sink_file.h
```diff
@@ -0,0 +1,113 @@
...
+#include "util/posix.h"
+
+namespace cvmfs {
+
+class FileSink : public Sink {
+ public:
+  explicit FileSink(FILE *destination_file) : Sink(true),
+                                              file_(destination_file) { }
+
+  virtual ~FileSink() { }
```
jblomer: If we are owner, we should close the `FILE *` pointer here. Same for `Purge()` ("release all resources")

cvmfs/network/sink_path.cc
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Purges all resources leaving the sink in an invalid state.
+ * More aggressive version of Reset().
+ * For some sinks it might do the same as Reset().
+ * 
+ * @returns Success = 0
+ *          Failure = -errno
+ */
+int PathSink::Purge() {
+  int ret = Reset();
```
jblomer:
```suggestion
  int ret = sink_->Purge();
```

cvmfs/network/sink_file.h
```diff
@@ -0,0 +1,113 @@
...
+#include <string>
+
+#include "sink.h"
+#include "util/posix.h"
+
+namespace cvmfs {
+
+class FileSink : public Sink {
+ public:
+  explicit FileSink(FILE *destination_file) : Sink(true),
```
jblomer: I think usually we don't own the FILE pointer, except when used from the path sink. So we should have an `Adopt` call, similar to the memory sink.

```suggestion
  explicit FileSink(FILE *destination_file) : Sink(false),
```

cvmfs/network/download.cc
```diff
@@ -300,104 +296,39 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
...
-       LogCvmfs(kLogDownload, kLogSyslogErr,
-                 "downloading %s, IO failure: %s (errno=%d)",
-                 info->url->c_str(), strerror(errno), errno);
-        info->error_code = kFailLocalIO;
-        return 0;
-      }
+    int64_t written = info->sink->Write(ptr, num_bytes);
+    if (written < 0 || static_cast<uint64_t>(written) != num_bytes) {
+      LogCvmfs(kLogDownload, kLogDebug,
+        "Failed to perform write of %%zu bytes to sink %s with errno %d",
```
jblomer:
```suggestion
        "Failed to perform write of %zu bytes to sink %s with errno %d",
```

jblomer (changes requested):
Great! Few tiny remaining things and this is good to go IMO.

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/network/sink_file.cc
```diff
@@ -41,6 +41,23 @@ int FileSink::Reset() {
...
+ * More aggressive version of Reset().
+ * For some sinks and depending on owner status it might do 
+ * the same as Reset().
+ *
+ * @returns Success = 0
+ *          Failure = -errno
+ */
+int FileSink::Purge() {
+  if (is_owner_ && file_) {
+    return (-1 * fclose(file_));
```
jblomer: You need to check `fclose` for `EOF` return value and then return `-errno` from this method.

jblomer: Also, set `file_` to `NULL` after the `fclose()` call.

jblomer (changes requested):

jblomer (approved):

jblomer:
@cernvm-bot cloudtest

cvmfs/network/sink_mem.h
```diff
@@ -16,6 +16,14 @@
 
 namespace cvmfs {
 
+/**
+ * MemSink is a data sink that writes to a unsigned char* buffer.
+ * The buffer has a fixed size, as such reservation of space is necessary.
+ * It can use Adopt() to write to a different buffer.
+ *
+ * By default, MemSink is the owner of the buffer and takes care of its
+ * creationg and deletion.
```
vvolkl:
```suggestion
 * creation and deletion.
```

cvmfs/network/download.cc
```diff
@@ -290,6 +281,11 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
   const size_t num_bytes = size*nmemb;
   JobInfo *info = static_cast<JobInfo *>(info_link);
 
+  // TODO(heretherebedragons) remove if no error comes up
+  // as this means only jobinfo data request (and not header only)
+  // come here
+  assert(info->sink != NULL);
+
```
vvolkl, line 288: This assert can be removed?

vvolkl (changes requested):

cvmfs/network/download.cc
HereThereBeDragons: i am not sure.. i added it because it was not clear to us if this is always true. probably it is..

vvolkl: Ok, then let's leave it.

vvolkl (approved):

PR #3205, opened 2023-02-20
Trim very long log lines in logbuffer xattr

### LINKED ISSUE #2979: attr -g logbuffer can overflow
When CVMFS_DEBUGLOG is enabled (at least on ligo.osgstorage.org) and presumably if there are just a lot of messages accumulated over time, then attr gets an error with the logbuffer:
```
$ attr -g logbuffer /cvmfs/ligo.osgstorage.org
attr_get: Argument list too long
Could not get "logbuffer" for /cvmfs/ligo.osgstorage.org
$ attr -l /cvmfs/ligo.osgstorage.org|grep logbuffer
Attribute "logbuffer" has a 374925 byte value for /cvmfs/ligo.osgstorage.org
```
Apparently it went over a 64kb limit.  The attribute should show a tail of some number of lines.  I think about 128 should be sufficient and still safely below a 64kb limit.

jblomer:
Fixes #2979

### DIFF
--- a/cvmfs/magic_xattr.cc
+++ b/cvmfs/magic_xattr.cc
@@ -338,9 +338,15 @@ std::string LogBufferXattr::GetValue() {
   throttle_.Throttle();
   std::vector<LogBufferEntry> buffer = GetLogBuffer();
   std::string result;
-  for (unsigned i = 0; i < buffer.size(); ++i) {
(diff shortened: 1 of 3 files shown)


jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

PR #3219, opened 2023-02-24
Improve watchdog startup procedure

### LINKED ISSUE #2895: Failure of the crash monitor process causes mount process to abort

### LINKED ISSUE #3089: watchdog process RSS a function of CVMFS_MEMCACHE_SIZE

jblomer:
Fork the watchdog before creation of the file system and the mount point objects. Besides less resource consumption, this change significantly reduces the state/complexity of the watchdog process and may solve the "sudden disappearance" issues.

Fixes: #3089
Maybe fixes: #2895
Replaces: #3092

### DIFF
--- a/cvmfs/cache_plugin/libcvmfs_cache.cc
+++ b/cvmfs/cache_plugin/libcvmfs_cache.cc
@@ -361,7 +361,7 @@ void cvmcache_spawn_watchdog(const char *crash_dump_file) {
     return;
   g_watchdog = Watchdog::Create((crash_dump_file != NULL)
                                 ? string(crash_dump_file)
-                                : "");
+                                : "", NULL);
   assert(g_watchdog != NULL);
   g_watchdog->Spawn();
 }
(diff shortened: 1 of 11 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

PR #3223, opened 2023-03-01
Bump min required go version

vvolkl:
This is needed for some dependency updates such as https://github.com/cvmfs/cvmfs/pull/3220 . Since go 1.18 is in the externals it should not be a problem.


vvolkl:
@cernvm-bot cloudtest

jblomer (commented):
I've two questions:
  - Can we bump go to 1.18 (which is the version we ship)
  - Can we avoid the explicit changes to the dependencies in `go.mod`? I though go can manage this completely on its own.

vvolkl:
> Can we bump go to 1.18 (which is the version we ship)

Sure, I put 1.17 because that would still be enough for #3220 , but I don't see any downside in a stricter requirement.

> Can we avoid the explicit changes to the dependencies in go.mod? I though go can manage this completely on its own.

Here I am not yet sure - there is definitely tooling to help manage this but I'd have to check if there is a way of avoiding checking this into the repository.

jblomer:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

jblomer (approved):

PR #3238, opened 2023-03-14
Prevent long fail-over chains when certificate is invalid

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3244, opened 2023-03-23
Don't assert on memory allocation failure

mharvey-jt:
(as discussed offline - finally validated in production)

If CVMFS_SUPPRESS_ASSERTS is defined, keep retrying memory allocation/map failures on failure. This lets CVMFS ride out temporary OOM situations.

Steps to repro:
1) Start with a version of CVMFS that can survive the loss of the the monitor process
2) make a trivial program to consume memory: 
```
 #include <string.h>
#include <stdlib.h>


void main(void) {
for(;;) {
 void *buf = malloc(1024*1024*1024);
 if(buf) { memset(buf, 0, 1024*1024*1024);}
}
}
```
3) mount a repo, and run several concurrent `find /cvmfs/repo ` over it
4) run many copies of the memory hog
5) Wait for the machine to start ooming
6) Observe the cvmfs mount will eventually abort, from one of the assert  in alloc/mmap functions in `smalloc.h`

Retry with this patch in place. Note that the mount process is much more resilient throughout the OOMs.


HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/util/smalloc.h
```diff
@@ -39,19 +39,42 @@ static inline uint64_t RoundUp8(const uint64_t size) {
...
 
 static inline void * __attribute__((used)) smalloc(size_t size) {
-  void *mem = malloc(size);
+  void *mem = NULL;
+#ifdef CVMFS_SUPPRESS_ASSERTS
+  do {
+#endif
+    mem = malloc(size);
+#ifdef CVMFS_SUPPRESS_ASSERTS
+  } while (mem == NULL);
```
jblomer:
```suggestion
  } while ((size > 0) && (mem == NULL));
```

cvmfs/util/smalloc.h
```diff
@@ -39,19 +39,42 @@ static inline uint64_t RoundUp8(const uint64_t size) {
...
 static inline void * __attribute__((used)) srealloc(void *ptr, size_t size) {
-  void *mem = realloc(ptr, size);
+  void *mem = NULL;
+
+#ifdef CVMFS_SUPPRESS_ASSERTS
+  do {
+#endif
+    mem = realloc(ptr, size);
+#ifdef CVMFS_SUPPRESS_ASSERTS
+  } while (mem == NULL);
```
jblomer:
```suggestion
  } while ((size > 0) && (mem == NULL));
```

cvmfs/util/smalloc.h
```diff
@@ -39,19 +39,42 @@ static inline uint64_t RoundUp8(const uint64_t size) {
...
 static inline void * __attribute__((used)) scalloc(size_t count, size_t size) {
-  void *mem = calloc(count, size);
+  void *mem = NULL;
+
+#ifdef CVMFS_SUPPRESS_ASSERTS
+  do {
+#endif
+    mem = calloc(count, size);
+#ifdef CVMFS_SUPPRESS_ASSERTS
+  } while (mem == NULL);
```
jblomer:
```suggestion
  } while ((count * size > 0) && (mem == NULL));
```

jblomer (changes requested):
LGTM, except for handling of `size == 0`.

cvmfs/util/smalloc.h
mharvey-jt: changed as suggested

cvmfs/util/smalloc.h
mharvey-jt: ditto

cvmfs/util/smalloc.h
mharvey-jt: ditto

jblomer (approved):

vvolkl (approved):

PR #3246, opened 2023-03-30
ducc convert: sleep briefly between wishes to avoid load spikes on registries

vvolkl:


jblomer:
How much longer will it take to process the wishlist with this patch? Just to make sure we stay comfortably in the 10 minute window.

vvolkl:
@jblomer it should add about 7.5 minutes of sleep time, is that too much?

jblomer:
It will probably still work but I think it should be seen a stopgap solution. The wish list will most likely grow.

vvolkl:
Superseded by #3248

PR #3248, opened 2023-04-02
[ducc] make requests to registries more robust

vvolkl:
Launches one thread per registry instead of one per request as before, avoiding load spikes and rate limits. This also has the advantage of allowing some registry-specific treatments. In order to do that, the loop over `recipeYamlV1.Input ` is split up, in a first step the input lines are parsed and filled into a map, so all images in one repository can be easily adressed. The second step is as before, but moves the loop over the input image into the goroutine, and launches one goroutine per registry instead.

- [ ] the input list could be shuffled to get a bit more resilience in situations where the registry stops responding halfway through the wishlist

### DIFF
--- a/ducc/cmd/convert_single_image.go
+++ b/ducc/cmd/convert_single_image.go
@@ -62,7 +62,7 @@ var convertSingleImageCmd = &cobra.Command{
 			os.Exit(RepoNotExistsError)
 		}
 
-    input, err := lib.ParseImage(inputImage)
+		input, err := lib.ParseImage(inputImage)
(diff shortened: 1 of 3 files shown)


jblomer (approved):
Very nice!  As follow ups, we can consider
- Not hardcoding the sleep time and the target registry (but taking it from env variables or config file)
- Still using some parallelism per registry.  I think we may get higher throughput due to the parallel connections.

vvolkl:
Thanks for the suggestions, this will be much easier to add with the refactoring in https://github.com/cvmfs/cvmfs/issues/3294 - I'm merging this PR now and have added a note there.

Issue #3192, opened 2023-02-10
sqlite readahead optimisation 1/2


mharvey-jt:
1) The cache readahead does so in 4kiB blocks  - this should be coalesced into a single large read
2) The readahead should be omitted altogether when the cache is on a tmpfs, as it's already in the page cache (see https://stackoverflow.com/questions/26635846/how-to-tell-if-a-file-is-on-tmpfs-given-its-path-on-linux for how to determine)

mharvey-jt:
https://github.com/cvmfs/cvmfs/pull/3250 addresses point 2

jblomer:
@HereThereBeDragons Is the first point also addressed?

HereThereBeDragons:
nope not yet. reopen

HereThereBeDragons:
Related #3315 and #3316

HereThereBeDragons:
closing this part as the missing improvement (point 1 .. the 4kiB reads) are tracked in issue #3335

PR #3250, opened 2023-04-03
Cache_posix: Omit prefetch if fd is on a tmpfs

mharvey-jt:
If the cache is on a tmpfs then the prefetch is redundant as the file's already in page cache. (Linux only)

Ref https://github.com/cvmfs/cvmfs/issues/3192 point 2


vvolkl:
Great!  We'll take care of this. We should probably move the "Is Fd on Tmpfs" function to platform_[linux|osx].h.

HereThereBeDragons:
Superseded by #3254

PR #3253, opened 2023-04-04
support lazy unmounts

mharvey-jt:
We'd like to be able to lazy unmount a repo, and then remount it before the first mount has terminated. This needs a change to `mount.cvmfs` to permit concurrent mounts, and to the cvmfs talk code to prevent the terminating mount from removing the new talk sockets created by the new mount.

This change: 
1) Upon termination check to see whether `/var/run/cvmfs/cvmfs.[repo]` and `/var/cache/cvmfs/cvmfs_io.[repo]` sockets have changed before deleting them.
A change indicates that a `umount -l` and remount has probably happened, and the sockets are now owned by a new mount. In this case, don't delete them.

2) 
* Modify (linux) platform_mountlist() to optionally list only fuse mountpoints
* control the ReattachMount() functionality with a configuration flag, so - even if the kernel has the capability - we can get a new mount rather than a reattach to a `zombie` one

An example of use: 
```
$ cd /cvmfs/repo
$ umount -l .  # unmounted, but mount process won't go until /cvmfs/repo is not in use
$ cvmfs_talk -i repo pid # returns the PID of the now unmounted mount process
$ ls /cvmfs/repo # trigger an automount, starting a new mount process
$ cvmfs_talk -i repo pid # returns the PID of the new mount process
$ cd /cvmfs/repo # this releases the first mount procss which will now exit

### DIFF
--- a/cvmfs/loader_talk.cc
+++ b/cvmfs/loader_talk.cc
@@ -96,7 +96,20 @@ void Spawn() {
 
 
 void Fini() {
-  unlink(socket_path_->c_str());
+  bool changed = false;
+  struct stat current_stat;
+  struct stat new_stat;
+  int ret1, ret2;
+  ret1 = fstat(socket_fd_, &current_stat);
+  ret2 = stat(socket_path_->c_str(), &new_stat);
+  if (!ret1 && !ret2
+     && (current_stat.st_dev != new_stat.st_dev
+     || current_stat.st_ino != new_stat.st_ino)) {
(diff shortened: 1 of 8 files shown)


mharvey-jt:
@cernvm-bot cloudtest

mharvey-jt:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

Issue #3211, opened 2023-02-23
Authz helper's process invokation should reset signal handlers


jblomer:
When the authz helper is executed, the signal handling is still configured according to the watchdog.  The watchdog is not functional however after the `fork()`.  We should reset the signal handlers to their original values.

HereThereBeDragons:
Should be part of refactor download manager #3095

PR #3256, opened 2023-04-14
Reset signals in Authz helper

HereThereBeDragons:
Issue #3211 

First version, just proof of concept to see if it works.


HereThereBeDragons:
@cernvm-bot cloudtest

jblomer:
The watchdog (monitor.cc) already has a record of the old signal handlers. You should be able to let that class do the swapping of the handlers.

HereThereBeDragons:
Superseded by #3260

Issue #2792, opened 2021-10-06
Cache cleanup does not discard currently open files


mcb30:
We have a slightly unusual issue with an application that operates with large numbers of open file handles.  This results in the CVMFS client holding open file handles to the cached chunks that correspond to the current file positions (via the file handle cache in `chunk_tables->handle2fd`).

When attempting a cache cleanup via `cvmfs_talk cleanup 0` or similar, the cached chunk files are therefore unlinked from the cache directory but are not freed until the application subsequently closes its file handle (or reads to the next chunk).  This prevents `cvmfs_talk cleanup` from freeing the expected amount of disk space and, in extreme circumstances, can leave the cache device with insufficient space to download new files.

The CVMFS client is already capable of (re)opening the underlying cached chunk file as necessary, but there is currently no logic that would cause it to close the cached chunk file in response to a cache cleanup event.

I see two viable approaches to this problem:

1. Extend the `FileWatcher` mechanism to be able to monitor a whole directory tree (including subdirectories), watch the cache directory, and react to deletions by closing any corresponding cached chunk file handles.
2. Close all cached chunk file handles whenever a cache cleanup operation completes.

Option 1 has the advantage of working regardless of how the cache gets cleaned (e.g. it would correctly handle a manual `rm -rf`).  Option 2 is probably a lot simpler to implement.

@jblomer Do you have any thoughts on which method you would prefer?

jblomer:
I think among these two options I'd certainly go for the second one.  But I'm still not sure if any of the two options would really help.  If an application keeps the file handles open, my expectation would be that it is going to read data from the open files.  So even if the open files (or chunks) get removed, they'd quickly come back an fill the cache again. Or is the problem rather that the cache eviction removes the wrong files the open ones instead of other, not opened files that are also sitting in the cache? In this case, we should probably pin the open files to make them unavailable to cache eviction.

jblomer:
Thinking further about it, another, possibly cleaner option would be a cache manager implementation that uses the download manager to stream data on read (instead of pre-placing data in the cache). That would fit nicely because the cache manager is already prepared to be pluggable; it could even be an [external process](https://cvmfs.readthedocs.io/en/stable/cpt-plugins.html#cache-plugins). Such a streaming cache manager could still keep a small, internal on-disk or memory cache of course.

mcb30:
[...]

PR #3263, opened 2023-04-24
Streaming cache manager

jblomer:
This is a first draft of a streaming cache manager. The streaming cache manager implements the `CacheManager` interface. It uses a download manager and a backing cache manager to deliver data. Pinned files and catalogs use the backing cache manager. Regular data blocks are downloaded on read, the required data window copied to the user.  

To turn it on, just set `CVMFS_STREAMING_CACHE=yes`.

For small files, we probably don't need a memory cache for downloaded objects. The kernel page cache does all the work.  It gets problematic for large chunks that are read in several small pieces, because each of them trigger a re-download of the entire chunk.

TODOs:

- [x] Fix use of external files
- [x] Save and restore the file table to support client reload
- [x] Testing

### DIFF
--- a/cvmfs/cache_stream.cc
+++ b/cvmfs/cache_stream.cc
@@ -1,0 +1,89 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include "cvmfs_config.h"
+#include "cache_stream.h"
+
+#include <cstdlib>
+#include <string>
+
+#include "util/murmur.hxx"
+#include "util/mutex.h"
+#include "util/smalloc.h"
+
+
+namespace {
+
+static uint32_t hasher_int(const int &key) {
+  return MurmurHash2(&key, sizeof(key), 0x07387a4f);
+}
+
+}  // anonymous namespace
+
+
+StreamingCacheManager::StreamingCacheManager(
+  CacheManager *cache_mgr, download::DownloadManager *download_mgr)
+  : cache_mgr_(cache_mgr)
+  , download_mgr_(download_mgr)
+  , min_stream_size_(kMinStreamSize)
+{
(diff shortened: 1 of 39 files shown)


vvolkl:
Related issue: https://github.com/cvmfs/cvmfs/issues/2792

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

cvmfs/cache_stream.cc
```diff
@@ -0,0 +1,320 @@
...
+
+  if (info.fd_in_cache_mgr >= 0)
+    return cache_mgr_->Pread(info.fd_in_cache_mgr, buf, size, offset);
+
+  uint64_t nbytes_streamed = Stream(info, buf, size, offset);
+  if (nbytes_streamed < offset)
+    return 0;
+  if (nbytes_streamed > (offset + size))
+    return static_cast<int64_t>(size);
+  return static_cast<int64_t>(offset + size - nbytes_streamed);
```
vvolkl: Not sure I understand this line. Why is nbytes_streamed subtracted here?

jblomer: Thanks, that's a good catch!  We should be returning `nbytes_streamed - offset`. Patch coming.

jblomer:
@cernvm-bot cloudtest

PR #3283, opened 2023-05-31
Fix force abort after client crash

### LINKED ISSUE #3278: Unable to recover from publish failures (again)
Similar to https://github.com/cvmfs/cvmfs/issues/3088, publish failures still aren't recoverable by aborting if the FUSE mount state is broken:

```bash
$ ls /cvmfs/lhcbdev.cern.ch
ls: reading directory /cvmfs/lhcbdev.cern.ch: Transport endpoint is not connected

$ cvmfs_server abort lhcbdev.cern.ch
You are about to DISCARD ALL CHANGES OF THE CURRENT TRANSACTION for lhcbdev.cern.ch!  Are you sure (y/N)? y
HTTP Uploader - Could not open session token file.
HTTP Uploader - Could not open session token file.
cannot retrieve root hash from read-only mount point
```

It seems to happen about once a week.

I currently have two lhcbdev publishers in this state if you want investigate.

jblomer:
Fixes #3278

### DIFF
--- a/cvmfs/publish/repository_managed.cc
+++ b/cvmfs/publish/repository_managed.cc
@@ -107,16 +107,22 @@ int Publisher::ManagedNode::Check(bool is_quiet) {
     std::string root_hash_str;
     bool retval = platform_getxattr(rdonly_mnt, root_hash_xattr,
                                     &root_hash_str);
-    if (!retval)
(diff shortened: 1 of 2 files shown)


vvolkl (approved):

vvolkl:
@cernvm-bot cloudtest

Issue #3067, opened 2022-09-16
Open libraries count as real files when opened via CVMFS


ktf:
When a library is loaded from disk, it does not count against the open file descriptors because it gets opened, mmaped, and then the file descriptor is closed.

This is not the case when loading a library in CVMFS. What happens is that the CVMFS fuse daemon keeps an open fd to the file in the cache, even after the file has been closed on the user side. This happens on a per process basis, so that with an handful of processes loading a few library you could have a severe impact on the amount of open file descriptors for the cvmfs deamon, which itself is limited by the CVMFS_NFILE limit.

Is there any particular reason why that needs to be the case? Why can't the fuse daemon simply close the file when the close syscall is issued by dlopen?

jblomer:
That's mandated by the semantics of the Linux file system layer. For the file system, it doesn't matter if the process has the file open for reading or mmaped the file. In both cases, the inode is active.

We currently set a 128k limit for file descriptors in cvmfs. Did you hit that limit?

ktf:
Ciao, yes, we hit that limit because we have a number of processes and the .pcm files of root or shared libraries all count as one, while on other filesystems they get multiplexed. I can understand that 1 inode is active per file, but why one per process? After all they are all opened as readonly.

jblomer:
I see, thanks for the additional information!

The fastest fix, when you control the client, is to just increase `CVMFS_NFILES`. It should be safe to go as high as 1M. We can also generally discuss increasing the default limit (not sure how much you typically overshoot our current default).

Regarding multiplexing: that happens for the memory mappings, i.e. every mmaped file (inode) consumes memory only once even if read by multiple processes.  For file descriptors, the cvmfs fuse module (process) reads all data from the local disk cache, and to do so it "duplicates" all file descriptors of applications using cvmfs.  Native file systems just access block data directly.

It may be possible to re-use the file descriptor for a file already open in the local cache. That would only help for non-chunked files though, and I expect many of the C++ libraries are "large" (> few MB) and therefore chunked.  For chunked files, different process may ready different parts of the file at a given moment, so they each trigger another chunk to be opened.

One could also consider that cvmfs closes file descriptors when it does not currently read from them. This is tricky to get right though, e.g. when should a descriptor be closed, what happens if re-opening fails etc.

[...]

PR #3288, opened 2023-06-07
Re-use the file descriptor for a file already open in the local cache

vvolkl:
This PR adresses https://github.com/cvmfs/cvmfs/issues/3067 , currently just adding an integration test that checks that the cvmfs client process indeed opens the local cache file several times.


vvolkl:
Still to do:

- [x] Implement SaveState and RestoreState
- [x] Erase closed objects from map
- [ ] Documentation
- [ ] Unittests

cvmfs/cache_posix.h
```diff
@@ -110,7 +110,6 @@ class PosixCacheManager : public CacheManager {
   virtual int DoRestoreState(void *data);
   virtual bool DoFreeState(void *data);
 
- private:
```
jblomer, line 113: If we only need the `Transaction` class in the derived class, we could move the `private` section further down instead of removing it entirely.

cvmfs/mountpoint.cc
```diff
@@ -563,7 +564,7 @@ string FileSystem::MkCacheParm(
 {
   assert(HasPrefix(generic_parameter, "CVMFS_CACHE_", false));
 
-  if (instance == kDefaultCacheMgrInstance) {
+  if (instance == kDefaultCacheMgrInstance || instance == "posix_refcount") {
```
jblomer: I think we need a slightly different configuration mechanism because users may have already set `CVMFS_CACHE_PRIMARY` and we don't want to overwrite that. The configuration should say "if you are using the posix cache manager, use it with refcounting".  I'd suggest `CVMFS_CACHE_REFCOUNT=yes/no`

cvmfs/mountpoint.cc
```diff
@@ -711,6 +716,36 @@ CacheManager *FileSystem::SetupPosixCacheMgr(const string &instance) {
   return cache_mgr.Release();
 }
 
+CacheManager *FileSystem::SetupPosixRefcountCacheMgr(const string &instance) {
```
jblomer: I think that can be combined with the regular posix cache manager setup. Depending on the refcount option, you can then either create the regular or the refcount flavor.

test/src/099-reusefd/main
```diff
@@ -0,0 +1,38 @@
...
+
+cvmfs_run_test() {
+  logfile=$1
+
+  cvmfs_mount alice.cern.ch CVMFS_DEBUGLOG=/tmp/debug.log CVMFS_CACHE_PRIMARY=posix_refcount || return 10
+
+  ############
+  # open some test file in several simultaneously in different processes
+
+  testfile="/cvmfs/alice.cern.ch/el6-x86_64/Packages/AliDPG/prod-202306-01-1/README.md"
```
jblomer: Use `find /cvmfs/alice.cern.ch/ -type f | head -n1` to make the test code independent from the repo content.

test/src/099-reusefd/main
```diff
@@ -0,0 +1,38 @@
...
+  # testfile has the following cvmfs hash:
+  testfile_hash="72a19643e2d198ccf96b7eb5f870dd0deaa6ade0"
```
jblomer: You can query the hash through extended attributes

test/src/099-reusefd/main
```diff
@@ -0,0 +1,38 @@
...
+
+  for i in {1..10}; do
+    timeout 2s tail -f ${testfile} > /dev/null &
```
jblomer: This is brittle, just use bash job control to kill the sub processes before the test ends.

test/src/099-reusefd/main
```diff
@@ -0,0 +1,38 @@
...
+  echo $alice_pid
+  num_open_fds=$(ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2} | wc -l)
+
+
+  echo $(ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2})
+  if [ "$num_open_fds" -gt "1" ]; then
+    echo "The cvmfs process has several open file descriptors for the same open test file" && return 20
+  fi
+
+  sleep 3
```
jblomer:
```suggestion
```

test/src/099-reusefd/main
jblomer, line 1: In addition, run the same test without refcounting (test that the same file is open multiple times) to make sure it actually shows the behavior we want to fix.

cvmfs/cache_posix_refc.h
```diff
@@ -0,0 +1,71 @@
...
+#include "crypto/signature.h"
+#include "fd_refcount_mgr.h"
+#include "file_chunk.h"
+#include "gtest/gtest_prod.h"
+#include "manifest_fetch.h"
+#include "shortstring.h"
+#include "statistics.h"
+#include "util/atomic.h"
+
+
```
jblomer:
```suggestion
```

cvmfs/cache_posix_refc.h
```diff
@@ -0,0 +1,71 @@
...
+    fd_mgr()
+  {
+    atomic_init32(&no_inflight_txns_);
+  }
+
+  virtual ~PosixRefcountCacheManager() {}
+
+  virtual int Open(const LabeledObject &object);
+  virtual int Close(int fd);
+  virtual std::string Describe();
```
jblomer: I think we also need a new cache manager ID (virtual method `id()`). This will automatically ensure that an error message is thrown at the attempt to switch between refcounted and non-refcounted cache manager. It then remains to handle the situation where the refcount option is set, but cvmfs was loaded with a regular cache manager. We may be able to do this in `CacheManager::RestoreState`, and then set a flag and the refcounted cache manager where it falls back to the base class for all functions (including save/restore/free state).

Also, in most/all other places where `kPosixCacheManager` is used, both `kPosixCacheManager` and the new id are valid. For that we should probably have a static function like `IsPosixManagerCompatible` or so.

cvmfs/cache_posix_refc.cc
jblomer, line 1: Here and in the header: the includes seem a bit excessive. Perhaps let's remove them entirely and add one by one as needed.

cvmfs/cache_posix_refc.cc
```diff
@@ -0,0 +1,154 @@
...
+    if (!MakeCacheDirectories(cache_path, 0700))
+      return NULL;
+  }
+
+  // TODO(jblomer): we might not need to look anymore for cvmfs 2.0 relicts
+  if (FileExists(cache_path + "/cvmfscatalog.cache")) {
+    LogCvmfs(kLogCache, kLogDebug | kLogSyslogErr,
+             "Not mounting on cvmfs 2.0.X cache");
+    return NULL;
+  }
```
jblomer: This can be factored out in a protected `bool InitCacheDirectory()` method.

cvmfs/cache_posix_refc.cc
```diff
@@ -0,0 +1,154 @@
...
+  int retval = fd_mgr.Close(fd);
+  if (retval != 0)
+    return -errno;
+  return 0;
+}
+
+int PosixRefcountCacheManager::Open(const LabeledObject &object) {
+  const string path = GetPathInCache(object.id);
+  int result = fd_mgr.Open(object.id, path);
+  if (result >= -1) {
```
jblomer:
```suggestion
  if (result >= 0) {
```

cvmfs/fd_refcount_mgr.h
jblomer, line 1: Since FdRefcountMgr is not a templated class, I'd suggest the usual split between header and compilation unit.

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  FdRefcountMgr* Clone() {
+    FdRefcountMgr* clone = new FdRefcountMgr(map_refcount_, map_fd_);
+    return clone;
+  }
+
+ private:
+  SmallHashDynamic<int, std::pair<int, shash::Any>> map_refcount_;
+  SmallHashDynamic<shash::Any, int> map_fd_;
```
jblomer: Add comments to these members

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+    lock_cache_refcount_ =
+      reinterpret_cast<pthread_mutex_t *>(smalloc(sizeof(pthread_mutex_t)));
+    int retval = pthread_mutex_init(lock_cache_refcount_, NULL);
+    assert(retval == 0);
+  }
+
+  FdRefcountMgr(
+        SmallHashDynamic<int, std::pair<int, shash::Any>> map_refcount,
+        SmallHashDynamic<shash::Any, int> map_fd):
+        map_refcount_(map_refcount), map_fd_(map_fd) {
```
jblomer: I'm not sure if the `SmallHashDynamic` can be correctly copy constructed. Is there another example in the code? If so, the input parameters should still be const ref.

Otherwise, we can iterate through the input and insert the key-value pairs one-by-one into the new hash maps.

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+
+  ~FdRefcountMgr() {
+    pthread_mutex_destroy(lock_cache_refcount_);
+    free(lock_cache_refcount_);
+  }
+
+  void AssignFrom(const FdRefcountMgr *other) {
+    map_fd_ = other->GetFdMap();
+    map_refcount_ = other->GetRefcountMap();
+  }
```
jblomer: See above

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+
+  SmallHashDynamic<shash::Any, int>  GetFdMap() const {
+    return map_fd_;
+  }
+
+  int Open(const shash::Any id, const std::string path) {
+    int result = -1;
+    {
+    MutexLockGuard lock_guard(lock_cache_refcount_);
```
jblomer: indent the block

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+    if (!map_fd_.Contains(id)) {
```
jblomer: The else branch can be done with a single `Lookup` call instead of `Contains() + Lookup()`

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+  return retval;
+  }
+
+  FdRefcountMgr* Clone() {
+    FdRefcountMgr* clone = new FdRefcountMgr(map_refcount_, map_fd_);
+    return clone;
+  }
+
+ private:
+  SmallHashDynamic<int, std::pair<int, shash::Any>> map_refcount_;
```
jblomer: I'd suggest to use a mini struct instead of a pair.  The struct adds "self documentation" through the naming of the members.

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+    } else {
+      map_fd_.Lookup(id, &result);
+    }
+    if (result >= 0) {
+       std::pair<int, shash::Any>* refc_info = new std::pair<int, shash::Any>();
+      if (map_refcount_.Lookup(result, refc_info)) {
+        map_refcount_.Insert(result, {refc_info->first + 1, id});
+        } else {
+        map_refcount_.Insert(result, {1, id});
+      }
```
jblomer: Indentation a bit clobbered

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+    }
+    }
+    return result;
+    }
+
+  int Close(int fd) {
```
jblomer: Indentation clobbered.

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+        if (refc_info->first > 0) {
+          map_refcount_.Insert(fd, {refc_info->first -1, refc_info->second});
+          retval = 0;
+        } else {
+          retval = close(fd);
+          map_fd_.Erase(refc_info->second);
+          map_refcount_.Erase(fd);
+        }
+    } else {
+        retval = close(fd);
```
jblomer: This should probably never happen. We should print a warning to syslog.

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,126 @@
...
+    return result;
+    }
+
+  int Close(int fd) {
+    int retval = -1;
+    {
+      MutexLockGuard lock_guard(lock_cache_refcount_);
+      std::pair<int, shash::Any>* refc_info = new std::pair<int, shash::Any>();
+      if (map_refcount_.Lookup(fd, refc_info)) {
+        if (refc_info->first > 0) {
```
jblomer: The refcount must never reach 0.
```suggestion
        if (refc_info->first > 1) {
```

jblomer (changes requested):
Cool!  Some comments inline.

cvmfs/cache_posix_refc.h
vvolkl: Hmmm, this is something I missed - does that mean we need an extra if clause or use this static function [here](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/talk.cc#L559) and [here](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/mountpoint.cc#L895)?

cvmfs/fd_refcount_mgr.h
vvolkl: To me this looked like it should be ok: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/smallhash.h#L285
But good point about the input parameters being const ref, will fix.

jblomer: Ah, you're right. Copy construction is already handled correctly.

vvolkl:
Making `cvmfs_config reload` work when going from the regular to the refcounted cache manager required a few more fixes. The ["magic number"](https://github.com/cvmfs/cvmfs/pull/3288/files#diff-ad5cdb74991ea7372ccdd11a5b85f20162d2c8f40db77f89996101384ded4450R76) in the saved state of the new cache manager seems a bit hacky, but I'm not sure if the dummy saved state of the old cache manager can be easily changed and I'm not sure there is another way to distinguish between the two?

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

cvmfs/fd_refcount_mgr.h
```diff
@@ -23,9 +23,16 @@ static inline uint32_t hasher_int(const int &key) {
   return MurmurHash2(&key, sizeof(key), 0x07387a4f);
 }
 
-
 class FdRefcountMgr {
  public:
+  /**
+   * Helper class containing the values for the map: fd -> refcount+id
+  */
+  struct FdRefcountInfo {
```
jblomer, line 30: Just for good measure, let's add a default constructor that sets recount to -1

cvmfs/fd_refcount_mgr.h
```diff
@@ -117,7 +117,10 @@ class FdRefcountMgr {
           map_refcount_.Erase(fd);
         }
       } else {
-      retval = close(fd);
+        // fd not present in our table - this should never happen!
+        LogCvmfs(kLogCache, kLogWarning, "WARNING: trying to close fd that "
+                                         " is not in refcount tables");
```
jblomer: The destination should be `kLogSyslogWarn | kLogDebug`

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
+/**
+ * This file is part of the CernVM File System.
+ *
```
jblomer: Remove empty comment line

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
...
+}
+
+SmallHashDynamic<shash::Any, int> FdRefcountMgr::GetFdMap() const {
+  return map_fd_;
+}
+
+SmallHashDynamic<int, FdRefcountMgr::FdRefcountInfo>
+    FdRefcountMgr::GetRefcountMap() const {
+  return map_refcount_;
+}
```
jblomer: Ok to keep them here, in general I'd say one-liners can stay in the header.

test/src/099-reusefd/main
```diff
@@ -6,7 +6,47 @@ cvmfs_test_suites="quick"
...
+  echo $(ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2})
+  if [ "$num_open_fds" -lt "2" ]; then
+    echo "The normal cache manager has $num_open_fds open file descriptors for the same open test file";
+    return 21
+  else
+    echo "The normal cache manager has $num_open_fds open file descriptors for the same open test file";
+  fi
+
+  jobs -p | xargs kill
+  sleep 1;
```
jblomer: Remove semicolon

cvmfs/cache_posix.h
```diff
@@ -163,6 +162,14 @@ class PosixCacheManager : public CacheManager {
    */
   atomic_int32 no_inflight_txns_;
 
+
+
+
+
```
jblomer: Remove extra empty lines

cvmfs/mountpoint.cc
jblomer, line 1: I think we should have a static function in the `CacheManager` class called, e.g., `IsPosixCompatible(CacheManagerIds id)` that returns true for the both IDs. We should then probably also use it in talk.cc where we test for `kPosixCacheManager`.

jblomer (changes requested):
I went through the commits one-by-one, I'll now take a complete look again.

I think cling-tidy also found something.

test/src/103-reloadcachemgr/main
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+cvmfs_test_name="Check that open cache file descriptors from different processes are not duplicated in cvmfs"
+cvmfs_test_suites="quick"
+
+cleanup() {
+  jobs -p | xargs kill || true
```
jblomer: I think we should also check with wait if the tail processes quit without error.

test/src/103-reloadcachemgr/main
```diff
@@ -0,0 +1,43 @@
...
+}
+
+cvmfs_run_test() {
+  logfile=$1
+
```
jblomer:
```suggestion
```

test/src/103-reloadcachemgr/main
```diff
@@ -0,0 +1,43 @@
...
+
+  ############### 1) Check that the normal cache manager 
+  ############### creates new fds for the same object 
+  cvmfs_mount alice.cern.ch CVMFS_CACHE_REFCOUNT=no  || return 10
+
+
```
jblomer, line 20:
```suggestion
```

test/src/103-reloadcachemgr/main
```diff
@@ -0,0 +1,43 @@
...
+  done;
+  trap cleanup EXIT HUP INT TERM
+  sleep 1;
+
+
+  sudo sh -c "echo \"CVMFS_CACHE_REFCOUNT=yes\" >> /etc/cvmfs/default.local" || return 2
+  
+  sudo cvmfs_config reload || return 20
+
+
```
jblomer:
```suggestion
```

test/src/103-reloadcachemgr/main
```diff
@@ -0,0 +1,43 @@
...
+  testfile=$(find /cvmfs/alice.cern.ch -type f | head -n1)
+  # testfile has the following cvmfs hash:
+  testfile_hash=$(get_xattr hash $testfile)
+
+  for i in {1..10}; do
+    tail -f ${testfile} > /dev/null 2>&1 &
+  done;
+  trap cleanup EXIT HUP INT TERM
+  sleep 1;
+
```
jblomer, line 39:
```suggestion
```

test/src/103-reloadcachemgr/main
```diff
@@ -0,0 +1,43 @@
...
+
+  sudo sh -c "echo \"CVMFS_CACHE_REFCOUNT=yes\" >> /etc/cvmfs/default.local" || return 2
+  
+  sudo cvmfs_config reload || return 20
+
+
+
+  cleanup
```
jblomer: Not needed, covered by the `trap`
```suggestion
```

test/src/103-reloadcachemgr/main
jblomer, line 1: Should we also test the reverse, from recounting to not refcounting?

test/src/103-reloadcachemgr/main
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+cvmfs_test_name="Check that open cache file descriptors from different processes are not duplicated in cvmfs"
```
jblomer:
```suggestion
cvmfs_test_name="Check changing CVMFS_CACHE_REFCOUNT between reloads"
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+cvmfs_test_name="Check that open cache file descriptors from different processes are not duplicated in cvmfs"
+cvmfs_test_suites="quick"
+
+cleanup() {
+  jobs -p | xargs kill || true
```
jblomer: We should also check that the processes quit without error.

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
...
+  done;
+  trap cleanup EXIT HUP INT TERM
+  sleep 1;
+
+  ###########
+  # check the open file descriptors of the cvmfs2 process 
+  alice_pid=$(sudo cvmfs_talk -i alice.cern.ch pid)
+  num_open_fds=$(ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2} | wc -l)
+
+
```
jblomer:
```suggestion
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
...
+  ############### 1) Check that the normal cache manager 
+  ############### creates new fds for the same object 
+  cvmfs_mount alice.cern.ch CVMFS_CACHE_REFCOUNT=no  || return 10
+
+  ############
+  # open some test file in several simultaneously in different processes
+
+  testfile=$(find /cvmfs/alice.cern.ch -type f | head -n1)
```
jblomer:
```suggestion
  local testfile=$(find /cvmfs/alice.cern.ch -type f | head -n1)
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
...
+  # testfile has the following cvmfs hash:
+  testfile_hash=$(get_xattr hash $testfile)
```
jblomer:
```suggestion
  local testfile_hash=$(get_xattr hash $testfile)
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
...
+
+  for i in {1..10}; do
+    tail -f ${testfile} > /dev/null 2>&1 &
+  done;
+  trap cleanup EXIT HUP INT TERM
+  sleep 1;
+
+  ###########
+  # check the open file descriptors of the cvmfs2 process 
+  alice_pid=$(sudo cvmfs_talk -i alice.cern.ch pid)
```
jblomer:
```suggestion
  local alice_pid=$(sudo cvmfs_talk -i alice.cern.ch pid)
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
...
+  num_open_fds=$(ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2} | wc -l)
```
jblomer:
```suggestion
  local num_open_fds=$(ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2} | wc -l)
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
...
+    tail -f ${testfile} > /dev/null 2>&1 &
+  done;
+  trap cleanup EXIT HUP INT TERM
+  sleep 1;
+
+  ###########
+  # check the open file descriptors of the cvmfs2 process 
+  alice_pid=$(cvmfs_talk -i alice.cern.ch pid)
+  num_open_fds=$(ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2} | wc -l)
+
```
jblomer, line 73:
```suggestion
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,82 @@
...
+
+    echo "The refcount cache manager has $num_open_fds open file descriptors for the same open test file"
+  if [ "$num_open_fds" -gt "1" ]; then
+    return 40
+  fi
+
+  cleanup
```
jblomer: Covered by the trap
```suggestion
```

vvolkl:
@cernvm-bot cloudtest

jblomer (changes requested):
This is the full round of the tests.  I'll now turn to the code, looking at the latest state.

cvmfs/cache.cc
```diff
@@ -194,10 +194,13 @@ int CacheManager::RestoreState(const int fd_progress, void *data) {
...
     abort();
   }
-  if (state->manager_type != id()) {
-    if (fd_progress >= 0)
-      SendMsg2Socket(fd_progress, "switching cache manager unsupported!\n");
-    abort();
+  // exception: we do allow switching from the non-refcounted
+  // to the refcounted posix cache manager
+  if (IsReloadCompatible(state->manager_type, id())) {
+      if (fd_progress >= 0) {
```
jblomer: Whitespacing

cvmfs/cache.cc
```diff
@@ -194,10 +194,13 @@ int CacheManager::RestoreState(const int fd_progress, void *data) {
       SendMsg2Socket(fd_progress, "unsupported state version!\n");
     abort();
   }
-  if (state->manager_type != id()) {
-    if (fd_progress >= 0)
-      SendMsg2Socket(fd_progress, "switching cache manager unsupported!\n");
-    abort();
+  // exception: we do allow switching from the non-refcounted
+  // to the refcounted posix cache manager
+  if (IsReloadCompatible(state->manager_type, id())) {
```
jblomer:
```suggestion
  if (!IsReloadCompatible(state->manager_type, id())) {
```

We should also add a comment before on the test of the state version.  It probably works because both the regular and the refcounted manager use version 0. Now that we can switch between them, we have entangled their state versions.

cvmfs/mountpoint.cc
```diff
@@ -682,17 +686,26 @@ CacheManager *FileSystem::SetupExternalCacheMgr(const string &instance) {
 }
 
 
-CacheManager *FileSystem::SetupPosixCacheMgr(const string &instance) {
+CacheManager *FileSystem::SetupPosixCacheMgr(const string &instance,
+                                             bool use_refcount) {
   PosixCacheSettings settings = DeterminePosixCacheSettings(instance);
+  CacheManager* cache_mgr;
```
jblomer: I think this needs to be a `UniquePtr<CacheManager>`. Or we need to call `Release()` on the object returned from `[PosixCacheManager|PosixRefcountCacheManager]::Create()`. Otherwise the cache manager will be immediately deconstructed.

cvmfs/cache_posix.cc
```diff
@@ -228,38 +228,27 @@ int PosixCacheManager::CommitTxn(void *txn) {
...
 
   if (fs_info.type == kFsTypeTmpfs) {
-    cache_manager->is_tmpfs_ = true;
+    is_tmpfs_ = true;
   }
 
-  if (cache_manager->alien_cache_) {
+  if (alien_cache_) {
     if (!MakeCacheDirectories(cache_path, 0770)) {
       return NULL;
```
jblomer: Here and further down: `return NULL` --> `return false`

cvmfs/cache_posix_refc.h
```diff
@@ -0,0 +1,58 @@
...
+  virtual CacheManagerIds id() { return kPosixRefcountCacheManager; }
+
+ protected:
+  struct SavedState {
+    SavedState() : version(0), magic_number(32123), fd_mgr(NULL) { }
+    unsigned int version;
+    /// this helps to distinguish from the SavedState of the normal
+    /// posix cache manager
+    unsigned int magic_number;
```
jblomer: Since the payload of the regular cache manager is just a char, let's use a char here as well. Nice idea, btw!

cvmfs/cache_posix_refc.cc
```diff
@@ -0,0 +1,101 @@
...
+  SavedState *state = new SavedState();
+  state->fd_mgr = fd_mgr.Clone();
+  return state;
+}
+
+
+int PosixRefcountCacheManager::DoRestoreState(void *data) {
+  fd_mgr = FdRefcountMgr();
+  assert(data);
+    SavedState *state = reinterpret_cast<SavedState *>(data);
```
jblomer: whitespace

cvmfs/cache_posix_refc.cc
```diff
@@ -0,0 +1,101 @@
...
+  if (state->magic_number == 32123) {
+      LogCvmfs(kLogCache, kLogDebug, "Restoring refcount cache manager from "
```
jblomer: whitespace

cvmfs/cache_posix_refc.cc
```diff
@@ -0,0 +1,101 @@
...
+                                     "refcounted posix cache manager");
+
+    fd_mgr.AssignFrom(state->fd_mgr);
+  } else {
+      LogCvmfs(kLogCache, kLogDebug, "Restoring refcount cache manager from "
```
jblomer: whitespace

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+             reinterpret_cast<const uint32_t *>(key.digest) + 1);
+}
+static inline uint32_t hasher_int(const int &key) {
+  return MurmurHash2(&key, sizeof(key), 0x07387a4f);
+}
+
+class FdRefcountMgr {
+ public:
+  /**
+   * Helper class containing the values for the map: fd -> refcount+id
```
jblomer, line 28: whitespace

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+
+  int Close(int fd);
+
+  FdRefcountMgr* Clone();
+
+ private:
+  /** map for fd -> refcount lookups. A backreference
+  * to the object id is included in FdRefcountInfo in order 
+  * to be able to remove the file descriptor from map_fd_.
+  */
```
jblomer:
```suggestion
  /** 
   * map for fd -> refcount lookups. A backreference
   * to the object id is included in FdRefcountInfo in order 
   * to be able to remove the file descriptor from map_fd_.
   */
```

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+  SmallHashDynamic<int, FdRefcountInfo> map_refcount_;
+  /** map for object id -> fd lookups, used when 
+   * opening files in the cache. The fd is used as key
+   * in the refcount map.
+  */
```
jblomer:
```suggestion
  /** 
   * map for object id -> fd lookups, used when 
   * opening files in the cache. The fd is used as key
   * in the refcount map.
   */
```

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
...
+      result = open(path.c_str(), O_RDONLY);
+      if (result >= 0) {
+        map_fd_.Insert(id, result);
+      }
+    }
+    if (result >= 0) {
+      FdRefcountInfo refc_info;
+      if (map_refcount_.Lookup(result, &refc_info)) {
+        refc_info.refcount++;
+        map_refcount_.Insert(result, refc_info);
```
jblomer: This and line 31 can be moved and merged after the if-else block

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
...
+
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <string>
+
+
+int FdRefcountMgr::Open(const shash::Any id, const std::string& path) {
+  int result = -1;
+  {
```
jblomer: I don't think we need the extra block since it spans all of the function.

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
...
+        map_refcount_.Insert(result, refc_info);
+      }
+    }
+  }
+  return result;
+}
+
+int FdRefcountMgr::Close(int fd) {
+  int retval = -1;
+  {
```
jblomer: No need for the extra block

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
...
+        retval = 0;
+      } else {
+        retval = close(fd);
+        map_fd_.Erase(refc_info.id);
+        map_refcount_.Erase(fd);
+      }
+    } else {
+      // fd not present in our table - this should only happen
+      // when reloading from the normal posix cache manager!
+      LogCvmfs(kLogCache, kLogSyslogWarn | kLogDebug,
```
jblomer: Actually, since this is now an expected result (moving from regular to refcounted cache manager), it should only be a debug log
```suggestion
      LogCvmfs(kLogCache, kLogDebug,
```

cvmfs/cache_posix_refc.h
jblomer: Yep, use the static function there

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+
+  FdRefcountMgr(
+        const SmallHashDynamic<int, FdRefcountInfo> &map_refcount,
+        const SmallHashDynamic<shash::Any, int> &map_fd);
+
+  ~FdRefcountMgr();
+
+  void AssignFrom(FdRefcountMgr *other);
+
+  SmallHashDynamic<int, FdRefcountInfo>* GetRefcountMap();
```
jblomer:
```suggestion
  SmallHashDynamic<int, FdRefcountInfo>* GetRefcountMapPtr();
```

cvmfs/fd_refcount_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+
+  SmallHashDynamic<shash::Any, int> * GetFdMap();
```
jblomer:
```suggestion
  SmallHashDynamic<shash::Any, int>* GetFdMapPtr();
```

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
...
+
+FdRefcountMgr::~FdRefcountMgr() {
+  pthread_mutex_destroy(lock_cache_refcount_);
+  free(lock_cache_refcount_);
+}
+
+FdRefcountMgr::FdRefcountMgr() {
+  const shash::Any hash_null;
+  map_fd_.Init(16, hash_null, hasher_any);
+  map_refcount_.Init(16, 0, hasher_int);
```
jblomer: Since zero can be a valid file descriptor, use
```suggestion
  map_refcount_.Init(16, -1, hasher_int);
```

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,112 @@
...
+  assert(retval == 0);
+}
+
+FdRefcountMgr::FdRefcountMgr(
+    const SmallHashDynamic<int, FdRefcountMgr::FdRefcountInfo> &map_refcount,
+                            const SmallHashDynamic<shash::Any, int> &map_fd)
+{
+  const shash::Any hash_null;
+  map_fd_.Init(16, hash_null, hasher_any);
+  map_refcount_.Init(16, 0, hasher_int);
```
jblomer:
```suggestion
  map_refcount_.Init(16, -1, hasher_int);
```

test/src/103-reloadcachemgr/main
vvolkl: We expect an error there, right? What would you test for?

jblomer (changes requested):
Looks good, I think we get close! My only remaining question is about the transition from refcounted to regular cache. Currently we crash, which is one option. In this way, however, we cannot turn off the refcount easily (we can unset the option but then we risk crashes during reload).

I don't have a good solution yet.

cvmfs/cache.cc
vvolkl:
```suggestion
  if (!IsReloadCompatible(state->manager_type, id())) {
```
typo

cvmfs/mountpoint.cc
vvolkl: `Create` does already `Release`: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/cache_posix.cc#L281

jblomer: Ah, of course

vvolkl:
> Looks good, I think we get close! My only remaining question is about the transition from refcounted to regular cache. Currently we crash, which is one option. In this way, however, we cannot turn off the refcount easily (we can unset the option but then we risk crashes during reload).
> 
> I don't have a good solution yet.

Now merged the inherited class into the PosixCacheManager, it already survives the reload from refcounted -> non-refcounted, but does not correctly keep the fd maps in that case. I don't know why the test passes, it probably needs  more sophisticated test logic to catch problems caused by closed file descriptors.

vvolkl:
@cernvm-bot cloudtest

cvmfs/cache.h
jblomer, line 1: I guess we don't need the extra ID anymore then, and also not the `IsPosixCompatible` and `IsReloadCompatible` helpers.

cvmfs/cache_posix.h
```diff
@@ -172,6 +195,11 @@ class PosixCacheManager : public CacheManager {
    * True if posixcache is on tmpfs (and with this already in RAM)
    */
   bool is_tmpfs_;
+  /**
+   * Refcount and return only unique file descriptors
+   */
+  const bool do_refcount_;
+  FdRefcountMgr* fd_mgr;
```
jblomer: Make it a UniquePtr so that you don't need extra handling in the destructor

vvolkl:
@cernvm-bot cloudtest

cvmfs/mountpoint.cc
```diff
@@ -684,15 +691,16 @@ CacheManager *FileSystem::SetupExternalCacheMgr(const string &instance) {
 
 CacheManager *FileSystem::SetupPosixCacheMgr(const string &instance) {
   PosixCacheSettings settings = DeterminePosixCacheSettings(instance);
+  CacheManager* cache_mgr;
   if (!CheckPosixCacheSettings(settings))
     return NULL;
-
-  UniquePtr<PosixCacheManager> cache_mgr(PosixCacheManager::Create(
+  cache_mgr = PosixCacheManager::Create(
```
jblomer: Let's keep it in a unique pointer. The advantage is that it will be properly cleaned up in the error paths, e.g. if the quota manager initialization fails in line 717.

test/common/openwaitprint.c
```diff
@@ -0,0 +1,62 @@
+/**
+ * This file is part of the CernVM File System.
+ * Helper for integration test 690 (streaming cache manager)
```
jblomer:
```suggestion
 * Helper for cache manager integration tests, e.g. 690 (streaming cache manager)
```

cvmfs/cache.cc
```diff
@@ -194,10 +194,13 @@ int CacheManager::RestoreState(const int fd_progress, void *data) {
       SendMsg2Socket(fd_progress, "unsupported state version!\n");
     abort();
   }
+  // exception: we do allow switching from the non-refcounted
+  // to the refcounted posix cache manager
```
jblomer: I think this comment is now obsolete
```suggestion
```

cvmfs/cache.cc
```diff
@@ -194,10 +194,13 @@ int CacheManager::RestoreState(const int fd_progress, void *data) {
...
   if (state->manager_type != id()) {
-    if (fd_progress >= 0)
+    if (fd_progress >= 0) {
       SendMsg2Socket(fd_progress, "switching cache manager unsupported!\n");
     abort();
+    }
```
jblomer: This changes the semantics subtly, should be reverted

cvmfs/cache_posix.h
```diff
@@ -109,8 +111,8 @@ class PosixCacheManager : public CacheManager {
   virtual void *DoSaveState();
   virtual int DoRestoreState(void *data);
   virtual bool DoFreeState(void *data);
+  bool InitCacheDirectory(const string &cache_path);
 
- private:
```
jblomer, line 113: We may now keep the `private` modifier at this point I guess.

cvmfs/cache_posix.h
```diff
@@ -134,21 +136,24 @@ class PosixCacheManager : public CacheManager {
...
+                    const bool do_refcount = false)
     : cache_path_(cache_path)
     , txn_template_path_(cache_path_ + "/txn/fetchXXXXXX")
     , alien_cache_(alien_cache)
     , rename_workaround_(kRenameNormal)
     , cache_mode_(kCacheReadWrite)
     , reports_correct_filesize_(true)
     , is_tmpfs_(false)
+    , do_refcount_(do_refcount)
+    , fd_mgr(0)
```
jblomer:
```suggestion
    , fd_mgr(new FdRefcountMgr())
```

cvmfs/cache_posix.h
```diff
@@ -172,6 +189,11 @@ class PosixCacheManager : public CacheManager {
    * True if posixcache is on tmpfs (and with this already in RAM)
    */
   bool is_tmpfs_;
+  /**
+   * Refcount and return only unique file descriptors
+   */
+  bool do_refcount_;
+  UniquePtr<FdRefcountMgr> fd_mgr;
```
jblomer: Style:
```suggestion
  UniquePtr<FdRefcountMgr> fd_mgr_;
```

cvmfs/cache_posix.h
```diff
@@ -134,21 +136,24 @@ class PosixCacheManager : public CacheManager {
...
     , alien_cache_(alien_cache)
     , rename_workaround_(kRenameNormal)
     , cache_mode_(kCacheReadWrite)
     , reports_correct_filesize_(true)
     , is_tmpfs_(false)
+    , do_refcount_(do_refcount)
+    , fd_mgr(0)
   {
     atomic_init32(&no_inflight_txns_);
+    fd_mgr = new FdRefcountMgr();
```
jblomer:
```suggestion
```

cvmfs/cache_posix.h
```diff
@@ -163,6 +168,18 @@ class PosixCacheManager : public CacheManager {
...
+  struct SavedState {
+    SavedState() : magic_number(123), version(0), fd_mgr(NULL) { }
+    /// this helps to distinguish from the SavedState of the normal
+    /// posix cache manager
+    char magic_number;
+    unsigned int version;
+    UniquePtr<FdRefcountMgr> fd_mgr;
+  };
+  int Rename(const char *oldpath, const char *newpath);
+  int Flush(Transaction *transaction);
```
jblomer: Can also go back to the old place (together with the revert of the `private` modified)

cvmfs/cache_posix.cc
```diff
@@ -138,7 +138,12 @@ bool PosixCacheManager::AcquireQuotaManager(QuotaManager *quota_mgr) {
 
 
 int PosixCacheManager::Close(int fd) {
-  int retval = close(fd);
+  int retval;
+  if (do_refcount_) {
+    retval = fd_mgr->Close(fd);
+  } else {
+    retval = close(fd);
+  }
```
jblomer: Perhaps:
```suggestion
  int retval = do_refcount_ ? fd_mgr->Close(fd) : close(fd);
```

cvmfs/cache_posix.h
```diff
@@ -163,6 +168,18 @@ class PosixCacheManager : public CacheManager {
    */
   atomic_int32 no_inflight_txns_;
 
+ private:
+  struct SavedState {
+    SavedState() : magic_number(123), version(0), fd_mgr(NULL) { }
```
jblomer: Let's make the magic number a named constant, like `static const char kMagicRefcount = 123;`

cvmfs/cache_posix.cc
```diff
@@ -310,14 +335,44 @@ void *PosixCacheManager::DoSaveState() {
...
-  free(data);
+  assert(data);
+  SavedState *state = reinterpret_cast<SavedState *>(data);
+  if (state->magic_number == 123) {
+    delete state;
+  } else {
+    // this should be the dummy SavedState
+    // of the regular posix cache manager
+    free(data);
+  }
```
jblomer: In both cases we need to `free(data)`
```suggestion
  } 
  // If not refcounted, the state is the dummy SavedState
  // of the regular posix cache manager
  free(data);
```

cvmfs/cache_posix.cc
```diff
@@ -310,14 +335,44 @@ void *PosixCacheManager::DoSaveState() {
...
+    } else {
+      LogCvmfs(kLogCache, kLogDebug, "Restoring refcount cache manager from "
+                                    "non-refcounted posix cache manager");
+    }
+    return -1;
+  }
+
   char *c = reinterpret_cast<char *>(data);
-  assert(*c == '\0');
+  assert(*c == '\0' || *c == 123);
```
jblomer: Maybe also add a named constant for the "magic number" of the regular cache manager.

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,107 @@
...
+
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <string>
+
+
+int FdRefcountMgr::Open(const shash::Any id, const std::string& path) {
+  int result = -1;
+  MutexLockGuard lock_guard(lock_cache_refcount_);
```
jblomer, line 18: Include util/mutex.h for the `MutexLockGuard`

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,107 @@
...
+}
+
+FdRefcountMgr::FdRefcountMgr() {
+  const shash::Any hash_null;
+  map_fd_.Init(16, hash_null, hasher_any);
+  map_refcount_.Init(16, -1, hasher_int);
+  lock_cache_refcount_ =
+      reinterpret_cast<pthread_mutex_t*>(smalloc(sizeof(pthread_mutex_t)));
+  int retval = pthread_mutex_init(lock_cache_refcount_, NULL);
+  assert(retval == 0);
```
jblomer, line 113: Include cassert header

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,107 @@
...
+  pthread_mutex_destroy(lock_cache_refcount_);
+  free(lock_cache_refcount_);
+}
+
+FdRefcountMgr::FdRefcountMgr() {
+  const shash::Any hash_null;
+  map_fd_.Init(16, hash_null, hasher_any);
+  map_refcount_.Init(16, -1, hasher_int);
+  lock_cache_refcount_ =
+      reinterpret_cast<pthread_mutex_t*>(smalloc(sizeof(pthread_mutex_t)));
```
jblomer, line 111: Include smalloc.h

test/src/102-reusefd/main
```diff
@@ -0,0 +1,91 @@
...
+  cvmfs_mount alice.cern.ch CVMFS_CACHE_REFCOUNT=no  || return 10
+
+  ############
+  # open some test file in several simultaneously in different processes
+
+  local testfile=$(find /cvmfs/alice.cern.ch -type f | head -n1)
+  # testfile has the following cvmfs hash:
+  local testfile_hash=$(get_xattr hash $testfile)
+
+    gcc -Wall -pthread -o openwaitprint $src_location/openwaitprint.c || return 60
```
jblomer:
```suggestion
   gcc -Wall -pthread -o openwaitprint $src_location/openwaitprint.c || return 60
```

test/src/102-reusefd/main
```diff
@@ -0,0 +1,91 @@
...
+  local pid_helper3=$!
+
+
+  ###########
+  # check the open file descriptors of the cvmfs2 process 
+  local alice_pid=$(sudo cvmfs_talk -i alice.cern.ch pid)
+  num_open_fds=$(sudo ls -l /proc/$alice_pid/fd/ | grep /var/lib/cvmfs/shared/${testfile_hash:0:2}/${testfile_hash:2} | wc -l)
+
+  if [ "$num_open_fds" -lt "2" ]; then
+    echo "The normal cache manager has $num_open_fds open file descriptors for the same open test file";
```
jblomer: Move out of the if-else block (and remove the else block)

test/src/103-reloadcachemgr/main
jblomer, line 1: After the reload operations, also add a test for reading a complete file (`cat $testfile`). So to test a complete open-read-close cycle on the new (restored) cache manager.

jblomer (changes requested):
This looks pretty good! Some minor comments.

jblomer:
There are actually two more methods we overlooked and which should ideally take care of refcounting: `Dup()` and `OpenFromTxn()`. Perhaps we can also add at least a basic unit test in `t_cache.cc` for the refcounting. The unit test may, e.g., access the `fd_mgr_` to check if things work as expected.

vvolkl:
Now just missing unittests

@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

cvmfs/cache_posix.cc
```diff
@@ -293,38 +301,80 @@ void PosixCacheManager::CtrlTxn(
...
 bool PosixCacheManager::DoFreeState(void *data) {
+  assert(data);
+  SavedState *state = reinterpret_cast<SavedState *>(data);
+  if (state->magic_number == kMagicRefcount) {
+    delete state;
+  } else {
+  // If not refcounted, the state is the dummy SavedState
+  // of the regular posix cache manager
   free(data);
+  }
```
jblomer: I think here something got accidentally reverted
```suggestion
  }
  // If not refcounted, the state is the dummy SavedState
  // of the regular posix cache manager
  free(data);
```

cvmfs/cache_posix.cc
```diff
@@ -293,38 +301,80 @@ void PosixCacheManager::CtrlTxn(
...
+      LogCvmfs(kLogCache, kLogDebug, "Restoring refcount cache manager from "
+                                    "non-refcounted posix cache manager");
+    }
+    return -1;
+  }
+
   char *c = reinterpret_cast<char *>(data);
-  assert(*c == '\0');
+  assert(*c == kMagicNoRefcount || *c == kMagicRefcount);
+  if (*c == 123) {
```
jblomer:
```suggestion
  if (*c == kMagicRefcount) {
```

cvmfs/fd_refcount_mgr.cc
```diff
@@ -0,0 +1,128 @@
...
+  FdRefcountInfo refc_info;
+  if (map_refcount_.Lookup(fd, &refc_info)) {
+      refc_info.refcount += 1;
+      map_refcount_.Insert(fd, refc_info);
+      retval = fd;
+  } else {
+    // fd not present in our table - this should never happen!
+    LogCvmfs(kLogCache, kLogWarning | kLogSyslogWarn,
+              "WARNING: trying to dup fd that "
+              " is not in refcount tables");
```
jblomer: Can't this happen when we upgrade from regular to refcounted manager? Probably not in the way we use `Dup` currently, but in principle we may use the API in such a way at some point.
```suggestion
    // fd not present in our table, may
    LogCvmfs(kLogCache, kLogDebug,
              "WARNING: trying to dup fd that "
              " is not in refcount tables");
```

jblomer (changes requested):
Some final comments.

cvmfs/cache_posix.cc
vvolkl: Ah, no, I haven't made this clear, but this is intentional, the reload does not work otherwise. I think it's because only the non-refcounted path in `DoSaveState` calls `smalloc`.

jblomer: Oh, of course! One version calls `new`, the other one `malloc`. Very well then. Then it just requires fixing the indentation in the else block.

test/cloud_testing/platforms/centos7_x86_64-EXCLCACHE_test.sh
jblomer, line 1: On the RAM cache, I understand.  But the exclusive cache manager test should work. That affects only the quota manager in fact and the question whether or not it is shared.

vvolkl: OK, then I need to double check what is going wrong here: https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/1855/CVMFS_PLATFORM=el7,CVMFS_PLATFORM_CONFIG=x86_64_exlcache,label=trampoline/testReport/junit/(root)/ClientIntegrationTests/Check_that_open_cache_file_descriptors_from_different_processes_are_not_duplicated_in_cvmfs/

vvolkl:
@cernvm-bot cloudtest

test/cloud_testing/platforms/centos7_x86_64-EXCLCACHE_test.sh
vvolkl: just needed a fix in the test logic.

jblomer:
The [macOS](https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsUnittest/1829/) unit test failures might be real.

vvolkl:
@cernvm-bot unittest

jblomer (approved):

PR #3289, opened 2023-06-12
Use the ISO ordering for the date in cvmfs_server_gc.sh

jmcarcell:
It's less confusing. This only changes the ordering of the date, the rest should remain the same most of the time (I don't know if changing the locale the format would be different though). With `en_US.UTF-8` it only changes the ordering


vvolkl:
That makes sense to me, thanks!

vvolkl (approved):

vvolkl:
@cernvm-bot cloudtest

jblomer (approved):

PR #3291, opened 2023-06-14
Fix race in parallel catalog traversal

### LINKED ISSUE #3171: CatalogTraversal can get stuck
Multi-threaded version of CatalogTraversal needs investigation as it can hang forever

Can be seen with T_CatalogTraversal/1.TraverseDepthFirstParallelStressSlow does not always succeed. if it fails it hangs forever
with the refactored loadcatalog #3156 this unittest always hangs

see also https://sft.its.cern.ch/jira/browse/CVM-2019

jblomer:
Fixes #3171


jblomer:
@cernvm-bot cloudtest

vvolkl (approved):

PR #3293, opened 2023-06-15
Update filter options in cvmfs_rsync to allow cvmfscatalog files to be created

duncanmmacleod:
This PR modifies the `--filter` options automatically applied in `cvmfs_rsync` to allow `.cvmfs{auto}catalog` files to be _created_ during the rsync operation, but not (accidentally) deleted.

I'm not sure if the `Pp` filter should only apply to `.cvmfscatalog` and not `.cvmfsautocatalog` (i.e. don't allow rsync to create or delete `.cvmfsautocatalog`) - I don't really understand the standard use cases well enough.

cc @DrDaveD

### DIFF
--- a/cvmfs/cvmfs_rsync
+++ b/cvmfs/cvmfs_rsync
@@ -9,7 +9,7 @@ ME=cvmfs_rsync
 usage()
 {
     echo "Usage: $ME [rsync_parameters]" >&2
-    echo "  Rsync to cvmfs repositories.  Avoids .cvmfscatalog and .cvmfsautocatalog" >&2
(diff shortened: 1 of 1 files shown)


DrDaveD:
It is rather unlikely that a source will create any `.cvmfsautocatalog` files, but I think it's easier to explain if the script treats both of the files the same way.

DrDaveD (approved):

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

jblomer:
The failing tests are unrelated.

vvolkl (approved):

PR #3302, opened 2023-07-03
Initialize logging according to options in mount helper

### LINKED ISSUE #2962: Mount helper should use CVMFS_USYSLOG setting
The mount helper does ignore the `CVMFS_USYSLOG` setting. Especially when a mountpoint is attached, it's important to have the mount helper logs in the same place than the logs from the fuse module. Linked to #2956

jblomer:
Fixes #2962


jblomer:
@cernvm-bot cloudtest

vvolkl (approved):

PR #3303, opened 2023-07-03
Warn about potential incorrect use of cvmfs_talk

### LINKED ISSUE #3075: `cvmfs_talk` does not warn if `-i repo` is needed but not given
Commands in `cvmfs_talk` that need `-i repo` finish without any output. This creates confusion if the execution was successful or not. 

It would be better if `cvmfs_talk` gives feedback that arguments are missing.

jblomer:
Fixes #3075

### DIFF
--- a/cvmfs/cvmfs_talk.cc
+++ b/cvmfs/cvmfs_talk.cc
@@ -244,7 +244,7 @@ int main(int argc, char *argv[]) {
       command.push_back(' ');
   }
   if (command.empty()) {
-    LogCvmfs(kLogCvmfs, kLogStderr, "Command missing");
+    Usage(argv[0]);
     return 1;
   }
 
--- a/cvmfs/cvmfs_talk.cc
+++ b/cvmfs/cvmfs_talk.cc
@@ -143,7 +143,8 @@ bool SendCommand(const std::string &command, InstanceInfo instance_info) {
 static void Usage(const std::string &exe) {
   LogCvmfs(kLogCvmfs, kLogStdout,
     "Usage: %s [-i instance | -p socket] <command>                     \n"
-    "   By default, iteratate through all instances.                   \n"
+    "   By default, iteratate through all instances defined in         \n"
+    "   CVMFS_REPOSITORIES                                             \n"
     "\n"
     "Example:                                                          \n"
     "  %s -i atlas.cern.ch pid                                         \n"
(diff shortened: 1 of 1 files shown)


cvmfs/cvmfs_talk.cc
```diff
@@ -143,7 +143,8 @@ bool SendCommand(const std::string &command, InstanceInfo instance_info) {
 static void Usage(const std::string &exe) {
   LogCvmfs(kLogCvmfs, kLogStdout,
     "Usage: %s [-i instance | -p socket] <command>                     \n"
-    "   By default, iteratate through all instances.                   \n"
+    "   By default, iteratate through all instances defined in         \n"
```
vvolkl:
```suggestion
    "   By default, iterate through all instances defined in         \n"
```

cvmfs/cvmfs_talk.cc
```diff
@@ -254,14 +255,23 @@ int main(int argc, char *argv[]) {
...
+      is_empty_repo_list = false;
       instance_info.instance_name = repos[i];
       LogCvmfs(kLogCvmfs, kLogStdout, "%s:", repos[i].c_str());
       bool retval = SendCommand(command, instance_info);
       if (!retval) retcode = 1;
     }
+    if (is_empty_repo_list) {
+      LogCvmfs(kLogCvmfs, kLogStdout,
+               "Warning: no instance was specified. In this case, the command "
+               "is executed for all instances defined in CVMFS_REPOSITORES but "
```
HereThereBeDragons: typo in CVMFS_REPOSITORIES

cvmfs/cvmfs_talk.cc
```diff
@@ -254,14 +255,23 @@ int main(int argc, char *argv[]) {
...
+               "this list is empty. Did you mean running \n\n"
+               "    cvmfs_talk -i <repository name> <command>\n");
```
HereThereBeDragons, line 273: do you also want to add the option with `-p <socket>`?

HereThereBeDragons (changes requested):
only a small but important typo

during testing i also found the following: 
```
cvmfs_talk
> Command missing
```

so maybe we can add that just calling cvmfs_talk falls back to `--help`? (could be another PR)

cvmfs/cvmfs_talk.cc
jblomer: I think I'd rather not advertise the `-p` interface. Under normal circumstances, `-i` should be used.

vvolkl (approved):

Issue #3213, opened 2023-02-23
HTTPS connections always trigger authz helper


jblomer:
The authz client authentication should _only_ be used if `CVMFS_AUTHZ_HELPER` is set. Currently, there is an unconditional (failed) attempt to use the authz helper when the client connects through HTTPS.

HereThereBeDragons:
Should be part of download manager refactor #3095

HereThereBeDragons:
#3333 will most likely superseed #3307

HereThereBeDragons:
#3333 is the merged PR

PR #3307, opened 2023-07-05
Allow authz helper to be ignored for HTTPS

HereThereBeDragons:
Issue #3213 

Authz is always created, but only used for HTTPS downloads. 
As such change `mountpoint.cc` that no Authz is created when no CVMFS param is given

Change of behavior:
`CVMFS_AUTHZ_HELPER` must always be given now. If the default authz helper should be used (found automatically by cvmfs) then `CVMFS_AUTHZ_HELPER=default` must be set (as such a custom authz helper cannot be called "default")

Still needs verification with a cloud test

### DIFF
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1283,8 +1283,15 @@ MountPoint *MountPoint::Create(
 void MountPoint::CreateAuthz() {
   string optarg;
   string authz_helper;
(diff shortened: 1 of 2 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

jblomer:
Replaced by #3333

PR #3145, opened 2022-12-21
HTTP Tracing

mharvey-jt:
When enabled, this feature adds additional headers to to all HTTP requests, allowing tracing of requests from FUSE request through HTTP caching layers. New configuration options:
```
CVMFS_HTTP_TRACING=on  (default off)
CVMFS_HTTP_TRACING_HEADERS="X-CVMFS-key1:  val1|X-CVMFS-key2: val2|...."     #'|'-delimited list of headers to include in trace
```
The headers added are:
```
X-CVMFS-PID: pid of process originating the fuse request
X-CVMFS-GID: gid "
X-CVMFS-UID: uid "
```
as well as any defined in the `CVMFS_HTTP_TRACING_HEADERS` configuration option

Testing this has the same challenge as https://github.com/cvmfs/cvmfs/pull/3102 - we'd need to log the headers. Could possibly do this by adding a feature to log all http requests at the Curl level with https://curl.se/libcurl/c/CURLOPT_VERBOSE.html

Issue #3094, opened 2022-11-08
Custom headers on HTTP requests


mharvey-jt:
For observability purposes we would like to be able to annotate all HTTP requests with headers that indicate the source of read. This will be a mix of statically-assigned config data, and request-specific, eg the user id, pid, which are available from the fuse request.

eg in config file
```
CUSTOM_HEADERS=h1=foo1,h2=foo2
```
and headers in the HTTP request: 
```
X-CVMFS-HEADER-h1: foo1
X-CVMFS-HEADER-h2: foo2
X-CVMFS-HEADER-uid: <uid>
X-CVMFS-HEADER-gid: <gid>
X-CVMFS-HEADER-pid: <pid>
```

PR #3308, opened 2023-07-06
Custom http tracing headers

HereThereBeDragons:
Issue #3094 

Superseeds #3145 

client params:
```bash
CVMFS_HTTP_TRACING=on #(default off)

CVMFS_HTTP_TRACING_HEADERS='h1:test|CVMFS-X-h2:ff|X-CVMFS-h3:12_ad |  h4  : 12fs_?'
# only illegal header is: CVMFS-X-h2:ff
# all other will be expanded to valid ones 


# example debug output
(download) CURL Header for URL: /data/81/7c882d4a2e9dd7f9c5c2bfb4e04ff316e436dfC is:
 Connection: Keep-Alive
Pragma:
User-Agent: cvmfs Fuse 2.11.0
X-CVMFS-h1: test
X-CVMFS-h3: 12_ad
X-CVMFS-h4: 12fs_?
X-CVMFS-PID: 561710
X-CVMFS-GID: 0
X-CVMFS-UID: 0
```

requirements for the key: alphanumeric (a-z, A-Z, 0-9) and can start with `X-CVMFS-`, no other - or _ allowed
the value can be any sequences (obvisouly no quotes and the separators below)
separator between key and value: `:`
separator between different key value pairs: `|`

Included: new client integration test 099
**Question:** is there any better way how to set the `CVMFS_HTTP_TRACING_HEADERS` with white space etc. bash really does not like it and the workaround is horrible

Missing: Documentation

Edit:
final version only accepts alphanumerical keys and does not accept already added `X-CVMFS-<restofkey>`

### DIFF
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1455,6 +1455,53 @@ bool MountPoint::CreateDownloadManagers() {
     download_mgr_->ShardProxies();
   }
 
+  // configure http tracing header
+  if (options_mgr_->GetValue("CVMFS_HTTP_TRACING", &optarg) &&
+      options_mgr_->IsOn(optarg)) {
+    download_mgr_->EnableHTTPTracing();
+    if (options_mgr_->GetValue("CVMFS_HTTP_TRACING_HEADERS", &optarg)) {
+      std::vector<std::string> tokens = SplitString(optarg, '|');
+      sanitizer::AlphaNumSanitizer sanitizer;
+
+      for (unsigned int i = 0; i < tokens.size(); i++) {
+        std::string token = Trim(tokens[i]);
+
+        std::vector<std::string> key_val = SplitString(token, ':');
+
+        if (key_val.size() != 2) {
+          LogCvmfs(kLogCvmfs, kLogSyslog | kLogDebug,
+             "Http tracing header: Skipping current token part of "
+             "CVMFS_HTTP_TRACING_HEADERS! Invalid "
+             "<key:value> pair. Token: %s", token.c_str());
+          continue;
+        }
+
+        std::string prefix = "X-CVMFS-";
+        std::string key = Trim(key_val[0]);
+
+        if (!sanitizer.IsValid(key)) {
+          // check if the key starts already with the X-CVMFS- prefix and that
+          // no other "-" are part of it. If yes accept it as valid
+          // tracing header
+          if (key.compare(0, prefix.size(), prefix) == 0 &&
+              SplitString(key, '-').size() == 3) {
+            prefix = "";
+          } else {
+            LogCvmfs(kLogCvmfs, kLogSyslog | kLogDebug,
+            "Http tracing header: Skipping current token part of "
+            "CVMFS_HTTP_TRACING_HEADERS! Invalid key. Only alphanumeric keys "
+            "are allowed (a-z, A-Z, 0-9). Token: %s", token.c_str());
+            continue;
+          }
+        }
+
+        std::string final_token = prefix + key + ": " + Trim(key_val[1]);
+
+        download_mgr_->AddHTTPTracingHeader(final_token);
+      }
+    }
+  }
+
   return SetupExternalDownloadMgr(do_geosort);
 }
 
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1460,10 +1460,10 @@ bool MountPoint::CreateDownloadManagers() {
       options_mgr_->IsOn(optarg)) {
     download_mgr_->EnableHTTPTracing();
     if (options_mgr_->GetValue("CVMFS_HTTP_TRACING_HEADERS", &optarg)) {
-      if (optarg.size() > 200) {
+      if (optarg.size() > 1000) {
         LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
-             "CVMFS_HTTP_TRACING_HEADERS too large ( max 200 chars, given %d )",
-             optarg.size());
+            "CVMFS_HTTP_TRACING_HEADERS too large ( max 1000 chars, given %d )",
+            optarg.size());
       } else {
         std::vector<std::string> tokens = SplitString(optarg, '|');
         sanitizer::AlphaNumSanitizer sanitizer;
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1485,19 +1485,11 @@ bool MountPoint::CreateDownloadManagers() {
           std::string key = Trim(key_val[0]);
 
           if (!sanitizer.IsValid(key)) {
-            // check if the key starts already with the X-CVMFS- prefix and that
-            // no other "-" are part of it. If yes accept it as valid
-            // tracing header
-            if (key.compare(0, prefix.size(), prefix) == 0 &&
-                SplitString(key, '-').size() == 3) {
-              prefix = "";
-            } else {
-              LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
-              "Http tracing header: Skipping current token part of "
-              "CVMFS_HTTP_TRACING_HEADERS! Invalid key. Only alphanumeric keys "
-              "are allowed (a-z, A-Z, 0-9). Token: %s", token.c_str());
-              continue;
-            }
+            LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
+            "Http tracing header: Skipping current token part of "
+            "CVMFS_HTTP_TRACING_HEADERS! Invalid key. Only alphanumeric keys "
+            "are allowed (a-z, A-Z, 0-9). Token: %s", token.c_str());
+            continue;
           }
 
           std::string final_token = prefix + key + ": " + Trim(key_val[1]);
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1460,44 +1460,50 @@ bool MountPoint::CreateDownloadManagers() {
       options_mgr_->IsOn(optarg)) {
     download_mgr_->EnableHTTPTracing();
     if (options_mgr_->GetValue("CVMFS_HTTP_TRACING_HEADERS", &optarg)) {
-      std::vector<std::string> tokens = SplitString(optarg, '|');
-      sanitizer::AlphaNumSanitizer sanitizer;
+      if (optarg.size() > 200) {
+        LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
+             "CVMFS_HTTP_TRACING_HEADERS too large ( max 200 chars, given %d )",
+             optarg.size());
+      } else {
+        std::vector<std::string> tokens = SplitString(optarg, '|');
+        sanitizer::AlphaNumSanitizer sanitizer;
 
-      for (unsigned int i = 0; i < tokens.size(); i++) {
-        std::string token = Trim(tokens[i]);
+        for (unsigned int i = 0; i < tokens.size(); i++) {
+          std::string token = Trim(tokens[i]);
 
-        std::vector<std::string> key_val = SplitString(token, ':');
+          std::vector<std::string> key_val = SplitString(token, ':');
 
-        if (key_val.size() != 2) {
-          LogCvmfs(kLogCvmfs, kLogSyslog | kLogDebug,
-             "Http tracing header: Skipping current token part of "
-             "CVMFS_HTTP_TRACING_HEADERS! Invalid "
-             "<key:value> pair. Token: %s", token.c_str());
-          continue;
-        }
-
-        std::string prefix = "X-CVMFS-";
-        std::string key = Trim(key_val[0]);
-
-        if (!sanitizer.IsValid(key)) {
-          // check if the key starts already with the X-CVMFS- prefix and that
-          // no other "-" are part of it. If yes accept it as valid
-          // tracing header
-          if (key.compare(0, prefix.size(), prefix) == 0 &&
-              SplitString(key, '-').size() == 3) {
-            prefix = "";
-          } else {
-            LogCvmfs(kLogCvmfs, kLogSyslog | kLogDebug,
-            "Http tracing header: Skipping current token part of "
-            "CVMFS_HTTP_TRACING_HEADERS! Invalid key. Only alphanumeric keys "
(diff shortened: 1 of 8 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
note: pid/gid/uid are only set if the download request came from `fetch.cc`

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/network/download.h
```diff
@@ -290,6 +292,9 @@ class DownloadManager {  // NOLINT(clang-analyzer-optin.performance.Padding)
    */
   bool ignore_signature_failures_;
 
+  bool enable_http_tracing_;
+  std::vector<std::string> *http_tracing_headers_;
```
jblomer:
```suggestion
  std::vector<std::string> http_tracing_headers_;
```

cvmfs/network/download.cc
```diff
@@ -2716,6 +2756,10 @@ DownloadManager *DownloadManager::Clone(
   clone->opt_backoff_init_ms_ = opt_backoff_init_ms_;
   clone->opt_backoff_max_ms_ = opt_backoff_max_ms_;
   clone->enable_info_header_ = enable_info_header_;
+  clone->enable_http_tracing_ = enable_http_tracing_;
+  if (enable_http_tracing_) {
```
jblomer: I don't think we should check in this function. Let's just clone the headers.

cvmfs/network/download.cc
```diff
@@ -1676,6 +1693,21 @@ Failures DownloadManager::Fetch(JobInfo *info) {
...
+    const std::string str_uid = "X-CVMFS-UID: " + StringifyUint(info->uid());
+
+    // will be auto freed at the end of this function Fetch(JobInfo *info)
+    info->SetTracingHeaderPid(static_cast<char *>(alloca(str_pid.size() + 1)));
+    info->SetTracingHeaderGid(static_cast<char *>(alloca(str_gid.size() + 1)));
+    info->SetTracingHeaderUid(static_cast<char *>(alloca(str_uid.size() + 1)));
+
+    memcpy(info->tracing_header_pid(), str_pid.c_str(), str_pid.size() + 1);
+    memcpy(info->tracing_header_gid(), str_gid.c_str(), str_gid.size() + 1);
+    memcpy(info->tracing_header_uid(), str_uid.c_str(), str_uid.size() + 1);
```
jblomer, line 1708: Since we are anyway using (short-lived) `std::string` instances here, there is not much point in using `alloca` in addition.  I think we can either go for `std::string::c_str()` or we'd use `alloca()` and C string manipulation, which should be a bit faster but may not be worth the effort.

cvmfs/mountpoint.cc
```diff
@@ -1455,6 +1455,53 @@ bool MountPoint::CreateDownloadManagers() {
...
+      std::vector<std::string> tokens = SplitString(optarg, '|');
+      sanitizer::AlphaNumSanitizer sanitizer;
+
+      for (unsigned int i = 0; i < tokens.size(); i++) {
+        std::string token = Trim(tokens[i]);
+
+        std::vector<std::string> key_val = SplitString(token, ':');
+
+        if (key_val.size() != 2) {
+          LogCvmfs(kLogCvmfs, kLogSyslog | kLogDebug,
```
```suggestion
          LogCvmfs(kLogCvmfs, kLogSyslogError | kLogDebug,
```

cvmfs/mountpoint.cc
```diff
@@ -1455,6 +1455,53 @@ bool MountPoint::CreateDownloadManagers() {
...
+        std::string prefix = "X-CVMFS-";
+        std::string key = Trim(key_val[0]);
+
+        if (!sanitizer.IsValid(key)) {
+          // check if the key starts already with the X-CVMFS- prefix and that
+          // no other "-" are part of it. If yes accept it as valid
+          // tracing header
+          if (key.compare(0, prefix.size(), prefix) == 0 &&
+              SplitString(key, '-').size() == 3) {
+            prefix = "";
```
jblomer: Should we go through the hassle to support this? We could simply require headers to be specified without the X-CVMFS- part

cvmfs/mountpoint.cc
```diff
@@ -1455,6 +1455,53 @@ bool MountPoint::CreateDownloadManagers() {
...
+          } else {
+            LogCvmfs(kLogCvmfs, kLogSyslog | kLogDebug,
```
```suggestion
            LogCvmfs(kLogCvmfs, kLogSyslogError | kLogDebug,
```

cvmfs/mountpoint.cc
```diff
@@ -1455,6 +1455,53 @@ bool MountPoint::CreateDownloadManagers() {
...
+             "Http tracing header: Skipping current token part of "
+             "CVMFS_HTTP_TRACING_HEADERS! Invalid "
+             "<key:value> pair. Token: %s", token.c_str());
+          continue;
+        }
+
+        std::string prefix = "X-CVMFS-";
+        std::string key = Trim(key_val[0]);
+
+        if (!sanitizer.IsValid(key)) {
```
jblomer: Let's also set a maximal length (there may be something on length in the HTTP standard, otherwise: 100B?)

test/test_functions
```diff
@@ -509,6 +509,10 @@ cvmfs_mount() {
     sudo sh -c "echo \"CVMFS_DEBUGLOG=$CVMFS_TEST_DEBUGLOG\" >> /etc/cvmfs/default.local" || return 100
   fi
 
+  if [ "x$CVMFS_TEST_QUOTED_PARAM_KEY_VAL" != "x" ]; then
+    sudo sh -c "echo '${CVMFS_TEST_QUOTED_PARAM_KEY_VAL}' >> /etc/cvmfs/default.local" || return 100
+  fi
```
jblomer, line 514: Is this needed? You can pass multiple `key=value` parameters to `cvmfs_mount`, like `cvmfs_mount lhcb.cern.ch a=b c=d`

test/src/099-http_tracing/setup_teardown
```diff
@@ -0,0 +1,21 @@
...
+  local params=$2
+
+  cvmfs_mount $repo $params
+}
+
+test099_unmount() {
+  local repo=$1
+
+  cvmfs_umount $repo
+}
```
jblomer: These seem unnecessary wrappers around `cvmfs_mount` and `cvmfs_umount`.

test/src/099-http_tracing/setup_teardown
```diff
@@ -0,0 +1,21 @@
...
+
+cleanup() {
+  echo "cleanup $cvmfs_test_name"
+  if [ "x$CVMFS_TEST_099_TMPFILE" != "x" ]; then
+    rm $CVMFS_TEST_099_TMPFILE
```
jblomer: Stuff created in the current working directory does not need cleanup, it's being removed by the test driver.
```suggestion
```

test/src/099-http_tracing/setup_teardown
```diff
@@ -0,0 +1,21 @@
...
+    sudo rm /tmp/$CVMFS_TEST_099_TMPFILE
```
jblomer:
```suggestion
    sudo rm -f /tmp/$CVMFS_TEST_099_TMPFILE
```

test/src/099-http_tracing/main
```diff
@@ -0,0 +1,80 @@
...
+
+  echo "   ... Success"
+}
+
+cvmfs_run_test() {
+  logfile=$1
+  trap cleanup HUP INT TERM EXIT || return $?
+
+  echo "create temporary directory"
+  CVMFS_TEST_099_TMPFILE=$(mktemp ./test099.log.XXXXXXXX)
```
jblomer:
```suggestion
  CVMFS_TEST_099_TMPFILE=$(mktemp /tmp/test099.log.XXXXXXXX)
```

test/src/099-http_tracing/main
```diff
@@ -0,0 +1,80 @@
...
+  CVMFS_TEST_DEBUGLOG="/tmp/$CVMFS_TEST_099_TMPFILE"
+  CVMFS_TEST_QUOTED_PARAM_KEY_VAL='CVMFS_HTTP_TRACING_HEADERS="h1:test|CVMFS-X-h2:ff|X-CVMFS-h3:12_ad |  h4  : 12fs_? "'
+
+  echo "   Mount $TEST099_REPO"
+  cvmfs_mount $TEST099_REPO "CVMFS_HTTP_TRACING=on" || return 1
+
+  echo "   Client config $TEST099_REPO"
+  cat /etc/cvmfs/default.local
+
+
```
jblomer:
```suggestion
```

test/src/099-http_tracing/main
```diff
@@ -0,0 +1,80 @@
...
+
+  echo "   Access data to send HTTP requests"
+  cat $TEST099_MOUNTPOINT/*.sh > /dev/null
```
jblomer: This may break. You can use `find  $TEST099_MOUNTPOINT -type f | tail -n1` to get a file to print.

test/src/099-http_tracing/main
```diff
@@ -0,0 +1,80 @@
...
+  cat $TEST099_MOUNTPOINT/etc/grid-security/*.pem > /dev/null
```
jblomer: One file should be enough.
```suggestion
```

test/src/099-http_tracing/main
```diff
@@ -0,0 +1,80 @@
...
+  # used and not the actual fuse pid/gid/uid
+  echo "      Unique occurence of pid correct? $pid > 1"
+  [ "$pid" -gt "1" ] || return 20
+
+  echo "      Unique occurence of gid correct? $gid > 1"
+  [ "$pid" -gt "1" ] || return 20
+
+  echo "      Unique occurence of uid correct? $uid > 1"
+  [ "$pid" -gt "1" ] || return 20
+
```
jblomer:
```suggestion
```

test/src/099-http_tracing/main
```diff
@@ -0,0 +1,80 @@
...
+  echo "      Occurence of h4 correct? $h4 > 0"
+  [ "$h4" -gt "0" ] || return 13
+
+  # if unique occurence would be 1 this means only the default values would be
+  # used and not the actual fuse pid/gid/uid
+  echo "      Unique occurence of pid correct? $pid > 1"
+  [ "$pid" -gt "1" ] || return 20
+
+  echo "      Unique occurence of gid correct? $gid > 1"
+  [ "$pid" -gt "1" ] || return 20
```
jblomer:
```suggestion
  [ "$gid" -gt "1" ] || return 20
```

test/src/099-http_tracing/main
```diff
@@ -0,0 +1,80 @@
...
+
+  echo "      Unique occurence of uid correct? $uid > 1"
+  [ "$pid" -gt "1" ] || return 20
```
jblomer:
```suggestion
  [ "$uid" -gt "1" ] || return 20
```

jblomer (changes requested):
Very nice! Almost there.

cvmfs/mountpoint.cc
HereThereBeDragons: the code is there and it is only called once per mounting.. whatever you decide ;) i would just leave it and not advertise it.

cvmfs/network/download.cc
HereThereBeDragons: `alloca` we decided to use because of the scope of the variables life

cvmfs/network/download.h
HereThereBeDragons: i just copied the schema used for `std::vector<std::string> *opt_host_chain_;`

cvmfs/network/download.cc
HereThereBeDragons: see above

cvmfs/mountpoint.cc
HereThereBeDragons: https://stackoverflow.com/questions/686217/maximum-on-http-header-values old article but smallest size is 4k for *sum of the request line and ALL header fields*. so i would say max 200 chars?

test/test_functions
HereThereBeDragons: im open to suggestions but for me this was  the only way how i could get 
`CVMFS_HTTP_TRACING_HEADERS="h1:test|CVMFS-X-h2:ff|X-CVMFS-h3:12_ad |  h4  : 12fs_? "` correctly into the config. with the white spaces and the quotes around.

jblomer: I see. That's indeed a special case. Ok, let's keep it as is then.

cvmfs/mountpoint.cc
HereThereBeDragons: decision was to remove it for the moment

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/mountpoint.cc
```diff
@@ -1460,44 +1460,42 @@ bool MountPoint::CreateDownloadManagers() {
...
     if (options_mgr_->GetValue("CVMFS_HTTP_TRACING_HEADERS", &optarg)) {
-      std::vector<std::string> tokens = SplitString(optarg, '|');
-      sanitizer::AlphaNumSanitizer sanitizer;
-
-      for (unsigned int i = 0; i < tokens.size(); i++) {
-        std::string token = Trim(tokens[i]);
+      if (optarg.size() > 200) {
+        LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
+             "CVMFS_HTTP_TRACING_HEADERS too large ( max 200 chars, given %d )",
+             optarg.size());
```
jblomer: If we look at the entire CVMFS_HTTP_TRACING_HEADERS, I think we should go higher, let's say 1000 characters.

jblomer (approved):
Regarding the limit on CVMFS_HTTP_TRACING_HEADERS: up to you if you'd like to change the current limit or not.

vvolkl (approved):

PR #3309, opened 2023-07-07
Fix performance bug introduced by commit d67bfbd

jblomer:
Fixes the performance bug found in the CHEP preparation. For one of the possible execution paths, filling the cache was missing.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer:
@cernvm-bot cloudtest

vvolkl (approved):

PR #3314, opened 2023-07-10
Use dedicated log files for mount helper

jblomer:
PR #3302 introduced a problem when the log file is created by the mount helper: it is then owned by root and the fuse module (running as cvmfs user) cannot use it.  This patch adds the `.mount` suffix for log files, to separate them from the fuse module's logs. This should fix the recent failures seen in test 043.

### DIFF
--- a/mount/mount.cvmfs.cc
+++ b/mount/mount.cvmfs.cc
@@ -416,9 +416,9 @@ int main(int argc, char **argv) {
   if (options_manager_.GetValue("CVMFS_SYSLOG_FACILITY", &optarg))
     SetLogSyslogFacility(String2Int64(optarg));
   if (options_manager_.GetValue("CVMFS_USYSLOG", &optarg))
-    SetLogMicroSyslog(optarg);
+    SetLogMicroSyslog(optarg + ".mount");
   if (options_manager_.GetValue("CVMFS_DEBUGLOG", &optarg))
-    SetLogDebugFile(optarg);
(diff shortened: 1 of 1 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
Looks good, the test failures are unrelated.

HereThereBeDragons:
i agree. only question is: why are the unittest t_quotamanager so slow on arm?

jblomer:
> i agree. only question is: why are the unittest t_quotamanager so slow on arm?

This is due to counting open files.  Fixed here: https://github.com/cvmfs/cvmfs/pull/3263/commits/c1dc997e8ada5c047c4af27767a9d98a8e789a7d as part of the streaming cache manager PR.

HereThereBeDragons:
ok some other questions:
- do we want the logs to be in the same file? if yes then i guess the other option would be to change the access rights of the file? 
- the naming i guess will end up to be like `<somelog>.log.mount` for the mounthelper log?

jblomer:
>     * do we want the logs to be in the same file? if yes then i guess the other option would be to change the access rights of the file?
> 
>     * the naming i guess will end up to be like `<somelog>.log.mount` for the mounthelper log?

Correct. I don't think we necessarily need to have them in the same file. We already create multiple log files based on the base name (e.g., for the cache manager, also as a shadow file for log rotation).

HereThereBeDragons (approved):

vvolkl (approved):

PR #3262, opened 2023-04-21
Readahead skip when on tmpfs

HereThereBeDragons:
Ref Issue https://github.com/cvmfs/cvmfs/issues/3192 point 2

Replaces #3254

Issue #3315, opened 2023-07-10
Tmpfs detection not working correctly


HereThereBeDragons:
#3262 is not complete/correct:
- currently only alien cache checks if on tmpfs
- GetFileSystemInfo does not check if on tmpfs

PR #3316, opened 2023-07-10
fix tmpfs recognition to skip readahead

HereThereBeDragons:
Issue  #3315 

Extends/fixes #3262

### DIFF
--- a/test/unittests/t_util.cc
+++ b/test/unittests/t_util.cc
@@ -407,6 +407,8 @@ TEST_F(T_Util, GetFileSystemInfo) {
   EXPECT_EQ(kFsTypeProc, fs_info.type);
   fs_info = GetFileSystemInfo("/");
   EXPECT_EQ(kFsTypeUnknown, fs_info.type);
+  fs_info = GetFileSystemInfo("/run");
+  EXPECT_EQ(kFsTypeTmpfs, fs_info.type);
 }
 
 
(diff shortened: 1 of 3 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

test/unittests/t_util.cc
```diff
@@ -407,6 +407,8 @@ TEST_F(T_Util, GetFileSystemInfo) {
   EXPECT_EQ(kFsTypeProc, fs_info.type);
   fs_info = GetFileSystemInfo("/");
   EXPECT_EQ(kFsTypeUnknown, fs_info.type);
+  fs_info = GetFileSystemInfo("/run");
```
jblomer: That only works on Linux, and perhaps not on all distributions.  I'd check if `/run` exists first.

jblomer (changes requested):

test/unittests/t_util.cc
HereThereBeDragons: seeing that this also fails on ubuntu i guess we either find some better tmpfs that is always there or we just remove the test?

jblomer: Probably best to remove it then.  It's a pity but it's tough to test it reliably without being root.

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot unittest

jblomer (approved):

vvolkl (approved):

PR #3323, opened 2023-07-14
Add integration test for issue #2844

### LINKED ISSUE #2844: Aborting a transaction should clear local lock files when the lease has expired on the gateway
Calling `cvmfs_server abort` should clear any local transaction lock files on the publisher (`/var/spool/cvmfs/<REPO>/{in_transaction.lock,session_token}`, even when the transaction lease is expired on the gateway.

jblomer:
The issue has already been fixed by the transition to the C++ implementation of the abort command. This just adds a test.

Closes #2844

### DIFF
--- a/test/src/814-expired_lease/main
+++ b/test/src/814-expired_lease/main
@@ -1,0 +1,34 @@
+cvmfs_test_name="Automatic cleanup when a lease has expired"
+cvmfs_test_autofs_on_startup=false
+cvmfs_test_suites="quick"
+
+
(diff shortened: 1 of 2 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
The test can be made more reliable, I think

jblomer:
@cernvm-bot cloudtest

vvolkl (approved):

Issue #3128, opened 2022-12-05
Gateway restart prevents removal of active leases


jblomer:
When the gateway restarts, it should continue to work with the leases in the database. Currently it fails to use the leases because the (in-memory) statistics is missing.  Either put statistics in the db or ignore missing stats.

PR #3325, opened 2023-07-14
Fix transaction abort after gateway restart

jblomer:
This is a partial fix of #3128. While publishing still fails if the gateway was restarted, the publisher can at least abort the transaction gracefully now.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

vvolkl (approved):

PR #3327, opened 2023-07-14
Allow change to/from debug mode during cvmfs_config reload

HereThereBeDragons:
issue #2897 

based on #2933 

- [x] Feature implementation
- [x] ~~Test for hot patching of versions that do not support it without breaking~~ only needed at a later time when hotpatching between versions that all support it.
- [x] Test of reloading and switching debug mode on and off

right now: only on remount the new option will be available as the loader belongs to the base software, that is not touched by hot patching. it's new version is only loaded when remounting cvmfs

### DIFF
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -543,10 +543,15 @@ static CvmfsExports *LoadLibrary(const bool debug_mode,
 }
 
 Failures Reload(const int fd_progress, const bool stop_and_go,
-                const bool debug_mode) {
+                const ReloadMode reload_mode) {
   int retval;
 
-  debug_mode_ = debug_mode;
+  // for legacy call we take the current state of debug_mode_
+  if (reload_mode == kReloadDebug) {
+    debug_mode_ = true;
+  } else if (reload_mode == kReloadNoDebug) {
+    debug_mode_ = false;
+  }
 
   retval = cvmfs_exports_->fnMaintenanceMode(fd_progress);
   if (!retval)
@@ -618,9 +623,6 @@ int FuseMain(int argc, char *argv[]) {
 
   int retval;
 
-  // check if already mounted in debug mode
-  debug_mode_ = getenv("__CVMFS_DEBUG_MODE__") != NULL;
-
   // Jump into alternative process flavors (e.g. shared cache manager)
   // We are here due to a fork+execve (ManagedExec in util.cc) or due to
   // utility calls of cvmfs2
@@ -634,8 +636,17 @@ int FuseMain(int argc, char *argv[]) {
 
       // always last param of the cvmfs2 __RELOAD__ command
       // check if debug mode is requested
+      // NOTE:
+      // debug mode is decided based on CVMFS_DEBUGLOG being set or not
+      // this means: reloading is now always based on CVMFS_DEBUGLOG, and
+      // reload ignores the current state
+      //
+      // if you mount with debug but do not set CVMFS_DEBUGLOG and reload,
+      // you will reload with
       if (std::string(argv[argc - 1]) == std::string("--debug")) {
         debug_mode_ = true;
+      } else {
+        debug_mode_ = false;
       }
       retval = loader_talk::MainReload(argv[2], stop_and_go, debug_mode_);
 
@@ -682,6 +693,7 @@ int FuseMain(int argc, char *argv[]) {
       return 0;
     }
 
+    debug_mode_ = getenv("__CVMFS_DEBUG_MODE__") != NULL;
     cvmfs_exports_ = LoadLibrary(debug_mode_, NULL);
     if (!cvmfs_exports_)
       return kFailLoadLibrary;
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -634,7 +634,7 @@ int FuseMain(int argc, char *argv[]) {
 
       // always last param of the cvmfs2 __RELOAD__ command
       // check if debug mode is requested
-      if(!strcmp("--debug", argv[argc-1])) {
+      if (!strcmp("--debug", argv[argc-1])) {
         debug_mode_ = true;
       }
       retval = loader_talk::MainReload(argv[2], stop_and_go, debug_mode_);
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -497,7 +497,7 @@ static CvmfsExports *LoadLibrary(const bool debug_mode,
   library_name = platform_libname(library_name);
   string error_messages;
 
-  static vector<string> library_paths;  // TODO(rmeusel): C++11 initializer
+  vector<string> library_paths;  // TODO(rmeusel): C++11 initializer
   if (library_paths.empty()) {
     library_paths.push_back(local_lib_path + library_name);
     library_paths.push_back("/usr/lib/"   + library_name);
@@ -542,9 +542,11 @@ static CvmfsExports *LoadLibrary(const bool debug_mode,
   return *exports_ptr;
 }
 
-
-Failures Reload(const int fd_progress, const bool stop_and_go) {
+Failures Reload(const int fd_progress, const bool stop_and_go,
+                const bool debug_mode) {
   int retval;
+
+  debug_mode_ = debug_mode;
 
   retval = cvmfs_exports_->fnMaintenanceMode(fd_progress);
   if (!retval)
@@ -616,6 +618,9 @@ int FuseMain(int argc, char *argv[]) {
 
   int retval;
 
+  // check if already mounted in debug mode
+  debug_mode_ = getenv("__CVMFS_DEBUG_MODE__") != NULL;
+
   // Jump into alternative process flavors (e.g. shared cache manager)
   // We are here due to a fork+execve (ManagedExec in util.cc) or due to
   // utility calls of cvmfs2
@@ -626,7 +631,14 @@ int FuseMain(int argc, char *argv[]) {
       bool stop_and_go = false;
       if ((argc > 3) && (string(argv[3]) == "stop_and_go"))
         stop_and_go = true;
-      retval = loader_talk::MainReload(argv[2], stop_and_go);
+
+      // always last param of the cvmfs2 __RELOAD__ command
+      // check if debug mode is requested
+      if(!strcmp("--debug", argv[argc-1])) {
+        debug_mode_ = true;
+      }
+      retval = loader_talk::MainReload(argv[2], stop_and_go, debug_mode_);
+
       if ((retval != 0) && (stop_and_go)) {
         CreateFile(string(argv[2]) + ".paused.crashed", 0600);
       }
@@ -670,7 +682,6 @@ int FuseMain(int argc, char *argv[]) {
       return 0;
     }
 
-    debug_mode_ = getenv("__CVMFS_DEBUG_MODE__") != NULL;
     cvmfs_exports_ = LoadLibrary(debug_mode_, NULL);
     if (!cvmfs_exports_)
       return kFailLoadLibrary;
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -634,7 +634,7 @@ int FuseMain(int argc, char *argv[]) {
 
       // always last param of the cvmfs2 __RELOAD__ command
       // check if debug mode is requested
-      if (!strcmp("--debug", argv[argc-1])) {
+      if (std::string(argv[argc - 1]) == std::string("--debug")) {
         debug_mode_ = true;
       }
       retval = loader_talk::MainReload(argv[2], stop_and_go, debug_mode_);
--- a/cvmfs/loader_talk.cc
+++ b/cvmfs/loader_talk.cc
@@ -74,7 +74,7 @@ static void *MainTalk(void *data __attribute__((unused))) {
       SetLogMicroSyslog(*usyslog_path_);
       bool debug = (command == 'S' || command == 'R');
       LogCvmfs(kLogCvmfs, kLogSyslog, "reloading Fuse module. Debug=%d", debug);
-      int retval = Reload(con_fd, (command == 'S' || command == 's'), debug); 
+      int retval = Reload(con_fd, (command == 'S' || command == 's'), debug);
       SendMsg2Socket(con_fd, "~");
       (void)send(con_fd, &retval, sizeof(retval), MSG_NOSIGNAL);
       if (retval != kFailOk) {
@@ -124,7 +124,7 @@ int MainReload(const std::string &socket_path, const bool stop_and_go,
   }
   LogCvmfs(kLogCvmfs, kLogStdout, "done");
 
-  //S,R if debug=true, s,r if debug=false
+  // S,R if debug=true, s,r if debug=false
   char command = debug ? (stop_and_go ? 'S' : 'R') : (stop_and_go ? 's' : 'r');
 
   WritePipe(socket_fd, &command, 1);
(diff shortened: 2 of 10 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

test/src/0100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+  if [[ -s $CVMFS_TEST_0100_TMPFILE ]]
+  then
+    echo "     *** True"
+  else
+    echo "     *** False"
+    return 11
+  fi
+
+  echo "   ** Remove CVMFS_DEBUGLOG from config"
+  sudo sed -i '/^CVMFS_DEBUGLOG=/d' /etc/cvmfs/default.local
```
vvolkl: In the macos test there is a 
```
sed: 1: "/etc/cvmfs/default.local": command c expects \ followed by text
```
error, could be due to https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux/14813278#14813278

Needs doublechecking if this is what causes the test to fail

vvolkl: I'd go for
```suggestion
  sudo sed -i'' -e '/^CVMFS_DEBUGLOG=/d' /etc/cvmfs/default.local
```

vvolkl:
The integration test can be renamed to `100-...`

HereThereBeDragons:
@cernvm-bot unittest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

test/src/100-reload-switch-debug/setup_teardown
```diff
@@ -0,0 +1,7 @@
+cleanup() {
+  echo "cleanup $cvmfs_test_name"
+  if [ "x$CVMFS_TEST_100_TMPFILE" != "x" ]; then
+    sudo rm $CVMFS_TEST_100_TMPFILE
```
jblomer:
```suggestion
    sudo rm -f $CVMFS_TEST_100_TMPFILE
```

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+  echo "   ** Success"
+}
+
+cvmfs_run_test() {
+  logfile=$1
+  local_script_dir=$2
+  trap cleanup HUP INT TERM EXIT || return $?
+
+  echo "create temporary directory"
+  CVMFS_TEST_100_TMPFILE=$(mktemp /tmp/test100.log.XXXXXXXX)
```
jblomer:
```suggestion
  CVMFS_TEST_100_TMPFILE=$(mktemp /tmp/cvmfs_test100.log.XXXXXXXX)
```

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+  echo "tmpfile is $CVMFS_TEST_100_TMPFILE"
+
+  # delete it because of wrong ownership, let cvmfs create it itself
+  rm $CVMFS_TEST_100_TMPFILE
```
jblomer:
```suggestion
  rm -f $CVMFS_TEST_100_TMPFILE
```

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+TEST100_REPO=lhcb.cern.ch
+TEST100_MOUNTPOINT=/cvmfs/lhcb.cern.ch
+
+CVMFS_TEST_100_TMPFILE=
+
+mount_with_debuglog() {
+  CVMFS_TEST_DEBUGLOG=$(echo "\"${CVMFS_TEST_100_TMPFILE}\"")
+
+  echo "   ** Mount_with_debuglog"
+  cvmfs_mount $TEST100_REPO || return $?
```
jblomer: The `CVMFS_TEST_DEBUGLOG` is meant to be used "from the outside" to run the integration tests in debug mode. Here, I'd use this:

```suggestion
  echo "   ** Mount_with_debuglog"
  cvmfs_mount $TEST100_REPO CVMFS_DEBUGLOG=${CVMFS_TEST_100_TMPFILE} || return $?
```

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+}
+
+mount_without_debuglog() {
+  CVMFS_TEST_DEBUGLOG=""
+
```
jblomer:
```suggestion
```

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+  echo "   ** Mount_without_debuglog"
+  cvmfs_mount $TEST100_REPO || return $?
+}
+
+reload_without_debuglog() {
+  echo "   ** Access some data at $TEST100_REPO"
+  cat /cvmfs/$TEST100_REPO/*.sh > /dev/null || return 10
```
jblomer: That may change in the future. Better find a regular file like this: `find /cvmfs/$TEST100_REPO/ -type f | head -n1`

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+
+  # assumption: debug log is already there
+  echo "   ** Check non-empty existence of debug log"
+  if [[ -s $CVMFS_TEST_100_TMPFILE ]]
```
jblomer: Doesn't work on `dash`, change to:
```suggestion
  if [ -s $CVMFS_TEST_100_TMPFILE ]
```

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+  sudo cvmfs_config reload > /dev/null
+
+  echo "   ** Wipe cache"
+  sudo cvmfs_config wipecache > /dev/null
+
+  echo "   ** Get last line of debug log"
+  local last_line=$(sudo tail -1 $CVMFS_TEST_100_TMPFILE)
+
+  echo "   ** Access some data at $TEST100_REPO"
+  cat /cvmfs/$TEST100_REPO/*.sh > /dev/null || return 11
```
jblomer: See above

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+
+  echo "   ** Make sure last line is the same"
+  [ "$last_line" = "$new_last_line" ] || return 12
+
+  echo "   ** Success"
+}
+
+reload_with_debuglog() {
+  echo "   ** Access some data at $TEST100_REPO"
+  cat /cvmfs/$TEST100_REPO/*.sh > /dev/null || return 20
```
jblomer: See above

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+
+  # assumption: debug log is not there
+  echo "   ** Check that debug log does not exist"
+  if [[ -f $CVMFS_TEST_100_TMPFILE ]]
```
jblomer: See above

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+  then
+    echo "     *** False: File exists"
+    return 21
+  else
+    echo "     *** True: File does not exist"
+  fi
+
+  echo "   ** Add CVMFS_DEBUGLOG to config"
+  sudo echo "CVMFS_DEBUGLOG=\"$CVMFS_TEST_100_TMPFILE\"" | sudo tee -a /etc/cvmfs/default.local
```
jblomer:
```suggestion
  echo "CVMFS_DEBUGLOG=\"$CVMFS_TEST_100_TMPFILE\"" | sudo tee -a /etc/cvmfs/default.local
```

test/src/100-reload-switch-debug/main
```diff
@@ -0,0 +1,128 @@
...
+  sudo cvmfs_config reload > /dev/null
+
+  echo "   ** Wipe cache"
+  sudo cvmfs_config wipecache > /dev/null
+
+  echo "   ** Get last line of debug log"
+  local last_line=$(sudo tail -1 $CVMFS_TEST_100_TMPFILE)
+
+  echo "   ** Access some data at $TEST100_REPO"
+  cat /cvmfs/$TEST100_REPO/*.sh > /dev/null || return 21
```
jblomer: See above

cvmfs/loader_talk.cc
```diff
@@ -121,7 +124,9 @@ int MainReload(const std::string &socket_path, const bool stop_and_go) {
   }
   LogCvmfs(kLogCvmfs, kLogStdout, "done");
 
-  const char command = stop_and_go ? 'S' : 'R';
+  // S,R if debug=true, s,r if debug=false
+  char command = debug ? (stop_and_go ? 'S' : 'R') : (stop_and_go ? 's' : 'r');
```
jblomer: I think that needs to be handled differently.  There is in fact a problem if a new cvmfs version reloads the currently active old fuse module. The old fuse module won't understand the small letters `s` and `r` and will not reload.  So for an installation without debug mode enabled, this breaks.

I'd suggest instead to send two letters. First one indicates if there is a debug log or not, the second is the known `S` or `R`. In this case, old fuse modules will ignore the first letter and then correctly behave on the second.

test/src/100-reload-switch-debug/main
jblomer, line 1: We should also test targeted reload of a single repository (`cvmfs_config reload <reponame>`)

cvmfs/loader.cc
```diff
@@ -626,7 +631,14 @@ int FuseMain(int argc, char *argv[]) {
       bool stop_and_go = false;
       if ((argc > 3) && (string(argv[3]) == "stop_and_go"))
         stop_and_go = true;
-      retval = loader_talk::MainReload(argv[2], stop_and_go);
+
+      // always last param of the cvmfs2 __RELOAD__ command
+      // check if debug mode is requested
+      if (!strcmp("--debug", argv[argc-1])) {
```
jblomer: For consistency:
```suggestion
      if (std::string(argv[argc - 1]) == std::string("--debug")) {
```

cvmfs/loader.cc
```diff
@@ -616,6 +618,9 @@ int FuseMain(int argc, char *argv[]) {
 
   int retval;
 
+  // check if already mounted in debug mode
+  debug_mode_ = getenv("__CVMFS_DEBUG_MODE__") != NULL;
```
jblomer: Moving this from line 673 to the front seems unnecessary. Let's try without.

jblomer (changes requested):
Nice! Some changes required.

cvmfs/loader.cc
HereThereBeDragons: no we have to move it to the front because of line 
https://github.com/cvmfs/cvmfs/blob/1030e2e0a27c37bc0703acbafb9f34b1502c90b7/cvmfs/loader.cc#L640

test/src/100-reload-switch-debug/main
HereThereBeDragons: does this really make a difference? what do you want? mount several and only reload one? 
at the moment the debug log is written into `default.local` that would have to change then...

cvmfs/loader.cc
jblomer: But if you look into the three lines before, we overwrite `debug_mode_` anyway.

test/src/100-reload-switch-debug/main
jblomer: Yes it does. It probes a different code path in `cvmfs_config`, and in the C++ code `stop_and_go` is false. You don't need to change anything to the mounting. It's sufficient to call `cvmfs_config reload <repo>` instead of `cvmfs_config reload`.

cvmfs/loader_talk.cc
HereThereBeDragons: i thought only 
https://github.com/cvmfs/cvmfs/blob/1030e2e0a27c37bc0703acbafb9f34b1502c90b7/cvmfs/loader.cc#L545-L546
with
https://github.com/cvmfs/cvmfs/blob/1030e2e0a27c37bc0703acbafb9f34b1502c90b7/cvmfs/loader.cc#L578
interacts with fuse.

`'S'` is used in `loader_talk.cc` to communicate between `MainTalk` (which calls the `Reload` mentioned above) and `MainReload` 

or am i missunderstanding which of these threads are part of fuse?

test/src/100-reload-switch-debug/main
HereThereBeDragons: this is the `group_login.sh` of lhcb.. i would not think this would change any time soon?

cvmfs/loader_talk.cc
jblomer: The `MainReload` is part of a short-lived process that triggers a fuse module reload during `cvmfs_config reload`. It talks to the fuse module process through a pipe, and the fuse module process receives the commands in its `MainTalk()` function.  I.e., these two functions are in the same compilation unit but they run in different processes.

test/src/100-reload-switch-debug/main
jblomer: Maybe not, but it's still better if we can avoid relying on repository specific information.

cvmfs/loader.cc
HereThereBeDragons: you mean the string compare if `--debug` is set?

as far as i understood there are 2 cases how you can be in debug mode:
- you mounted in debug mode
- you set the debuglog in the client config

the `getenv` checks if we already are in debug mode and sets `debug_mode_` true in that case. and `--debug` sets `_debug_mode_` true if in the reload command it is requested to switch to debug mode.

in the old location it will definitely overwrite the request to change from non-debug to debug mode (as `__CVMFS_DEBUG_MODE__` is only set in `ManagedExec` in `posix.cc`

but again.. maybe i dont fully understand how everything interacts with each other.

HereThereBeDragons: i think you are right with putting the line back to where it was. looking closer at the function this is a totally different code path and not part of the `reload` functionality

HereThereBeDragons:
do we also need another version of this for backward compatibility (`loader_talk.cc`)?
```c++
int MainReload(const std::string &socket_path, const bool stop_and_go); // old version i removed
int MainReload(const std::string &socket_path, const bool stop_and_go,
               const bool debug);
 ```

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer:
> do we also need another version of this for backward compatibility (`loader_talk.cc`)?
> 
> ```c++
> int MainReload(const std::string &socket_path, const bool stop_and_go); // old version i removed
> int MainReload(const std::string &socket_path, const bool stop_and_go,
>                const bool debug);
> ```

Not needed, because `MainReload` is in the short-lived process triggered by `cvmfs_config`. I.e., there is no possible mix between old and new version. Old and new version communicate only through the socket.

HereThereBeDragons:
the cloudtest fails on mac - but only during the cleanup, after the test run successfully. not sure what is wrong there.

vvolkl (changes requested):
On macos the test is very slow, as any time you do a reload  or wipecache we need to wait the full 60s timeout for the cache to evict. It's about 10 min overall-  I think you can just increase the timeout by setting `cvmfs_test_timeout=900` in your test, but ideally this could be split in a shorter quick test and a more complete test in the full suite. (Can be in a future PR though)

jblomer:
I suggest that for macOS we set the `CVMFS_KCACHE_TIMEOUT` parameter to something small, e.g. 3 (seconds) instead of the default 60 seconds.

HereThereBeDragons:
i shorten it now for any platform to 5sec. lets see what the cloudtest does

HereThereBeDragons:
@cernvm-bot cloudtest

vvolkl (approved):

jblomer (approved):

PR #3332, opened 2023-07-18
ci-scripts: accept also test numbers 1* as  client tests

### LINKED ISSUE #3328: Improve structure/filenames for integration tests
We are running out of 3-digit numbers for client integration tests as tests are only considerred to be client tests if their name starts with `src/0*`

As there are also a lot of tests by now in a single folder, scrolling through them is quite exhausting.

There are multiple possible solutions:
1. Split tests in separate folders (client, server, gateway, ...)
2. Accept 4-digit numbers
3. Also declare 1** numbers as client tests

I personally would prefer option 1

vvolkl:
Fixes #3328 until larger reorganization of the integration tests


vvolkl:
@cernvm-bot  cloudtest

jblomer (approved):

vvolkl:
@cernvm-bot cloudtest

PR #3333, opened 2023-07-18
Use http client auth only if membership is set

jblomer:
Suggested replacement for #3307

### DIFF
--- a/cvmfs/authz/authz_curl.cc
+++ b/cvmfs/authz/authz_curl.cc
@@ -158,6 +158,10 @@ bool AuthzAttachment::ConfigureCurlHandle(
 {
   assert(info_data);
 
+  // File catalog has no membership requirement, no tokens to attach
+  if (membership_.empty())
+    return false;
+
   // We cannot rely on libcurl to pipeline (yet), as cvmfs may
   // bounce between different auth handles.
   curl_easy_setopt(curl_handle, CURLOPT_FRESH_CONNECT, 1);


jblomer:
@cernvm-bot cloudtest

HereThereBeDragons:
if this is the solution then we should remove this code section, as it will never be called. credential_attachement is always set in the mountpoint.

https://github.com/cvmfs/cvmfs/blob/8ed16ca371884fac90d2be44e5b76d13ec2cb7b3/cvmfs/network/download.cc#L935-L937

or replace the `if` with  `if (!credential_attachement.membership())` (just need to add it to the class, not sure how much effort that would be)

jblomer:
This is only true if the download manager is created by the `Mountpoint` class. Otherwise (e.g., server), the attachment may or may not be set.

HereThereBeDragons:
ok so should we extend the if? `if (credentials_attachment_ == NULL || (!credential_attachement.membership())`?

jblomer:
> ok so should we extend the if? `if (credentials_attachment_ == NULL || (!credential_attachement.membership())`?

It may actually be a bit nicer to read that way, but it's also a slightly larger change because the membership information is only present in the derived class `AuthzAttachment`, and the download manager at this point has the base class `CredentialsAttachment` in its hand.  I'd suggest to leave it as is and perhaps come back to the credentials management at a later point.

HereThereBeDragons:
ok should we then add some comment there?

> > ok so should we extend the if? `if (credentials_attachment_ == NULL || (!credential_attachement.membership())`?
> 
> It may actually be a bit nicer to read that way, but it's also a slightly larger change because the membership information is only present in the derived class `AuthzAttachment`, and the download manager at this point has the base class `CredentialsAttachment` in its hand. I'd suggest to leave it as is and perhaps come back to the credentials management at a later point.

jblomer:
> ok should we then add some comment there?

I think that would be a bit of an arbitrary spot for what should be a more general review of the handling of client credentials. Issue #3342 created.

HereThereBeDragons (approved):

vvolkl (approved):

Issue #2922, opened 2022-04-25
xattr `chunk_list` can overflow for large files


mharvey-jt:
There is a 64kiB system limit on the size of an an extended attribute's value.
For large files, the value of the `chunk_list` attribute can exceed this limit.

desired behaviour:
* for large files, split the chunk list over as many additional `chunk_list_{1...N}` attributes as necessary
* optionally retain breaking breaking behaviour of `chunk_list` for compatibility

PR #3336, opened 2023-07-20
Add paging to xattr chunk_list that can overflow the 64kiB limit

HereThereBeDragons:
Issue #2922 

add functionality that xattr `chunk_list` will return in pages if otherwise an overflow would occur.
pages can be accessed by `chunk_list_<num>`

based on the file size, dynamically calculates how many entries per page to be shown

`chunk_list`, `chunk_list_0` and `chunk_list_1` default to the same page: page 1

output:
on error: starts with `ERROR`
on success:
```
Access different pages with 'chunk_list_<page>' ( 2 pages )
max_pages,curr_page,idx,hash,offset,size
2,1,0,39008434907e0d569a0131f024f1a39822cdd0be,0,7393679
2,1,1,ae4c82590fd6a5cfea1a325a0dc1394f0af13dab,7393679,69786
...
```

10 GB file created with 1MB blocks of `/dev/urandom` creates 2 pages with the first page having 907 chunks.
`Attribute "chunk_list" had a 61820 byte value for /cvmfs/my.test.repo/bigfile.blob:`

file that has no chunks but is valid:
```
Attribute "chunk_list" had a 92 byte value for /cvmfs/my.test.repo/text1.txt:
max_pages,curr_page,idx,hash,offset,size
1,1,0,e8ec3d88b62ebf526e4e5a4ff6162a3aa48a6b78,0,0
```

Possible future  improvements:
- right now it only works for `chunk_list` -- could be made as a vector of paged attributes if needed and selected from the vector which xattr should be paged
- we now have 2 functions. in theory with little refactoring it could be just one and xattr that dont need the `attr_page_num` ignore it
```c++
  virtual bool PrepareValueFenced() { return true; }
  virtual bool PrepareValueFenced(int32_t attr_page_num) { return true; }
````

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -112,6 +112,7 @@
 #include "util/logging.h"
 #include "util/platform.h"
 #include "util/smalloc.h"
+#include "util/string.h"
 #include "util/uuid.h"
 #include "wpad.h"
 #include "xattr.h"
@@ -1662,7 +1663,24 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   }
   TraceInode(Tracer::kEventGetXAttr, ino, "getxattr()");
 
-  const string attr = name;
+  string attr = name;
+  int32_t attr_page_num = -1;
+
+  // split xattr if of format <name>_<num>
+  // for different "pages" of xattr to be returned
+  // if no <num> is valid, the first page is printed 
+  // as String2Uint64 failure defaults to 0
+  string paged_attr = "user.chunk_list";
+  if (attr.rfind(paged_attr, 0) == 0) {
+    attr = paged_attr;
+
+    vector<string> tokens = SplitString(name, '_');
+
+    attr_page_num = String2Uint64(tokens[tokens.size() - 1]);
+  }
+
+
+
   catalog::DirectoryEntry d;
   const bool found = GetDirentForInode(ino, &d);
   bool retval;
@@ -1696,7 +1714,7 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     attr, path, &d));
   if (!magic_xattr.IsNull()) {
     magic_xattr_success = magic_xattr->
-                              PrepareValueFencedProtected(fuse_ctx->gid);
+                      PrepareValueFencedProtected(fuse_ctx->gid, attr_page_num);
   }
 
   fuse_remounter_->fence()->Leave();
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1668,7 +1668,7 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
 
   // split xattr if of format <name>_<num>
   // for different "pages" of xattr to be returned
-  // if no <num> is valid, the first page is printed 
+  // if no <num> is valid, the first page is printed
   // as String2Uint64 failure defaults to 0
   string paged_attr = "user.chunk_list";
   if (attr.rfind(paged_attr, 0) == 0) {
@@ -1676,7 +1676,8 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
 
     vector<string> tokens = SplitString(name, '_');
 
-    attr_page_num = String2Uint64(tokens[tokens.size() - 1]);
+    attr_page_num =
+                 static_cast<int32_t>(String2Uint64(tokens[tokens.size() - 1]));
   }
 
 
--- a/cvmfs/magic_xattr.cc
+++ b/cvmfs/magic_xattr.cc
@@ -245,9 +245,9 @@ bool ChunkListMagicXattr::PrepareValueFenced(int32_t attr_page_num) {
                 "'chunked', but no chunks found.", path_.c_str());
       return false;
     } else {
-      const int32_t kB_limit = 64 * 1024;
-      const int32_t header_size = 120;
-      const int32_t max_digits_offset =
+      const size_t kB_limit = 64 * 1024;
+      const size_t header_size = 120;
+      const size_t max_digits_offset =
                       StringifyInt(chunks.At(chunks.size()-1).offset()).size();
 
       // entry_size is calculated to find out how many entries we can show per
@@ -264,12 +264,13 @@ bool ChunkListMagicXattr::PrepareValueFenced(int32_t attr_page_num) {
       // +      max offset within file (this should be the max size of any
       //                             chunk, but for simplicity we use this here.
       //                             this should be for sure larger.)
-      const int32_t entry_size = 3 + 3 + StringifyUint(chunks.size()).size() +
-                                 chunks.At(0).content_hash().ToString().size() +
-                                 max_digits_offset + max_digits_offset;
-      const int32_t entries_per_page = (kB_limit - header_size) / entry_size;
+      const size_t entry_size = (3ul * 2) +
+                                StringifyUint(chunks.size()).size() +
+                                chunks.At(0).content_hash().ToString().size() +
+                                max_digits_offset + max_digits_offset;
+      const size_t entries_per_page = (kB_limit - header_size) / entry_size;
 
-      const int32_t num_pages = chunks.size() / entries_per_page +
+      const size_t num_pages = chunks.size() / entries_per_page +
                               ((chunks.size() % entries_per_page != 0) ? 1 : 0);
 
       chunk_list_  = "Access different pages with 'chunk_list_<page>' ";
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by vvolkl** [comment]
The test failure on macos could be related, will have to rerun it to get the logs

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

**Comment by HereThereBeDragons** [comment]
@cernvm-bot cloudtest

File: cvmfs/cvmfs.cc [hunk 1]
```diff
@@ -1662,7 +1663,25 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+  // for different "pages" of xattr to be returned
+  // if no <num> is valid, the first page is printed
+  // as String2Uint64 failure defaults to 0
+  string paged_attr = "user.chunk_list";
+  if (attr.rfind(paged_attr, 0) == 0) {
+    attr = paged_attr;
+
+    vector<string> tokens = SplitString(name, '_');
+
+    attr_page_num =
```
jblomer, line 1679: Need to check for number of tokens equal to 2

cvmfs/cvmfs.cc
```diff
@@ -1662,7 +1663,25 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
+
+  // split xattr if of format <name>_<num>
+  // for different "pages" of xattr to be returned
+  // if no <num> is valid, the first page is printed
+  // as String2Uint64 failure defaults to 0
+  string paged_attr = "user.chunk_list";
+  if (attr.rfind(paged_attr, 0) == 0) {
+    attr = paged_attr;
+
+    vector<string> tokens = SplitString(name, '_');
```
jblomer, line 1677: As we use the underscore frequently in attribute names, how about the tilde (~) as a separator for the page number.

cvmfs/cvmfs.cc
```diff
@@ -1662,7 +1663,25 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
-  const string attr = name;
+  string attr = name;
+  int32_t attr_page_num = -1;
+
+  // split xattr if of format <name>_<num>
+  // for different "pages" of xattr to be returned
+  // if no <num> is valid, the first page is printed
+  // as String2Uint64 failure defaults to 0
+  string paged_attr = "user.chunk_list";
+  if (attr.rfind(paged_attr, 0) == 0) {
```
jblomer, line 1674: We have the `HasPrefix` utility function for that.

cvmfs/magic_xattr.h
```diff
@@ -96,6 +96,7 @@ class BaseMagicXattr {
   * inside FuseRemounter::fence(), which should prevent data races.
   */
   virtual bool PrepareValueFenced() { return true; }
+  virtual bool PrepareValueFenced(int32_t /*attr_page_num*/) { return true; }
```
jblomer, line 99: How about combining these two and giving `attr_page_num` a default value of 0. The can then also become an `unsigned int`, I don't think we need then negative values any more.

cvmfs/magic_xattr.cc
```diff
@@ -222,28 +227,75 @@ std::string CatalogCountersMagicXattr::GetValue() {
...
     return false;
   }
+
   if (dirent_->IsChunkedFile()) {
     FileChunkList chunks;
     if (!xattr_mgr_->mount_point()->catalog_mgr()
                      ->ListFileChunks(path_, dirent_->hash_algorithm(), &chunks)
         || chunks.IsEmpty())
     {
+      chunk_list_ += "File marked as chunked, but no chunks found\n";
```
jblomer, line 243: We return `false` (value could not be gathered), so we should not set `chunk_list_`. It will never be shown to the user.

cvmfs/magic_xattr.cc
```diff
@@ -222,28 +227,75 @@ std::string CatalogCountersMagicXattr::GetValue() {
   return res;
 }
 
-bool ChunkListMagicXattr::PrepareValueFenced() {
-  chunk_list_ = "hash,offset,size\n";
+bool ChunkListMagicXattr::PrepareValueFenced(int32_t attr_page_num) {
+  chunk_list_ = "ERROR: No chunks! ";
   if (!dirent_->IsRegular()) {
+    chunk_list_ += "Not a regular file\n";
```
jblomer, line 233: See below

cvmfs/magic_xattr.cc
```diff
@@ -222,28 +227,75 @@ std::string CatalogCountersMagicXattr::GetValue() {
   return res;
 }
 
-bool ChunkListMagicXattr::PrepareValueFenced() {
-  chunk_list_ = "hash,offset,size\n";
+bool ChunkListMagicXattr::PrepareValueFenced(int32_t attr_page_num) {
+  chunk_list_ = "ERROR: No chunks! ";
```
jblomer, line 231: Why not keeping the initial value being the header line. The way, we don't need to replace it with the header further down.

cvmfs/magic_xattr.cc
```diff
@@ -222,28 +227,75 @@ std::string CatalogCountersMagicXattr::GetValue() {
...
                      ->ListFileChunks(path_, dirent_->hash_algorithm(), &chunks)
         || chunks.IsEmpty())
     {
+      chunk_list_ += "File marked as chunked, but no chunks found\n";
       LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogErr, "file %s is marked as "
                 "'chunked', but no chunks found.", path_.c_str());
       return false;
     } else {
-      for (size_t i = 0; i < chunks.size(); ++i) {
+      const size_t kB_limit = 64 * 1024;
```
jblomer, line 248: Let's add a good safety margin for characters that may be added elsewhere. I'd suggest 40000 characters payload max.

cvmfs/magic_xattr.cc
```diff
@@ -222,28 +227,75 @@ std::string CatalogCountersMagicXattr::GetValue() {
...
+
+      attr_page_num = attr_page_num == 0 ? 1 : attr_page_num;
+      size_t attr_page_num_fixed = static_cast<size_t>(attr_page_num) - 1;
+      size_t start_entry = entries_per_page * attr_page_num_fixed;
+      for (size_t i = start_entry;
+                  i < chunks.size() && i < start_entry + entries_per_page;
+                  ++i) {
+        chunk_list_ += StringifyUint(num_pages) + ",";
+        chunk_list_ += StringifyInt(attr_page_num) + ",";
+        chunk_list_ += StringifyUint(i) + ",";
```
jblomer, line 290: I don't think we need this level of sophistication. Here is my suggestion for a simpler way:
- Let's not show the page number and number of pages in the result, i.e. let's not modify the payload (it's ok to keep the information about pagination in the header)
- Let's create a vector of strings for the attribute value. Every element of the vector corresponds to a page. We add line by line to the string of the current page. If that page gets bigger than the limit, we append a new element to the vector. At the end (or when we reach the page number), we return that vector element.

Actually, I think we should prepare this vector entirely in `PrepareValueFenced()` and then return the vector element that corresponds to the page in `GetValue()`. This may come as a follow-up refactoring.

cvmfs/magic_xattr.cc
```diff
@@ -222,28 +227,75 @@ std::string CatalogCountersMagicXattr::GetValue() {
...
+      const size_t num_pages = chunks.size() / entries_per_page +
+                              ((chunks.size() % entries_per_page != 0) ? 1 : 0);
+
+      chunk_list_  = "Access different pages with 'chunk_list_<page>' ";
+      chunk_list_ += "( " + StringifyInt(num_pages);
+      chunk_list_ += num_pages > 1 ? " pages )\n" : " page )\n";
+
+      chunk_list_ += "max_pages,curr_page,idx,hash,offset,size\n";
+
+      attr_page_num = attr_page_num == 0 ? 1 : attr_page_num;
```
jblomer, line 282: It's ok to have the pages zero-indexed. So `attr_name` is an alias to `attr_name~0` and `attr_name~1` is the 2nd page.

test/src/690-streamingcache/main
```diff
@@ -50,9 +50,12 @@ cvmfs_run_test() {
     [ $nchunks -gt 1 ] || return 30
     mkdir /srv/cvmfs/${CVMFS_TEST_REPO}/data/external || return 31
     for c in $chunk_list; do
-      local hash=$(echo $c | cut -d, -f1)
-      cat /srv/cvmfs/${CVMFS_TEST_REPO}/data/${hash:0:2}/${hash:2}P >> \
-        /srv/cvmfs/${CVMFS_TEST_REPO}/data/external/foo || return 32
+      # ignore header
+      if [ "x$c" != "xmax_pages,curr_page,idx,hash,offset,size" ]; then
```
jblomer, line 54: We should not need this as we cut the header line from the result earlier.

jblomer (changes requested):

HereThereBeDragons:
Superseeded by #3355

PR #3337, opened 2023-07-21
Use cmake to build libarchive externals on macos

### LINKED ISSUE #3319: Build failure in libarchive on macos 13
Probably recently introduced due to an update of clang on the platform. Building the externals fails on libarchive with:

```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/_types/_intmax_t.h:32:25: error: cannot combine with previous 'long long' declaration specifier
typedef __INTMAX_TYPE__ intmax_t;
                        ^
./config.h:1342:23: note: expanded from macro 'intmax_t'
#define intmax_t long long int
```
The fix likely involves using cmake instead of the configure script (https://github.com/libarchive/libarchive/blob/master/INSTALL) but care needs  to be taken to make sure the build options stay the same.

EDIT: this seems unrelated to the system architecture, but is more likely a feature of macos 13/clang 14

vvolkl:
Fixes #3319


jblomer (changes requested):
I think we don't need libarchive at all on mac. Can we try to disable the dependency there?

vvolkl:
Right, I checked and it is indeed not needed.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
Ah, that was the problem: It seems actually easier to build libarchive than to remove it from the unittests - it is included in a bunch of places via the common headers in https://github.com/cvmfs/cvmfs/blob/devel/test/common/

jblomer:
> Ah, that was the problem: It seems actually easier to build libarchive than to remove it from the unittests - it is included in a bunch of places via the common headers in https://github.com/cvmfs/cvmfs/blob/devel/test/common/

Fair enough, then let's go with the original patch.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
The failing unittests are still from the previous patch.

PR #3338, opened 2023-07-21
Fix publication of uncompressed files through gateway

jblomer:
Suggested replacement of #2919 

I think in the original issue, the workaround is to set `CVMFS_COMPRESSION_ALGORITHM=zlib` on the gateway.

### DIFF
--- a/cvmfs/receiver/params.cc
+++ b/cvmfs/receiver/params.cc
@@ -70,22 +70,10 @@ bool GetParamsFromFile(const std::string& repo_name, Params* params) {
   params->hash_alg = shash::ParseHashAlgorithm(hash_algorithm_str);
   params->hash_alg_str = hash_algorithm_str;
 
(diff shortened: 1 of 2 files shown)


HereThereBeDragons:
dont we only want the compression for the catalog?
in theory you could have no compression on the files but the catalog still needs compression?

HereThereBeDragons:
ah just saw your comment: if its true that only the catalog is stored: then yes

jblomer:
@cernvm-bot cloudtest

jblomer:
> ah just saw your comment: if its true that only the catalog is stored: then yes

That's a good point: if the gateway is also used for normal publications, then the workaround doesn't work.

jblomer:
@cernvm-bot cloudtest

HereThereBeDragons (approved):

Issue #3290, opened 2023-06-13
Add packages and CI for Debian 12 ("bookworm")


jblomer:

HereThereBeDragons:
So i tested it on debian with fuse3. For the packages, most can be taken from here: https://github.com/cvmfs/cvmfs/blob/devel/.travis.yml
some others are `apache2` and `zlib1g-dev`  for `zlib`.

following unittests failed and had to be fixed:
```
[----------] Global test environment tear-down
[==========] 1151 tests from 128 test cases ran. (275349 ms total)
[  PASSED  ] 1145 tests.
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] T_Dns.CaresResolverMany                   // in dns.cc ares_search incorrectly returns ARES_EBADNAME instead of ARES_ENODATA
[  FAILED  ] T_Dns.CaresResolverSearchDomainSlow       // in dns.cc ares_search incorrectly returns ARES_EBADNAME instead of ARES_ENODATA
[  FAILED  ] T_Dns.NormalResolverCombinedSlow          // in dns.cc ares_search incorrectly returns ARES_EBADNAME instead of ARES_ENODATA
[  FAILED  ] T_GlueBuffer.StringHeap                   // out of memory crashes child, comment at unittests states this can happen
                                                       // /home/debian/cvmfs/cvmfs/util/smalloc.h:108: void* smmap(size_t): Assertion `(mem != MAP_FAILED) && "Out Of Memory"' failed.
[  FAILED  ] T_Util.ManagedExecExecuteBinaryDoubleFork // solved by installing valgrind
[  FAILED  ] T_Util.ManagedExecExecuteBinaryAsChild    // solved by installing valgrind

 6 FAILED TESTS
```

 it is not clear why on debian `ares_search `returns a different value. source-wise it seems identical to a working version on ubuntu 22


i also have run some client and server integration tests and those were all happy.

vvolkl:
todo: add to build containers

vvolkl:
Available for a while already, closing.

PR #3343, opened 2023-07-27
Fix packaging for debian 12

vvolkl:


vvolkl:
Partially fixes #3290

jblomer (approved):
(It's probably time to revert the if statement for fuse3: single out the platforms where fuse 3 is _not_ available.)

vvolkl:
Agreed -- will update this in the script along with python2/3 dependencies in a future PR.

PR #3113, opened 2022-11-21
Make better use of cvmfs-libs throughout the project

jblomer:

PR #3344, opened 2023-08-01
Fix test 602: static library renamed

vvolkl:
See https://github.com/cvmfs/cvmfs/pull/3113

### DIFF
--- a/test/cloud_testing/platforms/centos7_x86_64_s3_test.sh
+++ b/test/cloud_testing/platforms/centos7_x86_64_s3_test.sh
@@ -66,6 +66,7 @@ if [ $s3_retval -eq 0 ]; then
                                src/595-geoipdbupdate                        \
                                src/599-removehardlinks                      \
                                src/600-securecvmfs                          \
+                               src/602-libcvmfs                             \
                                src/605-resurrectancientcatalog              \
                                src/607-noapache                             \
                                src/608-infofile                             \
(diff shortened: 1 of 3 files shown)


jblomer (changes requested):
While we are at it, let's add this test to the quick test suite.

jblomer:
@cernvm-bot cloudtest

jblomer (approved):

vvolkl:
There's a segfault only on el7, will have to look into what is happening there.

vvolkl:
The segfault is coming from libcrypto on el7:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff709cf05 in OPENSSL_ia32_cpuid () from /lib64/libcrypto.so.10
```
Do we want to support libcvmfs on this platform or  do we just skip the test there?

vvolkl:
@cernvm-bot cloudtest

jblomer:
That's strange, it seems to only happen on the S3 platform.

vvolkl:
> That's strange, it seems to only happen on the S3 platform.

No, I already excluded the regular el7, (thinking partial test coverage now  is preferable and wanting to move this to an issue), I just forgot about el7_s3.

jblomer:
See #3351: I think there was an underlying problem with the symbol visibility in libcvmfs_crypto

PR #3345, opened 2023-08-01
update of cvmfs snapshotter

marcoverl:
update to containerd 1.7.0 and golang 1.20, and fix to avoid mounting twice the same layer


vvolkl:
Thanks! Any changes to the vendored dependencies or CI should go in a separate PR, please include only the changes to `cvmfs.go` here.

marcoverl:
ok, i've closed this and made another PR only for cvmfs.go

PR #3346, opened 2023-08-02
Fix test 705-servercheck-hashmap

vvolkl:
The test checks for the log, expecting a higher verbosity that was not set.

### DIFF
--- a/test/src/705-servercheck-hashmap/main
+++ b/test/src/705-servercheck-hashmap/main
@@ -1,6 +1,7 @@
 #!/bin/bash
 cvmfs_test_name="Hashmap to avoid duplicated lookups in cvmfs_server check"
 cvmfs_test_autofs_on_startup=false
+cvmfs_test_suites="quick"
 
 
 cvmfs_run_test() {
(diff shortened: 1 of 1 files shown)


jblomer (changes requested):
Should also become a quick test

vvolkl:
It's not that quick though; takes around 10 min. Can be made faster, so I'd propose to do this in a future PR.

EDIT, scratch that, it's 10 sec, so yes, will add it to quick suite too.

vvolkl:
@cernvm-bot cloudtest

jblomer (approved):

vvolkl:
This check here is flaky: https://github.com/cvmfs/cvmfs/pull/3346/files#diff-b4ab84b9d199714f8d17be0c0117bacbdf7dfcb5b5a13612dfda926a943332c1R41

It's not always testdata/b that is checked second. Will update the regexp/

vvolkl:
@cernvm-bot  cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3347, opened 2023-08-02
Keep fuse state during reload

jblomer:
Required to remember if the fuse connection supports expiring dentries.

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -178,6 +178,18 @@ class FuseInterruptCue : public InterruptCue {
   fuse_req_t *req_ptr_;
 };
 
+/**
+ * Options related to the fuse kernel connection. The capabilities are
+ * determined only once at mount time. If the capability trigger certain
+ * behavior of the cvmfs fuse module, it needs to be re-triggered on reload.
+ * Used in SaveState and RestoreState to store the details of symlink caching.
+ */
+struct FuseState {
+  FuseState() : version(0), has_dentry_expire(false) {}
+  unsigned version;
+  bool has_dentry_expire;
+};
+
 
 static inline double GetKcacheTimeout() {
   if (!fuse_remounter_->IsCaching())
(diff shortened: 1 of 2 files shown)


vvolkl (approved):

cvmfs/cvmfs.cc
```diff
@@ -2693,6 +2715,15 @@ static bool RestoreState(const int fd_progress,
         cvmfs::file_system_->RemapCatalogFd(0, new_root_fd);
       }
     }
+
+    if (saved_states[i]->state_id == loader::kStateFuse) {
+      SendMsg2Socket(fd_progress, "Restoring fuse state... ");
+      cvmfs::FuseState *fuse_state =
+        static_cast<cvmfs::FuseState *>(saved_states[i]->state);
+      if (fuse_state->has_dentry_expire)
```
HereThereBeDragons, line 2727: i personally prefer to have `{}` even around single line `if`s

jblomer:
@cernvm-bot cloudtest

cvmfs/cvmfs.cc
```diff
@@ -178,6 +178,18 @@ class FuseInterruptCue : public InterruptCue {
...
+/**
+ * Options related to the fuse kernel connection. The capabilities are
+ * determined only once at mount time. If the capability trigger certain
+ * behavior of the cvmfs fuse module, it needs to be re-triggered on reload.
+ * Used in SaveState and RestoreState to store the details of symlink caching.
+ */
+struct FuseState {
+  FuseState() : version(0), has_dentry_expire(false) {}
+  unsigned version;
+  bool has_dentry_expire;
```
HereThereBeDragons, line 191: i think we also need to save `cache_symlinks_` . it is set at the same location in `cvmfs.cc` as the `fuse_expire_entry_` and it also depends on if the capability is available

(the default is `cache_symlinks_ i = false`, if `CVMFS_CACHE_SYMLINKS` is set it is true, but only in `cvmfs.cc` it will double-check and disable if the capapbility is not supported)

jblomer: Yep, thanks! That has already be [done](https://github.com/cvmfs/cvmfs/pull/3347/commits/92f4fa0c66c46499d55b343ea028e7066f204efe)

cvmfs/cvmfs.cc
```diff
@@ -2693,6 +2717,17 @@ static bool RestoreState(const int fd_progress,
         cvmfs::file_system_->RemapCatalogFd(0, new_root_fd);
       }
     }
+
+    if (saved_states[i]->state_id == loader::kStateFuse) {
+      SendMsg2Socket(fd_progress, "Restoring fuse state... ");
+      cvmfs::FuseState *fuse_state =
+        static_cast<cvmfs::FuseState *>(saved_states[i]->state);
+      if (!fuse_state->cache_symlinks)
+        cvmfs::mount_point_->DisableCacheSymlinks();
```
HereThereBeDragons, line 2726: i think we need also an `EnableCacheSymlink` function, as the default is `false`. i cannot see here how we can set it to `true`

cvmfs/cvmfs.cc
HereThereBeDragons: my bad.. should have f5

HereThereBeDragons (changes requested):

cvmfs/cvmfs.cc
jblomer: Ok, for consistency with the surrounding code I'd leave it as is here.

jblomer:
@cernvm-bot cloudtest

PR #3350, opened 2023-08-03
Gracefully ignore toggling streaming cache on reload

jblomer:
Similar to what was discussed for the refcounted cache manager, this patch lets cvmfs ignore toggling of the streaming cache manager during reload. The patch is, unfortunately, relatively ugly but I think it's still better than crashing fuse modules when the streaming cache option is used.


jblomer:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

vvolkl (approved):

PR #3353, opened 2023-08-09
Fix debug reload

HereThereBeDragons:
switching which signal is send first so that legacy versions can cope with it

### DIFF
--- a/cvmfs/loader_talk.cc
+++ b/cvmfs/loader_talk.cc
@@ -64,7 +64,8 @@ static void *MainTalk(void *data __attribute__((unused))) {
       break;
     }
 
-    char command;
+    char command = '?';
+    char second_cmd = '?';
     ReloadMode reload_mode = kReloadLegacy;
     if (recv(con_fd, &command, 1, 0) > 0) {
       bool unkown_command = true;
@@ -85,7 +86,6 @@ static void *MainTalk(void *data __attribute__((unused))) {
             unkown_command = false;
           break;
           default:  // Version that can set debug on/off
-            char second_cmd;
             if (recv(con_fd, &second_cmd, 1, 0) > 0) {
               if ((second_cmd == 'd') || (second_cmd == 'n')) {
                 reload_mode = second_cmd == 'd' ? kReloadDebug : kReloadNoDebug;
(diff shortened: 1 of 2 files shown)


cvmfs/loader_talk.cc
```diff
@@ -66,22 +67,38 @@ static void *MainTalk(void *data __attribute__((unused))) {
...
+            }
+          break;
         }
-      } else if ((command != 'R') && (command != 'S')) {  // legacy support
+      }
+      if (unkown_command) {
         SendMsg2Socket(con_fd, "unknown command\n");
         continue;
-      }
+       }
```
vvolkl:
```suggestion
      }
```
whitespace

cvmfs/loader_talk.cc
```diff
@@ -66,22 +67,38 @@ static void *MainTalk(void *data __attribute__((unused))) {
...
+                reload_mode = second_cmd == 'd' ? kReloadDebug : kReloadNoDebug;
+                unkown_command = false;
+              }
+            }
+          break;
         }
-      } else if ((command != 'R') && (command != 'S')) {  // legacy support
+      }
+      if (unkown_command) {
         SendMsg2Socket(con_fd, "unknown command\n");
```
vvolkl: printing the unknown command char could be helpful

HereThereBeDragons: it is printed thats what `sendmsg2socket` does. (you might remember the error message wie had before `<repo>.cern.ch: unknown command` ... REPO CRASHED etc. that came from there

vvolkl (approved):

jblomer (commented):
@vvolkl @HereThereBeDragons  Please don't tag yet, I think this needs another small change. I'll submit a longer comment later tonight. The short version is: the timeout can trigger for various reasons, so this approach is (too) brittle.

jblomer (commented):
Apparently with this fix the order of the two commands gets swapped: first 'S/R' then 'd/n'. 

Is there a reason why this is necessary? In the previous version, the advantage was that the first command already indicated if a new cvmfs or an old cvmfs version sends it. So we don't need the timeout. It should also be backwards compatible, since an old loader would ignore the first `d/n` and then pick up the following `S/R`.

jblomer:
A problem with the previous approach is that we have a spurious "unknown command" message displayed. That should only be a presentation error though. I think we can just announce it in the release notes.

HereThereBeDragons:
it is not only a spurious `unkown command` all follow up calls end with `Reload CRASHED!` etc. It does not seem that the reload is successfully executed.

Why is the timeout not ok? it is 1sec timeout. the 2 cmds written to the pipe are immediately done one after each other. also if the 2nd cmd does not reach, it falls back to the legacy mode: just reloading in the current mode the libraries are. (unless it is an error)

jblomer:
Interesting. The problem is that we don't know for sure that the two commands are sent atomically. I think they are not, because it's two different `WritePipe` operations. Even if it was one, I think there is not a guarantee because the socket is stream oriented and not datagram oriented. So, in between the two commands, the sending process can wait an arbitrary amount of time (it could be swapped out, for instance). That is unlikely, but we have seen in the past that the cvmfs deployment scale is big enough to trigger unlikely events.

It's true though that the worst case effect is limited.  Still, we designed the original approach in a backwards compatible way. And apparently it isn't. So it would be good to understand what we got wrong. Perhaps there is another issue hidden there.

PR #3355, opened 2023-08-09
V2: Add paging to xattr 

### LINKED ISSUE #2922: xattr `chunk_list` can overflow for large files

HereThereBeDragons:
Issue #2922

Superseeds #3336 

-Page request now moved to `GetValue()`.
-New xattr wide variable `std::vector<std::string>> result_pages_` to save results. 
-Limit per page is `kMaxCharsPerPage`
-`GetValue()` is not pure virtual anymore, instead `GetValue()` of basexattr can automatically constructs output, single or multi-page xattr depending on `result_pages_` (so in theory, xattr that just get all their data in `preparedfencedvalue()` dont need their own `getvalue()` anymore

there might be more that can be extended. eg. i guess logbuffer? so far i only did the "simple" ones

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1800,6 +1800,7 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
       fuse_reply_err(req, ENOATTR);
       return;
     }
+    attribute_result.first = true;
   }
 
   if (!attribute_result.first) {
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1695,19 +1695,27 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   vector<string> tokens_mode_machine = SplitString(name, '~');
   vector<string> tokens_mode_human = SplitString(name, '@');
 
-  uint32_t attr_req_page;
+  int32_t attr_req_page;
   MagicXattrMode xattr_mode;
   string attr;
 
   if (tokens_mode_human.size() > 1) {
-    attr_req_page = static_cast<uint32_t>(String2Uint64(
+    if (tokens_mode_human[tokens_mode_human.size() - 1] == "?") {
+      attr_req_page = -1;
+    } else {
+      attr_req_page = static_cast<int32_t>(String2Uint64(
                               tokens_mode_human[tokens_mode_human.size() - 1]));
+    }
     xattr_mode = kXattrHumanMode;
     attr = tokens_mode_human[0];
   } else {
-    attr_req_page = tokens_mode_machine.size() > 1 ?
-            static_cast<uint32_t>(String2Uint64(
+    if (tokens_mode_machine[tokens_mode_machine.size() - 1] == "?") {
+      attr_req_page = -1;
+    } else {
+      attr_req_page = tokens_mode_machine.size() > 1 ?
+            static_cast<int32_t>(String2Uint64(
                       tokens_mode_machine[tokens_mode_machine.size() - 1])) : 0;
+    }
     xattr_mode = kXattrMachineMode;
     attr = tokens_mode_machine[0];
   }
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1692,12 +1692,26 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   }
   TraceInode(Tracer::kEventGetXAttr, ino, "getxattr()");
 
-  vector<string> tokens = SplitString(name, '~');
+  vector<string> tokens_mode_machine = SplitString(name, '~');
+  vector<string> tokens_mode_human = SplitString(name, '@');
 
-  const uint32_t attr_req_page = tokens.size() > 1 ?
-            static_cast<uint32_t>(String2Uint64(tokens[tokens.size() - 1])) : 0;
+  uint32_t attr_req_page;
+  MagicXattrMode xattr_mode;
+  string attr;
 
-  const string attr = tokens[0];
+  if (tokens_mode_human.size() > 1) {
+    attr_req_page = static_cast<uint32_t>(String2Uint64(
+                              tokens_mode_human[tokens_mode_human.size() - 1]));
+    xattr_mode = kXattrHumanMode;
+    attr = tokens_mode_human[0];
+  } else {
+    attr_req_page = tokens_mode_machine.size() > 1 ?
+            static_cast<uint32_t>(String2Uint64(
+                      tokens_mode_machine[tokens_mode_machine.size() - 1])) : 0;
+    xattr_mode = kXattrMachineMode;
+    attr = tokens_mode_machine[0];
+  }
+
   catalog::DirectoryEntry d;
   const bool found = GetDirentForInode(ino, &d);
   bool retval;
@@ -1749,7 +1763,7 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   string attribute_value;
 
   if (!magic_xattr.IsNull()) {
-    attribute_value = magic_xattr->GetValue(attr_req_page);
+    attribute_value = magic_xattr->GetValue(attr_req_page, xattr_mode);
   } else {
     if (!xattrs.Get(attr, &attribute_value)) {
       fuse_reply_err(req, ENOATTR);
@@ -1757,7 +1771,9 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
     }
   }
 
-  if (size == 0) {
+  if (attribute_value == "ENOENT") {
+    fuse_reply_err(req, ENOENT);
+  } else if (size == 0) {
     fuse_reply_xattr(req, attribute_value.length());
   } else if (size >= attribute_value.length()) {
     fuse_reply_buf(req, &attribute_value[0], attribute_value.length());
@@ -2083,40 +2099,42 @@ class ExpiresMagicXattr : public BaseMagicXattr {
     return true;
   }
 
-  virtual std::string GetValue(uint32_t /*requested_page*/) {
+  virtual void FinalizeValue() {
     if (catalogs_valid_until_ == MountPoint::kIndefiniteDeadline) {
-      return "never (fixed root catalog)";
+      result_pages_.push_back("never (fixed root catalog)");
+      return;
     } else {
       time_t now = time(NULL);
-      return StringifyInt( (catalogs_valid_until_ - now) / 60);
+      result_pages_.push_back(StringifyInt((catalogs_valid_until_ - now) / 60));
     }
   }
 };
 
 class InodeMaxMagicXattr : public BaseMagicXattr {
-  virtual std::string GetValue(uint32_t /*requested_page*/) {
-    return StringifyInt(
+  virtual void FinalizeValue() {
+    result_pages_.push_back(StringifyInt(
       cvmfs::inode_generation_info_.inode_generation +
-      xattr_mgr_->mount_point()->catalog_mgr()->inode_gauge());
+      xattr_mgr_->mount_point()->catalog_mgr()->inode_gauge()));
   }
 };
 
 class MaxFdMagicXattr : public BaseMagicXattr {
-  virtual std::string GetValue(uint32_t /*requested_page*/) {
-    return StringifyInt(cvmfs::max_open_files_ - cvmfs::kNumReservedFd);
+  virtual void FinalizeValue() {
+    result_pages_.push_back(StringifyInt(
+                               cvmfs::max_open_files_ - cvmfs::kNumReservedFd));
   }
 };
 
 class PidMagicXattr : public BaseMagicXattr {
-  virtual std::string GetValue(uint32_t /*requested_page*/) {
-    return StringifyInt(cvmfs::pid_); }
+  virtual void FinalizeValue() {
+    result_pages_.push_back(StringifyInt(cvmfs::pid_)); }
 };
 
 class UptimeMagicXattr : public BaseMagicXattr {
-  virtual std::string GetValue(uint32_t /*requested_page*/) {
+  virtual void FinalizeValue(){
     time_t now = time(NULL);
     uint64_t uptime = now - cvmfs::loader_exports_->boot_time;
-    return StringifyInt(uptime / 60);
+    result_pages_.push_back(StringifyInt(uptime / 60));
   }
 };
 
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1692,7 +1692,12 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   }
   TraceInode(Tracer::kEventGetXAttr, ino, "getxattr()");
 
-  const string attr = name;
+  vector<string> tokens = SplitString(name, '~');
+
+  const uint32_t attr_req_page = tokens.size() > 1 ?
+            static_cast<uint32_t>(String2Uint64(tokens[tokens.size() - 1])) : 0;
+
+  const string attr = tokens[0];
   catalog::DirectoryEntry d;
   const bool found = GetDirentForInode(ino, &d);
   bool retval;
@@ -1744,7 +1749,7 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   string attribute_value;
 
   if (!magic_xattr.IsNull()) {
-    attribute_value = magic_xattr->GetValue();
+    attribute_value = magic_xattr->GetValue(attr_req_page);
   } else {
     if (!xattrs.Get(attr, &attribute_value)) {
       fuse_reply_err(req, ENOATTR);
@@ -2078,7 +2083,7 @@ class ExpiresMagicXattr : public BaseMagicXattr {
     return true;
   }
 
-  virtual std::string GetValue() {
+  virtual std::string GetValue(uint32_t /*requested_page*/) {
     if (catalogs_valid_until_ == MountPoint::kIndefiniteDeadline) {
       return "never (fixed root catalog)";
     } else {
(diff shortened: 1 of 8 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/magic_xattr.cc
```diff
@@ -482,42 +514,93 @@ std::string ProxyMagicXattr::GetValue() {
...
 
-std::string PubkeysMagicXattr::GetValue() {
-  return pubkeys_;
+std::string PubkeysMagicXattr::GetValue(uint32_t requested_page) {
+  result_pages_.clear();
+
+  if (pubkeys_.size() <= kMaxCharsPerPage) {
+    result_pages_.push_back(pubkeys_);
+  } else {
+    // we need to creat our own pages, no need to do it in the fuse fence
```
vvolkl:
```suggestion
    // we need to create our own pages, no need to do it in the fuse fence
```
typo

cvmfs/magic_xattr.cc
```diff
@@ -195,11 +195,33 @@ void MagicXattrManager::SanityCheckProtectedXattrs() {
...
+std::string BaseMagicXattr::HeaderMultipage(uint32_t max_pages,
+                                            uint32_t requested_page) {
+  return "# Access page at idx: " + StringifyUint(requested_page) + ". " +
+         "Total num pages: " + StringifyUint(max_pages) +
+         " (access other pages: xattr~<page_num>, starting " +
+         " with 0)\n\n";
+}
+
+std::string BaseMagicXattr::GetValue(uint32_t requested_page) {
+  if (result_pages_.size() == 1) {
```
jblomer: I think this can fail if `result_pages_.size()` is zero, e.g. if the proxy list is empty.

cvmfs/magic_xattr.cc
```diff
@@ -482,42 +514,93 @@ std::string ProxyMagicXattr::GetValue() {
...
+std::string PubkeysMagicXattr::GetValue(uint32_t requested_page) {
+  result_pages_.clear();
+
+  if (pubkeys_.size() <= kMaxCharsPerPage) {
+    result_pages_.push_back(pubkeys_);
+  } else {
+    // we need to create our own pages, no need to do it in the fuse fence
+    result_pages_.push_back("pubkey size " + StringifyUint(pubkeys_.size()));
+    uint64_t start_idx = 0;
+    uint64_t end_idx = kMaxCharsPerPage - 400;  // find a hit for the end of a
```
jblomer: What if we don't find the marker within these bounds?

I think it is fine to cut the attribute in-between a key. That makes our code simpler.

jblomer (changes requested):
Almost there, but some minor fixes needed. Also two more unit test:
- Multi-page attributes with multiple pages
- Empty multi-pages

cvmfs/magic_xattr.cc
HereThereBeDragons: why?
if the `size != 0` it goes to the `HeaderMultipage` but only prints a page if it really exists. if not you just get an unnecessary `HeaderMultipage` back without any content?

cvmfs/magic_xattr.cc
HereThereBeDragons: the else case is only hit if we have at least a string of all pubkeys >40000 symbols. otherwise we just return the string we get (can be empty)
i found it much nicer to cut after `-----END PUBLIC KEY-----` as you might want to compare pub keys directly (with an easy `ctrl+c`) and it is way nicer to copy. i dont think this extra logic creates so much overhead. it is already written and i tested it by hand.

HereThereBeDragons:
for the unittests:
- ok for testing multiple pages for multipage attribute
- i am not sure how we should test an empty multi page attribute? this should never happen. the logic around it should not create it. a multi page attribute should only be created if the output is larger `kMaxCharsPerPage`. For this we would need to have an MagicXattr that is faulty implemented to allow this. Also In the end i dont think it makes any differences because it will just return the string that is available at `std::vector<std::string> result_pages_[<mypage>]` called from `BaseMagicXattr::GetValue(uint32_t requested_page)` even if it is just an empty string (or ignored if the requested page is of larger number than the `result_pages_` large is)

cvmfs/magic_xattr.cc
jblomer: So, if I understand correctly, the output is
  - Only the multi page header if `result_pages_` is empty
  - The page without header, independent of the requested page, if `result_pages_` contains a single page
  - Multi page header and the requested page if `result_pages_` has more than 1 page. In this case, only the header if the requested page is out of range.

That looks inconsistent to me. I think the result should be empty in the first case, i.e. if `result_pages_` is empty. I think we should check the requested page also in the case that `result_pages_` has only a single page.

cvmfs/magic_xattr.cc
jblomer: The extended attribute interface is open to regular users, and unforeseen requests have crashed the client in the past. Parsing the key files correctly will require more (automated) testing. I doubt that it is worthwhile. Scripts using the attribute anyway need to combine information from all pages. If we cut in the middle of the key, users need to combine the pages without extra whitespace (newline etc.). In the documentation of the attribute, we can suggest a command to do so.

HereThereBeDragons:
following was decided:
- `attr` is always without multipage header

distinguish between
- `attr~0`  - machine readable output, no multipage header, ENOENT if out-of-bounds page is requested
- `attr@0` - human readable output, multipage header, error message if out-of-bounds page is requested
- `attr~?` returns how many pages exist

`SignatureManager::GetActivePubkeys` will get a version that returns a vector to more easily split the pages at key boundary

Start counting pages with 0 (already done)

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/magic_xattr.h
```diff
@@ -302,69 +338,72 @@ class LogBufferXattr : public BaseMagicXattr {
...
-  virtual std::string GetValue();
+  virtual void FinalizeValue();
 };
 
 class PubkeysMagicXattr : public BaseMagicXattr {
-  std::string pubkeys_;
+  FRIEND_TEST(T_MagicXattr, MultiPageMachineModeXattr);
+  FRIEND_TEST(T_MagicXattr, MultiPageHumanModeXattr);
+
+  std::vector<std::string> pubkeys_;
```
jblomer, line 407: Should this be private? Otherwise there is no point in declaring the friend tests, I guess.

cvmfs/cvmfs.cc
```diff
@@ -1744,15 +1771,17 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
...
   } else {
     if (!xattrs.Get(attr, &attribute_value)) {
       fuse_reply_err(req, ENOATTR);
       return;
     }
   }
 
-  if (size == 0) {
+  if (attribute_value == "ENOENT") {
+    fuse_reply_err(req, ENOENT);
```
jblomer: As per the documentation of `getxattr`, this should return `ENODATA`.

cvmfs/magic_xattr.h
```diff
@@ -38,7 +44,21 @@ class MagicXattrManager;  // needed for BaseMagicXattr
...
+ * Implementation notes:
+ * - Each MagicXattr must implement it's own FinalizeValue() function which
+ *   has to write its value into result_pages_.
+ * - result_pages_ is a vector of strings. Each string is <= kMaxCharsPerPage
+ * - BaseMagicXattr::GetValue() takes care of clearing result_pages_ before each
+ *   new call of FinalizeValue(), and takes care of adding a header for the
+ *   human readable version
+ * - In case an invalid request is done in machine-readable mode, a string with
+ *   "ENOENT" will be returned, while in human-readable mode a more verbose
+ *   error message will be returned
```
jblomer: Like this we won't be able to return `ENOENT` as an attribute value. I don't think that we currently do, but I still find this not ideal.

I would suggest returning an `std::pair<std::string, bool>` from `GetValue`. The boolean can be used to indicate if the call was successful or not.

cvmfs/magic_xattr.cc
```diff
@@ -195,12 +195,55 @@ void MagicXattrManager::SanityCheckProtectedXattrs() {
   }
 }
 
+std::string BaseMagicXattr::HeaderMultipageHuman(uint32_t max_pages,
```
jblomer: We don't need to pass `max_pages`, I think we know it is `result_pages_.size()`.

cvmfs/magic_xattr.cc
```diff
@@ -195,12 +195,55 @@ void MagicXattrManager::SanityCheckProtectedXattrs() {
...
+                                            uint32_t requested_page) {
+  return "# Access page at idx: " + StringifyUint(requested_page) + ". " +
+         "Total num pages: " + StringifyUint(max_pages) +
+         " (access other pages: xattr~<page_num>, starting " +
+         " with 0)\n";
```
jblomer: Maybe also explain how to retrieve the number of pages.

cvmfs/magic_xattr.cc
```diff
@@ -195,12 +195,55 @@ void MagicXattrManager::SanityCheckProtectedXattrs() {
...
+                                     const MagicXattrMode mode) {
+  result_pages_.clear();
+  FinalizeValue();
+
+  std::string res = "";
+  if (mode == kXattrMachineMode) {
+    if (requested_page >= static_cast<int32_t>(result_pages_.size())) {
+      return "ENOENT";
+    }
+    if (requested_page == -1) {
```
jblomer, line 216: This will segfault if `requested_page = -2)`. Check for `<0`.

cvmfs/magic_xattr.cc
```diff
@@ -195,12 +195,55 @@ void MagicXattrManager::SanityCheckProtectedXattrs() {
...
+      return "num_pages, " + StringifyUint(result_pages_.size());
+    }
+  } else {
```
jblomer: Be specific with `else if (mode == ...)` and add a final else that panics. This way we notice immediately if we ever going to extend the `MagicXattrMode` enum.

cvmfs/magic_xattr.cc
```diff
@@ -195,12 +195,55 @@ void MagicXattrManager::SanityCheckProtectedXattrs() {
...
+             + "Access them with xattr~<page_num> (machine-readable mode) "
+             + "or xattr@<page_num> (human-readable mode).\n"
+             + "Use xattr@? or xattr~? to get extra info about the attribute";
+    } else if (requested_page == -1) {
+      return "Access xattr with xattr~<page_num> (machine-readable mode) or "
+             + std::string(" xattr@<page_num> (human-readable mode).\n")
+             + "Pages available: " + StringifyUint(result_pages_.size());
+    } else {
+      res = HeaderMultipageHuman(result_pages_.size(), requested_page);
+      result_pages_[requested_page];
```
jblomer: This line seems like a leftover.

cvmfs/magic_xattr.cc
```diff
@@ -195,12 +195,55 @@ void MagicXattrManager::SanityCheckProtectedXattrs() {
...
+      return "num_pages, " + StringifyUint(result_pages_.size());
+    }
+  } else {
+    if (requested_page >= static_cast<int32_t>(result_pages_.size())) {
+      return "Page requested does not exists. There are "
+             + StringifyUint(result_pages_.size()) + " pages available.\n"
+             + "Access them with xattr~<page_num> (machine-readable mode) "
+             + "or xattr@<page_num> (human-readable mode).\n"
+             + "Use xattr@? or xattr~? to get extra info about the attribute";
+    } else if (requested_page == -1) {
```
jblomer, line 228: Also here: check for `<0` otherwise it's prone to segfaults.

cvmfs/magic_xattr.cc
```diff
@@ -214,16 +257,20 @@ bool CatalogCountersMagicXattr::PrepareValueFenced() {
...
 
 bool ChunkListMagicXattr::PrepareValueFenced() {
-  chunk_list_ = "hash,offset,size\n";
+  chunk_list_.clear();
+
+  const std::string header = "hash,offset,size\n";
+  std::string chunk_list_page_(header);
   if (!dirent_->IsRegular()) {
     return false;
   }
```
jblomer: In this case we would have previously returned the header, but I think we don't anymore. We should restore that behavior.

cvmfs/magic_xattr.cc
```diff
@@ -492,33 +556,77 @@ static std::string ListProxy(download::DownloadManager *dm) {
...
+
+    if (buf.size() > BaseMagicXattr::kMaxCharsPerPage) {
+      result_pages->push_back(buf);
+      buf = "";
+    }
+  }
+
+  if (buf.size() > 0) {
+    result_pages->push_back(buf);
   }
```
jblomer: If we have no proxies defined, we now return an error for the corresponding attributes instead of an empty string. I think  we should maintain the current behavior.

cvmfs/magic_xattr.cc
```diff
@@ -492,33 +556,77 @@ static std::string ListProxy(download::DownloadManager *dm) {
...
+
+  if (full_size == 0) {
+    return;
+  }
+
+  if (full_size <= kMaxCharsPerPage) {
+    std::string res = "";
+
+    for (size_t i = 0; i < pubkeys_.size(); i++) {
+      res += pubkeys_[i] + "\n";
```
jblomer: Are you sure that this doesn't add an extra blank line between the pubkeys?

cvmfs/magic_xattr.cc
```diff
@@ -492,33 +556,77 @@ static std::string ListProxy(download::DownloadManager *dm) {
...
+    }
+
+    result_pages_.push_back(res);
+  } else {
```
jblomer: It seems like we don't need the `if` branch. This `else` branch already covers all cases.

test/src/087-xattrs/main
```diff
@@ -117,6 +117,22 @@ cvmfs_run_test() {
   [ "x$proxy_list" != "x" ] || return 21
   [ "x$proxy_list_external" != "x" ] || return 22
 
+  # check different access modes 
+  echo "*** Check different access modes: human vs machine"
+  runc_chunk_list_machine=$(get_xattr chunk_list~0 /cvmfs/grid.cern.ch/vc/containers/runc)
+  runc_chunk_list_human=$(get_xattr chunk_list@0 /cvmfs/grid.cern.ch/vc/containers/runc)
```
jblomer, line 124: Also test `~?` and `@?`

jblomer, line 124: Also add checks for mistyped attr names: page numbers `-1`, `-2`, and `<string that is not a number>`, and nothing (`chunk_list~`)

jblomer (changes requested):

cvmfs/magic_xattr.h
HereThereBeDragons: by default visibility within a class is private. so the `FRIEND_TEST` is needed. ( i also checked it just to make sure and it gives compilation errors)

jblomer: Oh, of course! I got confused.

cvmfs/magic_xattr.cc
HereThereBeDragons: done

cvmfs/magic_xattr.cc
HereThereBeDragons: done

cvmfs/magic_xattr.cc
HereThereBeDragons: as discussed added `assert`

cvmfs/magic_xattr.cc
HereThereBeDragons: done

cvmfs/magic_xattr.cc
HereThereBeDragons: true, removed it

cvmfs/magic_xattr.cc
HereThereBeDragons: handled by assert in the beginning of the function

cvmfs/magic_xattr.cc
HereThereBeDragons: good catch

cvmfs/magic_xattr.cc
HereThereBeDragons: ok removed it

cvmfs/magic_xattr.cc
HereThereBeDragons: i added it. but in the end i dont think we will manage ever to get in this position, unless you magically can mount it with a broken config.

set proxy="direct"
```
attr -g proxy_list /cvmfs/my.test.repo
Attribute "proxy_list" had a 7 byte value for /cvmfs/my.test.repo:
DIRECT
```

replacing current proxys with ""
```
sudo cvmfs_talk proxy set ' '
lhcb.cern.ch:
OK


 $ attr -g proxy_list /cvmfs/lhcb.cern.ch/
Attribute "proxy_list" had a 120 byte value for /cvmfs/lhcb.cern.ch/:
http:// 
http://128.142.161.84:3126
http://128.142.248.156:3126
http://131.225.188.245:3126
http://131.225.188.246:3126
```

set proxy=""
```
ls /cvmfs/my.test.repo/
ls: cannot access '/cvmfs/my.test.repo/': No such file or directory
```

cvmfs/magic_xattr.h
HereThereBeDragons: do you really think this is realistic? so far if an attribute is valid but is empty it returns an empty string.

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/magic_xattr.h
HereThereBeDragons: changed it but to `std::pair<bool, std::string>`

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

PR #3359, opened 2023-08-11
Fix reloading on pre-2.11 socket

vvolkl:


vvolkl:
Fixes #2897

cvmfs/loader_talk.cc
```diff
@@ -136,19 +136,58 @@ int MainReload(const std::string &socket_path, const bool stop_and_go,
...
+  // both the commands and react according to what we read from the socket.
+
+  ssize_t retval;
+  do {
+    retval = send(socket_fd, commands, 2, MSG_NOSIGNAL);
+  } while ((retval <= 0) && (errno == EINTR));
+
+  if (retval <= 0) {
+    LogCvmfs(kLogCvmfs, kLogStderr, "Sending reload command failed!");
+    return 103;
```
HereThereBeDragons, line 155: do we have here some constant for the return type? PANIC or something like this?

cvmfs/loader_talk.cc
```diff
@@ -136,19 +136,58 @@ int MainReload(const std::string &socket_path, const bool stop_and_go,
...
 
   char buf;
-  int retval;
+  std::string first_line;
+  bool past_first_line = false;
   while ((retval = read(socket_fd, &buf, 1)) == 1) {
     if (buf == '~')
       break;
+
+    if (first_line == "unknown command") {
```
HereThereBeDragons, line 165: `std::string::compare`?

cvmfs/loader_talk.cc
```diff
@@ -136,19 +136,58 @@ int MainReload(const std::string &socket_path, const bool stop_and_go,
...
+      // We have a pre-2.11 loader, reconnect to the socket
+      LogCvmfs(kLogCvmfs, kLogStdout,
+               "Connecting in backwards compatibility mode");
+      close(socket_fd);
+      socket_fd = ConnectSocket(socket_path);
+      if (socket_fd < 0) {
+        LogCvmfs(kLogCvmfs, kLogStderr, "reconnecting failed!");
+        return 104;
```
HereThereBeDragons, line 173: same here: constant for the return type? PANIC or something like this?

cvmfs/loader_talk.cc
vvolkl: `==` is more readable, compare only makes sense if you are interested in the lexical ordering

cvmfs/loader_talk.cc
vvolkl: Introducing named constants would be nice, but should be done consistently for the whole file, I'd propose to do this in a future PR.

cvmfs/loader_talk.cc
```diff
@@ -136,19 +136,58 @@ int MainReload(const std::string &socket_path, const bool stop_and_go,
...
+      }
+      WritePipe(socket_fd, &commands[1], 1);
+      first_line.clear();
+      past_first_line = true;
+      continue;
+    }
+
+    if (!past_first_line) {
```
HereThereBeDragons, line 183: maybe add a comment that the socket only receives char by char so we need to construct the entire string and check if it is equal to "unknown command"?

cvmfs/loader_talk.cc
```diff
@@ -136,19 +136,58 @@ int MainReload(const std::string &socket_path, const bool stop_and_go,
...
+      if (buf == '\n') {
+        LogCvmfs(kLogCvmfs, kLogStdout, "%s", first_line.c_str());
```
HereThereBeDragons, line 185: if we print it here: i think we will print double  `unkown command unkown command\n\n`  due to the logcvmfs in line 191 (also note because of this if here we will print the full command again before printing another `\n`

vvolkl: I don't think so: if we're not past the first line, there is the `continue` on line 188, and if we are past the first line this log won't be executed - but let me double check.

HereThereBeDragons (changes requested):
looks good.
i also build on and switched between old version that did not have the reload switch mechanics and this one. and vice versa. seems to be no issues there.

some small improvements for the documentation would be nice (maybe in next PR if too much work), otherwise its good for me

cvmfs/loader_talk.cc
HereThereBeDragons: ah thats my c/java heritage

cvmfs/loader_talk.cc
HereThereBeDragons: on the second look i think you are right..
but then i suggest deleteing the logcvmfs in line 191 and the continue statements in 178 and 188. they seem unnecessary.

vvolkl: It's necessary: for printing everything that comes over the socket after "Connecting in backwards compatibility mode"
```
alice.cern.ch: Debug: using library /usr/lib/libcvmfs_fuse3_stub.so
alice.cern.ch: Connecting to CernVM-FS loader... done
alice.cern.ch: Connecting in backwards compatibility mode
alice.cern.ch: Entering maintenance mode
alice.cern.ch: Draining out kernel caches (up to 60s)
alice.cern.ch: Blocking new file system calls
alice.cern.ch: Waiting for active file system calls
alice.cern.ch: Saving inode tracker
alice.cern.ch: Saving negative entry cache
alice.cern.ch: Saving page cache entry tracker
alice.cern.ch: Saving chunk tables
alice.cern.ch: Saving inode generation
alice.cern.ch: Saving fuse state
alice.cern.ch: Saving open files table
alice.cern.ch: Unloading Fuse module
alice.cern.ch: Waiting for the delivery of SIGUSR1...
alice.cern.ch: Re-Loading Fuse module
alice.cern.ch: Restoring inode tracker...  done
alice.cern.ch: Restoring dentry tracker...  done
alice.cern.ch: Restoring page cache entry tracker...  done
alice.cern.ch: Restoring chunk tables...  done
alice.cern.ch: Restoring inode generation...  done
alice.cern.ch: Restoring fuse state...  done
alice.cern.ch: Restoring open files table... done
alice.cern.ch: Restoring open files counter...  done
alice.cern.ch: Releasing saved glue buffer
alice.cern.ch: Releasing saved dentry tracker
alice.cern.ch: Releasing saved page cache entry cache
alice.cern.ch: Releasing chunk tables
alice.cern.ch: Releasing saved inode generation info
alice.cern.ch: Releasing fuse state
alice.cern.ch: Releasing saved open files table
alice.cern.ch: Releasing open files counter
alice.cern.ch: Activating Fuse module
```

cvmfs/loader_talk.cc
vvolkl: done

PR #3367, opened 2023-08-18
Add BuildRequires: git to the main rpm package

DrDaveD:
When building from the source rpm for osg I found this new BuildRequires was missing.


DrDaveD:
I patched it for osg so we do not require another release just for this, but this fix is for the next release.

vvolkl:
Hi Dave,

thanks for spotting this! It looks like the git dependency crept in with the upgrade of pacparser. It's really not needed there though, and ca ben easily patched out, which I could also do for the next version. Let me know if you prefer to  merge this or remove it in the  pacparser external.

DrDaveD:
It would probably be better to remove the dependency in the external.

vvolkl:
Superseded by #3376

PR #3372, opened 2023-08-25
Performance benchmark + visualization

HereThereBeDragons:
currently user-defined parameters must be changed in `start_... .py`

if everything looks good we should go should add default settings and let the rest be overwritten by command line arguments


test/performance-benchmark/util_benchmark/benchmark_cmds.py
```diff
@@ -0,0 +1,117 @@
...
+      for line in cvmfs_config:
+        print(line, end="")
+
+
+
+################################################################################
+################################################################################
+################################################################################
+
+#        ANYTHING BELOW HERE: DEPREACTED AND NOT USED ANYMORE
```
vvolkl: Please remove

test/performance-benchmark/util_visualization/visualization_time.py
```diff
@@ -0,0 +1,197 @@
...
+  "inode_cache.n_forget": "Inode cache: #Forgets",
+  "inode_cache.n_hit": "Inode cache: #Hits",
+  "inode_cache.n_insert": "Inode cache: #Inserts",
+  "inode_cache.n_insert_negative": "Inode cache: #Negative inserts",
+  "inode_cache.n_miss": "Inode cache: #Misses",
+  "inode_cache.n_replace": "Inode cache: #Replaces",
+  "inode_cache.n_update": "Inode cache: #Updates",
+  "inode_cache.n_update_value": "Inode cache: #Value changes",
+  "inode_cache.sz_allocated": "Inode cache: #Allocated bytes",
+  "inode_cache.sz_size": "Inode cache: Total size", # TODO which size? bytes?
```
vvolkl: Yes, I believe this is given in bytes. Same below.

vvolkl (commented):
Nice, thanks! I think this PR could benefit from one more round of cleanup, to remove stale/commented code.

We'll also want to move this in a `client` subfolder as we're adding server performance benchmark and more complicated setups, but it can be done in a future PR.

test/performance-benchmark/util_visualization/visualization_time.py
HereThereBeDragons: done

test/performance-benchmark/util_benchmark/benchmark_cmds.py
HereThereBeDragons: done

HereThereBeDragons:
also moved everything to subdir `performance-benchmark/client`

PR #3376, opened 2023-08-29
Patch out the `git` build dependency of pacparser

vvolkl:
Updating to pacparser 1.4.2 accidentally introduced a new build dependency on git. This just records the git revision if present, which isn't needed in our use case, so removing it via patch lets us keep a smaller set of build dependencies, and, accidentally, also remove a prior patch that fixed a python version issue.


vvolkl:
@cernvm-bot  cloudtest

DrDaveD (approved):
Looks good to me, although I haven't tested it

vvolkl:
All our build images have git installed, but I did run a manual test on a machine without git where it compiled fine.

Issue #3384, opened 2023-09-06
cvmfs repository stopped getting updates after cvmfs client 2.11.0 update


smuzaffar:
He,
  On few (4 out of 50) of our OpenStack based build nodes at CERN, we noticed that after the update to `cvmfs-2.11.0` some of repositories stop getting updates. System logs show errors messages like [a]. All our build nodes are identical so no idea why this happened on only few nodes. Running `cvmfs_config killall; cvmfs_config reload` fixed the issue though

[a]
```
Sep  4 17:40:15 cmsbuild126.cern.ch cvmfs2[15200]: (singularity.opensciencegrid.org) switching host from http://klei.nikhef.nl:8000/cvmfs/singularity.opensciencegrid.org to http://cvmfs-egi.gridpp.rl.ac.uk:8000/cvmfs/singularity.opensciencegrid.org (reset host)
Sep  4 17:40:15 buildnode cvmfs2[6596]: (unpacked.cern.ch) switching host from http://cernvmfs.gridpp.rl.ac.uk/cvmfs/unpacked.cern.ch to http://cvmfs-stratum-one.cern.ch/cvmfs/unpacked.cern.ch (reset host)
Sep  4 17:40:15 buildnode cvmfs2[6596]: (unpacked.cern.ch) whitelist verification failed (6): invalid whitelist signature
Sep  4 17:40:15 buildnode cvmfs2[6596]: (unpacked.cern.ch) failed to fetch manifest (8 - bad whitelist), trying another stratum 1
Sep  4 17:40:15 buildnode cvmfs2[6596]: (unpacked.cern.ch) switching host from http://cvmfs-stratum-one.cern.ch/cvmfs/unpacked.cern.ch to http://cernvmfs.gridpp.rl.ac.uk/cvmfs/unpacked.cern.ch (manually triggered)
Sep  4 17:40:18 buildnode cvmfs2[6325]: (cms-ib.cern.ch) whitelist verification failed (6): invalid whitelist signature
Sep  4 17:40:18 buildnode cvmfs2[6325]: (cms-ib.cern.ch) failed to fetch manifest (8 - bad whitelist), trying another stratum 1
Sep  4 17:40:18 buildnode cvmfs2[6325]: (cms-ib.cern.ch) switching host from http://cvmfs-stratum-one.cern.ch/cvmfs/cms-ib.cern.ch to http://cvmfs.fnal.gov/cvmfs/cms-ib.cern.ch (manually triggered)
[...]

PR #3387, opened 2023-09-08
Make mount/reload fail if CVMFS_CONFIG_REPOSITORY unreachable

HereThereBeDragons:
Issue #3384 (explanation of failure will be there)

- if `CVMFS_CONFIG_REPOSITORY` is unreachable: fallback to `default`.
- if default `CVMFS_CONFIG_REPOSITORY` not reachable: PANIC/exit process

- added note to look in `syslog`
- added other output when pubkeys are empty

- 2nd commit just consists of maybe TODOs if we also want to add some checking around those `ParseDefault()`

tested on:
- mounting with invalid default CVMFS_CONFIG_REPOSITORY
- mounting with valid default CVMFS_CONFIG_REPOSITORY and then changing it to invalid one and reloading

### DIFF
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1589,8 +1589,14 @@ bool MountPoint::CreateSignatureManager() {
     boot_status_ = loader::kFailSignature;
     return false;
   }
-  LogCvmfs(kLogCvmfs, kLogDebug, "CernVM-FS: using public key(s) %s",
+  if (public_keys.size() > 0) {
+    LogCvmfs(kLogCvmfs, kLogDebug, "CernVM-FS: using public key(s) %s",
            public_keys.c_str());
+  } else {
+    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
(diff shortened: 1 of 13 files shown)


cvmfs/mountpoint.cc
```diff
@@ -1589,8 +1589,14 @@ bool MountPoint::CreateSignatureManager() {
     boot_status_ = loader::kFailSignature;
     return false;
   }
-  LogCvmfs(kLogCvmfs, kLogDebug, "CernVM-FS: using public key(s) %s",
+  if (public_keys.size() > 0) {
+    LogCvmfs(kLogCvmfs, kLogDebug, "CernVM-FS: using public key(s) %s",
            public_keys.c_str());
+  } else {
+    LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
+                                              "CernVM-FS: NO PUBLIC KEY GIVEN");
```
jblomer, line 1597:
```suggestion
                                              "no public key loaded");
```

HereThereBeDragons:
Superseeded by #3390

PR #3390, opened 2023-09-12
fix client reload if remote config repo is unavailable

HereThereBeDragons:
Issue #3384 

Superseeds #3387 

parameter `CVMFS_CONFIG_REPO_DEFAULT_ENV` is set in the client config by the remote config repo. if the remote config repo is not available a local config should be loaded that is guarded behind `CVMFS_CONFIG_REPO_DEFAULT_ENV`. as such `/etc/cvmfs/default.conf` must reset `CVMFS_CONFIG_REPO_DEFAULT_ENV` to know if remote repo is available or not

+ add output if public keys are not set

### DIFF
--- a/packaging/debian/config-default/changelog
+++ b/packaging/debian/config-default/changelog
@@ -1,4 +1,4 @@
-cvmfs-config-default (2.0-2) lucid; urgency=low
+cvmfs-config-default (2.1-0) lucid; urgency=low
 
   * During reload/mount: fallback to local config repo if remote repo is unavailable
 
--- a/packaging/debian/config-default/changelog
+++ b/packaging/debian/config-default/changelog
@@ -1,4 +1,4 @@
-cvmfs-config-default (2.1-0) lucid; urgency=low
+cvmfs-config-default (2.1-1) lucid; urgency=low
 
   * During reload/mount: fallback to local config repo if remote repo is unavailable
 
--- a/packaging/debian/config-default/changelog
+++ b/packaging/debian/config-default/changelog
@@ -1,3 +1,9 @@
+cvmfs-config-default (2.0-2) lucid; urgency=low
+
+  * During reload/mount: fallback to local config repo if remote repo is unavailable
+
+ -- HereThereBeDragons <HereThereBeDragons@users.noreply.github.com>  Wed, 13 Sep 2023 13:51:10 2023 +0200
+
 cvmfs-config-default (2.0-1) lucid; urgency=low
 
   * Various fixes, use config repository
--- a/ci/cvmfs/deb.sh
+++ b/ci/cvmfs/deb.sh
@@ -19,7 +19,7 @@ CVMFS_SOURCE_LOCATION="$1"
 CVMFS_RESULT_LOCATION="$2"
 CVMFS_NIGHTLY_BUILD_NUMBER="${3-0}"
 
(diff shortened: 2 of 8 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
just as a note: when using `wipecache` or `reload` the `/proc/<pid>/environ` is not updated and can show wrong information

DrDaveD:
Consider setting `CVMFS_CONFIG_REPO_DEFAULT_ENV=` in mount/default.d/50-cern.conf instead of mount/default.conf, because the former goes into the cvmfs-config-default rpm and the latter goes into the cvmfs client rpm.  The use of `CVMFS_CONFIG_REPO_DEFAULT_ENV` is entirely a concept of the cvmfs-config-default rpm so logically it belongs there.  On the other hand, putting it in default.conf should cause no harm anywhere because nobody else uses it.

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
cloudtest all green but macos that didnt start. so rerunning cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer (approved):

jblomer (commented):
One think I forgot: since we want to release a new cvmfs-config-default RPM, please update the package version / changelog accordingly.

HereThereBeDragons:
please have a look if i changed all the correct files (and not too much)

packaging/debian/config-default/changelog
```diff
@@ -1,3 +1,9 @@
+cvmfs-config-default (2.0-2) lucid; urgency=low
```
DrDaveD: Shouldn't that be 2.0-3?

jblomer:
> please have a look if i changed all the correct files (and not too much)

In principle it's the right files. The common convention is to increase the version number if the content changes, and the release number if the packaging itself changes. So in this case, I think we should have a new minor version number, release number 1.

packaging/debian/config-default/changelog
HereThereBeDragons: if it is synchronous with the non-debian package, yes. but the changelog here does not include the 2.0-2 which the other one has. not sure if forgotten or not?

HereThereBeDragons: i tried to find the commit as the change log for the non-debian says 
```
* Sat Apr 10 2021 Jakob Blomer <jblomer@cern.ch> - 2.0-2
- Fix RPM linter errors
```

but i could not find it
```
commit 9660032838612c6f95ba201069175c02a34b0307
Author: Dave Dykstra <2129743+DrDaveD@users.noreply.github.com>
Date:   Mon Apr 12 09:14:43 2021 -0500

    add common SanitizeParameterAssignment function

commit 93179d74a6a631853d4042a3ace8dcc95db500cb
Author: Dave Dykstra <2129743+DrDaveD@users.noreply.github.com>
Date:   Thu Apr 8 17:02:38 2021 -0500

    skip attempting to parse parameters with spaces in them
```

HereThereBeDragons:
maybe `cvmfs/packaging/debian/config-default/maintscript` needs adjustment?

jblomer:
Sorry, one more change: the initial release number for a new version should be 1. Release 0 is by convention reserved for pre-releases of new versions.

vvolkl (approved):

PR #3408, opened 2023-10-06
Limit initscripts dependency to rhel <=7

vvolkl:
The `initscripts` dependency has already been removed on debian, and should not be needed after el 6. Removing this fixes an issue for IT, where yum check reports a spurious error, related to: https://access.redhat.com/solutions/6969215

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -222,7 +222,9 @@ Requires: insserv
 Requires: libcap-progs
   %endif
 %else
+  %if 0%{?rhel} < 8
 Requires: initscripts
+  %endif
 %endif
 Requires: bash
 Requires: coreutils


jblomer (approved):

jblomer:
(Another way for patch release PR is to open them against the version branch, i.e. 2.11. After merging, you can then simply to `git merge 2.11` on devel to upstream the patch. Otherwise, you'd need to cherry-pick the commits from devel to the 2.11 branch.)

DrDaveD:
I note that this change is on the cvmfs-server subpackage only.

What in fact is using initscripts on el7?  It's not clear to me that anything is.

In addition, I see chkconfig is still required in the cvmfs client package on el7 & el8 (and at first glance suse versions, except that that check is in a `%else` case that isn't used on suse).
```
%if 0%{?suse_version}
...
%else
%if 0%{?el7}%{?el8}%{?suse_version}
Requires: chkconfig
%endif
```
I don't see where chkconfig is used either. 

(Note that chkconfig is a dependency of initscripts.  Both packages have ownership of the `/etc/init.d` and `/etc/rc*.d` directories according to rpm -ql.)

PR #3422, opened 2023-10-17
Make sure release of rpm subpackages match in addition to version

DrDaveD:
I noticed that when I have made multiple patched releases of cvmfs versions that required subpackages don't automatically get upgraded by yum when moving from one release to the next.  This fixes that.

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -165,7 +165,7 @@ Requires: util-linux
   %endif
 %endif
 Requires: cvmfs-config
-Requires: cvmfs-libs = %{version}
+Requires: cvmfs-libs = %{version}-%{release}
 
 # SELinux integration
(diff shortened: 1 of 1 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer (approved):

jblomer:
Thank you!  That issue is not great... should we unpublish the 2.11.1 release?

Hm, probably for this patch release the `cvmfs-libs-2.11.0` should work together with `cvmfs-2.11.0`.

DrDaveD:
No, the `.1` of `2.11.1` is part of the `%{version}`, that's not a problem; it's not as bad as you were thinking.  The problem is only when we make a different Release number, for example at OSG we have made a `1.2.osg36` and `1.3.osg36` already with patches applied, in addition to the `1.osg36` without a patch.

PR #3423, opened 2023-10-19
[deb] let cvmfs-fuse3 depend on cvmfs of same version

### LINKED ISSUE #3421: `cvmfs` package should block `cvmfs-fuse3` with non-identical version
Installing 2.11.0 on a Debian system (including `cvmfs-fuse3`, which is listed as a `recommends`) and subsequently upgrading the `cvmfs` package only to `2.11.1` leads to all `cvmfs` mounts failing with `transport endpoint not connected` since loading the `cvmfs-fuse3` library fails:
```
CernVM-FS: loading Fuse module... failed to load cvmfs library, tried: './libcvmfs_fuse3.so' '/usr/lib/libcvmfs_fuse3.so' '/usr/lib64/libcvmfs_fuse3.so
...
root@machine:/usr/lib# ls libcvmfs_
libcvmfs_cache.so                libcvmfs_client.so.2.11.1        libcvmfs_crypto.so               libcvmfs_fuse3_debug.so.2.11.0   libcvmfs_fuse3_stub.so           libcvmfs_fuse_debug.so.2.11.1    libcvmfs_fuse_stub.so            libcvmfs_util_debug.so.2.11.1
libcvmfs_cache.so.2.11.1         libcvmfs_crypto_debug.so         libcvmfs_crypto.so.2.11.1        libcvmfs_fuse3.so                libcvmfs_fuse3_stub.so.2.11.0    libcvmfs_fuse.so                 libcvmfs_fuse_stub.so.2.11.1     libcvmfs_util.so
libcvmfs_client.so               libcvmfs_crypto_debug.so.2.11.1  libcvmfs_fuse3_debug.so          libcvmfs_fuse3.so.2.11.0         libcvmfs_fuse_debug.so           libcvmfs_fuse.so.2.11.1          libcvmfs_util_debug.so           libcvmfs_util.so.2.11.1
root@machine:/usr/lib# ldd '/usr/lib/libcvmfs_fuse3.so'
        linux-vdso.so.1 (0x00007ffe95bb1000)
        libcvmfs_crypto.so.2.11.0 => not found
        libcvmfs_util.so.2.11.0 => not found
        libfuse3.so.3 => /lib/x86_64-linux-gnu/libfuse3.so.3 (0x00007fa020cb6000)
        libssl.so.1.1 => /lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007fa020c23000)
        libcrypto.so.1.1 => /lib/x86_64-linux-gnu/libcrypto.so.1.1 (0x00007fa02092f000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa02090d000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fa020905000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fa020738000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fa0205f4000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fa0205da000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa020406000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fa02104a000)
```

While of course orchestration (e.g. Puppet) can synchronize the upgrade of those versions, it would be best if this broken situation would be prevented by the package dependencies themselves. 

So I think the best solution would be for `cvmfs` to block `cvmfs-fuse3` with non-identical version number (so enforcing the package to be upgraded at the same time or uninstalled). Other ideas are of course welcome ;-). 

Since the `cvmfs-fuse3` package is `recommends` on recent Debian-based distros, this may catch some users by surprise when doing an upgrade of the `cvmfs` package only.

jblomer:
@olifre Please let us know if you have a comment

Fixes #3421

### DIFF
--- a/packaging/debian/cvmfs/control.in
+++ b/packaging/debian/cvmfs/control.in
@@ -69,7 +69,7 @@ Description: CernVM File System Unit Tests
 Package: cvmfs-fuse3
 Architecture: i386 amd64 armhf arm64
 #Pre-Depends: ${misc:Pre-Depends}   (preparation for multiarch support)
-Depends: fuse3, libfuse3-3
+Depends: fuse3, libfuse3-3, cvmfs (= ${binary:Version})
 #Multi-Arch: same   (preparation for multiarch support)
 Homepage: http://cernvm.cern.ch
 Description: CernVM File System fuse3 libraries


jblomer:
@cernvm-bot cloudtest

olifre:
@jblomer Many thanks, this should be the correct / best approach IIUC (c.f. https://www.debian.org/doc/manuals/debmake-doc/ch05.en.html#binnmu which recommends the same).

PR #3425, opened 2023-10-19
Fix ducc webhook

marcoverl:
fix the webhook listener scripts and make them work with wsgi-express


vvolkl:
Will merge now, but once #3427 is merged, the listener script can be revisited, this can then make use of the new capabilities of the daemon.

PR #3426, opened 2023-10-22
Suppress CVMFS_NFILES checking for refcounted cache manager

jblomer:


jblomer:
@cernvm-bot cloudtest

vvolkl (approved):

DrDaveD:
Is this fix related to an issue?  If so, which one?

vvolkl:
This solves the problem discussed in https://github.com/cvmfs/cvmfs/issues/3067#issuecomment-1761192128 - even if  the refcounting cache manager actually uses fewer file descriptors, one could not use a lower value for CVMFS_NFILES because of internal checks that were not aware of the refcounting.

PR #3429, opened 2023-10-25
parallel make externals

mharvey-jt:
Speeds up end-to-end build time when build environment has multiple cores.


vvolkl:
Thanks! I think `make -j` always tries to parallelize, regardless of core count - probably this is still better even on single core machines, but maybe adding an env var like  `make -j ${CVMFS_BUILD_EXTERNALS_NJOBS}` can be useful in some situations, and you can just leave it unset to get `make -j`. I'll add the change to this PR.

vvolkl (approved):

jblomer (approved):

Issue #3277, opened 2023-05-23
RFE: Cache recent catalogs in gateway


mharvey-jt:
To improve the latency of publication, the gateway should use a local cache for catalogs, reading and writing through it. This will avoid successive updates of a given catalog object having to make a round-trip all the way back to the backing object store.

PR #3431, opened 2023-10-31
Cache catalogs for cvmfs_receiver

HereThereBeDragons:
Issue #3277

works (manually tested) but still needs some unittest or integration test


HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer:
@HereThereBeDragons if this is ready, please rebase request a review

cvmfs/catalog_mgr_ro.cc
```diff
@@ -60,6 +102,22 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
+      PANIC(kLogStderr, "failed to create temp file when loading %s",
+                        url.c_str());
+    }
+    ctlg_context->SetSqlitePath(tmp_path);
+
+    CopyPath2File(cache_path, fcatalog);
+    fclose(fcatalog);
+  }
+
+  ctlg_context->SetHash(effective_hash);
```
HereThereBeDragons: i think we can remove this because `effective_hash` is only set in the beginning and no download/whatever action changes it?

HereThereBeDragons:
the linter complaints in `/t_catalog_merge_tool.cc` are due to the commented out function where i tried to test somehow this feature. it did not work.

if you have any good idea - aside from adding debug messages and grepping for those - please let me know

cvmfs/catalog_mgr_ro.cc
HereThereBeDragons: this was the somewhat tricky part of the function before the refactor to use `catalog_context` which replaces `catalog_path` and `catalog_hash`

```c++
 FILE *fcatalog;

  if (use_local_cache_) {
    std::string tmp_path = local_cache_dir_ + "/"
                   + effective_hash.MakePathWithoutSuffix();

    *catalog_path = local_cache_dir_ + "/"
                   + effective_hash.MakePathWithoutSuffix();

    // file is cached
    if (FileExists(catalog_path->c_str())) {
      if (!copy_to_tmp_dir_) {
        *catalog_hash = effective_hash;
        return kLoadNew;
      } else {  // for writable catalog create copy in dir_temp_
        std::string cache_path = *catalog_path;
        fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w",
                                                                  catalog_path);
        if (!fcatalog) {
          PANIC(kLogStderr, "failed to create temp file when loading %s",
                url.c_str());
        }

        CopyPath2File(cache_path, fcatalog);
        fclose(fcatalog);

        *catalog_hash = effective_hash;
        return kLoadNew;
      }
    }

    fcatalog = fopen(catalog_path->c_str(), "w");
  } else {  // use tmp file
    fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w", catalog_path);
    if (!fcatalog) {
      PANIC(kLogStderr, "failed to create temp file when loading %s",
            url.c_str());
    }
  }
```

HereThereBeDragons:
@cernvm-bot cloudtest

test/unittests/t_catalog_merge_tool.cc
```diff
@@ -157,6 +163,105 @@ TEST_F(T_CatalogMergeTool, CRUD) {
   EXPECT_FALSE(output_spec.Item("dir/dir"));
 }
 
+// TEST_F(T_CatalogMergeTool, CRUDwithCache) {
```
jblomer: What about this test? Should it be removed, and if so why?

test/common/catalog_test_tools.h
```diff
@@ -183,8 +183,9 @@ class CatalogTestTool : public ServerTool {
 
   static catalog::WritableCatalogManager* CreateCatalogMgr(
       const shash::Any& root_hash, const std::string stratum0,
-      const std::string& temp_dir, upload::Spooler* spooler,
-      download::DownloadManager* dl_mgr, perf::Statistics* stats);
+      const std::string name, const std::string& temp_dir,
```
jblomer:
```suggestion
      const std::string &name, const std::string& temp_dir,
```

test/common/catalog_test_tools.h
```diff
@@ -183,8 +183,9 @@ class CatalogTestTool : public ServerTool {
...
+      upload::Spooler* spooler, download::DownloadManager* dl_mgr,
+      perf::Statistics* stats, const std::string local_cache_dir);
```
jblomer:
```suggestion
      perf::Statistics* stats, const std::string &local_cache_dir);
```

test/common/catalog_test_tools.cc
```diff
@@ -483,12 +482,13 @@ manifest::Manifest* CatalogTestTool::CreateRepository(
 
 catalog::WritableCatalogManager* CatalogTestTool::CreateCatalogMgr(
     const shash::Any& root_hash, const std::string stratum0,
-    const std::string& temp_dir, upload::Spooler* spooler,
-    download::DownloadManager* dl_mgr, perf::Statistics* stats) {
+    const std::string name, const std::string& temp_dir,
```
jblomer: Why do we pass `name`? It seems unused.

cvmfs/catalog_mgr_ro.h
```diff
@@ -37,13 +39,33 @@ class SimpleCatalogManager : public AbstractCatalogManager<Catalog> {
...
+        if (!success) {
+          LogCvmfs(kLogCatalog, kLogStdout | kLogSyslog,
+                 "Failure during creation of local cache directory for server");
+          use_local_cache_ = false;
+          copy_to_tmp_dir_ = false;
+        }
+      } else {
+        copy_to_tmp_dir_ = false;
+        use_local_cache_ = false;
+      }
```
jblomer: Move to .cc file

cvmfs/catalog_mgr_ro.h
```diff
@@ -66,6 +88,16 @@ class SimpleCatalogManager : public AbstractCatalogManager<Catalog> {
     return (relative_path == "") ? "" : "/" + relative_path;
   }
 
+ protected:
+  bool                       use_local_cache_;  // use local cache directory
+                                                // for faster lookup of catalogs
```
jblomer: We can safe that extra state variable by using an empty `local_cache_dir_` for "nocache"

cvmfs/catalog_mgr_rw.cc
```diff
@@ -41,9 +41,10 @@ WritableCatalogManager::WritableCatalogManager(
   perf::Statistics          *statistics,
   bool                       is_balanceable,
   unsigned                   max_weight,
-  unsigned                   min_weight)
+  unsigned                   min_weight,
+  const                      std::string &dir_cache)
   : SimpleCatalogManager(base_hash, stratum0, dir_temp, download_manager,
-      statistics)
+      statistics, false, dir_cache, true)
```
jblomer:
```suggestion
      statistics, false, dir_cache, true /* copy to tmpdir */)
```

cvmfs/catalog_mgr_ro.h
```diff
@@ -37,13 +39,33 @@ class SimpleCatalogManager : public AbstractCatalogManager<Catalog> {
     const std::string          &dir_temp,
     download::DownloadManager  *download_manager,
     perf::Statistics           *statistics,
-    const bool                  manage_catalog_files = false)
+    const bool                  manage_catalog_files = false,
+    std::string                 dir_cache = "",
+    bool                        copy_to_tmp_dir = false)
```
jblomer: We always have this turned on. So remove?

cvmfs/catalog_mgr_rw.cc
```diff
@@ -1253,6 +1254,23 @@ void WritableCatalogManager::CatalogUploadCallback(
...
+  if (use_local_cache_) {
+    std::string tmp_catalog_path;
+    std::string cache_catalog_path = local_cache_dir_ + "/"
+                                  + result.content_hash.MakePathWithoutSuffix();
+    FILE *fcatalog = CreateTempFile(local_cache_dir_ + "/txn/catalog", 0666,
+                                                        "w", &tmp_catalog_path);
+    CopyPath2File(result.local_path.c_str(), fcatalog);
+    fclose(fcatalog);
+
+    rename(tmp_catalog_path.c_str(), cache_catalog_path.c_str());
```
jblomer: Check return values

cvmfs/catalog_mgr_rw.cc
```diff
@@ -1404,6 +1422,23 @@ void WritableCatalogManager::CatalogUploadSerializedCallback(
...
+  }
```
jblomer: Second use of this block. Factor out.

cvmfs/publish/repository.cc
```diff
@@ -691,7 +691,8 @@ void Publisher::ConstructSyncManagers() {
       statistics_,
       settings_.transaction().use_catalog_autobalance(),
       settings_.transaction().autobalance_max_weight(),
-      settings_.transaction().autobalance_min_weight());
+      settings_.transaction().autobalance_min_weight(),
+      "");
```
jblomer, line 695: Ok for now but eventually we do want that feature enabled in libcvmfs_server (follow-up PR).

cvmfs/catalog_mgr_ro.cc
```diff
@@ -39,13 +48,46 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
-  std::string tmp;
-
-  FILE *fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w", &tmp);
-  ctlg_context->SetSqlitePath(tmp);
-  if (!fcatalog) {
-    PANIC(kLogStderr, "failed to create temp file when loading %s",
-          url.c_str());
+  FILE *fcatalog;
+
+  // use local cache
```
jblomer:
```suggestion
```

cvmfs/catalog_mgr_ro.cc
```diff
@@ -39,13 +48,46 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
+        ctlg_context->SetSqlitePath(tmp_path);
+
+        CopyPath2File(cache_path, fcatalog);
+        fclose(fcatalog);
+
+        return kLoadNew;
+      }
+    }
+
+    // file not cache yet
```
jblomer:
```suggestion
    // file not cached yet
```

cvmfs/catalog_mgr_ro.cc
```diff
@@ -39,13 +48,46 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
+    // open file to download into "cache_dir/" + standard cvmfs file hierarchy
+    fcatalog = fopen(ctlg_context->sqlite_path().c_str(), "w");
```
jblomer: Check return value

cvmfs/catalog_mgr_ro.cc
```diff
@@ -39,13 +48,46 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
+        std::string cache_path = tmp_path;
+        fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w", &tmp_path);
+        if (!fcatalog) {
+          PANIC(kLogStderr, "failed to create temp file when loading %s",
+                            url.c_str());
+        }
+        ctlg_context->SetSqlitePath(tmp_path);
+
+        CopyPath2File(cache_path, fcatalog);
+        fclose(fcatalog);
```
jblomer: Return value

cvmfs/catalog_mgr_ro.cc
```diff
@@ -60,6 +102,22 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
+    fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w", &tmp_path);
+    if (!fcatalog) {
+      PANIC(kLogStderr, "failed to create temp file when loading %s",
+                        url.c_str());
+    }
+    ctlg_context->SetSqlitePath(tmp_path);
+
+    CopyPath2File(cache_path, fcatalog);
+    fclose(fcatalog);
+  }
```
jblomer: Factor out the copy to temp dir.

cvmfs/receiver/params.cc
```diff
@@ -59,6 +59,13 @@ bool GetParamsFromFile(const std::string& repo_name, Params* params) {
     }
   }
 
+  std::string use_local_cache_str;
+  if (parser.GetValue("CVMFS_RECEIVER_USE_LOCAL_CACHE", &use_local_cache_str)) {
```
jblomer:
```suggestion
  if (parser.GetValue("CVMFS_RECEIVER_CACHE", &use_local_cache_str)) {
```

jblomer (changes requested):
For future PRs: please split unrelated changes in different PRs (here: change of logging in ingest/sync).

Some comments inline. It's not yet clear to me if we have to think about eviction after all.

test/unittests/t_catalog_merge_tool.cc
HereThereBeDragons: "the linter complaints in `/t_catalog_merge_tool.cc` are due to the commented out function where i tried to test somehow this feature. it did not work."

it was one try to create a unittest with the mock services as you suggested. but it did not work. so please, if you have a better idea, let me know

cvmfs/publish/repository.cc
HereThereBeDragons: ok

cvmfs/catalog_mgr_ro.h
HereThereBeDragons: no this is the read-only catalog mgr - with logic that needs to work for both read only and writeable.
and on failure or empty cache dir it will be turned off.

for writeable catalog it is always on.

cvmfs/catalog_mgr_ro.h
HereThereBeDragons: i thought about it but would prefer not to. it makes the if's etc easier.

HereThereBeDragons: agreement is to just use the string and have a function that returns a bool to know if to use cache or not

cvmfs/receiver/commit_processor.cc
```diff
@@ -163,10 +163,16 @@ CommitProcessor::Result CommitProcessor::Process(
...
+  if (params.use_local_cache) {
+    local_cache_dir_ = "/var/spool/cvmfs/" + repo_name + "/cache.server";
+  }
+
   const std::string spooler_temp_dir =
-      GetSpoolerTempDir(params.spooler_configuration);
+                                GetSpoolerTempDir(params.spooler_configuration);
   assert(!spooler_temp_dir.empty());
-  assert(MkdirDeep(spooler_temp_dir + "/receiver", 0666, true));
+  assert(MkdirDeep(spooler_temp_dir + "/receiver", 0755, true));
```
HereThereBeDragons, line 173: are we happy with this set of permissions?

test/common/catalog_test_tools.cc
HereThereBeDragons: as i now switched to an integration test i fully removed those changes from the unittests

HereThereBeDragons:
the new commits include now the support of `swissknife sync` used by normal transactions to do the caching.
needed to have a "simple" integration test (new one with number 630)
because of this i changed the variable to `CVMFS_SERVER_CATALOG_CACHE`


( i checked if this resolved the unittest issue but it did not - more or less the download manager using the mockproxy did not receive any download requests)

HereThereBeDragons:
@cernvm-bot cloudtest

mharvey-jt:
This seems to work well - the other half of the puzzle is to cache the new catalogs on upload.

HereThereBeDragons:
hi @mharvey-jt 
this is already done here in  `void WritableCatalogManager::CatalogUploadCallback()`
https://github.com/cvmfs/cvmfs/blob/df323caa15936f3ce0244ba8075f77894b34f058/cvmfs/catalog_mgr_rw.cc#L1290-L1292

mharvey-jt:
so it is I missed that. Thanks! This PR looks good. 
 guess the periodic cleanup could be done by a gateway thread

mharvey-jt:
Seems like  something deletes an open catalog file on cleanup. So, when a catalog is served directly from cache copy, without first making a temporary file copy, it gets deleted.

cvmfs/catalog_mgr_ro.cc
```diff
@@ -60,6 +147,13 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
                       url.c_str(), retval, download::Code2Ascii(retval));
   }
 
+  // for writable catalog make copy in dir_temp_ that can be modified
+  if (useLocalCache() && copy_to_tmp_dir_) {
```
mharvey-jt: for some reason, catalog files are getting deleted on close, removing them from the cache. Pending fixing the root cause, had to change this to always copy to a tmp file.

cvmfs/catalog_mgr_ro.h
```diff
@@ -66,7 +64,18 @@ class SimpleCatalogManager : public AbstractCatalogManager<Catalog> {
     return (relative_path == "") ? "" : "/" + relative_path;
   }
 
+  bool useLocalCache() { return !local_cache_dir_.empty(); }
```
jblomer:
```suggestion
  bool UseLocalCache() { return !local_cache_dir_.empty(); }
```

cvmfs/catalog_mgr_ro.h
```diff
@@ -14,6 +14,8 @@
 #include <string>
 
 #include "catalog_mgr.h"
+#include "util/logging.h"
+#include "util/posix.h"
```
jblomer: The includes are not required in the header, it seems.
```suggestion
```

cvmfs/catalog_mgr_ro.h
```diff
@@ -66,7 +64,18 @@ class SimpleCatalogManager : public AbstractCatalogManager<Catalog> {
     return (relative_path == "") ? "" : "/" + relative_path;
   }
 
+  bool useLocalCache() { return !local_cache_dir_.empty(); }
+
+  std::string                local_cache_dir_;  // absolute path to local cache
```
jblomer: Consistent naming in this class:
```suggestion
  std::string                dir_cache_;  // absolute path to local cache
```

cvmfs/catalog_mgr_ro.cc
```diff
@@ -14,7 +14,38 @@ using namespace std;  // NOLINT
...
+  if (!dir_cache.empty()) {
+    const bool success = MakeCacheDirectories(local_cache_dir_, 0755);
+
+    if (!success) {
+      LogCvmfs(kLogCatalog, kLogStdout | kLogSyslog,
+              "Failure during creation of local cache directory for server."
+              "Continue, but no local cache will be used.");
+      local_cache_dir_ = "";
+      copy_to_tmp_dir_ = false;
+    }
```
jblomer: I'd rather make this a fatal error. Warnings get overlooked more easily.

cvmfs/catalog_mgr_ro.cc
```diff
@@ -26,11 +57,37 @@ LoadReturn SimpleCatalogManager::GetNewRootCatalogContext(
...
+  if (!fcatalog) {
+    PANIC(kLogStderr, "failed to create temp file when loading %s",
+                      cache_path.c_str());
+  }
+
+  const bool retval = CopyPath2File(cache_path, fcatalog);
+  if (!retval) {
+    PANIC(kLogStderr, "failed to read %s", cache_path.c_str());
+  }
+  (void) fclose(fcatalog);
```
jblomer: Cleanup:
```suggestion
  const bool retval = CopyPath2File(cache_path, fcatalog);
  (void) fclose(fcatalog);
  if (!retval) {
    unlink(tmp_path.c_str());
    PANIC(kLogStderr, "failed to read %s", cache_path.c_str());
  }
```

cvmfs/catalog_mgr_ro.cc
```diff
@@ -39,13 +96,43 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
+        tmp_path = CopyCatalogToTempFile(cache_path);
+        ctlg_context->SetSqlitePath(tmp_path);
+
+        return kLoadNew;
+      }
+    }
+
+    // file not cached yet
+    // open file to download into "cache_dir/" + standard cvmfs file hierarchy
+    fcatalog = fopen(ctlg_context->sqlite_path().c_str(), "w");
```
jblomer: This is not safe under parallelism. Why not downloading the file into a temporary location and then copying it into the cache?

cvmfs/catalog_mgr_ro.cc
```diff
@@ -39,13 +96,43 @@ LoadReturn SimpleCatalogManager::LoadCatalogByHash(
...
-  FILE *fcatalog = CreateTempFile(dir_temp_ + "/catalog", 0666, "w", &tmp);
-  ctlg_context->SetSqlitePath(tmp);
-  if (!fcatalog) {
-    PANIC(kLogStderr, "failed to create temp file when loading %s",
-          url.c_str());
+  FILE *fcatalog;
+
+  if (useLocalCache()) {
+    std::string tmp_path = local_cache_dir_ + "/"
+                           + effective_hash.MakePathWithoutSuffix();
```
jblomer: Why calling a path in the cache directory `tmp_path`? It makes it look like the path is in the tmp directory.
```suggestion
    std::string cache_path = local_cache_dir_ + "/"
                           + effective_hash.MakePathWithoutSuffix();
```

cvmfs/catalog_mgr_ro.cc
jblomer: @HereThereBeDragons can you confirm this problem?

cvmfs/catalog_mgr_rw.cc
```diff
@@ -1253,6 +1286,11 @@ void WritableCatalogManager::CatalogUploadCallback(
   assert(catalog_size > 0);
 
   SyncLock();
+
+  if (useLocalCache()) {
+    CopyCatalogToLocalCache(result);
+  }
```
jblomer: Copying can happen outside the lock.
```suggestion
  if (useLocalCache()) {
    CopyCatalogToLocalCache(result);
  }
  
  SyncLock();
```

cvmfs/catalog_mgr_rw.cc
```diff
@@ -1229,6 +1230,38 @@ void WritableCatalogManager::ScheduleCatalogProcessing(
...
+  std::string tmp_catalog_path;
+  const std::string cache_catalog_path = local_cache_dir_ + "/"
+                                  + result.content_hash.MakePathWithoutSuffix();
+  FILE *fcatalog = CreateTempFile(local_cache_dir_ + "/txn/catalog", 0666,
+                                                      "w", &tmp_catalog_path);
+  if (!fcatalog) {
+    LogCvmfs(kLogCatalog, kLogDebug | kLogStderr,
+                               "Creating file for temporary catalog failed: %s",
+                               tmp_catalog_path.c_str());
+    return false;
```
jblomer: Either make it a fatal error or check the return value. (Also further down)

cvmfs/catalog_mgr_rw.cc
```diff
@@ -1229,6 +1230,38 @@ void WritableCatalogManager::ScheduleCatalogProcessing(
...
+  }
+  CopyPath2File(result.local_path.c_str(), fcatalog);
```
jblomer, line 1253: It may be possible that we can directly rename from `result.local_path`  to `cache_catalog_path`. We should try that first and only if it fails (different file systems or source and destination) make the additional copy.

cvmfs/catalog_diff_tool.h
```diff
@@ -26,19 +26,23 @@ class CatalogDiffTool {
...
         needs_setup_(false) {}
 
-  CatalogDiffTool(const std::string& repo_path, const shash::Any& old_root_hash,
+  CatalogDiffTool(const std::string& repo_path,
+                  const shash::Any& old_root_hash,
                   const shash::Any& new_root_hash,
                   const std::string& temp_dir_prefix,
-                  download::DownloadManager* download_manager)
+                  download::DownloadManager* download_manager,
+                  const std::string& local_cache_dir = "")
```
jblomer:
```suggestion
                  const std::string& cache_dir = "")
```

cvmfs/catalog_diff_tool.h
```diff
@@ -26,19 +26,23 @@ class CatalogDiffTool {
...
       : repo_path_(repo_path),
         old_root_hash_(old_root_hash),
         new_root_hash_(new_root_hash),
         temp_dir_prefix_(temp_dir_prefix),
         download_manager_(download_manager),
+        local_cache_dir_(local_cache_dir),
```
jblomer:
```suggestion
        cache_dir_(cache_dir),
```

cvmfs/catalog_diff_tool.h
```diff
@@ -117,7 +121,8 @@ class CatalogDiffTool {
                                    const std::string& temp_dir,
                                    const shash::Any& root_hash,
                                    download::DownloadManager* download_manager,
-                                   perf::Statistics* stats);
+                                   perf::Statistics* stats,
+                                   const std::string& local_cache_dir);
```
jblomer:
```suggestion
                                   const std::string& cache_dir);
```

cvmfs/catalog_diff_tool.h
```diff
@@ -127,6 +132,7 @@ class CatalogDiffTool {
   std::string temp_dir_prefix_;
 
   download::DownloadManager* download_manager_;
+  const std::string local_cache_dir_;  // path if local caching of catalogs
```
jblomer:
```suggestion
  const std::string cache_dir_;  // path if local caching of catalogs
```

cvmfs/receiver/catalog_merge_tool.h
```diff
@@ -112,6 +118,7 @@ class CatalogMergeTool : public CatalogDiffTool<RoCatalogMgr> {
   bool CreateNewManifest(std::string* new_manifest_path);
 
   std::string repo_path_;
+  const std::string local_cache_dir_;  // path if local cache repo if used
```
jblomer:
```suggestion
  const std::string cache_dir_;  // path if local cache repo if used
```

cvmfs/receiver/commit_processor.cc
jblomer: I think yes, for the moment. It's consistent with the other directory permissions. We may want to revisit them altogether though, not sure if it isn't too open. That's for another PR though.

cvmfs/receiver/commit_processor.cc
```diff
@@ -163,10 +163,16 @@ CommitProcessor::Result CommitProcessor::Process(
            lease_path.c_str(),
            manifest->catalog_hash().ToString(false).c_str());
 
+
+  std::string local_cache_dir_ = "";
```
jblomer:
```suggestion
  std::string local_cache_dir_;
```

cvmfs/receiver/params.cc
```diff
@@ -59,6 +59,13 @@ bool GetParamsFromFile(const std::string& repo_name, Params* params) {
     }
   }
 
+  std::string use_local_cache_str;
+  if (parser.GetValue("CVMFS_SERVER_CATALOG_CACHE", &use_local_cache_str)) {
```
jblomer:
```suggestion
  if (parser.GetValue("CVMFS_SERVER_USE_CATALOG_CACHE", &use_local_cache_str)) {
```

cvmfs/receiver/params.cc
```diff
@@ -59,6 +59,13 @@ bool GetParamsFromFile(const std::string& repo_name, Params* params) {
     }
   }
 
+  std::string use_local_cache_str;
```
jblomer, line 62:
```suggestion
  std::string use_cache;
```

cvmfs/swissknife_sync.h
```diff
@@ -79,6 +80,7 @@ struct SyncParameters {
   std::string tar_file;
   std::string base_directory;
   std::string to_delete;
+  std::string local_cache_dir;
```
jblomer:
```suggestion
  std::string cache_dir;
```

cvmfs/swissknife_sync.h
```diff
@@ -292,6 +295,11 @@ class CommandSync : public Command {
     r.push_back(Parameter::Optional('S',
                                     "virtual directory options "
                                     "[snapshots, remove]"));
+    r.push_back(
+        Parameter::Optional('G', "Use persistent caching for all catalogs "
```
jblomer: For switches, we already have infrastructure code (see next parameter).
```suggestion
        Parameter::Switch('G', "Use persistent caching for all catalogs "
```

cvmfs/swissknife_sync.cc
```diff
@@ -723,6 +736,16 @@ int swissknife::CommandSync::Main(const swissknife::ArgumentList &args) {
...
 
+  if (args.find('G') != args.end()) {
+    // parser just for the IsOn() func
+    const SimpleOptionsParser parser = SimpleOptionsParser(
+                           new DefaultOptionsTemplateManager(params.repo_name));
+    if (parser.IsOn(*args.find('G')->second)) {
+      params.local_cache_dir =
+                       "/var/spool/cvmfs/" + params.repo_name + "/cache.server";
+    }
+  }
```
jblomer: Not necessary to parse an argument, use `-G` as a switch.

cvmfs/server/cvmfs_server_publish.sh
```diff
@@ -201,6 +201,7 @@ cvmfs_server_publish() {
         -C $trusted_certs                              \
         -N $name                                       \
         -K $CVMFS_PUBLIC_KEY                           \
+        -G $CVMFS_SERVER_CATALOG_CACHE                 \
```
jblomer: Use as a switch: if `$CVMFS_SERVER_CATALOG_CACHE` is set, add the `-G` switch.

test/src/630-publish_with_local_cache/main
```diff
@@ -0,0 +1,108 @@
...
+  # new transaction to have some cached catalogs
+  start_transaction "$CVMFS_TEST_REPO" || return $?
+  mkdir -p /cvmfs/"$CVMFS_TEST_REPO"/foo/bar
+  touch /cvmfs/"$CVMFS_TEST_REPO"/foo/.cvmfscatalog
+  touch /cvmfs/"$CVMFS_TEST_REPO"/foo/bar/.cvmfscatalog
+  publish_repo "$CVMFS_TEST_REPO" || return 2
+
+  # mount repo to extract hash
+  echo "*** Mount repo to extract catalog hash of /cvmfs/$CVMFS_TEST_REPO/foo/bar/.cvmfscatalog"
+  sudo cvmfs_server mount $CVMFS_TEST_REPO
```
jblomer: This should not be necessary (already mounted).
```suggestion
```

test/src/630-publish_with_local_cache/main
```diff
@@ -0,0 +1,108 @@
...
+  echo "   Umount"
+
+
+  # delete nested catalog from backend --> caching must work otherwise it fails
+  echo "*** Delete /cvmfs/$CVMFS_TEST_REPO/foo/bar/.cvmfscatalog"
+  local retval
+  ### add C flag at the end to hash to find proper file
+  nested_ctlg_hash="${nested_ctlg_hash}C"
+  retval=$(peek_backend "$CVMFS_TEST_REPO" "$nested_ctlg_hash")
+  [[ $retval =~ "available" ]] || return 3
```
jblomer: Double-breackets are a bashism. You can check the return value of the command, which should anyway be more robust. Also further down.

test/src/630-publish_with_local_cache/main
```diff
@@ -0,0 +1,108 @@
...
+  # delete_from_backend "$CVMFS_TEST_REPO" "$nested_ctlg_hash" || return 4 <--- this does NOT work!
```
jblomer: What exactly does not work? Shouldn't this be fixed before merging? Or removed from the PR?

test/src/630-publish_with_local_cache/main
```diff
@@ -0,0 +1,108 @@
...
+  # Test the reverse
+  echo "*** NOW TEST THE REVERSE: fail if no caching is available"
+  echo "*** Remove local caching for publishing from config."
+  sudo sed -i '/CVMFS_SERVER_CATALOG_CACHE/d' "$repo_server_config"
+  cat $repo_server_config
+
+
+  # mount repo to extract hash
+  echo "*** Mount repo to extract catalog hash of /cvmfs/$CVMFS_TEST_REPO/foo/bar/.cvmfscatalog"
+  sudo cvmfs_server mount $CVMFS_TEST_REPO
```
jblomer:
```suggestion
```

test/src/630-publish_with_local_cache/main
```diff
@@ -0,0 +1,108 @@
...
+  echo "*** Delete /cvmfs/$CVMFS_TEST_REPO/foo/bar/.cvmfscatalog"
+  local retval
+  ### add C flag at the end to hash to find proper file
+  nested_ctlg_hash="${nested_ctlg_hash}C"
+  retval=$(peek_backend "$CVMFS_TEST_REPO" "$nested_ctlg_hash")
+  [[ $retval =~ "available" ]] || return 8
+  # delete_from_backend "$CVMFS_TEST_REPO" "$nested_ctlg_hash" || return 4 <--- this does NOT work!
+  echo "    Check if file was really removed"
+  # retval=$(peek_backend "$CVMFS_TEST_REPO" "$nested_ctlg_hash")
+  # echo $retval
```
jblomer: See above

jblomer (changes requested):
Looks generally good to me, some smaller comments.

test/src/630-publish_with_local_cache/main
HereThereBeDragons: `delete_from_backend "$CVMFS_TEST_REPO" "$nested_ctlg_hash"`
this does not work for catalogs while `peek_backend "$CVMFS_TEST_REPO" "$nested_ctlg_hash"` works... which took me quite a while to understand.

i tried different things but `delete_from_backend` just does not take catalogs

you have to manually delete it `sudo rm -f /srv/cvmfs/$CVMFS_TEST_REPO/$(make_path $nested_ctlg_hash)`

mharvey-jt:
One further optimisation - the catalogs accepted and forwarded on by the payload processor could also be added to the cache

cvmfs/swissknife_sync.cc
HereThereBeDragons: is there a particular reason why you would like to change the current implementation?

with the current implementation you can say `CVMFS_SERVER_USE_CATALOG_CACHE=off` or `CVMFS_SERVER_USE_CATALOG_CACHE=on` in the config file and it works.

if i understand  `cvmfs_server_publish.sh` correctly, for flags/switches it only looks if `CVMFS_*` is defined, or at most something like `f [ "x$CVMFS_USE_FILE_CHUNKING" = "xtrue" ]; then`.  it does not check for all the different other values that can be used by the client config and which the options parser understands: 
```
bool OptionsManager::IsOn(const std::string &param_value) const {
  const string uppercase = ToUpper(param_value);
  return ((uppercase == "YES") || (uppercase == "ON") || (uppercase == "1") ||
          (uppercase == "TRUE"));
}
```

cvmfs/catalog_mgr_rw.cc
HereThereBeDragons: this should not be a fatal error.
yes it is annoying if the caching does not work, but it should not fail the overall publication process.
i suggest just adding a debug/stderr message

test/src/630-publish_with_local_cache/main
jblomer: I this case, `delete_from_backend` needs to be fixed. Otherwise the test does not work in the S3 configuration.

cvmfs/catalog_mgr_rw.cc
HereThereBeDragons: is this safe?
`const upload::SpoolerResult &result` is not used/reused somewhere else afterwards?
for sure some `ifs` will be added to prevent `unlink(result.local_path)` in other code sections...

cvmfs/catalog_mgr_ro.cc
HereThereBeDragons: i had been thinking about this but in the end i decided to go with the following:
`tmp_path` is the path to the new catalog (for writeable catalog it is a copy to a temproary file that is editable)
`cache_path` is the path to the catalog in the cache directory. they are not allowed to edit that file.

we can rename it if you have better ideas. but throughout the method this should be consistent

cvmfs/catalog_mgr_ro.h
HereThereBeDragons: i would prefer either `local_cache_dir_` or `cache_dir_`

cvmfs/catalog_mgr_ro.cc
mharvey-jt: Just encountered exactly this problem, when populating the cache from the payload processor.

HereThereBeDragons: ok was a bit larger rewrite

cvmfs/catalog_mgr_ro.cc
HereThereBeDragons: update: after the rewrite for comment above, this should be more clear now

cvmfs/catalog_diff_tool.h
HereThereBeDragons: see above, lets find a uniform solution for this naming

test/src/630-publish_with_local_cache/main
HereThereBeDragons: `#!/bin/bash` ? i guess i miss what is wrong with double brackets?
i change it to return codes.. but i really wish swissknife would be better documented... and that each command gets its own file. why is command `peek` in `swissknife_sync.cc`?

cvmfs/catalog_mgr_rw.cc
HereThereBeDragons: --> make PANIC

cvmfs/catalog_mgr_rw.cc
HereThereBeDragons: --> nevermind

cvmfs/swissknife_sync.cc
HereThereBeDragons: --> use Switch and make extra Issue to add on/off/true/false/... option parsing in bash

test/src/630-publish_with_local_cache/main
HereThereBeDragons: --> `delete_from_backend` must be fixed before this merge

jblomer:
> One further optimisation - the catalogs accepted and forwarded on by the payload processor could also be added to the cache

@mharvey-jt Given the early return optimizations in the receiver, would there actually be a cache reuse for the nested catalogs from the payloads?

mharvey-jt:
> @mharvey-jt Given the early return optimizations in the receiver, would there actually be a cache reuse for the nested catalogs from the payloads?

In fact, all that needs to enter the cache are the catalogs from the root to the top of the commit. 

We've a patch in the gateway that does this just before entering the commit critical section. (We aren't using the payload processor right now - the granting clients push catalogs directly to the object store, bypassing the gateway)

cvmfs/catalog_mgr_ro.h
HereThereBeDragons: renamed it to `dir_cache_` to keep file consistency

cvmfs/catalog_diff_tool.h
HereThereBeDragons: `cache_dir` for all related to the diff/merge tools

test/src/630-publish_with_local_cache/main
HereThereBeDragons: `delete_from_backend` works correct but not the same as `peek_backend`. a new function is introduced `delete_hash_from_backend` that does the same as `peek_backend`

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot unittest

HereThereBeDragons:
rebuild of unittest: https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsUnittest/2173/ all good now

HereThereBeDragons:
@cernvm-bot cloudtest

jblomer (approved):
Very nice! Thank you!

vvolkl (approved):

PR #3438, opened 2023-11-08
Also support Yubikey devices that are named `YubiKey` instead of `Yubikey`

### LINKED ISSUE #3437: YubiKey 5 Nano not recognized by `cvmfs_server masterkeycard`
I'm trying to use a YubiKey 5 Nano with `cvmfs_server masterkeycard`, but this doesn't seem to work:
```
$ cvmfs_server masterkeycard -a
opensc-tool -l has no device matching "Yubikey.*CCID"
```

Looking at the output of `opensc-tool -l`, it's actually due to the uppercase `K` in `YubiKey` for my device:
```
$ opensc-tool -l
# Detected readers (pcsc)
Nr.  Card  Features  Name
0    No              Broadcom Corp 58200 [Contacted SmartCard] (0123456789ABCD) 00 00
1    Yes             Yubico YubiKey OTP+FIDO+CCID 01 00
```

Perhaps this was changed for newer YubiKey devices? Would it be possible to change the pattern that's being used in `cvmfs_server` to something like `"Yubi[kK]ey.*CCID"`? That seems to fix the issue for me:
```
$ ./cvmfs_server masterkeycard -a
masterkeycard is available
```

bedroge:
Fixes #3437.

Edit: I just found that there's a yubikey-related test as well in https://github.com/cvmfs/cvmfs/blob/devel/test/src/643-masterkeycard/main, but it looks like that doesn't need to be changed: for some reason, `lsusb` does name my Yubikey `Yubico.com Yubikey 4/5 OTP+U2F+CCID`, while `opensc-tool -l` shows `Yubico YubiKey OTP+FIDO+CCID 01 00`.

### DIFF
--- a/cvmfs/server/cvmfs_server_masterkeycard.sh
+++ b/cvmfs/server/cvmfs_server_masterkeycard.sh
@@ -15,7 +15,7 @@
 # If not, the reason is sent to stdout and false is returned,
 # otherwise there's nothing to stdout and true is returned
 masterkeycard_available() {
-  local pattern="Yubikey.*CCID"
+  local pattern="Yubi[kK]ey.*CCID"
   local reason=""
   if ! lsusb | grep -q "$pattern"; then
     reason="USB device matching \"$pattern\" not present"


vvolkl:
Thanks for the PR!

vvolkl:
@cernvm-bot cloudtest

Issue #3441, opened 2023-11-14
CVMFS client for RISC-V


julianmorillo:
Hello, 
I can see in [Getting Started](https://cvmfs.readthedocs.io/en/stable/cpt-quickstart.html) that there is experimental support for RISC-V architecture. 
Are there any prebuilt CVMFS client package for RISC-V available somewhere? Or a way/instructions to build it from source?
Thank you,
Julián

vvolkl:
No binary packages at the moment, unfortunately. I'd encourage you to build from source (see https://cvmfs.readthedocs.io/en/stable/cpt-quickstart.html#building-from-source) and report any problems you might encounter. You need some basic build-dependencies, for RHEL9 this should be exhaustive (see https://gitlab.cern.ch/cernvm/build-images/-/tree/master/docker?ref_type=heads for other distros):
```
        dnf -y install                     \
                       clang-analyzer          \
                       clang-tools-extra       \
                       cmake                   \
                       curl-devel              \
                       fuse-devel              \
                       fuse3-devel             \
                       gcc-c++                 \
                       gdb                     \
                       git                     \
                       golang                  \
                       hardlink                \
                       libattr-devel           \
                       libcap-devel            \
                       libffi-devel            \
                       libuuid-devel           \
                       make                    \
                       openssl-devel           \
                       python3-devel           \
                       rpm-build               \
                       ruby-devel              \
                       selinux-policy-devel    \
                       selinux-policy-targeted \
                       systemd                 \
                       which                   \
                       valgrind-devel          \
                       voms-devel              \
                       zlib-devel
```
`attr fuse3 autofs httpd` are also typically needed as runtime dependencies.

julianmorillo:
Trying to build cvmfs from source in a RISC-V Fedora 37 system, I found out that the config.guess included in libressl-3.5.3 (as distributed with cvmfs) fails to detect the build system type.
Downloading libressl-3.8.2.tar.gz into "externals/libcrypto" and changing both SSL_VERSION and CRYPTO_VERSION in **bootstrap.sh** file:
```
SSL_VERSION=3.8.2
CRYPTO_VERSION=3.8.2
```
worked for me, so probably it would be nice to update the libressl version distributed with cvmfs.

julianmorillo:
I had the same problem with **protobuf-2.6.1** (as distributed with cvmfs). In this case, the solution was: 
- download protobuf-21.10-all.tar.gz into "**externals/protobuf**"
[...]

PR #3446, opened 2023-11-20
autoreconf needed for RISC-V build

julianmorillo:
Updating the config.guess files as distributed within libcrypto and protobuf dependencies is needed in order to recognize the system build type when building in a RISCV Fedora 37 system (see Issue #3441 ).

### DIFF
--- a/externals/libcrypto/src/configureHook.sh
+++ b/externals/libcrypto/src/configureHook.sh
@@ -9,6 +9,7 @@ if [ "$IS_64_BIT" = "FALSE" ]; then
   DISABLE_ASM="--disable-asm"
 fi
 
+autoreconf -vfi
 mkdir build && cd build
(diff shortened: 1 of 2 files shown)


vvolkl:
Thanks! Like this there should be no problem at all. I'd merge this but would prefer it the `autoreconf` call is optional - for most platforms we don't need it. Maybe you can add a cmake option or an if that checks /proc/cpuinfo for RISC-V?

julianmorillo:
I have just committed and pushed a version following your suggestion: now the autoreconf will only be executed for RISC-V architectures.

vvolkl:
@cernvm-bot cloudtest

boegel:
Seems like this fix didn't get included yet in the CernVM-FS 2.11.5 release?

vvolkl:
No, but it'll be in the upcoming 2.12.

PR #2772, opened 2021-08-17
Add "noautofs" option to cvmfs_config setup

aandvalenzuela:

PR #3450, opened 2023-11-28
mention `noautofs` option in usage of `cvmfs_config`

boegel:
The `noautofs` option of `cvmfs_config setup`, which is a combination of `nocfgmod` and `nostart`, was implemented in #2772, but the `cvmfs_config` help output doesn't mention it (and neither does the CernVM-FS documentation BTW).


vvolkl:
Not only are those options not documented, I just realized that `nouser` is actually broken, so thanks for pointing this out! I'll update the documentation.

PR #3452, opened 2023-11-29
Fix nouser option for cvmfs_config

### LINKED ISSUE #3451: `cvmfs config setup nouser` option is broken
`cvmfs_config` anyway tries to make the cvmfs user owner  of /var/lib/cvmfs, resulting in the error:

```
# cvmfs_config setup nouser
chown: invalid user: ‘cvmfs:cvmfs’
```

vvolkl:
Fixes #3451

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -279,13 +279,11 @@ cvmfs_setup() {
     fi
   fi
 
-  if [ $nouser -eq 0 ]; then
-    # if group fuse exists, add user cvmfs
-    if check_group "fuse"; then
-      if ! add_user_to_group_fuse; then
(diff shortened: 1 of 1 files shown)


vvolkl:
@cernvm-bot cloudtest

jblomer (approved):

cvmfs/cvmfs_config
```diff
@@ -279,11 +279,13 @@ cvmfs_setup() {
     fi
   fi
 
-  # if group fuse exists, add user cvmfs
```
vvolkl: In order to stay consistent with previous behavior, I'll revert the change in this block, so if a cvmfs user exists already it is still added to the fuse group. There is no error here if no user exists.

PR #3475, opened 2024-01-10
Avoid possible race in mount helper

### LINKED ISSUE #3471: mount.cvmfs intermittent hang
`ended` needs to be checked in the while() condition
https://github.com/cvmfs/cvmfs/blob/4a477cc8eae9ffe9f6570d027b312781c7fb8001/mount/mount.cvmfs.cc#L698
without this there can be a race with the output of the child cvmfs2 which leaves mount.cvmfs stuc k in the while() loop.

vvolkl:
Fixes #3471

### DIFF
--- a/mount/mount.cvmfs.cc
+++ b/mount/mount.cvmfs.cc
@@ -694,6 +694,12 @@ int main(int argc, char **argv) {
           return 32;
       }
     }
+
+    // TODO(vvolkl): This block has seen several rounds of iterations,
+    // that tried to adress races and incorrect return codes.
+    // Should be looked at once again to make sure the cause of the
+    // race is understood, and if the timeout is really needed.
+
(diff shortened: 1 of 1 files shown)


vvolkl:
@cernvm-bot cloudtest

jblomer:
Is it understood where exactly the race comes from?

jblomer (approved):
It's possible that the race is connected to the timeout in the select call. I would actually think that we don't need the timeout at all. I can't see that the change would change anything for the worse.

Perhaps let's add a comment to this code block. It is apparently tricky and has kept us busy for a while.

vvolkl:
Agreed. Added a comment to revisit this issue in the future. I think there's also some potential to refactor and break this block up a bit.

PR #3476, opened 2024-01-11
Reduce write lock contention in catalog mgr

jblomer:
First draft at improved lock management in the catalog manager.

This patch stages a nested catalog in the cache while the cache manager holds the read lock. The following write-locked phase should then be quick since the catalog is attached from the local cache.

This patch should allow a first performance evaluation. There are the following caveats:
- Staging the catalog inspects the nested catalogs (again). This lookup should somehow be combined with the one from `MountSubtree`.
- There is a race between giving up on the read lock and taking the write lock, in which the file system revision (root catalog) may change. If this change also affects the nested catalog at hand, we staged an outdated nested catalog. That's not ideal but not fatal either and it should be rare.
- The cache may cleanup the staged nested catalog before it is attached. Again, not ideal but also not fatal and should be rare.
- The `MountSubtree()` method allows for "deep mounts" of catalog hierarchies. Only the first nested catalog is staged. In the context of the fuse module, however, that should not matter due to POSIX enforced incremental path loopups.

@mharvey-jt fyi.

### DIFF
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -334,6 +334,19 @@ class AbstractCatalogManager : public SingleCopy {
   const std::vector<CatalogT*>& GetCatalogs() const { return catalogs_; }
 
   /**
+   * Opportunistic optimization: the client catalog manager uses this method
+   * to preload into the cache a nested catalog that is likely to be required
+   * next. Likely, because there is a race with the root catalog reload which
+   * may result in the wrong catalog being staged. That's not a fault though,
+   * the correct catalog will still be loaded with the write lock held.
+   */
+  virtual void StageNestedCatalogByHash(const shash::Any & /*hash*/,
+                                        const PathString & /*mountpoint*/)
+  { }
+  void StageNestedCatalog(const PathString &path, const CatalogT *parent,
+                          bool is_listable);
+
+  /**
    * Create a new Catalog object.
    * Every derived class has to implement this and return a newly
    * created (derived) Catalog structure of it's desired type.
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -344,8 +344,13 @@ class AbstractCatalogManager : public SingleCopy {
   virtual void StageNestedCatalogByHash(const shash::Any & /*hash*/,
                                         const PathString & /*mountpoint*/)
   { }
-  void StageNestedCatalog(const PathString &path, const CatalogT *parent,
-                          bool is_listable);
+  /**
+   * Called within the ReadLock(), which will be released before downloading
+   * the catalog (and before leaving the method)
+   */
+  void StageNestedCatalogAndUnlock(const PathString &path,
+                                   const CatalogT *parent,
+                                   bool is_listable);
 
   /**
    * Create a new Catalog object.
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -23,8 +23,10 @@
 #include "file_chunk.h"
 #include "manifest_fetch.h"
 #include "statistics.h"
+#include "util/algorithm.h"
 #include "util/atomic.h"
 #include "util/logging.h"
+#include "util/platform.h"
 
 class XattrList;
 namespace catalog {
@@ -178,6 +180,9 @@ struct Statistics {
   perf::Counter *n_listing;
   perf::Counter *n_nested_listing;
   perf::Counter *n_detach_siblings;
+  perf::Counter *n_write_lock;
+  perf::Counter *ns_write_lock;
+
   perf::Counter *catalog_revision;
 
   explicit Statistics(perf::Statistics *statistics) {
@@ -196,6 +201,10 @@ struct Statistics {
         "Number of listings of nested catalogs");
     n_detach_siblings = statistics->Register("catalog_mgr.n_detach_siblings",
         "Number of times the CVMFS_CATALOG_WATERMARK was hit");
+    n_write_lock = statistics->Register("catalog_mgr.n_write_lock",
+                                        "number of write lock calls");
+    ns_write_lock = statistics->Register("catalog_mgr.ns_write_lock",
+        "time spent in WriteLock() [ns]");
     catalog_revision = statistics->Register("catalog_revision",
                                     "Revision number of the root file catalog");
   }
@@ -392,8 +401,12 @@ class AbstractCatalogManager : public SingleCopy {
     assert(retval == 0);
   }
   inline void WriteLock() const {
+    uint64_t timestamp = platform_monotonic_time_ns();
     int retval = pthread_rwlock_wrlock(rwlock_);
     assert(retval == 0);
+    perf::Inc(statistics_.n_write_lock);
+    uint64_t duration = platform_monotonic_time_ns() - timestamp;
+    perf::Xadd(statistics_.ns_write_lock, duration);
   }
   inline void Unlock() const {
     int retval = pthread_rwlock_unlock(rwlock_);
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -339,6 +339,7 @@ class AbstractCatalogManager : public SingleCopy {
    * next. Likely, because there is a race with the root catalog reload which
    * may result in the wrong catalog being staged. That's not a fault though,
    * the correct catalog will still be loaded with the write lock held.
+   * Note that this method is never used for root catalogs.
    */
   virtual void StageNestedCatalogByHash(const shash::Any & /*hash*/,
                                         const PathString & /*mountpoint*/)
--- a/cvmfs/catalog_mgr_client.cc
+++ b/cvmfs/catalog_mgr_client.cc
@@ -267,6 +267,16 @@ LoadReturn ClientCatalogManager::GetNewRootCatalogContext(
   return success_code;
 }
 
+std::string ClientCatalogManager::GetCatalogDescription(
+  const PathString &mountpoint, const shash::Any &hash)
+{
+  return "file catalog at " + repo_name_ + ":" +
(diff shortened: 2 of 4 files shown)


cvmfs/catalog_mgr.h
```diff
@@ -333,6 +333,19 @@ class AbstractCatalogManager : public SingleCopy {
   virtual void ActivateCatalog(CatalogT *catalog) { }
   const std::vector<CatalogT*>& GetCatalogs() const { return catalogs_; }
 
+  /**
+   * Opportunistic optimization: the client catalog manager uses this method
+   * to preload into the cache a nested catalog that is likely to be required
+   * next. Likely, because there is a race with the root catalog reload which
+   * may result in the wrong catalog being staged. That's not a fault though,
+   * the correct catalog will still be loaded with the write lock held.
+   */
```
HereThereBeDragons, line 352: maybe add a comment that this really should only be used for nested catalogs - the output log message does not do the proper replacement for root catalogs

cvmfs/catalog_mgr_client.cc
```diff
@@ -365,6 +365,22 @@ LoadReturn ClientCatalogManager::FetchCatalogByHash(
...
 
+void ClientCatalogManager::StageNestedCatalogByHash(
+  const shash::Any &hash,
+  const PathString &mountpoint)
+{
+  string catalog_descr = "file catalog at " + repo_name_ + ":" +
+    string(mountpoint.GetChars(), mountpoint.GetLength()) +
+    " (" + hash.ToString() + ")";
+
+  assert(hash.suffix == shash::kSuffixCatalog);
```
HereThereBeDragons, line 378: should this assert be at the very top of the func?

cvmfs/catalog_mgr_client.cc
```diff
@@ -365,6 +365,22 @@ LoadReturn ClientCatalogManager::FetchCatalogByHash(
...
+  CacheManager::Label label;
+  label.path = catalog_descr;
+  label.flags = CacheManager::kLabelCatalog;
+  int fd = fetcher_->Fetch(CacheManager::LabeledObject(hash, label));
+  if (fd >= 0)
+    fetcher_->cache_mgr()->Close(fd);
```
HereThereBeDragons, line 385: i would at some { } around the if

cvmfs/catalog_mgr_impl.h
```diff
@@ -841,6 +844,43 @@ bool AbstractCatalogManager<CatalogT>::IsAttached(const PathString &root_path,
 }
 
 
+template <class CatalogT>
+void AbstractCatalogManager<CatalogT>::StageNestedCatalog(
```
HereThereBeDragons: i was wondering if we could combine it with `findcatalog` or `mountsubtree` but i think this is not possible.  especially the `if (MountSubtree(path, best_fit, true /* is_listable */, NULL)) {` seems interesting but i guess you get too muhc latency in case the `mountsubtree` should fail

cvmfs/catalog_mgr_impl.h
```diff
@@ -841,6 +844,43 @@ bool AbstractCatalogManager<CatalogT>::IsAttached(const PathString &root_path,
...
+
+  perf::Inc(statistics_.n_nested_listing);
+  typedef typename CatalogT::NestedCatalogList NestedCatalogList;
+  const NestedCatalogList& nested_catalogs = parent->ListNestedCatalogs();
+
+  for (typename NestedCatalogList::const_iterator i = nested_catalogs.begin(),
+       iEnd = nested_catalogs.end(); i != iEnd; ++i)
+  {
+    if (!path.StartsWith(i->mountpoint))
+      continue;
```
HereThereBeDragons, line 861: { }

cvmfs/catalog_mgr_impl.h
```diff
@@ -841,6 +844,43 @@ bool AbstractCatalogManager<CatalogT>::IsAttached(const PathString &root_path,
...
+
+    // in this case the path doesn't start with
+    // the mountpoint in a file path sense
+    // (e.g. path is /a/bc and mountpoint is /a/b), and will be ignored
+    const unsigned mountpoint_len = i->mountpoint.GetLength();
+    if (path_len > mountpoint_len && path.GetChars()[mountpoint_len] != '/')
+      continue;
```
HereThereBeDragons, line 868: { }

cvmfs/catalog_mgr_impl.h
```diff
@@ -841,6 +844,43 @@ bool AbstractCatalogManager<CatalogT>::IsAttached(const PathString &root_path,
...
+
+    // Found a nested catalog transition point
+    if (!is_listable && (path_len == mountpoint_len))
+      break;
```
HereThereBeDragons, line 872: { }

HereThereBeDragons (commented):
think this is a nice idea to try out.
one question i have is if this should also be applied to `lookupcounters`?

mharvey-jt:
We see a substantial improvement with this.

cvmfs/catalog_mgr_impl.h
```diff
@@ -284,6 +284,7 @@ bool AbstractCatalogManager<CatalogT>::LookupPath(const PathString &path,
   if (!found && MountSubtree(path, best_fit, false /* is_listable */, NULL)) {
     LogCvmfs(kLogCatalog, kLogDebug, "looking up '%s' in a nested catalog",
              path.c_str());
+    StageNestedCatalog(path, best_fit, false /* is_listable */);
```
mharvey-jt: The performance is even better if the read lock is released before the catalog stage: 
```
Unlock();
StageNestedCatalog(path, best_fit, false /* is_listable */);
WriteLock();
```

jblomer: I think that would make it incorrect though, because in this case `best_fit` may become invalid in the meantime.

mharvey-jt: In that case, how about the Unlock happens down at  line 879,  after the last reference to `best_fit` and immediately before the  call to `StageNestedCatalogByHash(hash, mountpoint)`, copying the args; and with a check to unconditionally Unlock() if no staging is done.

jblomer:
At the moment, I can only think of ugly ways to reuse the lookup in the nested catalogs. Ready for review.

@HereThereBeDragons your comments are addressed, except the stylistic ones (for those, cpplint defines the baseline).

jblomer:
@cernvm-bot cloudtest

cvmfs/catalog_mgr_impl.h
HereThereBeDragons: but do we really care if it becomes invalid? we are just staging it. yes maybe its a "dead" object that trashes the cache but as long as the quota manager is aware of the object it shouldnt be a problem? the staging is on "best effort"  basis anyway?

mharvey-jt:
@jblomer  see PR https://github.com/cvmfs/cvmfs/pull/3490 as mentioned in today's meeting

cvmfs/catalog_mgr.h
```diff
@@ -373,8 +401,12 @@ class AbstractCatalogManager : public SingleCopy {
     assert(retval == 0);
   }
   inline void WriteLock() const {
+    uint64_t timestamp = platform_monotonic_time_ns();
```
HereThereBeDragons, line 404: i guess we want to keep those timers?

cvmfs/catalog_mgr_impl.h
```diff
@@ -841,6 +841,45 @@ bool AbstractCatalogManager<CatalogT>::IsAttached(const PathString &root_path,
 }
 
 
+template <class CatalogT>
+void AbstractCatalogManager<CatalogT>::StageNestedCatalogAndUnlock(
```
HereThereBeDragons, line 845: i find this function a bit confusing so i tried to rewrite it to understand it better.
maybe you want to use this (just needs a double check that i didnt do any mistake)

```c++
 for (typename NestedCatalogList::const_iterator i = nested_catalogs.begin(),
       iEnd = nested_catalogs.end(); i != iEnd; ++i)
  {
    // only stage the mountpoint that is a correct (partial) match
    // of the requested path (e.g. if path is /a/bc then /a/b is not a match).
    if (path.StartsWith(i->mountpoint) &&
          (path_len == i->mountpoint.GetLength()
          || (path_len > i->mountpoint.GetLength()
              && path.GetChars()[mountpoint_len] == '/'))) {

        // No need to stage, parent with the transaction point is sufficient
        if (!is_listable && (path_len == mountpoint_len)) {
          break;
        }

        Unlock();
        LogCvmfs(kLogCatalog, kLogDebug, "staging nested catalog at %s (%s)",
             i->mountpoint.c_str(), i->hash.ToString().c_str());
        StageNestedCatalogByHash(i->hash, i->mountpoint);
        return;
      }
    }
  Unlock();
}  // just here for readability <<this is the end bracket of the function>>
```

HereThereBeDragons (approved):
have a look at the 2 comments, otherwise its fine for me.

cvmfs/catalog_mgr.h
jblomer: With `n_write_lock` and `ns_write_lock` we get the average time spent in `WriteLock()`. If we need the distribution, we can use the [Log2Histogram](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/util/algorithm.h#L141)

vvolkl (approved):

mharvey-jt (approved):
This looks good now the read lock is also released before the staging  - we have this rolling to prod now and see a substantial speedup on a particularly catalog-load intensive workload.

cvmfs/catalog_mgr.h
mharvey-jt: Right now we are ok with a simple arithmetic mean.

jblomer:
@cernvm-bot cloudtest

PR #3478, opened 2024-01-11
Consistently use 64bit ints for catalog revision

### LINKED ISSUE #3469: treat repository revision as a uint64
the repository revision is processed in various places as an int32. On continuously updated repositories, it's not implausible that this might overflow. Ideally, this would be treated as a (u)int64.

jblomer:
Fixes #3469

### DIFF
--- a/cvmfs/catalog_traversal.h
+++ b/cvmfs/catalog_traversal.h
@@ -45,7 +45,7 @@ struct CatalogTraversalData {
                        const shash::Any   &catalog_hash,
                        const unsigned      tree_level,
                        const size_t        file_size,
-                       const unsigned int  history_depth)
+                       const uint64_t      history_depth)
   : catalog(catalog)
   , catalog_hash(catalog_hash)
   , tree_level(tree_level)
@@ -56,7 +56,7 @@ struct CatalogTraversalData {
   const shash::Any    catalog_hash;
   const unsigned int  tree_level;
   const size_t        file_size;
-  const unsigned int  history_depth;
+  const uint64_t      history_depth;
 };
 
 
@@ -142,13 +142,13 @@ class CatalogTraversalBase
       , num_threads(8)
       , serialize_callbacks(true) {}
 
-    static const unsigned int kFullHistory;
-    static const unsigned int kNoHistory;
+    static const uint64_t     kFullHistory;
+    static const uint64_t     kNoHistory;
     static const time_t       kNoTimestampThreshold;
 
     ObjectFetcherT *object_fetcher;
 
-    unsigned int    history;
+    uint64_t        history;
     time_t          timestamp;
     bool            no_repeat_history;
     bool            no_close;
@@ -275,7 +275,7 @@ class CatalogTraversalBase
     CatalogJob(const std::string  &path,
                const shash::Any   &hash,
                const unsigned      tree_level,
-               const unsigned      history_depth,
+               const uint64_t      history_depth,
                      CatalogTN    *parent = NULL) :
       path(path),
       hash(hash),
@@ -299,7 +299,7 @@ class CatalogTraversalBase
     const std::string   path;
     const shash::Any    hash;
     const unsigned      tree_level;
-    const unsigned      history_depth;
+    const uint64_t      history_depth;
           CatalogTN    *parent;
 
     // dynamic processing state (used internally)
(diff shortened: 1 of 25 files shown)


cvmfs/garbage_collection/garbage_collector_impl.h
```diff
@@ -93,7 +93,7 @@ void GarbageCollector<CatalogTraversalT, HashFilterT>::PreserveDataObjects(
     if (configuration_.verbose) {
       const int    rev   = data.catalog->revision();
```
vvolkl:
```suggestion
      const uint64_t rev = data.catalog->revision();
```

cvmfs/swissknife_notify.cc
```diff
@@ -81,7 +81,7 @@ int CommandNotify::Main(const ArgumentList& args) {
   } else {  // subscribe
     std::string topic = *args.find('t')->second;
     bool continuous = args.count('c') > 0;
-    int revision = GetMinRevision(args);
+    uint64_t revision = GetMinRevision(args);
```
vvolkl:
```suggestion
    const uint64_t revision = GetMinRevision(args);
```
clang tidy complains: 
```
/home/sftnight/jenkins/workspace/CvmfsTidy/label/docker-x86_64/cvmfs/cvmfs/swissknife_notify.cc:84:5: error: variable 'revision' of type 'uint64_t' (aka 'unsigned long') can be declared 'const' [misc-const-correctness,-warnings-as-errors]
    uint64_t revision = GetMinRevision(args);
    ^
             const 
```

cvmfs/catalog_traversal.h
```diff
@@ -299,15 +299,15 @@ class CatalogTraversalBase
...
+    const uint64_t      history_depth;
           CatalogTN    *parent;
 
     // dynamic processing state (used internally)
     std::string   catalog_file_path;
     size_t        catalog_file_size;
     bool          ignore;
     CatalogTN    *catalog;
-    unsigned int  referenced_catalogs;
+    uint64_t      referenced_catalogs;
```
HereThereBeDragons, line 310: if you make this `uint64_t` should then also `tree_level` be `uint64_t`?

HereThereBeDragons (changes requested):
i did a search for "revision" myself and found the following that might be missing:


--> `StringifyUint`
https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/swissknife_history.cc#L776
https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/swissknife_history.cc#L806
https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/swissknife_info.cc#L230
https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/talk.cc#L357


https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/swissknife_history.cc#L844-L846


--> probably want to use `%" PRIu64 "`
https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/swissknife_history.cc#L856-L860



other aspects:
1)
`perf::Counter::Set()` takes `int64_t` and does not support `uint64_t`
there might have been some static casting being lost:
https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/swissknife_ingest.cc#L223

2)
more correct comparison would be `1ul`
https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/test/unittests/t_catalog_mgr.cc#L157

3)
it would probably help if it was clearer to distinguish in `history.h/cc`  : 
- `initial_revision` being a number
- and `SetPreviousRevision(hash)` wants the hash (if that is changed then it should also be changed in all the catalogs to keep it consistent)

cvmfs/catalog_traversal.h
jblomer: That should be independent, `tree_level` specifies the "height" of the tree in terms of nested catalogs. The revision represents the timeline of root catalogs.

jblomer:
Thanks for the cross-check! In this PR, I ignore possible issues to to signedness. It's just about 32/64bit.

HereThereBeDragons (approved):
ok for me. but  for consistency it might be nice to have a a follow up PR to fix the uint/int discrepancy

vvolkl (approved):

vvolkl (approved):

PR #3489, opened 2024-01-24
Add compiler checks to LogCvmfs() and fix formatting

rezan:
It looks like `LogCvmfs()` is a printf like format wrapper, but that is lost to the compiler. This adds an attribute to the logger declaration (`cvmfs/util/logging.h`) which tells the compiler to do format checking on the format string and parameter list:

https://github.com/rezan/cvmfs/blob/4ecbb5f339fc7436f0bf12287436f8973f4d082c/cvmfs/util/logging.h#L21

When I added this, it exposed close to 45 formatting warnings. Most of them were harmless field width mismatches, but there were a handful of completely broken format strings. I kept the specifiers as simple as possible given there were so many formatting warnings to get thru.

I also added a flag to ignore empty format strings (`-Wno-format-zero-length`). It looks like there are valid uses of this which the compiler doesn't like:

https://github.com/cvmfs/cvmfs/blob/d9962e3ec093bb7edd282ef4d554855e76ee152f/cvmfs/publish/cmd_help.cc#L32


vvolkl:
@cernvm-bot cloudtest

vvolkl:
Great, thanks a lot! Test failures are unrelated (some platforms had a network issue when setting up dependencies).

vvolkl (approved):

PR #3490, opened 2024-01-25
Improve nested catalog staging

mharvey-jt:
* Add StageNestedCatalog to other if(MountSubtree()) clauses
* Carry the read lock down into StageNestedCatalog and release immediately before download
* Add write lock acquisition latency logging
* Add envvar control for runtime disablement of staging

### DIFF
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -334,6 +334,19 @@ class AbstractCatalogManager : public SingleCopy {
   const std::vector<CatalogT*>& GetCatalogs() const { return catalogs_; }
 
   /**
+   * Opportunistic optimization: the client catalog manager uses this method
+   * to preload into the cache a nested catalog that is likely to be required
+   * next. Likely, because there is a race with the root catalog reload which
+   * may result in the wrong catalog being staged. That's not a fault though,
+   * the correct catalog will still be loaded with the write lock held.
+   */
+  virtual void StageNestedCatalogByHash(const shash::Any & /*hash*/,
+                                        const PathString & /*mountpoint*/)
+  { }
+  void StageNestedCatalog(const PathString &path, const CatalogT *parent,
+                          bool is_listable);
+
+  /**
    * Create a new Catalog object.
    * Every derived class has to implement this and return a newly
    * created (derived) Catalog structure of it's desired type.
--- a/cvmfs/catalog_mgr.h
+++ b/cvmfs/catalog_mgr.h
@@ -38,7 +38,8 @@ static time_t tick(void) {
 static void tock(time_t tick) {
   time_t delta = platform_monotonic_time_ns() - tick;
   float delta_ms =  delta /1000000.;
-  LogCvmfs(kLogCatalog, kLogDebug, "Catalog Writelock delay %.3f ms", delta_ms );
+  LogCvmfs(kLogCatalog, kLogDebug,
+		  "Catalog Writelock delay %.3f ms", delta_ms);
 }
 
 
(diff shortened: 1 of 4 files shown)


cvmfs/catalog_mgr_impl.h
```diff
@@ -876,9 +878,13 @@ void AbstractCatalogManager<CatalogT>::StageNestedCatalog(
 
     LogCvmfs(kLogCatalog, kLogDebug, "staging nested catalog at %s (%s)",
              i->mountpoint.c_str(), i->hash.ToString().c_str());
-    StageNestedCatalogByHash(i->hash, i->mountpoint);
-    break;
+    shash::Any hash       = shash::Any(i->hash);
+    PathString mountpoint = PathString(i->mountpoint);
+    Unlock();
```
HereThereBeDragons, line 883: If I see it correctly `Unlock()` is called here in a for loop? I dont think `Unlock()` should be called multiple times while in the former implementation it was only called once.

The underlying `pthread_rwlock_unlock` has the following description:
The pthread_rwlock_unlock() function shall release a lock held on the read-write lock object referenced by rwlock. Results are **undefined** if the read-write lock rwlock is not held by the calling thread.

cvmfs/catalog_mgr.h
```diff
@@ -31,6 +31,17 @@ namespace catalog {
 
 const unsigned kSqliteMemPerThread = 1*1024*1024;
 
+static time_t tick(void) {
```
HereThereBeDragons, line 34: I guess this should be maximum be part of the debug build. If not totally removed for default builds

cvmfs/catalog_mgr_impl.h
```diff
@@ -285,7 +285,6 @@ bool AbstractCatalogManager<CatalogT>::LookupPath(const PathString &path,
     LogCvmfs(kLogCatalog, kLogDebug, "looking up '%s' in a nested catalog",
              path.c_str());
     StageNestedCatalog(path, best_fit, false /* is_listable */);
-    Unlock();
```
jblomer, line 288: AFAIK one cannot upgrade a read lock to a write lock. The unlock before taking a write lock is necessary.

cvmfs/catalog_mgr_impl.h
```diff
@@ -544,7 +543,6 @@ bool AbstractCatalogManager<CatalogT>::Listing(const PathString &path,
   CatalogT *catalog = best_fit;
   if (MountSubtree(path, best_fit, true /* is_listable */, NULL)) {
     StageNestedCatalog(path, best_fit, true /* is_listable */);
-    Unlock();
```
jblomer, line 547: see above

cvmfs/catalog_mgr_impl.h
```diff
@@ -582,7 +580,6 @@ bool AbstractCatalogManager<CatalogT>::ListingStat(const PathString &path,
...
```
jblomer, line 585: see above

cvmfs/catalog_mgr_impl.h
jblomer: Looks correct to me. At this point the code path exits the for loop/method.

jblomer (commented):
The optimization in `StageNestedCatalogByHash` is nice, I think!

I'll integrate that #3476. We can also add a stopwatch for the time spent in `WriteLock`.

cvmfs/catalog_mgr_impl.h
```diff
@@ -853,6 +850,11 @@ void AbstractCatalogManager<CatalogT>::StageNestedCatalog(
   assert(parent);
   const unsigned path_len = path.GetLength();
 
+  if (getenv("_CVMFS_NO_CATALOG_STAGING")) {
+    Unlock();
+    return;
+  }
```
jblomer, line 856: @mharvey-jt any particular reason why this is needed?

jblomer:
The improvements except the envvar guard are now in #3476

Issue #3493, opened 2024-02-02
Remove bottleneck:  Allow decompression of download requests in parallel


HereThereBeDragons:
At the moment, decompression is performed in `CallbackCurlData()` function - which is sequentially executed by `MainDownload` thread.

Better would be if the threads executing the `Fetch()` request to download an object would perform the decompression. This would allow for decompression to run in parallel for multiple `Fetch()` requests, while within a single `Fetch()` it would be still sequential. (parallel decompression into the same file is a bit nasty).

Data transfer/communication between `Fetch()` thread and `MainDownload` should be done with some form of bounded queue to prevent memory problems.

Part of bigger issue #3095

HereThereBeDragons:
Final choice of implementation is in PR #3556

PR #3499, opened 2024-02-05
Introduction of data tube to DownloadMgr

HereThereBeDragons:
Issue #3493 

At the moment only supports: creating the tube and when the download finished to sync with `Fetch()`

Next steps:
- finalize tube and surrounding logic to support decompression in `Fetch()` and remove it from `CallbackCurlData()`
- extended `tube.h` to allow for weighted elements to calculate size of tube

### DIFF
--- a/cvmfs/network/jobinfo.h
+++ b/cvmfs/network/jobinfo.h
@@ -25,10 +25,38 @@
 #include "network/sink_mem.h"
 #include "network/sink_path.h"
 #include "util/pipe.h"
+#include "util/tube.h"
 
 class InterruptCue;
 
 namespace download {
+
+enum DataTubeAction {
+  kActionStop = 0,
+  kActionContinue,
+  kActionDecompress
+};
+
+/**
+ * Wrapper for the data tube to transfer data from CallbackCurlData() that is
+ * executed in MainDownload() Thread to Fetch() called by a fuse thread
+ * 
+ * TODO(heretherebedragons): do we want to have a pool of those datatubeelements?
+ */
+struct DataTubeElement {
+  char* data;
+  size_t size;
+  DataTubeAction action;
+
+  explicit DataTubeElement(DataTubeAction xact) :
+                                         data(NULL), size(0), action(xact) { };
+  DataTubeElement(char* mov_data, size_t xsize, DataTubeAction xact) :
+                                  data(mov_data), size(xsize), action(xact) { };
+  
+  ~DataTubeElement() {
+    delete data;
+  }
+};
 
 /**
  * Contains all the information to specify a download job.
@@ -39,6 +67,9 @@ class JobInfo {
   int64_t id_;
   /// Pipe used for the return value
   UniquePtr<Pipe<kPipeDownloadJobsResults> > pipe_job_results;
+  /// Tube (bounded thread-safe queue) to transport data from CURL callback
+  /// to be decompressed in Fetch() instead of MainDownload()
+  UniquePtr<Tube<DataTubeElement> > data_tube_;
   const std::string *url_;
   bool compressed_;
   bool probe_hosts_;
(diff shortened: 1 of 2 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/network/jobinfo.h
```diff
@@ -25,18 +25,49 @@
...
+  kActionDecompress
+};
+
+/**
+ * Wrapper for the data tube to transfer data from CallbackCurlData() that is
+ * executed in MainDownload() Thread to Fetch() called by a fuse thread
+ *
+ * TODO(heretherebedragons): do we want to have a pool of those datatubeelements?
+ */
+struct DataTubeElement {
```
jblomer: Should inherit from `SingleCopy`

cvmfs/network/jobinfo.h
```diff
@@ -98,6 +129,9 @@ class JobInfo {
     if (pipe_job_results.IsValid()) {
       pipe_job_results.Destroy();
     }
+    if (data_tube_.IsValid()) {
+      data_tube_.Destroy();
+    }
```
jblomer: This check is part of `Destroy()`
```suggestion
    data_tube_.Destroy();
```

cvmfs/network/jobinfo.h
```diff
@@ -126,6 +169,8 @@ class JobInfo {
   shash::ContextPtr *GetHashContextPtr() { return &hash_context_; }
   Pipe<kPipeDownloadJobsResults> *GetPipeJobResultWeakRef() {
                                            return pipe_job_results.weak_ref(); }
+  Tube<DataTubeElement> *GetDataTubeWeakRef() {
+                                                 return data_tube_.weak_ref(); }
```
jblomer: I think `weak_ref` is the wrong word in the `UniquePtr` class to begin with. But let's not perpetuate that.
```suggestion
  Tube<DataTubeElement> *GetDataTubePtr() {
                                                 return data_tube_.weak_ref(); }
```

jblomer (changes requested):
Looks generally good to me, some minor comments.

cvmfs/network/jobinfo.h
HereThereBeDragons: i oriented myself at the pipe stuff... i will rename both of them

HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/network/jobinfo.h
```diff
@@ -25,11 +25,39 @@
...
+  kActionStop = 0,
+  kActionContinue,
+  kActionDecompress
+};
+
+/**
+ * Wrapper for the data tube to transfer data from CallbackCurlData() that is
+ * executed in MainDownload() Thread to Fetch() called by a fuse thread
+ *
+ * TODO(heretherebedragons): do we want to have a pool of those datatubeelements?
```
jblomer, line 44: The `Tube` collects them (i.e. I don't see where a pool fits in)

jblomer (approved):

cvmfs/network/jobinfo.h
HereThereBeDragons: in #3501 you see that in each callbackcurldata we need a `malloc` as we need to copy the curl buffer
```
      char *data = static_cast<char*>(malloc(num_bytes));
      memcpy(data, ptr, num_bytes);
      DataTubeElement *ele = new DataTubeElement(data, num_bytes,
                                                         kActionDecompressZlib);
      info->GetDataTubePtr()->EnqueueBack(ele);
```

and in the fetch() we always do `delete ele` after retrieving any element from the tube which includes the freeing of the data buffer

that s why i was wondering if we want like a pool of those elements that are allocated to size `CURL_MAX_WRITE_SIZE` - at least for the flow between curlcallbackdata and the decompress section in fetch()

see doc:
```
The maximum amount of body data that is passed to the write callback is defined in the curl.h header file: CURL_MAX_WRITE_SIZE (the usual default is 16K). If [CURLOPT_HEADER](https://curl.se/libcurl/c/CURLOPT_HEADER.html) is enabled, which makes header data get passed to the write callback, you can get up to CURL_MAX_HTTP_HEADER bytes of header data passed into it. This usually means 100K.
```

PR #3503, opened 2024-02-08
Update webhook to alma9, add service

marcoverl:
updating the example to Almalinux and adding the registry-webhook.service systemd unit


vvolkl (approved):
Thanks!

PR #3501, opened 2024-02-06
WIP: Add parallel decompression

HereThereBeDragons:
Build on top of PR #3499

PR #3504, opened 2024-02-08
WIP: Add parallel decompression V2

HereThereBeDragons:
Issue #3493 

Replaces #3501

### DIFF
--- a/cvmfs/network/download.cc
+++ b/cvmfs/network/download.cc
@@ -273,10 +273,10 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
       memcpy(data, ptr, num_bytes);
       DataTubeElement *ele = new DataTubeElement(data, num_bytes,
                                                          kActionDecompressZlib);
-      info->GetDataTubeWeakRef()->EnqueueBack(ele);
+      info->GetDataTubePtr()->EnqueueBack(ele);
     } else { // TODO(heretherebedragons) i think we need this here to support
              // the non-multihreaded version?
-      zlib::StreamStates retval =
+      const zlib::StreamStates retval =
         zlib::DecompressZStream2Sink(ptr, static_cast<int64_t>(num_bytes),
                                     info->GetZstreamPtr(), info->sink());
       if (retval == zlib::kStreamDataError) {
@@ -707,8 +707,8 @@ void *DownloadManager::MainDownload(void *data) {
         // decompressing the data so that VerifyAndFinalize executes correctly
         if (info->IsValidDataTube()) {
           DataTubeElement *ele = new DataTubeElement(kActionEndOfData);
-          info->GetDataTubeWeakRef()->EnqueueBack(ele);
-          info->GetDataTubeWeakRef()->Wait();
+          info->GetDataTubePtr()->EnqueueBack(ele);
+          info->GetDataTubePtr()->Wait();
         }
 
         if (download_mgr->VerifyAndFinalize(curl_error, info)) {
@@ -1655,7 +1655,7 @@ bool DownloadManager::VerifyAndFinalize(const int curl_error, JobInfo *info) {
     if (failover_indefinitely_) {
       // try again, breaking if there's a cvmfs reload happening and we are in a
       // proxy failover. This will EIO the call application.
-      bool interrupted = Interrupted(fqrn_, info);
+      const bool interrupted = Interrupted(fqrn_, info);
       if (!interrupted) {
         info->SetErrorCode(kFailOk);
       }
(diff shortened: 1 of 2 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/network/jobinfo.h
```diff
@@ -34,7 +34,9 @@ namespace download {
 enum DataTubeAction {
   kActionStop = 0,
   kActionContinue,
-  kActionDecompress
+  kActionEndOfData,
+  kActionDecompressZlib,
```
jblomer, line 38:
```suggestion
  kActionDecompress,
```

cvmfs/network/download.cc
```diff
@@ -1888,15 +1916,68 @@ Failures DownloadManager::Fetch(JobInfo *info) {
     //          info->wait_at[0], info->wait_at[1]);
     pipe_jobs_->Write<JobInfo*>(info);
 
+    bool is_running = true;
     do {
```
jblomer, line 1967: This loop should go in its own method

cvmfs/network/download.cc
```diff
@@ -1888,15 +1916,68 @@ Failures DownloadManager::Fetch(JobInfo *info) {
...
+      switch (ele->action) {
+        case kActionStop:
+          is_running = false;
+        break;
+        case kActionEndOfData:
+          /*
+           * End of data processing.
+           * Used for synchronization before calling VerifyAndFinalize()
+           * Next element action received should be kActionStop
+           */
```
jblomer, line 1979: If we do nothing, why does this command need to be sent?

cvmfs/network/download.cc
```diff
@@ -1888,15 +1916,68 @@ Failures DownloadManager::Fetch(JobInfo *info) {
...
+               info->id(), num_bytes, info->sink()->Describe().c_str(),
+               written);
+            }
+          }
+        }
+        break;
+        default:
+          LogCvmfs(kLogDownload, kLogSyslogErr | kLogDebug,
+               "(id %" PRId64 ") FAILURE - Unknown DataTube Element Action: %d",
+               info->id(), ele->action);
```
jblomer, line 2026: Should be a PANIC

cvmfs/network/download.cc
```diff
@@ -255,37 +255,49 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
 
   // LogCvmfs(kLogDownload, kLogDebug, "Data callback,  %d bytes", num_bytes);
 
-  if (num_bytes == 0)
+  // the check for kFailOk is to check when using the DataTube that there was
+  // not early some cancellation of the download due to error
+  // TODO(heretherebedragons) we might want to have this as an atomic variable?
+  if (num_bytes == 0 || info->error_code() != kFailOk) {
```
jblomer: Now that multiple threads work concurrently on the info object, we have to make sure to lock common members, such as the error code.

cvmfs/network/download.cc
```diff
@@ -255,37 +255,49 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
...
-      return 0;
-    } else if (retval == zlib::kStreamIOError) {
-      LogCvmfs(kLogDownload, kLogSyslogErr,
+  if (info->IsValidDataTube()) {
+    char *data = static_cast<char*>(malloc(num_bytes));
+    memcpy(data, ptr, num_bytes);
+    DataTubeElement *ele = new DataTubeElement(data, num_bytes, kActionData);
+    info->GetDataTubePtr()->EnqueueBack(ele);
+  } else {  // TODO(heretherebedragons) i think we need this here to support
+            // the non-multihreaded version?
```
jblomer, line 271: That is true. So this block should be factored out.

cvmfs/network/download.cc
```diff
@@ -255,37 +255,49 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
...
-    if (retval == zlib::kStreamDataError) {
-      LogCvmfs(kLogDownload, kLogSyslogErr,
-                                     "(id %" PRId64 ") failed to decompress %s",
-                                     info->id(), info->url()->c_str());
-      info->SetErrorCode(kFailBadData);
-      return 0;
-    } else if (retval == zlib::kStreamIOError) {
-      LogCvmfs(kLogDownload, kLogSyslogErr,
+  if (info->IsValidDataTube()) {
+    char *data = static_cast<char*>(malloc(num_bytes));
```
jblomer: Since this is `malloc`d, `DataTubeElement` needs to `free()` it (not `delete`).

jblomer: Also: please use our `smalloc()`

cvmfs/network/download.cc
```diff
@@ -689,6 +701,15 @@ void *DownloadManager::MainDownload(void *data) {
                  "Number of CURL redirects %" PRId64 , info->id(), redir_count);
 
         curl_multi_remove_handle(download_mgr->curl_multi_, easy_handle);
+
+        // let's notify CURL is done and wait for the finishing of
+        // decompressing the data so that VerifyAndFinalize executes correctly
+        if (info->IsValidDataTube()) {
+          DataTubeElement *ele = new DataTubeElement(kActionEndOfData);
+          info->GetDataTubePtr()->EnqueueBack(ele);
+          info->GetDataTubePtr()->Wait();
```
jblomer: It may be necessary to let the other side of the tube signal when it's done. So that we don't need to wait here. For that signalling to work, we would actually need again the `kActionEndOfData`, which could trigger another bit to be send into the `poll` event loop. Let's discuss.

cvmfs/network/download.cc
HereThereBeDragons: because the "gate"/synchronization of when to run `VerifyAndFinalize` is based on the empty data tube. 
but you will remove an element before decompressing it. during decompression errors can occur.  as such we need one last item to tell us "we are done with all data processing"

cvmfs/network/download.cc
HereThereBeDragons: true.. all of it needs some clean up if we really found a solution that is at least equal and better (at least in some situations) then the old version

cvmfs/network/jobinfo.h
HereThereBeDragons: cleanup later on

cvmfs/network/download.cc
HereThereBeDragons: oops true. my mistake. thanks for catching this

cvmfs/network/download.cc
HereThereBeDragons: in the follow up commit 21b2c9d i have now a vector in the maindownload that takes all the "download finished but data still processing" elements to not block here

cvmfs/network/download.cc
HereThereBeDragons: in the final version i will do a nice cleanup putting it in a function

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
Superseeded by #3556

PR #3505, opened 2024-02-08
Use fuse3 loop_config if available

### LINKED ISSUE #3479: Configuration options for FUSE min_idle, max threads
libfuse >3.12 changes the behaviour of worker thread creation, and gives two new configuration control, min_idle_threads and max_threads. It's useful to be able to control these, as the default max_threads appears rather low.   The patch to add configuration options is simple, but it requires bumping the `FUSE_USE_VERSION` from 31 to 312  and that's definitely going to be a problem for OS support. Probably best to have this a compile-time option? 

```
@@ -1085,7 +1085,24 @@ int FuseMain(int argc, char *argv[]) {
#if CVMFS_USE_LIBFUSE == 2
    retval = fuse_session_loop_mt(session);
#else
   #if FUSE_VERSION >= 312
      struct fuse_loop_config *fuse_loop_cfg =  fuse_loop_cfg_create();
      fuse_loop_cfg_set_clone_fd( fuse_loop_cfg, 1 );
      if (options_manager->GetValue("CVMFS_FUSE3_MAX_THREADS", &parameter) ) {
         int val = String2Int64(parameter);
         fuse_loop_cfg_set_max_threads( fuse_loop_cfg, val );
        LogCvmfs(kLogCvmfs, kLogStdout, "FUSE3: Loop config: max_threads=%d", val );
      }
      if (options_manager->GetValue("CVMFS_FUSE3_IDLE_THREADS", &parameter) ) {
         int val = String2Int64(parameter);
         fuse_loop_cfg_set_idle_threads( fuse_loop_cfg, val );
        LogCvmfs(kLogCvmfs, kLogStdout, "FUSE3: Loop config: min_idle_threads=%d", val );
      } 

      retval = fuse_session_loop_mt(session, fuse_loop_cfg );
   #else
    retval = fuse_session_loop_mt(session, 1 /* use fd per thread */);
   #endif 
#endif
  }
  SetLogMicroSyslog(*usyslog_path_);
  ```

jblomer:
Applies/fixes #3479

Adds support for the new `CVMFS_FUSE3_IDLE_THREADS` and `CVMFS_FUSE3_MAX_THREADS` parameters.


jblomer:
@cernvm-bot cloudtest

vvolkl (approved):
Looks good.

vvolkl:
The new parameters should still be added to the docs - I'll update them.

PR #3506, opened 2024-02-09
DownloadManager: Parallel decomp - global pool of unused elements

HereThereBeDragons:
Based on PR #3504  -- possibly a performance improvement
 
PR #3504 has in each `CallbackCurlData` a malloc and in `Fetch()` a delete for each received element.

This PR here adds a "global pool" (another tube) of unused elements that `CallbackCurlData` can reuse. If none is available, a new element is created. 

During Spawn of DownloadManager 10k elements are created with CURL_MAX_WRITE_SIZE buffer size to give it a headstart.

All elements are destroyed during deconstruction of DownloadManager

### DIFF
--- a/cvmfs/network/download.cc
+++ b/cvmfs/network/download.cc
@@ -273,10 +273,10 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
       memcpy(data, ptr, num_bytes);
       DataTubeElement *ele = new DataTubeElement(data, num_bytes,
                                                          kActionDecompressZlib);
-      info->GetDataTubeWeakRef()->EnqueueBack(ele);
+      info->GetDataTubePtr()->EnqueueBack(ele);
     } else { // TODO(heretherebedragons) i think we need this here to support
              // the non-multihreaded version?
-      zlib::StreamStates retval =
+      const zlib::StreamStates retval =
         zlib::DecompressZStream2Sink(ptr, static_cast<int64_t>(num_bytes),
                                     info->GetZstreamPtr(), info->sink());
       if (retval == zlib::kStreamDataError) {
@@ -707,8 +707,8 @@ void *DownloadManager::MainDownload(void *data) {
         // decompressing the data so that VerifyAndFinalize executes correctly
         if (info->IsValidDataTube()) {
           DataTubeElement *ele = new DataTubeElement(kActionEndOfData);
-          info->GetDataTubeWeakRef()->EnqueueBack(ele);
-          info->GetDataTubeWeakRef()->Wait();
+          info->GetDataTubePtr()->EnqueueBack(ele);
+          info->GetDataTubePtr()->Wait();
         }
 
         if (download_mgr->VerifyAndFinalize(curl_error, info)) {
@@ -1655,7 +1655,7 @@ bool DownloadManager::VerifyAndFinalize(const int curl_error, JobInfo *info) {
     if (failover_indefinitely_) {
       // try again, breaking if there's a cvmfs reload happening and we are in a
       // proxy failover. This will EIO the call application.
-      bool interrupted = Interrupted(fqrn_, info);
+      const bool interrupted = Interrupted(fqrn_, info);
       if (!interrupted) {
         info->SetErrorCode(kFailOk);
       }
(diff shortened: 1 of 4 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

cvmfs/network/download.cc
```diff
@@ -70,6 +70,31 @@ using namespace std;  // NOLINT
...
+Tube<DataTubeElement>* DownloadManager::data_tube_empty_elements_ =
+                                              new Tube<DataTubeElement>(200000);
+atomic_int32 DownloadManager::counter_use_data_tube_ = 0;
+
+static DataTubeElement* GetUnusedDataTubeElement() {
+  DataTubeElement* ele = DownloadManager::data_tube_empty_elements_->
+                                                                  TryPopFront();
+
+  if (ele == NULL) {
+    char *data = static_cast<char*>(malloc(CURL_MAX_HTTP_HEADER));
```
rezan: You have `CURL_MAX_WRITE_SIZE` down below, which seems like the more correct value.

cvmfs/network/download.cc
```diff
@@ -1667,6 +1770,13 @@ DownloadManager::~DownloadManager() {
     health_check_.Reset();
   }
 
+  int32_t old_count =
+                    atomic_xadd32(&DownloadManager::counter_use_data_tube_, -1);
+  if (old_count == 1) {
+    delete data_tube_empty_elements_;
```
rezan: This breaks the test suite because the above count hits zero and this object is deleted. But then a subsequent test is spawned and it reads freed memory. So either no delete or you need to re-allocate on demand.

cvmfs/network/download.cc
```diff
@@ -255,37 +280,50 @@ static size_t CallbackCurlData(void *ptr, size_t size, size_t nmemb,
...
-      LogCvmfs(kLogDownload, kLogSyslogErr,
-                                     "(id %" PRId64 ") failed to decompress %s",
-                                     info->id(), info->url()->c_str());
-      info->SetErrorCode(kFailBadData);
-      return 0;
-    } else if (retval == zlib::kStreamIOError) {
-      LogCvmfs(kLogDownload, kLogSyslogErr,
+  if (info->IsValidDataTube()) {
+    DataTubeElement *ele = GetUnusedDataTubeElement();
+    memcpy(ele->data, ptr, num_bytes);
```
rezan, line 294: I think there should be some kind of assert here where we make sure `DataTubeElement.data` has enough capacity for the memcpy().

cvmfs/network/download.cc
HereThereBeDragons: thanks thats an old left over. it should be everywhere `CURL_MAX_WRITE_SIZE`

cvmfs/network/download.cc
HereThereBeDragons: We can add it.

but i am not sure if it is needed: in https://curl.se/libcurl/c/CURLOPT_WRITEFUNCTION.html it states 
"The callback function is passed as much data as possible in all invokes, but you must not make any assumptions. It may be one byte, it may be thousands. The **maximum** amount of body data that is passed to the write callback is defined in the curl.h header file: CURL_MAX_WRITE_SIZE (the usual default is 16K)."

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
@cernvm-bot cloudtest nodestroy

rezan:
I did some benchmarks using a synthetic workload, results looks great:

```
DownloadManager (normal)

 2 threads - 1 core - 480 MB/s
 4 threads - 1 core - 540 MB/s
 8 threads - 1 core - 560 MB/s
12 threads - 1 core - 600 MB/s
16 threads - 1 core - 640 MB/s

DownloadManager (parallel)

 2 threads - 2 cores -  800 MB/s
 4 threads - 4 cores - 1400 MB/s
 8 threads - 5 cores - 1800 MB/s
12 threads - 5 cores - 1900 MB/s
16 threads - 5 cores - 1900 MB/s
```

HereThereBeDragons:
@cernvm-bot cloudtest

HereThereBeDragons:
Superseeded by #3556

PR #3508, opened 2024-02-12
[cleanup] fix syntax highlighting in test scripts

vvolkl:
This is a purely decorative change, in the sense that it should have no effect on any functionality, but fixes syntax highlighting. Since the integration test scripts have no .sh suffix, most editors will display them as text files, in particular also the github web page renders them in this not-very-readable way (VSCode manages to infer that it's a script).

Adding a #!/bin/bash throughout seemed the easiest way of fixing this (adding a .sh suffix means changing a bunch more scripts) - it may be slightly confusing because none of these scripts are executed (all of them are sourced, which ignores the #!/bin/bash, and then cvmfs_run_test() is run by another script), but it provides enough of a hint to editors to make them more readable.

I also take the opportunity to harmonize file permissions on the main scripts - a few of them had +x set, I've removed this for all of them.

### DIFF
--- a/test/benchmarks/000-dummy/main
+++ b/test/benchmarks/000-dummy/main
@@ -1,3 +1,4 @@
+#!/bin/bash
 
 cvmfs_test_name="Dummy benchmark"
 cvmfs_test_autofs_on_startup=false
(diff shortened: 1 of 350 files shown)


vvolkl:
@cernvm-bot cloudtest

jblomer (approved):

vvolkl:
Test failure seems flaky.

vvolkl:
@cernvm-bot  cloudtest

PR #3509, opened 2024-02-13
Testbed scripts and tutorial: client installation/options

HereThereBeDragons:


test/testbed/rhel9_client_installation.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Further changes
+For using share memory as cache base for cvmfs create the following:
+```
+mkdir /dev/shm/cvmfs-cache
+```
+
+Set readahead to 1024 KiB
+```
+echo 1024 > /sys/class/bdi/<fsid>/read_ahead_kb
```
HereThereBeDragons, line 90: i dont know how we do this. fuse apparently only gets temporary fsid assigned? need to study a bit of each repo always gets the same fsid or if there is a global parameter.

mharvey-jt: `stat` the mount point of repo to find the `Device: XXh/YYd` id. The corresponding entry in bdi is  `0:YYd`

HereThereBeDragons: thanks. added now a script that should do it

Issue #3460, opened 2023-12-01
gateway 2.11 performance regression


mharvey-jt:
Hi,

Moving from gateway version 1.2 to 2.11.2 we see a big degradation in responsivity when the gateway is under heavy load, eg `api/v1/leases` can take tens of seconds to respond. 
The problem is that the CommitLease function holds a lock on the database while the receiver runs, blocking the frontend from servicing any other requests that take a db lock.

https://github.com/cvmfs/cvmfs/blob/76355d0043569c130fe2c0ffcd3ecae3280d14c5/gateway/internal/gateway/backend/lease_service.go#L269

Issue #3259, opened 2023-04-20
[gw] concurrency bug in lease database: "delete statement failed: database is locked"


vvolkl:
Publishing simultaneously can lead to errrors like

```
Lease end request - error reply: {"reason":"delete statement failed: database is locked","status":"error"}
```
This has surfaced on lhcbdev after updating to  2.10.1

PR #3258, opened 2023-04-18
[gw] fix for a concurrency issue in the leases db

vvolkl:
Closes #3259

PR #3510, opened 2024-02-15
[gw] add regression test for lease db locking

vvolkl:
This test reproduces the issue of lease end requests failing with

```
Lease end request - error reply: {"reason":"delete statement failed: database is locked","status":"error"}
```

as reported originally by LHCb (#3259).

The problem has been fixed in #3258 by liberally applying mutexes, but at the cost of performance regressions in other API operations, as reported by Matt (#3460). This patch supplements a regression test in order to be able to amend the original fix.

The test here handcrafts the API requests to the gateway instead of using transactions, which would work too, in order to avoid the overhead of running multiple publishers. It fairly consistently reproduces the error when testing a non-fixed gateway ( I tried to find a compromise between runtime and constency here. Sleeping between requests seems to had little effect)  and runs successfully with the fix.

Since it only gets and cancels leases, the test may not be exhaustive, but it's a start.

### DIFF
--- a/test/src/818-gateway-concurrent-leases/main
+++ b/test/src/818-gateway-concurrent-leases/main
@@ -27,7 +27,7 @@ cancel_lease () {
 }
 
 stresstest_lease() {
-  for i in {1..50}; do
+  for i in {1..100}; do
     local response=$(get_lease)
     local status=$(echo  ${response} | jq --raw-output .status)
     echo $response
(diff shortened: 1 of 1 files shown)


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot unittest

vvolkl:
@cernvm-bot cloudtest

PR #3513, opened 2024-02-23
Add support for high-precision timestamps

### LINKED ISSUE #3413: RFE: high precision mtime
Hi,

Currently CVMFS stores the mtime of a file within the catalog as an INTEGER, which is 32b. This isn't sufficient precision to store the exact timestamp of a source file, which we would like to use for comparing files. Could the mtime in the catalog be extended to 64bit in. backwards compatible way?

jblomer:
Extends the file catalog table by a `mtimens` column for the nanosecond part of the timestamp. The catalog revision increases to 7, the change is backwards and forwards compatible.

Introduces a new boolean server parameter `CVMFS_ENABLE_MTIME_NS`. Off by default, because nanosecond timstamps show a catalog size increase of a little under 10% compressed and a little under 5% uncompressed.

Fixes #3413

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -656,6 +656,7 @@ if (BUILD_SERVER)
        swissknife_sync.cc
        swissknife_zpipe.cc
        sync_item.cc
+       sync_item_dummy.cc
        sync_item_tar.cc
        sync_mediator.cc
        sync_union.cc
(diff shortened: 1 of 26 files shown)


jblomer:
@cernvm-bot cloudtest

jblomer:
@vvolkl I think this is now ready to be merged. We should probably run the integration tests again, we ran into a bunch of timeouts.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
`swissknife_migrate` is missing some updates I think - will take a look at that.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot unittest

vvolkl:
Remaining test failure unrelated

vvolkl (approved):

PR #3524, opened 2024-03-01
Containerize testing environment for "quick" suite

amalthundiyil:
### What

Adds ability to execute the "quick" test suite in a container. Example workflow:

1. Cloning
```sh
git clone https://github.com/cvmfs/cvmfs.git
cd tests/common/container
docker-compose up --build -d
docker exec -u sftnight -it cvmfs-dev bash
```
2. Building (from inside the docker environment):
```
cmake -S /home/sftnight/cvmfs -B /tmp/cvmfs-build -D EXTERNALS_PREFIX=/tmp/cvmfs-ext -D BUILD_SHRINKWRAP=ON
```

3. Testing
```sh
cd /home/sftnight/cvmfs/test
CVMFS_TEST_USER=sftnight ./run.sh /tmp/cvmfs-test.log -s "quick" ./src/0* ./src/1*
```
### Why

- Enables faster developer testing iterations (versus provisioning/destroying VMs)
- Serves as a good first step towards containerizing other elements of CVMFS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Docker
+name: CI
 
 on:
   push:
@@ -7,8 +7,8 @@ on:
     branches: [devel,master]
 
 jobs:
-  test:
-    name: "Test"
+  docker:
+    name: "Docker"
     runs-on: ubuntu-20.04
 
     steps:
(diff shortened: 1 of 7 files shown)


vvolkl (changes requested):
Great, thanks! This is a good first step indeed. I don't think you need to copy the alma9 Dockerfile with the build-dependencies - just pull from `gitlab-registry.cern.ch/cernvm/build-images/centos_x86_64:9` instead.

The entry point can be expanded to be a bit more flexible where cvmfs is taken from in a future PR. In this PR, maybe you can still add a short readme in test/common/container (basically with what you wrote in this PR. If possible you could also already add a rudimentary github actions workflow that runs this.

Next steps should then be  to enable the server tests in the container.

vvolkl:
@cernvm-bot cloudtest

.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
+name: CI
+
+on:
+  push:
+    branches: [devel,master]
+  pull_request:
+    branches: [devel,master]
```
vvolkl:
```suggestion
  push:
      branches: [devel]
  pull_request:
```
I don't think we need to restrict running this to any branches for pull requests - we sometimes have pull request targeting release branches. For push that seems sensible, but we don't really use the master branch.

vvolkl (approved):
Very nice, thanks!

Issue #3520, opened 2024-02-27
assert in `PageCacheTracker::Close`


mharvey-jt:
```
Program terminated with signal SIGABRT, Aborted.
#0  0x00007ffff6d05a9f in raise () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7fefa1ffb700 (LWP 33289))]
#1  0x00007ffff6cd8e05 in abort () from /lib64/libc.so.6
#2  0x00007ffff6cd8cd9 in __assert_fail_base.cold.0 () from /lib64/libc.so.6
#3  0x00007ffff6cfe3f6 in __assert_fail () from /lib64/libc.so.6
#4  0x00007ffff6398059 in glue::PageCacheTracker::Close (this=0x820a20, 
    inode=<optimized out>, inode@entry=206980125)
    at /data/build_modules/build/cvmfs-fuse3/cvmfs-fuse3/cvmfs/glue_buffer.cc:431
#5  0x00007ffff63eef21 in cvmfs::cvmfs_release (req=0x7fef901b9780, 
    ino=206980125, fi=<optimized out>)
    at /data/build_modules/build/cvmfs-fuse3/cvmfs-fuse3/cvmfs/mountpoint.h:539
#6  0x00007ffff6a93f1c in loader::stub_release (req=0x7fef901b9780, 
    ino=206980125, fi=0x7fefa1ffad30)
    at /data/build_modules/build/cvmfs-fuse3/cvmfs-fuse3/cvmfs/loader.cc:281
#7  0x00007ffff6863c4b in do_release () from /lib64/libfuse3.so.3
#8  0x00007ffff6865fea in fuse_session_process_buf_int ()
   from /lib64/libfuse3.so.3
#9  0x00007ffff6861310 in fuse_do_work () from /lib64/libfuse3.so.3
#10 0x00007ffff79b31cf in start_thread () from /lib64/libpthread.so.0
#11 0x00007ffff6cf0dd3 in clone () from /lib64/libc.so.6
```

https://github.com/cvmfs/cvmfs/blob/c0a48348df0066f89c7a7b52040a4ab1b4eaef83/cvmfs/glue_buffer.cc#L431

jblomer:
As a first step, I'll add some more information to this assert. PR coming soon.

mharvey-jt:
Thanks @jblomer! This is currently the last remaining client bug we are logging.

jblomer:
[...]

PR #3526, opened 2024-03-01
Improve error logging in page cache tracker

jblomer:
In support of #3520


cvmfs/glue_buffer.cc
```diff
@@ -415,6 +422,7 @@ void PageCacheTracker::Close(uint64_t inode) {
     return;
   }
 
+  int32_t old_open = entry.nopen;
```
vvolkl: From clang-tidy:
```suggestion
  const int32_t old_open = entry.nopen;
```

vvolkl (approved):

PR #3530, opened 2024-03-13
Fix some printf related warnings (for fedora38 build)

vvolkl:
This commit silences two types of  warnings, that need to be fixed since rpmbuild treats them as errors, starting with Fedora 38.

The first is:
```
cvmfs/cvmfs/util/exception.cc: In function ‘void Panic(const char*, LogSource, int, const char*, ...)’: cvmfs/cvmfs/util/logging.h:28:51: warning: format not a string literal and no format arguments [-Wformat-security]
   28 |     ((void)0) : LogCvmfs(source, mask, __VA_ARGS__));  // NOLINT
      |                                                   ^
cvmfs/cvmfs/util/exception.cc:47:3: note: in expansion of macro ‘LogCvmfs’
   47 |   LogCvmfs(source, mask, msg);
      |   ^~~~~~~~
```
where I add a "%s" format to fix it.

The second is -Wformat-zero-length, with more occurrences due to the use of
```
LogCvmfs(kLogCvmfs, kLogStdout, "");
```
to print a newline. The fix is to replace these with
```
LogCvmfs(kLogCvmfs, kLogStdout | kLogNoLinebreak, "\n");
```

### DIFF
--- a/cvmfs/cvmfs_fsck.cc
+++ b/cvmfs/cvmfs_fsck.cc
@@ -333,7 +333,7 @@ int main(int argc, char **argv) {
   }
   free(workers);
   if (!g_verbose)
-    LogCvmfs(kLogCvmfs, kLogStdout, "");
+    LogCvmfs(kLogCvmfs, kLogStdout | kLogNoLinebreak, "\n");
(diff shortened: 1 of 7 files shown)


vvolkl:
@cernvm-bot cloudtest

PR #3538, opened 2024-03-14
Improvements to the commit processor

jblomer:
Addresses several performance issues when committing catalogs through the gateway. 

There are more inefficiencies around the server tool and the signing tool. We should address them in several follow-up PRs, I think.

Fixes #3532
Fixes #3535

PR #3539, opened 2024-03-14
Remove 'trusted certs'

jblomer:
The "trusted_certs" were an early attempt to use X.509 certificates for the top-level signature (whitelist signature). This was never fully implemented nor used and can be removed.


cvmfs/server_tool.cc
```diff
@@ -42,7 +42,7 @@ bool ServerTool::InitDownloadManager(const bool follow_redirects,
 }
 
 bool ServerTool::InitVerifyingSignatureManager(
-    const std::string &pubkey_path, const std::string &trusted_certs) {
+    const std::string &pubkey_path, const std::string &/*trusted_certs*/) {
```
vvolkl: Why not remove the second argument alltogether? There's no interface to fullfill here, right?

jblomer: That's right, we'll actually do it with #3538 but we can do it here, too.

vvolkl:
@cernvm-bot cloudtest

jblomer:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3545, opened 2024-03-16
Fix some compiler warnings on macOS

jblomer:


jblomer:
@cernvm-bot cloudtest

PR #3546, opened 2024-03-18
Fixes and improvements to concurrency in SessionContext

jblomer:
The macOS unit test revealed a race in the `SessionContext` class. This PR
  - fixes the race
  - slightly simplifies the `SessionContext` class
  - fixes a pthread API misuse in the `Future` class
  - moves the `Future` class in its own header
  - replaces the `FifoChannel` class by the (newer) `Tube` in SessionContext


vvolkl:
@cernvm-bot cloudtest

vvolkl (approved):

PR #3549, opened 2024-03-18
Fix test using geodb api

vvolkl:
This is a workaround for the recent introduction of rate limiting. Letting CVMFS_GEO_DB_FILE be  a url makes it easier to distribute it for cloudtests, and setting it via env var is convenient too.

### DIFF
--- a/test/src/595-geoipdbupdate/main
+++ b/test/src/595-geoipdbupdate/main
@@ -81,15 +81,15 @@ cvmfs_run_test() {
     return 0
   fi
 
+  if [ -f /etc/cvmfs/server.local ]; then
+    . /etc/cvmfs/server.local
+  fi
   if [ -n "$CVMFS_GEO_DB_FILE" ]; then
     echo "Geo database update disabled, skipping tests"
     CVMFS_GENERAL_WARNING_FLAG=1
     return 0
   fi
 
-  if [ -f /etc/cvmfs/server.local ]; then
-    . /etc/cvmfs/server.local
-  fi
   if [ -z "$CVMFS_GEO_LICENSE_KEY" ]; then
     if [ -f /usr/share/GeoIP/`basename $CVMFS_TEST_595_GEODB` ]; then
       echo "System geo database installed, skipping tests"


vvolkl:
@cernvm-bot cloudtest

cvmfs/server/cvmfs_server_util.sh
```diff
@@ -1059,10 +1059,15 @@ _update_geodb() {
...
         return 1
       fi
-      rm -f $dbfile
-      echo "Linking GeoIP Database"
-      _to_syslog_for_geoip "linking db from $CVMFS_GEO_DB_FILE"
-      ln -s $CVMFS_GEO_DB_FILE $dbfile
+      if [ $(echo $CVMFS_GEO_DB_FILE | cut --bytes=1-7) = "http://" -o $(echo $CVMFS_GEO_DB_FILE | cut --bytes=1-8) = "https://" ]; then
+        rm -f $dbfile
+        echo "Downloading GeoIP Database"
+        curl $CVMFS_GEO_DB_FILE -o $dbfile
```
DrDaveD: I'm not sure I Iike this re-use of CVMFS_GEO_DB_FILE.  The CVMFS_GEO_DB_FILE is documented in readthedocs as being a file, not a URL, and the name implies that it is a FILE.  How about just downloading the file in the wrapping test code instead of in cvmfs_server?

I asked @cvuosalo to make an optional way to download a .mmdb file from db-ip.com instead of Maxmind and that's currently just a URL to download from so I think we will get a parameter to set a download URL, but I don't think this is a good way to do it.

DrDaveD (changes requested):

cvmfs/server/cvmfs_server_util.sh
vvolkl: Fair enough, this is a workaround that can be done in the test setup logic, I updated the PR.

vvolkl:
@cernvm-bot cloudtest

cvmfs/server/cvmfs_server_util.sh
vvolkl: Having an alternative provider would be very welcome. For our tests it makes sense in any case to have the option to use a locally cached db. I'll configure our weekly builds to fetch from Maxmind so we still have test coverage.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
(This is now done fully in ci-scripts)

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3558, opened 2024-03-28
Fix building of micro benchmarks

jblomer:
Note: after merging, incremental builds need to be reset or the `.bootstrapDone` files need to be fixed-up.


vvolkl:
@cernvm-bot unittest

vvolkl:
Thanks! Two small suggestions I added directly to this PR:

* move the check for BUILD_UBENCHMARK in bootstrap.sh . This avoids an error when first running cmake without the flag, and then adding it in a second go.
* googlebench seems to have added one more cmake flag for its internal testing, added that to the external configure hook (got an error looking for `python` otherwise)

vvolkl (approved):

vvolkl:
Unittest run with clean externals: https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsUnittest/2233/

Issue #3529, opened 2024-03-11
New MaxMind API for geoIP downloads


DrDaveD:
Maxmind has again [changed the API for geoIP database downloads](https://dev.maxmind.com/geoip/release-notes/2024#presigned-urls-for-database-downloads), and the old one will stop working May 1.   `cvmfs_server` needs to be updated to use the new API.

DrDaveD:
@cvuosalo found that new license keys are already requiring a `-L` option added to curl to respond to a redirect request.  Old license keys don't require that.   This means that anybody trying to set up a geodb with a new account between now and when this issue is fixed will need to manually hack the curl command in cvmfs_server.

vvolkl:
@DrDaveD We also see an issue with geodb in our integration tests. Do you have a fix already or should we take a look? I'm happy to do a patch release soon after this is fixed.

cvuosalo:
@vvolkl We have a fix in mind, but it will take a few work days to complete.

DrDaveD:
What issue do you see?  A temporary workaround might be to add the `-L` option to the curl line, if it's the same problem that Carl saw with a new license key.

jblomer:
There is now a rate limit that we hit.

DrDaveD:
Ah.  I think there has been a rate limit for quite some time, but maybe they reduced the limit.

cvuosalo:
PR #3566  addressing this issue has been merged. I think this issue can be closed.

PR #3559, opened 2024-04-01
Update method of downloading the MaxMind GeoIP database

cvuosalo:
This merge request addresses issue #3529 -- the need to update  `cvmfs_server` because MaxMind is changing the API for GeoIP database downloads. The new method requires the account ID.

`cvmfs` administrator documentation will have to be changed to say that the new variable `CVMFS_GEO_ACCOUT_ID` needs to be added to `/etc/cvmfs/server.local`. The administrator will need to check their MaxMind account to get their account ID.

These changes were tested to verify that the `cvmfs_server update-geodb` command successfully downloads the database file by the new method. Testing also confirms the command will now fail if `CVMFS_GEO_ACCOUT_ID` is not configured.


cvuosalo:
I'll close this request and push it to its own branch.

PR #3560, opened 2024-04-02
Fix ATLAS tightcache find test

vvolkl:
This test keeps failing with Input/Output errors as the repository grows. It is likely that this is just the increased size of pinned catalogs that do not fit anymore in the purposefully small cache, but it should be debugged with a recent root hash. For now, use an older root hash that is known to work in this test to fix the test suite.


vvolkl:
@cernvm-bot cloudtest full

vvolkl:
Still some other tests to fix, but this one passes.

Issue #3562, opened 2024-04-03
cvmfs_receiver: grafting inefficiency when adding new directories with nested catalogs


mharvey-jt:
This problem manifests when the gateway lease path does not already exist in the repo - new nested catalogs are unnecessarily recursed into during the merge.

consider a directory hierarchy `/TOPDIR/1/2/3/4/4/5/6/7/8/9/10` being created on an existing `/TOPDIR` with catalog. A catalog is added on each directory. The change is published via a gateway, with lease `TOPDIR/1`. The `cvmfs_receiver` commit processor will unexpectedly recurse down through all of the catalogs in the merge, down to `..../10`. 
In a subsequent commit that touches all of these directories, where the lease path already exists,  the commit processor will only wrangle the catalogs for `/`, `/TOPDIR`, `/TOPDIR/1`, as expected.

This initial, apparently unnecessary, recursion marked slows down the critical path commit of new directories into a repo

Looks like this might be a case not handled by the previous nested catalog optimisation https://github.com/cvmfs/cvmfs/issues/2707

mharvey-jt:
https://github.com/cvmfs/cvmfs/pull/3564/files is a quick attempt to fix it

jblomer:
A similar optimization could be done for the removal of nested catalog hierarchies. Is this a relevant use case?

EDIT: This is actually #3292

PR #3564, opened 2024-04-03
poc for non-recursive grafting at a  non-existent lease path

mharvey-jt:


jblomer:
Related to #3562

jblomer:
Replaced by #3565

PR #3575, opened 2024-04-10
Do only one gc -a at a time, and remove need for check -a to be run by root

### LINKED ISSUE #3574: Multiple swissknife pull and gc commands happening at once
The backup Nebraska stratum 1 hcc-cvmfs2.unl.edu is behaving slowly right now, we're not sure why yet, but I noticed that there were 4 copies of `cvmfs_swissknife pull unpacked.cern.ch` happening at once.  Looking at the open file descriptors to the lock file, three of them were marked `(deleted)` so something is deleting them.    I am quite sure that the problem is in the `gc -a` process, because that starts up once a day and each of the pull processes had started on subsequent days.

There were also multiple copies of the `gc -a` process happening, which also shouldn't happen.  I can see why that is, it is because there is no lock against it.  The `check -a` process has an `is_checking_all` lock to prevent that, and there should be a corresponding `is_gcing_all` lock for `gc -a` to prevent those from overlapping.  I haven't verified it yet but I suspect that overlapping `gc -a` processes may be the cause of those updating locks getting deleted while they're still in use.

DrDaveD:
Make sure only one `cvmfs_server gc -a` command is running at a time.
- Fixes #3574

Also, when comparing this behavior to `cvmfs_server check -a` I noticed that had inadvertently added a requirement that the command be run by root even if the repositories were owned by another user.  This removes that requirement as well.  For backward compatibility it checks for the old lock file as well, but without attempting to acquire it since it may not have the permission.  So there's a slight chance that the old lock file will exist without the corresponding active process still existing, but it should only happen at most once on a given machine and be found & remedied pretty easily on investigation.

### DIFF
--- a/cvmfs/server/cvmfs_server_check.sh
+++ b/cvmfs/server/cvmfs_server_check.sh
@@ -217,9 +217,18 @@ __do_all_checks() {
   fi
   [ -w /var/log/cvmfs ] || die "cannot write to /var/log/cvmfs"
 
-  local check_lock=/var/spool/cvmfs/is_checking_all
-  if ! acquire_lock $check_lock; then
-    to_syslog "skipping start of cvmfs_server check because $check_lock held by active process"
+  # This check is for backward compatibility, can eventually be removed
+  local old_check_all_lock=/var/spool/cvmfs/is_checking_all
+  if [ -f $old_check_all_lock ]; then
+    to_syslog "skipping starting cvmfs_server check -a because old $old_check_all_lock still exists"
+    return 1
+  fi
+
+  # Use /dev/shm for the lock because it is world-writable and goes away
+  # after reboot
+  local check_all_lock=/dev/shm/cvmfs_is_checking_all
+  if ! acquire_lock $check_all_lock; then
+    to_syslog "skipping starting cvmfs_server check -a because $check_all_lock held by active process"
     return 1
   fi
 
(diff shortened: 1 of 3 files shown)


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

DrDaveD:
I force-pushed after rebasing on the current cvmfs-2.11 branch

vvolkl:
@cernvm-bot cloudtest

vvolkl:
There's a unittest failure, which has been fixed in devel already. Will cherry-pick the fix and restart.

vvolkl:
@cernvm-bot help

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot unittest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
Remaining test failures seem due to flaky test. Merging and starting full cloudtests

DrDaveD:
This was merged into 2.11.3 but didn't get pulled forward to devel and so was not in 2.12.  Restored in [994f035](https://github.com/cvmfs/cvmfs/commit/994f035406047e8f2149b3ce3e9f8e55e2377f73) for 2.13.2.  See #3914.

Issue #3007, opened 2022-07-21
Overhaul signature manager


jblomer:
The signature manager could use a general code overhaul. Things that come to mind:
  - Remove `Init()`, `Fini()`, use a `UniquePtr<SignatureManager> Create()` factory method instead
  - Consistently use strings as containers for small memory buffers
  - Use enums to distinguish the different signing methods
  - Remove unused PKCS7 bits

PR #3576, opened 2024-04-11
Use libnettle as low-level crypto library

### LINKED ISSUE #3555: `libressl` sha1 performance regression vs `openssl`.
Looks like the sha1 code in `libressl` uses SSSE3 instructions, but not the newer SHA1 instruction extensions, which are used by `openssl`. This is showing up as a performance regression with 2.11 vs 2.8.

jblomer:
Changes from OpenSSL/LibreSSL to libnettle for libcvmfs_crypto. Libnettle is taken from the system on Linux and vendored for macOS. This PR replaces hashing and symmetric crypto.

For the public-key crypto, it is likely best to still depend on OpenSSL for the I/O of (PEM-)keys and X.509 certificates. For these operations, we can take stock OpenSSL (does not require LibreSSL). The actual RSA operations can also move to nettle.

Add micro benchmarks for RipeMD160 and SHAKE-128 and HMAC-SHA256.

Future improvements:
- Use libnettle for public-key crypto in signature manager; depends on #3007
- Use libnettle  for SHAKE-128  (needs libnettle 3.10+)

Based on #3558 (needs to be merged first for the CI to work)

Fixes #3555


vvolkl:
Rebased, and updated to use the same CMake FetchContent machinery as for all the other vendored externals. Previous state of the branch is at https://github.com/cvmfs/cvmfs/tree/archive/crypto-nettle

This PR now essentially rewrites the build system of libnettle in CMake. This is definitely somewhat brutal. A simpler alternative would be to keep using autotools to build with `ExternalProject_Add`, this would look something like:

```cmake
ExternalProject_Add(mylib_build
  SOURCE_DIR        ${mylib_SOURCE_DIR}
  CONFIGURE_COMMAND ${mylib_SOURCE_DIR}/configure
                       --prefix=${MYLIB_INSTALL_DIR}
                       --disable-shared
                       --enable-static
  BUILD_COMMAND     make -j${CMAKE_BUILD_PARALLEL_LEVEL}
  INSTALL_COMMAND   make install
  BUILD_IN_SOURCE   FALSE
  BUILD_BYPRODUCTS  ${MYLIB_INSTALL_DIR}/lib/libmylib.a
  ```

But I opted for the current approach because then the buildsystem for the externals is very uniform, and makes it easier to do  cross-compilations and other special builds by just setting CMake variables.

I've updated already to libnettle v3.10, which adds SHAKE-128 support, but I think the cvmfs crypto code still needs to be updated to use it (to be done in a future pull request). The packaging situation for libnettle is quite good, out of our targeted platforms I think only RHEL8 has a libnettle older than 3.10. I expect that most builds will actually use the system package.

I've also already updated the crypto code with libnettle 4.x compatibility.

PR #3587, opened 2024-04-26
FUSE-T support for macOS

YBelikov:
The PR contains added FUSE-T support for macOS client.
- Updated FindOSXFuse so it searches for FUSE-T
- Updated cvmfs_config so chksetup seeks for FUSE-T related files
- Updated mount_cvmfs.cc runtime check accrodingly
- Updated duplex_fuse.h since on macOS fuse.h file ends up being presented in /usr/local/include/fuse
- Remove requirements for macFUSE bundle from packaging configuration file


vvolkl (changes requested):
So as discussed, one issue here is that Fuse-t (wrongly?) installs its fuse.h header in /usr/local/include/fuse (where OSXFuse installs to /usr/local/include).

On a system which has both OSXFuse and Fuse-t installed (which should theoretically still work for CVMFS), the FindOSXFuse.cmake will pick up a different include directory, causing a build error for subsequent includes of "fuse/fuse_lowlevel.h" et. al.

I think it may even be OK for us to hardcode the include directory (or use a MACOS_FUSE_INCLUDE_DIRS with plural)

cmake/Modules/FindOSXFuse.cmake
```diff
@@ -1,43 +1,43 @@
...
+# MACOS_FUSE_FOUND       - True if either FUSE-T is found.
 
 # find includes
-FIND_PATH (OSXFUSE_INCLUDE_DIR fuse.h
-        /usr/local/include
+# NO_DEFAULT_PATH to avoid collisions with /Library/Frameworks/fuse_t.framework/Headers
+FIND_PATH (MACOS_FUSE_INCLUDE_DIR fuse.h
         /usr/include
-        /usr/local/include/osxfuse
+        /usr/local/include
```
vvolkl:
```suggestion
```
To avoid collision with MacFuse

vvolkl:
@cernvm-bot cloudtest

mount/mount.cvmfs.cc
```diff
@@ -68,21 +68,34 @@ static string MkFqrn(const string &repository) {
...
+  bool is_fuse_t_installed = IsFuseTInstalled();
+  if (!is_fuse_t_installed) {
+    LogCvmfs(kLogCvmfs, kLogStderr, "FUSE-T installation check failed"); 
   }
-  fuse_device = "/dev/macfuse0";
+  return is_fuse_t_installed;
 #else
   fuse_device = "/dev/fuse";
 #endif
+  string fuse_device;
```
vvolkl, line 98:
```suggestion
  string fuse_device;
```
For Linux, I think you need to move this declaration back up

vvolkl:
@cernvm-bot cloudtest

vvolkl:
`cvmfs_config probe` is still failing.  It seems because with fuse-t, `df` reports 
```
fuse-t:/cvmfs-config.cern.ch   20Gi  134Ki   20Gi     1%     644           0  100%   /Users/Shared/cvmfs/cvmfs-config.cern.ch
```
instead of 
```
cvmfs2   20Gi  134Ki   20Gi     1%     644           0  100%   /Users/Shared/cvmfs/cvmfs-config.cern.ch
```
I think we'll just need to update the check.

vvolkl:
@cernvm-bot cloudtest

YBelikov:
@vvolkl I see an issue with cvmfs_config stat <fqrn>.  It relies on macfuse filesystem in the internal command` mount -t <fs type>`, however, repositories relying on fuse-t are mounted as nfs type. I will update list_mounts() in cvmfs_config.sh, so it uses fuse_fs_type=nfs on macos and greps based on fuse-t regex (instead of cvmfs2.

YBelikov:
@cernvm-bot unittest

YBelikov:
> So as discussed, one issue here is that Fuse-t (wrongly?) installs its fuse.h header in /usr/local/include/fuse (where OSXFuse installs to /usr/local/include).
> 
> On a system which has both OSXFuse and Fuse-t installed (which should theoretically still work for CVMFS), the FindOSXFuse.cmake will pick up a different include directory, causing a build error for subsequent includes of "fuse/fuse_lowlevel.h" et. al.
> 
> I think it may even be OK for us to hardcode the include directory (or use a MACOS_FUSE_INCLUDE_DIRS with plural)

Updated with hardcoded path to /usr/local/include

YBelikov:
@vvolkl looks like there is another issue: executing `umount <fqrn>` leads to the following error: 
`umount(/Users/Shared/cvmfs/cvmfs-config.cern.ch): Resource busy -- try 'diskutil unmount'`. Gonna take a look at this closer. That's the reason why at least part of integration tests failed.

vvolkl:
:tada: The reduced suite of cloudtests is passing on GH actions!

@YBelikov  Can you comment here the list of currently excluded tests that are expected to pass with OSXFuse?

YBelikov:
@vvolkl here is the table of tests I have additionally excluded
| Test number | Error |
| ---- | ------ |
|   11  |   CURRENTLY FIXED. The issue in get_xattr returning a list of all xattr in the output; retval 20; cvmfs_mount fails for currently obscure reason 
|   15  |   retval 31; an issue with cache rebuild, apparently cachedb is not recreated on the final mount. PosixQuotaManager is not able to determine the presence of a crash guard. Also platform_readdir call returns only "." and ".." entries for all subdirectories in /var/lib/cvmfs/shared/ * |
|   20  |   CURRENTLY FIXED. retval 3; fails on open_http_server (line 46)  | 
|   28  |   retval 3;  Looks like an issue with NFS caching. The reply with error for the first dir lookup gets cached by NFS and this info persists for a time window (~20min). The subsequent attempt to do a lookup for the same directory fail without invocation of cvmfs_lookup;  seems like a system holds a cache and doesn't let this request to pass to our FS. Purging kernel cache doesn't work as well as attempts to mount FS with no-caching options | 
|   30  |   CURRENTLY FIXED. timeout; message: "mounting took too long with"; cvmfs_mount with FUSE-T requires more time to complete then predefined maximal timeout | 
|   32  |   CURRENTLY FIXED retval 10; getxattr pid and cvmfs_talk -i <repo> pid return different values due to the issue with get_xattr output    |
|   37  |   CURRENTLY FIXED issue with return status on cvmfs_mount try_automount; the test acquired status of "sleep".   | 
|   53  |   retval 20; subsequent mount after cvmfs_config wipecache trashes  cvmfs-config.cern.ch repository making public  key retrieval imposible for grid.cern.ch mount. Appears due to repos reload performed on "wipecache"|
|   60  |   NOT SUPPORTED. retval 30;  hidden attributes probably not supported   | 
|   66  |   CURRENTLY FIXED. retval 20;   sudo cvmfs_config killall (line 22) fails; no consistency, failed on CI, passed on a local machine launch |
|   69  |   retval 30;  double mount reports success which is not an expected result |
|   72  |   CURRENTLY FIXED. spits a lot of readonly filesystem errors; seems like a lack of sleep statement after mount (line 35). Indeed putting sleep for macOS works  | 
|   78  | retval 13; nioerr != 1 when remount error should fail (line 70) |
|   87  | retval 6; get_xattr issue; user.chunks attribute != 2 (line 87) |
|   90  | retval 30; get_xattr or cvmfs talk chroot issue; according to test tag_root != new_root (line 24, putting sleep before get_xattr doesn't help) |
|   92  | retval 32;  hitrate_3rd = 0 which is not expected) | 
|   96  | currently excluded |  
|   97  | excluded due to timeout, the reason is unclear,  probably CVMFS_KCACHE_TIMEOUT issue  | 
|   104  | test passes on CI but makes GitHub Actions stuck, the reason is unclear |

cvmfs/cvmfs_config
```diff
@@ -1256,7 +1264,11 @@ cvmfs_probe() {
         cvmfs_readconfig $fqrn || die "Failed to read CernVM-FS configuration"
         echo -n "Probing $CVMFS_MOUNT_DIR/$fqrn... "
         ls "$CVMFS_MOUNT_DIR/$fqrn" > /dev/null 2>&1  # workaround for FC28+
-        df -P "$CVMFS_MOUNT_DIR/$fqrn" 2>&1 | grep -q ^cvmfs2
+        if [ "$sys_arch" = "Darwin" ] && ! is_macfuse_enabled; then
+          df -P "$CVMFS_MOUNT_DIR/$fqrn" 2>&1 | grep -q ^fuse-t
```
vvolkl: It seems like this df output is not always consistent with fuse-t ( depending on how you install it ), this may be more robust
```suggestion
          df -P "$CVMFS_MOUNT_DIR/$fqrn" 2>&1 | grep -q $fqrn
```

.github/workflows/ci_macos.yml
```diff
@@ -0,0 +1,137 @@
...
+          which timeout
+          which greadlink
+          sudo mkdir /usr/local/include
+          ls /usr/local
+          
+      - name: Install Fuse-t and dependencies
+        run: |
+          brew install python-setuptools
+          wget -q https://github.com/macos-fuse-t/fuse-t/releases/download/1.0.41/fuse-t-macos-installer-1.0.41.pkg
+          sudo installer -pkg fuse-t-macos-installer-1.0.41.pkg -target /
```
vvolkl:
```suggestion
          wget -q https://github.com/macos-fuse-t/fuse-t/releases/download/1.0.43/fuse-t-macos-installer-1.0.43.pkg
          sudo installer -pkg fuse-t-macos-installer-1.0.43.pkg -target /
```
Switch to latest version

vvolkl:
I think we can use an ifdef to send an error reply in cvmfs_lookup, while waiting for the fix for https://github.com/macos-fuse-t/fuse-t/issues/72

.github/workflows/ci_macos.yml
```diff
@@ -0,0 +1,137 @@
...
+          which greadlink
+          sudo mkdir /usr/local/include
+          ls /usr/local
+          
+      - name: Install Fuse-t and dependencies
+        run: |
+          brew install python-setuptools
+          wget -q https://github.com/macos-fuse-t/fuse-t/releases/download/1.0.43/fuse-t-macos-installer-1.0.43.pkg
+          sudo installer -pkg fuse-t-macos-installer-1.0.43.pkg -target /
+          ls /usr/local/include
```
vvolkl:
```suggestion
          wget -q https://github.com/macos-fuse-t/fuse-t/releases/download/1.0.44/fuse-t-macos-installer-1.0.44.pkg
          sudo installer -pkg fuse-t-macos-installer-1.0.44.pkg -target /
          ls /usr/local/include
```
Fix for negative lookup reply is here

cvmfs/cvmfs_config
YBelikov: Done

YBelikov:
> I think we can use an ifdef to send an error reply in cvmfs_lookup, while waiting for the fix for [macos-fuse-t/fuse-t#72](https://github.com/macos-fuse-t/fuse-t/issues/72)

Since the patch was merged we can resolve this

.github/workflows/ci_macos.yml
```diff
@@ -0,0 +1,137 @@
...
+          export CVMFS_TEST_PROXY=DIRECT
+          cd $GITHUB_WORKSPACE/test
+          ./run.sh /tmp/cvmfs-client-test.log -o /tmp/cvmfs-client-test.xunit.xml \
+            -s quick                                          \
+            -x src/004-davinci                                \
+            src/005-asetup                                    \
+            src/006-buildkernel                               \
+            src/007-testjobs                                  \
+            src/008-default_domain                            \
+            src/016-dnsunreachable                            \
```
vvolkl, line 84:
```suggestion
            src/015-rebuild_on_crash                           \
            src/016-dnsunreachable                            \
```
Let's move test 15 to an issue

.github/workflows/ci_macos.yml
YBelikov: Now it is 1.0.44
I resolve this sub-thread

YBelikov:
> So as discussed, one issue here is that Fuse-t (wrongly?) installs its fuse.h header in /usr/local/include/fuse (where OSXFuse installs to /usr/local/include).
> 
> On a system which has both OSXFuse and Fuse-t installed (which should theoretically still work for CVMFS), the FindOSXFuse.cmake will pick up a different include directory, causing a build error for subsequent includes of "fuse/fuse_lowlevel.h" et. al.
> 
> I think it may even be OK for us to hardcode the include directory (or use a MACOS_FUSE_INCLUDE_DIRS with plural)

I guess we can resolve this one also, since the required changes are already in the branch

vvolkl (approved):
Thanks for all your hard work on this one, Yuriy! I think all remaining issues are well understood and not critical, and this is a massive improvement in terms of convenience.

PR #3588, opened 2024-04-26
Close page cache tracker entry if cvmfs_open() fails

mharvey-jt:
Ref https://github.com/cvmfs/cvmfs/issues/3520


jblomer:
Thank you very much! This is a good catch!

jblomer:
I think we do not need the fences. It is true that we have them in the "forget" callbacks -- but I think we should remove them there, too, together with the unnecessary call to `CatalogManager::MangleInode()`. We don't fence closing the inode in the page cache tracker in `cvmfs_release` either.

vvolkl:
@cernvm-bot cloudtest

PR #3604, opened 2024-05-22
External file direct download (WIP)

rezan:
This is the external file direct download patch. How it works is if `CVMFS_EXTERNAL_DIRECT` is enabled and an external file is opened and read, CVMFS will bypass the local cache and read directly from the external HTTP proxy.

This is still a work in progress, whats left todo:

- [x] Finish Fuse functionality
- [ ] Align proxy selection logic
- [ ] Add functionality to libcvmfs
- [ ] Test cases
- [ ] Review and feedback

Feel free to review and give feedback at any point during this process.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -77,6 +77,7 @@ if (BUILD_CVMFS OR BUILD_LIBCVMFS)
        manifest_fetch.cc
        monitor.cc
        mountpoint.cc
+       network/direct_download.cc
        network/dns.cc
        network/download.cc
        network/jobinfo.cc
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -96,6 +96,7 @@
 #include "monitor.h"
 #include "mountpoint.h"
 #include "network/download.h"
+#include "network/direct_download.h"
 #include "nfs_maps.h"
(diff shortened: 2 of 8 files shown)


rezan:
Some questions and comments before moving on as I believe the core Fuse functionality is done:

1. Im using a test bit in the `fuse_file_info.fh` to signal we want a direct proxy read (bit 61). Any concerns?
   https://github.com/cvmfs/cvmfs/pull/3604/files#diff-82b206f6673026f1cbb4ccd619f623288c6a83a87accdfeca9ee9c5e23b6c971R1192
2. Still need to come up with a solution for the proxy selection when doing a retry. Right now this logic is buried in `download.cc`. This will require a bit more work.
3. If you notice, I did not re-use much of the libcurl logic in `download.cc`. My longterm goal here is to get `DirectDownload` to splice HTTP data directly into Fuse and for this to work, we cannot use libcurl. So I avoided any libcurl integration between the 2 interfaces. I guess this PR would be a middle ground for the time being until the splice logic is ready, which I'm planning for later this year.

rezan:
Started testing this a bit, I believe I found the source of a bug that was mentioned. The DirectDownload bit was conflicting with the chunked file bits. The fix was just to use the abs_fd when checking for a DirectDownload.

PR #3607, opened 2024-05-29
remove external package ipaddress

### LINKED ISSUE #3599: Small bug in cvmfs/externals/ipaddress/ipaddress-1.0.22.tar.gz
`ipaddress.py` from `cvmfs/externals/ipaddress/ipaddress-1.0.22.tar.gz` has a small bug. Line 1106 has:
```
raise TypeError("%s and %s are not of the same version" (a, b))
```
but it should be:
```
raise TypeError("%s and %s are not of the same version"  % (a, b))
```
This bug causes a Python error when the line is executed.

Should the script in the tar archive be fixed? In which case, what version number should be used for the tar archive?

On the other hand, there is a newer version of `ipaddress` available here:
https://pypi.org/project/ipaddress/#description
The latest is 1.0.23, which does fix this bug. However, it has other changes and enhancements, which might cause unexpected and undesired effects.
How should this bug be fixed?

vvolkl:
This package is a backport of the ipaddress module of the python standard library for python 2 and < 3.3. Since there is a typo in version 1.0.22 on line 1106:
```
raise TypeError("%s and %s are not of the same version" (a, b))
```
it seems preferable to try to remove this external package altogether - there should be no platform anymore that needs it.

This implies that python 2 support is dropped ( which we still had through use of compatibility libraries).

Fixes #3599


vvolkl:
@cernvm-bot cloudtest

vvolkl:
Seems to be OK, save for one flaky tests.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

Issue #3611, opened 2024-06-05
Mount stuck in ReadHalfPIpe


DrDaveD:
At aglt2 with cvmfs-2.11.2 there was a case of a mount getting stuck.  I got a bugreport (after having to kill a few processes that got stuck), a stacktrace of the stuck cvmfs2 process, and an lsof of the stuck process.

The processes were:
```
root     17170 25302  0 Jun04 ?        00:00:00 /usr/bin/mount -n -t cvmfs atlas-condb.cern.ch /cvmfs/atlas-condb.cern.ch
root     17173 17170  7 Jun04 ?        02:17:12 /sbin/mount.cvmfs atlas-condb.cern.ch /cvmfs/atlas-condb.cern.ch -n -o rw
cvmfs    17271 17173  0 Jun04 ?        00:00:08 /usr/bin/cvmfs2 -o rw,system_mount,fsname=cvmfs2,allow_other,grab_mountpoint,uid=403,gid=402 atlas-condb.cern.ch /cvmfs/atlas-condb.cern.ch
```

The stack trace on 17173 was:
```
#0  0x00007fe4e14d6b23 in __select_nocancel () from /lib64/libc.so.6
#1  0x00007fe4e04b3f1d in SafeSleepMs (ms=ms@entry=256)
    at /builddir/build/BUILD/cvmfs-2.11.2/cvmfs/util/posix.cc:1976
#2  0x00007fe4e04b3f86 in ReadHalfPipe (fd=5, buf=buf@entry=0xba8308,
    nbyte=nbyte@entry=8)
    at /builddir/build/BUILD/cvmfs-2.11.2/cvmfs/util/posix.cc:540
#3  0x00007fe4e0b06a6e in PosixQuotaManager::GetLimits (
    this=this@entry=0xba82c0, limit=limit@entry=0xba8308,
    cleanup_threshold=cleanup_threshold@entry=0xba8310)
    at /builddir/build/BUILD/cvmfs-2.11.2/cvmfs/quota_posix.cc:605
#4  0x00007fe4e0b0a564 in PosixQuotaManager::CreateShared (
    exe_path="/usr/bin/cvmfs2", cache_workspace=".", limit=18874368000,
    cleanup_threshold=cleanup_threshold@entry=9437184000, foreground=false)
    at /builddir/build/BUILD/cvmfs-2.11.2/cvmfs/quota_posix.cc:295
[...]

PR #3614, opened 2024-06-21
Add timeout to ReadHalfPipe

vvolkl:
Adresses issue #3611 

- [x] some more replacements ReadHalfPipe -> ManagedReadHalfPipe

### DIFF
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -291,7 +291,7 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
     const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR, 0600);
-    ssize_t result = SafeRead(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
+    const ssize_t result = SafeRead(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
     close(fd_lockfile_rw);
     if ((result < 0) || (static_cast<size_t>(result) < sizeof(new_cachemgr_pid))) {
       LogCvmfs(kLogQuota, kLogDebug | kLogSysLogError,
@@ -1961,7 +1961,8 @@ void PosixQuotaManager::ManagedReadHalfPipe(int fd, void *buf, size_t nbyte) {
     result = ReadHalfPipe(fd, buf, nbyte, timeout_ms);
     // try only as long as the cachemgr is still alive
   } while (!result && getpgid(cachemgr_pid_) >= 0);
-  if (!result)
+  if (!result) {
     PANIC(kLogStderr, "Error: quota manager could not read from cachemanager pipe");
+  }
 
 }
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -291,15 +291,15 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
     const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR, 0600);
-    int result = SafeRead(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
-    if (result < sizeof(new_cachemgr_pid)) {
+    ssize_t result = SafeRead(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
+    close(fd_lockfile_rw);
+    if (result < (ssize_t) sizeof(new_cachemgr_pid)) {
       LogCvmfs(kLogQuota, kLogDebug, "could not read cache manager pid from lockfile");
       UnlockFile(fd_lockfile);
       delete quota_mgr;
       return NULL;
     }
     quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
-    close(fd_lockfile_rw);
 
 
     LogCvmfs(kLogQuota, kLogDebug, "connected to existing cache manager pipe");
@@ -1958,5 +1958,7 @@ void PosixQuotaManager::ManagedReadHalfPipe(int fd, void *buf, size_t nbyte) {
     result = ReadHalfPipe(fd, buf, nbyte, timeout_ms);
     // try only as long as the cachemgr is still alive
   } while (!result && getpgid(cachemgr_pid_) >= 0);
+  if (!result)
+    PANIC(kLogStderr, "Error: quota manager could not read from cachemanager pipe");
 
 }
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -1950,7 +1950,7 @@ void PosixQuotaManager::UnregisterBackChannel(
 }
 
 void PosixQuotaManager::ManagedReadHalfPipe(int fd, void *buf, size_t nbyte) {
-  unsigned timeout_ms = 1000;
+  const unsigned timeout_ms = 2000;
   bool result = false;
   do {
     result = ReadHalfPipe(fd, buf, nbyte, timeout_ms);
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -1954,6 +1954,7 @@ void PosixQuotaManager::UnregisterBackChannel(
 void PosixQuotaManager::ManagedReadHalfPipe(int fd, void *buf, size_t nbyte) {
   const unsigned timeout_ms = cachemgr_pid_ ? 1000 : 0;
   bool result = false;
+  Block2Nonblock(fd);
   do {
     result = ReadHalfPipe(fd, buf, nbyte, timeout_ms);
     // try only as long as the cachemgr is still alive
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -362,7 +362,8 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   command_line.push_back(StringifyInt(pipe_handshake[0]));
   command_line.push_back(StringifyInt(limit));
   command_line.push_back(StringifyInt(cleanup_threshold));
-  command_line.push_back(StringifyInt(foreground));
+  // ignore foreground in order to reliably get pid from exec
+  command_line.push_back(StringifyInt(true)); //foreground
   command_line.push_back(StringifyInt(GetLogSyslogLevel()));
   command_line.push_back(StringifyInt(GetLogSyslogFacility()));
   command_line.push_back(GetLogDebugFile() + ":" + GetLogMicroSyslog());
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -289,7 +289,7 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
 
-    quota_mgr->StoreCacheMgrPid(quota_mgr->GetPid());
+    quota_mgr->SetCacheMgrPid(quota_mgr->GetPid());
     LogCvmfs(kLogQuota, kLogDebug, "current cache manager pid  %d", quota_mgr->GetCacheMgrPid());
 
     LogCvmfs(kLogQuota, kLogDebug, "connected to existing cache manager pipe");
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -656,7 +656,7 @@ pid_t PosixQuotaManager::GetPid() {
   cmd.command_type = kPid;
   cmd.return_pipe = pipe_pid[1];
   WritePipe(pipe_lru_[1], &cmd, sizeof(cmd));
-  ReadHalfPipe(pipe_pid[0], &result, sizeof(result));
+  ManagedReadHalfPipe(pipe_pid[0], &result, sizeof(result));
   CloseReturnPipe(pipe_pid);
   return result;
 }
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -363,39 +363,28 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   command_line.push_back(StringifyInt(limit));
   command_line.push_back(StringifyInt(cleanup_threshold));
   // do not propagate foreground in order to reliably get pid from exec
-  // instead, daemonize right in the call to ManagedExec
+  // instead, daemonize right here
   command_line.push_back(StringifyInt(true)); //foreground
   command_line.push_back(StringifyInt(GetLogSyslogLevel()));
   command_line.push_back(StringifyInt(GetLogSyslogFacility()));
   command_line.push_back(GetLogDebugFile() + ":" + GetLogMicroSyslog());
 
   set<int> preserve_filedes;
-  map<int, int> map_filedes;
-  
-  int null_read = 0;
-  int null_write = 0;
-
-  if (foreground) {
-    preserve_filedes.insert(0);
-    preserve_filedes.insert(1);
-    preserve_filedes.insert(2);
-  } else {
-    null_read = open("/dev/null", O_RDONLY);
-    null_write = open("/dev/null", O_WRONLY);
-    map_filedes[null_read] = 0;  // Reading end of pipe_stdin
-    map_filedes[null_write] = 1;  // Writing end of pipe_stdout
-    map_filedes[null_write] = 2;  // Writing end of pipe_stderr
-  }
+  preserve_filedes.insert(0);
+  preserve_filedes.insert(1);
+  preserve_filedes.insert(2);
   preserve_filedes.insert(pipe_boot[1]);
   preserve_filedes.insert(pipe_handshake[0]);
 
-
-  retval = ManagedExec(command_line, preserve_filedes, map_filedes,
-                       /*drop_credentials*/ false,
-                       /*clear_env*/ false,
-                       /*double_fork*/  true,
-                       &new_cachemgr_pid,
-                       !foreground);
+  if (foreground) {
+    retval = ManagedExec(command_line, preserve_filedes, map<int, int>(),
+                         /*drop_credentials*/ false,
+                         /*clear_env*/ false,
+                         /*double_fork*/  true,
+                         &new_cachemgr_pid);
+  } else {
+    retval = ExecAsDaemon(command_line, &new_cachemgr_pid);
+  }
   if (!retval) {
     UnlockFile(fd_lockfile);
     ClosePipe(pipe_boot);
@@ -404,8 +393,6 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
     LogCvmfs(kLogQuota, kLogDebug, "failed to start cache manager");
     return NULL;
   }
-  if (null_read) close(null_read);
-  if (null_write) close(null_write);
   LogCvmfs(kLogQuota, kLogDebug, "new cache manager pid: %d", new_cachemgr_pid);
   quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
   const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR | O_TRUNC, 0600);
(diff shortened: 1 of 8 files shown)


cvmfs/util/posix.h
```diff
@@ -91,7 +91,7 @@ CVMFS_EXPORT int ConnectTcpEndpoint(const std::string &ipv4_address,
 CVMFS_EXPORT void MakePipe(int pipe_fd[2]);
 CVMFS_EXPORT void WritePipe(int fd, const void *buf, size_t nbyte);
 CVMFS_EXPORT void ReadPipe(int fd, void *buf, size_t nbyte);
-CVMFS_EXPORT void ReadHalfPipe(int fd, void *buf, size_t nbyte);
+CVMFS_EXPORT void ReadHalfPipe(int fd, void *buf, size_t nbyte, unsigned timeout_ms=0);
```
jblomer:
```suggestion
CVMFS_EXPORT void ReadHalfPipe(int fd, void *buf, size_t nbyte, unsigned timeout_ms = 0);
```

cvmfs/util/posix.cc
```diff
@@ -522,10 +522,11 @@ void ReadPipe(int fd, void *buf, size_t nbyte) {
 /**
  * Reads from a pipe where writer's end is not yet necessarily connected
  */
-void ReadHalfPipe(int fd, void *buf, size_t nbyte) {
+void ReadHalfPipe(int fd, void *buf, size_t nbyte, unsigned timeout_ms) {
```
jblomer: I think we need to indicate if the function quit due to a timeout. Perhaps by a bool return value.

cvmfs/util/posix.cc
```diff
@@ -538,8 +539,10 @@ void ReadHalfPipe(int fd, void *buf, size_t nbyte) {
     if ((i > 3000) && (num_bytes == 0)) {
       // The BackoffThrottle would pull in too many dependencies
       SafeSleepMs(backoff_ms);
+      waittime_ms += backoff_ms;
```
jblomer: As we have various contributors to the overall time spent in this method, perhaps let's measure with `platform_monotonic_time` how much time really passed.

For consistency of this interface, we may also want to use a non-blocking fd (see `Block2Nonblock()`) in the `read()` call, which may otherwise block if the pipe is available.

cvmfs/quota_posix.cc
```diff
@@ -1922,3 +1929,19 @@ void PosixQuotaManager::UnregisterBackChannel(
     ClosePipe(back_channel);
   }
 }
+
+void PosixQuotaManager::ManagedReadHalfPipe(int fd, void *buf, size_t nbyte, pid_t cachemgr_pid) {
+  unsigned timeout_ms = 100;
```
jblomer: We probably want to be more generous and line up with the backoff in `ReadHalfTime`. Perhaps 1s?

cvmfs/quota_posix.cc
```diff
@@ -1922,3 +1929,19 @@ void PosixQuotaManager::UnregisterBackChannel(
...
+  do {
+  ReadHalfPipe(fd, buf, nbyte, timeout_ms);
```
jblomer:
```suggestion
    ReadHalfPipe(fd, buf, nbyte, timeout_ms);
```
Also, we need to return if `ReadHalfPipe` was successful.

cvmfs/quota_posix.h
```diff
@@ -84,6 +84,11 @@ class PosixQuotaManager : public QuotaManager {
   virtual pid_t GetPid();
   virtual uint32_t GetProtocolRevision();
 
+  void ManagedReadHalfPipe(int fd, void *buf, size_t nbyte, pid_t cachemgr_pid);
+  void StoreCacheMgrPid(pid_t pid_);
+  pid_t GetCacheMgrPid();
```
jblomer: I don't think we need the getter and setter. If kept, let's inline and adjust to the style like
```suggestion
  void SetCacheMgrPid(pid_t pid_) { ... }
  pid_t cache_mgr_pid() const { return ... }
```

cvmfs/quota_posix.cc
```diff
@@ -595,14 +601,15 @@ uint64_t PosixQuotaManager::GetCapacity() {
 
 void PosixQuotaManager::GetLimits(uint64_t *limit, uint64_t *cleanup_threshold)
 {
+
```
jblomer:
```suggestion
```

cvmfs/quota_posix.h
```diff
@@ -311,6 +316,13 @@ class PosixQuotaManager : public QuotaManager {
    */
   bool async_delete_;
 
+
+  /**
+   * Record pid of current cache manager in order to check if its process
+   * disappeared.
+   */
+  pid_t cachemgr_pid_;
```
jblomer, line 329: Should be initialized in constructor

cvmfs/quota_posix.cc
```diff
@@ -288,6 +288,10 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   LogCvmfs(kLogQuota, kLogDebug, "trying to connect to existing pipe");
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
+
+    quota_mgr->StoreCacheMgrPid(quota_mgr->GetPid());
```
jblomer: This may be the crux of the approach: the `GetPid()` uses itself `ReadHalfPipe()`. For the fuse processes that connect to an existing cache manager, we probably want to find a different (reliable) way to get their pid.  Let's discuss.

cvmfs/quota_posix.cc
```diff
@@ -358,8 +362,8 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   preserve_filedes.insert(2);
   preserve_filedes.insert(pipe_boot[1]);
   preserve_filedes.insert(pipe_handshake[0]);
-
-  retval = ManagedExec(command_line, preserve_filedes, map<int, int>(), false);
+  pid_t newcachemgr_pid;
+  retval = ManagedExec(command_line, preserve_filedes, map<int, int>(), false, false, true, &newcachemgr_pid);
```
jblomer: Probably it would be nicer to have a config struct as an input to that function, so that all the parameters are named. In the meantime, we can annotate with comments the names of at least the boolean parameters.

jblomer (changes requested):
Additionally, for `Create()` (non-shared), we also need to store the pid.

cvmfs/quota_posix.cc
vvolkl: Yes indeed, however this already reduces the number of possibly blocking calls to ReadHalfPipe to one, and here also a (suitably long) timeout could be introduced.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
> Additionally, for `Create()` (non-shared), we also need to store the pid.

@jblomer Hmmm, which pid? `Create()`(non-shared) does not create a new process, right?

cvmfs/util/posix.cc
```diff
@@ -540,8 +543,11 @@ void ReadHalfPipe(int fd, void *buf, size_t nbyte) {
       SafeSleepMs(backoff_ms);
       if (backoff_ms < max_backoff_ms) backoff_ms *= 2;
     }
+
+  duration_ms = (platform_monotonic_time_ns() - timestamp) / 1000;
+  if (timeout_ms != 0 && duration_ms  > timeout_ms) return false;
```
jblomer: Indentation issue

jblomer: Maybe only call `platform_monotonic_time_ns()` if a timeout is used.

cvmfs/util/posix.cc
```diff
@@ -540,8 +543,11 @@ void ReadHalfPipe(int fd, void *buf, size_t nbyte) {
...
   } while (num_bytes == 0);
-  assert((num_bytes >= 0) && (static_cast<size_t>(num_bytes) == nbyte));
+  return ((num_bytes >= 0) && (static_cast<size_t>(num_bytes) == nbyte));
```
jblomer: I think we should leave the assert. At this point, we should really have all the bytes read.

test/unittests/t_util.cc
```diff
@@ -564,7 +564,7 @@ TEST_F(T_Util, ReadHalfPipe) {
 
   EXPECT_EQ(0,
     memcmp(const_cast<char *>(to_write.data()), buffer_output, size));
-  ASSERT_DEATH(ReadHalfPipe(-1, buffer_output, to_write.length()), ".*");
+  //ASSERT_DEATH(ReadHalfPipe(-1, buffer_output, to_write.length()), ".*");
```
jblomer: Let's restore, if we conclude that the assert in the function should remain.

cvmfs/quota_posix.cc
```diff
@@ -1922,3 +1948,13 @@ void PosixQuotaManager::UnregisterBackChannel(
     ClosePipe(back_channel);
   }
 }
+
+void PosixQuotaManager::ManagedReadHalfPipe(int fd, void *buf, size_t nbyte) {
+  const unsigned timeout_ms = cachemgr_pid_?  1000 : 0;
```
jblomer:
```suggestion
  const unsigned timeout_ms = cachemgr_pid_ ? 1000 : 0;
```

cvmfs/quota_posix.cc
```diff
@@ -1922,3 +1948,13 @@ void PosixQuotaManager::UnregisterBackChannel(
...
+  bool result = false;
+  do {
+    result = ReadHalfPipe(fd, buf, nbyte, timeout_ms);
+    // try only as long as the cachemgr is still alive
+  } while (!result && getpgid(cachemgr_pid_) >= 0);
+
```
jblomer, line 1987: If result is still false, we need to `PANIC`.

cvmfs/quota_posix.cc
```diff
@@ -288,6 +290,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   LogCvmfs(kLogQuota, kLogDebug, "trying to connect to existing pipe");
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
+    const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR, 0600);
+    int result = read(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
```
jblomer: We should use the `SafeRead` wrapper here

cvmfs/quota_posix.cc
```diff
@@ -288,6 +290,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
...
+    quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
+    if (result == 0) {
```
jblomer:
```suggestion
    if (result < sizeof(new_cachemgr_pid)) {
```

cvmfs/quota_posix.cc
```diff
@@ -288,6 +290,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   LogCvmfs(kLogQuota, kLogDebug, "trying to connect to existing pipe");
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
+    const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR, 0600);
+    int result = read(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
+    quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
```
jblomer: Nit: move below `if` block?

cvmfs/quota_posix.cc
```diff
@@ -288,6 +290,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
...
+    if (result == 0) {
+      LogCvmfs(kLogQuota, kLogDebug, "could not read cache manager pid from lockfile");
+      return NULL;
+    }
+    close(fd_lockfile_rw);
```
jblomer, line 297: Move up so that we don't leak the file descriptor in the error branch

cvmfs/quota_posix.cc
```diff
@@ -288,6 +290,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   LogCvmfs(kLogQuota, kLogDebug, "trying to connect to existing pipe");
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
+    const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR, 0600);
+    int result = read(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
+    quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
+    if (result == 0) {
+      LogCvmfs(kLogQuota, kLogDebug, "could not read cache manager pid from lockfile");
+      return NULL;
```
jblomer:
```suggestion
      UnlockFile(fd_lockfile);
      delete quota_mgr;
      return NULL;
```

cvmfs/quota_posix.cc
```diff
@@ -368,7 +383,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
     LogCvmfs(kLogQuota, kLogDebug, "failed to start cache manager");
     return NULL;
   }
+  LogCvmfs(kLogQuota, kLogDebug, "new cache manager pid: %d", new_cachemgr_pid );
+  quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
+  const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR | O_TRUNC, 0600);
+  int result = write(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
```
jblomer: Use `SafeWrite` wrapper

jblomer (changes requested):
Should we now also replace the `ReadHalfPipe` in `GetPid()` by the managed one?

jblomer:
@cernvm-bot unittest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

cvmfs/util/posix.cc
```diff
@@ -540,8 +545,15 @@ void ReadHalfPipe(int fd, void *buf, size_t nbyte) {
       SafeSleepMs(backoff_ms);
       if (backoff_ms < max_backoff_ms) backoff_ms *= 2;
     }
+  if (timeout_ms != 0) {
+    duration_ms = (platform_monotonic_time_ns() - timestamp) / 1000;
+    if (duration_ms  > timeout_ms)
+        return false;
+  }
```
jblomer: Whitespacing

cvmfs/util/posix.cc
```diff
@@ -540,8 +545,15 @@ void ReadHalfPipe(int fd, void *buf, size_t nbyte) {
       SafeSleepMs(backoff_ms);
       if (backoff_ms < max_backoff_ms) backoff_ms *= 2;
     }
+  if (timeout_ms != 0) {
+    duration_ms = (platform_monotonic_time_ns() - timestamp) / 1000;
```
jblomer: This is microseconds, not milliseconds

cvmfs/util/posix.cc
```diff
@@ -540,8 +545,15 @@ void ReadHalfPipe(int fd, void *buf, size_t nbyte) {
...
+    if (duration_ms  > timeout_ms)
+        return false;
+  }
   } while (num_bytes == 0);
-  assert((num_bytes >= 0) && (static_cast<size_t>(num_bytes) == nbyte));
+  if (timeout_ms == 0)
+    assert((num_bytes >= 0) && (static_cast<size_t>(num_bytes) == nbyte));
+  return ((num_bytes >= 0) && (static_cast<size_t>(num_bytes) == nbyte));
```
jblomer: At this point, read returned with some bytes or with an unexpected error. I think both short reads and unexpected errors have nothing to do with the timeout, so they should crash in any case.
```suggestion
  assert((num_bytes >= 0) && (static_cast<size_t>(num_bytes) == nbyte));
  return true;
```

cvmfs/quota_posix.cc
```diff
@@ -368,7 +385,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
     LogCvmfs(kLogQuota, kLogDebug, "failed to start cache manager");
     return NULL;
   }
+  LogCvmfs(kLogQuota, kLogDebug, "new cache manager pid: %d", new_cachemgr_pid );
```
jblomer: Nit:
```suggestion
  LogCvmfs(kLogQuota, kLogDebug, "new cache manager pid: %d", new_cachemgr_pid);
```

cvmfs/quota_posix.cc
```diff
@@ -368,7 +385,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
...
+  quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
+  const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR | O_TRUNC, 0600);
+  int result = SafeWrite(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
+  if (result == -1) {
```
jblomer:
```suggestion
  const bool result = SafeWrite(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
  if (!result) {
```

cvmfs/quota_posix.cc
```diff
@@ -368,7 +385,16 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
...
+    LogCvmfs(kLogQuota, kLogDebug | kLogSyslogErr,
```
jblomer: I think we should also `PANIC()` because it potentially leaves other clients in an undefined state.

cvmfs/quota_posix.cc
```diff
@@ -630,7 +656,7 @@ pid_t PosixQuotaManager::GetPid() {
   cmd.command_type = kPid;
   cmd.return_pipe = pipe_pid[1];
   WritePipe(pipe_lru_[1], &cmd, sizeof(cmd));
-  ReadHalfPipe(pipe_pid[0], &result, sizeof(result));
+  ManagedReadHalfPipe(pipe_pid[0], &result, sizeof(result));
```
jblomer: Contrary to my previous comment: do we need to make a call to the cache manager at all anymore? Actually we already know the PID, don't we...

cvmfs/quota_posix.cc
```diff
@@ -288,6 +290,18 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
   LogCvmfs(kLogQuota, kLogDebug, "trying to connect to existing pipe");
   quota_mgr->pipe_lru_[1] = open(fifo_path.c_str(), O_WRONLY | O_NONBLOCK);
   if (quota_mgr->pipe_lru_[1] >= 0) {
+    const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(), O_RDWR, 0600);
+    ssize_t result = SafeRead(fd_lockfile_rw, &new_cachemgr_pid, sizeof(new_cachemgr_pid));
+    close(fd_lockfile_rw);
+    if (result < (ssize_t) sizeof(new_cachemgr_pid)) {
```
jblomer:
```suggestion
    if ((result < 0) || (static_cast<size_t>(result) < sizeof(new_cachemgr_pid))) {
```

cvmfs/quota_posix.cc
```diff
@@ -288,6 +290,18 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
...
+      LogCvmfs(kLogQuota, kLogDebug, "could not read cache manager pid from lockfile");
```
jblomer:
```suggestion
      LogCvmfs(kLogQuota, kLogDebug | kLogSyslogErr, 
                         "could not read cache manager pid from lockfile");
```

jblomer (changes requested):

cvmfs/util/posix.cc
vvolkl: Good catch, thanks! I think this also explains the test failures - these were due to the PANIC in ReadHalfPipe being triggered on some reloads.

cvmfs/quota_posix.cc
vvolkl: That's true. I'm returning the cached pid now if available, which actually should always be the case.

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

cvmfs/quota_posix.cc
vvolkl: This has revealed another issue: The pid returned from the ManagedExec call may not be the final one due to the call to Daemonize here: https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/quota_posix.cc#L980. Will have to rethink how to get this pid

vvolkl: The easiest way will  certainly be to avoid the Daemonize, i.e running in `foreground`. Since the ManagedExec  already does a `double_fork`, this should be anyway somewhat redundant, right?

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot unittest

PR #3622, opened 2024-07-02
Add support for xattrs with values up to 64k

### LINKED ISSUE #3621: Xattr length limit too small
Xattr key and value length limits are hard coded to 256

https://github.com/cvmfs/cvmfs/blob/42e04529dc8eccb52bf62b27b220aa54b660681a/cvmfs/xattr.cc#L160-L179
with a total limit on the size of 512

In some cases, this may be too small, eg to encode a long ACL.

Could you increase this? In this particular instance, to avoid having to make a change to the fuse client, we're going to removing the individual key+value length constraints and impose a single check on their combined length.

**Edit**: we can't make that change as it would require changing the serialization scheme in the catalog.

jblomer:
Adds support in a backward and somewhat forward compatible way. If only small attributes are attached to a file, serialization will use the current format understood by old readers. If one of the attributes is a large one, the list of attributes is serialized in a new format that old clients don't understand. Old clients won't crash but they will show none of the attributes.

Fixes #3621

### DIFF
--- a/cvmfs/xattr.cc
+++ b/cvmfs/xattr.cc
@@ -164,11 +164,11 @@ string XattrList::ListKeysPosix(const string &merge_with) const {
 bool XattrList::Set(const string &key, const string &value) {
   if (key.empty())
     return false;
-  if (key.length() > 256)
+  if (key.length() > 255)
     return false;
   if (key.find('\0') != string::npos)
     return false;
-  if (value.length() > 256)
+  if (value.length() >= 64 * 1024)
     return false;
 
   const map<string, string>::iterator iter = xattrs_.find(key);
--- a/cvmfs/xattr.cc
+++ b/cvmfs/xattr.cc
@@ -20,7 +20,8 @@
 
 using namespace std;  // NOLINT
 
-const uint8_t XattrList::kVersion = 1;
+const uint8_t XattrList::kVersionSmall = 1;
+const uint8_t XattrList::kVersionBig = 2;  // As of cvmfs 2.12
 
 /**
  * Converts all the extended attributes of path into a XattrList.  Attributes
@@ -76,7 +77,7 @@ XattrList *XattrList::Deserialize(const unsigned char *inbuf,
     return NULL;
   XattrHeader header;
   memcpy(&header, inbuf, sizeof(header));
-  if (header.version != kVersion)
+  if (!IsSupportedVersion(header.version))
     return NULL;
   unsigned pos = sizeof(header);
   for (unsigned i = 0; i < header.num_xattrs; ++i) {
--- a/cvmfs/xattr.cc
+++ b/cvmfs/xattr.cc
@@ -53,15 +53,27 @@ XattrList *XattrList::CreateFromFile(const std::string &path) {
 
   // Retrieve extended attribute values
   XattrList *result = new XattrList();
-  char value[256];
+  char value_smallbuf[255];
   for (unsigned i = 0; i < keys.size(); ++i) {
     if (keys[i].empty())
       continue;
-    const ssize_t sz_value = platform_lgetxattr(path.c_str(), keys[i].c_str(),
-                                                value, 256);
-    if (sz_value < 0)
-      continue;
-    result->Set(keys[i], string(value, sz_value));
+
+    char *buffer = value_smallbuf;
+    size_t sz_buffer = 255;
+    ssize_t sz_value = platform_lgetxattr(path.c_str(), keys[i].c_str(), buffer,
+                                          sz_buffer);
+    while ((sz_value < 0) && (errno == ERANGE)) {
+      if (buffer != value_smallbuf)
+        free(buffer);
+      sz_buffer *= 2;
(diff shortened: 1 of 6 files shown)


jblomer:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl (changes requested):
I don't think this is ready yet. When I create a repository containing files with large and small xattrs with this patch, and then mount it with an older client, none of the xattrs, including the small ones, can be listed or got.

The debuglog shows:
```
unique: 8, opcode: LISTXATTR (23), nodeid: 258, insize: 48, pid: 63228
(cvmfs) cvmfs_listxattr on inode: 258, size 65536 [visibility 1]    [08-08-2024 11:18:27 CEST]
(lru) lookup inode --> dirent: 258 (miss)    [08-08-2024 11:18:27 CEST]
(catalog) looking up '/a' in catalog: ''    [08-08-2024 11:18:27 CEST]
(catalog) found entry '/a' in catalog ''    [08-08-2024 11:18:27 CEST]
(lru) insert inode --> dirent: 258 -> 'a'    [08-08-2024 11:18:27 CEST]
(lru) lookup inode --> path: 258 (miss)    [08-08-2024 11:18:27 CEST]
(cvmfs) MISS 258 - looking in inode tracker    [08-08-2024 11:18:27 CEST]
(lru) insert inode --> path 258 -> '/a'    [08-08-2024 11:18:27 CEST]
(catalog) refreshing nested catalog cache of ''    [08-08-2024 11:18:27 CEST]
(sql) successfully prepared statement 'SELECT path, sha1, size FROM nested_catalogs UNION ALL SELECT path, sha1, size FROM bind_mountpoints;'    [08-08-2024 11:18:27 CEST]
(sql) successfully prepared statement 'SELECT xattr FROM catalog WHERE (md5path_1 = :md5_1) AND (md5path_2 = :md5_2);'    [08-08-2024 11:18:27 CEST]
(catalog) corrupted xattr data    [08-08-2024 11:18:27 CEST]
   unique: 8, success, outsize: 16
```
Will take a look.

cvmfs/xattr.cc
```diff
@@ -50,15 +51,27 @@ XattrList *XattrList::CreateFromFile(const std::string &path) {
...
-      platform_lgetxattr(path.c_str(), keys[i].c_str(), value, 256);
-    if (sz_value < 0)
-      continue;
-    result->Set(keys[i], string(value, sz_value));
+
+    char *buffer = value_smallbuf;
+    size_t sz_buffer = 255;
+    ssize_t sz_value = platform_lgetxattr(path.c_str(), keys[i].c_str(), buffer,
+                                          sz_buffer);
+    while ((sz_value < 0) && (errno == ERANGE)) {
```
vvolkl: Instead of checking repeatedly for ERANGE, I think it should be possible to call lgetxattr with sz_buffer=0, which should then return the size needed for the buffer:

>        If size is specified as zero, these calls return the  current  size  of
>        the  named extended attribute (and leave value unchanged).  This can be
>        used to determine the size of the buffer that should be supplied  in  a
>        subsequent  call.   (But, bear in mind that there is a possibility that
>        the attribute value may change between the two calls,  so  that  it  is
>        still necessary to check the return status from the second call.)

 I'll try this out.

andrey-utkin: @vvolkl how about allocating 64KiB on the stack frame straight away? Untested yet: https://github.com/andrey-utkin/cvmfs/commit/25abbdbc03a9ff5aebc643aa87d60f4c482fc486

vvolkl: Possible, but that would consume more memory than necessary right?

andrey-utkin:
Yes, but it's bounded to 64 KB, not recursively, and time-based performance seems to be in higher appreciation than memory footprint austerity, across industry. We could ifdef an alternative implementation for "embedded" kind of targets :)

cvmfs/xattr.cc
vvolkl: I've implemented the additional lgetxattr call now and think I'll merge this PR like this after testing, but I'm open to accept patches with a build option to do allocation as you propose

vvolkl:
Something that still gives me pause is that with this patch  publishing a large xattr will remove all xattrs including small ones when viewing the repo with old clients. Maybe there should be a config option to force-constrain xattrs to 255 characters, if this situation needs to be avoided.

vvolkl (approved):

cvmfs/xattr.cc
andrey-utkin: > I've implemented the additional lgetxattr call now and think I'll merge this PR like this after testing, but I'm open to accept patches with a build option to do allocation as you propose

Submitted https://github.com/cvmfs/cvmfs/pull/4061

Issue #3629, opened 2024-07-08
CVMFS_STREAMING_CACHE=yes causes authorization errors and file corruption


DrDaveD:
In cvmfs-contrib/config-repo#266 we set `CVMFS_STREAMING_CACHE=yes` for all osgstorage.org and storage.igwn.org repositories when using the config-osg configuration repository.  Unfortunately, LIGO started noticing file corruption as reported in this [osg ticket](https://support.opensciencegrid.org/support/tickets/public/3b2c34dc8b2833f1a081e86e57c4fa0c0259f52baf686ba727bf3446c6df8848).

The first problem is that simply enabling this option causes authentication errors on all the authenticated storage.igwn.org repositories, for example ligo.storage.igwn.org.  I can provide a full debuglog trace if that's desired, but here's a subset of what all the connection attempts look like:
```
(curl) {header/out} GET //igwn/ligo/frames/O4/hoft_C00/H1/H-H1_HOFT_C00-140/H-H1_HOFT_C00-1401489994-2.gwf HTTP/1.1
Host: sc-cache.chtc.wisc.edu:8443
Accept: */*
Connection: Keep-Alive
User-Agent: cvmfs Fuse 2.11.3
cvmfs-info: /igwn/ligo/frames/O4/hoft_C00/H1/H-H1_HOFT_C00-140/H-H1_HOFT_C00-1401489994-2.gwf    [07-08-2024 13:48:33 CDT]
(curl) {ssldata/in} <snip>    [07-08-2024 13:48:33 CDT]
(curl) {header/in} HTTP/1.1 403 Forbidden    [07-08-2024 13:48:33 CDT]
(download) http status error code: HTTP/1.1 403 Forbidden [403]    [07-08-2024 13:48:33 CDT]
```

More concerning, despite the fact that all the external hosts are causing errors, corrupt data is provided to the user and the user is not given an error.  Here's how to reproduce that problem on a non-authenticated repository:
1.  First, make a copy of the test file to compare against with `cp /cvmfs/des.osgstorage.org/stash/kherner/load_testing/testfile.yujun /tmp/testfile.yujun.good`
2. Set `CVMFS_STREAMING_CACHE=yes` in /etc/cvmfs/config.d/des.osgstorage.org.local.
[...]

PR #3631, opened 2024-07-10
Fix streaming cache mgr with secure repos

jblomer:
Makes sure that the authentication token is used when the streaming cache manager downloads data.

Second part of the fix for #3629


jblomer:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl (approved):

Issue #3634, opened 2024-07-12
`/cvmfs/unpacked.cern.ch` unexpectedly unmounts with some commands after using the `cvmfs-snapshotter` with `nerdctl`


vvolkl:
Originally reported by @max-fatouros  in https://gitlab.cern.ch/clange/snapshotter-benchmarks/-/issues/4 :

## Issue

After pulling an image with `sudo nerdctl pull --snapshotter cvmfs-snapshotter.` The `/cvmfs/unpacked.cern.ch` directory is unmounted by either or the `sudo nerdctl rmi` or the `sudo systemctl stop cvmfs-snapshotter` commands.

### Steps to Reproduce

```bash
sudo nerdctl pull --snapshotter cvmfs-snapshotter python:3.9
```

Finally, either 

```bash
sudo nerdctl rmi 3b199ecfbdb5 # adapt hash
```

or

```bash
sudo systemctl stop cvmfs-snapshotter
```

unmounts `/cvmfs/unpacked.cern.ch` (as seen in the `journalctl` output below).

### `journalctl -f` output

In either case, I get something similar to

```bash
Jul 10 16:29:43 snapshotter-02 cvmfs2[5679]: (unpacked.cern.ch) CernVM-FS: unmounted /cvmfs/unpacked.cern.ch (unpacked.cern.ch)
```

## Additional Investigation

The snapshotter does a bind mount of the layer to the snapshots directory. A bind mount has the same device type as the directory you bind mount too, as can be seen with `findmnt`. A normal umount just removes the bind mount.

However, the snapshotter does a force unmount (equivalent to `umount -f`), which seems to actually unmount cvmfs in the repo.

This can also be reproduced without the snapshotter, just doing

```
mount --bind /cvmfs/unpacked.cern.ch /mnt/test
umount -f /mnt/test
ls /cvmfs/unpacked.cern.ch
```
Also private mount propagation for the bind mount does not change this behavior.

A fix is to just not do a force mount in the snapshotter, although this interaction of bind mounts with cvmfs should be investigated further, as this can be a source of issues.

vvolkl:
Fixed with the linded PR.

PR #3635, opened 2024-07-12
Snapshotter: fix cvmfs unmounting on `nerdctl rmi`

amalthundiyil:
What:
Replaces force unmounts in cvmfs snapshotter with normal a unmount (i.e no flags).

Why:
#3634


vvolkl (approved):
Thanks Amal!

Issue #3626, opened 2024-07-04
Race in symlink cache invalidation


mharvey-jt:
there's a race between cvmfs sending an expiry for a symlink and it beginning to return the new value. This means that a readlink() can end up retrieving the old value which then sticks around indefinitely in the kernel cache. The window seems to be a few hundred ms. Reproducer: 

compile and run:

```
#include <stdio.h>
#include <unistd.h>
#include <string.h>
void main(void){

  char buf[1024];
  char ref[1024];
  ref[0]=0;
 while(1) {
  memset(buf,0,1024); 
  ssize_t l = readlink("/cvmfs/repo/LINK", buf, 1024);
  if(l !=strlen(buf) ) {
    printf("Length mismatch %d != %d\n", l, strlen(buf));
  }
  if(strcmp(buf, ref) || strlen(buf)!=strlen(ref)) {
   printf("%s\n", buf);
   strcpy(ref,buf);
  } 
  sleep(1); // REMOVE ME
   
 }
}
```

then mutate the symlink `/cvmfs/repo/LINK`. you should see invalidation working as expected. Now remove the `sleep(1)` and try again - you should see the reported link value not change as expected. This can be further verified by instrumenting `cvmfs_readlink()` to log what is being returned.

HereThereBeDragons:
I can confirm there is odd behavior. I extended matts script to include timestamp, revision and a loop counter.
revision is read before readlink and printed if it is different to the previous one. (if wanted can share the script)

what you see below is the output when the sleep is removed and symlink caching is active. (in all other premutations of those 2 options it works correctly)

server was publishing a new revision (changing the symlink) every 13 sec, the client reloads with minimum TTL of 1min.

the observed behavior is different to the first issue #2949  when symlink caching is enabled but expire did not exist.

for #2949 new symlinks paths would be cut short if the old symlink paths were shorter.
this new issue here observes the opposite: new symlinks have the correct new symlink path length but show the old symlink path.

investigations are continued.. might be something due to the inode caching...

```
Fri Jul  5 13:51:44 2024
 0	revision 688
Fri Jul  5 13:51:44 2024
 0	/cvmfs/my2nd.test.repo/symlinktest/fileLLLLLLLLLLLLLLLLLLLL.txt

Fri Jul  5 13:52:30 2024
 2752452	revision 692
Fri Jul  5 13:52:30 2024
 2752452	/cvmfs/my2nd.test.repo/symlinktest/fileLL

[...]

PR #3636, opened 2024-07-12
fix bug: prevent symlink corruption during updating catalog revisions

HereThereBeDragons:
Issue #3626 

i tried the following:
- 1 step-lock that immediately locks at the top of `cvmfs_readlink` --> does not work, you stay forever in the sleep loop
- return `EAGAIN` --> does not work (forgot what exactly)

working solution: have a 2-step lock
- fuse_remounter requests lock, readlink can finish 1 more call and then locks/sleeps before any further call to readlink
- if no readlink is called, have a timeout of waiting for it for 100ms (see `WaitForPauseReadlink`)
- when fuse_remounter is finished, the lock is released and readlink can function correctly on the new data

note:
there might be a super tiny chance that a readlink appears between waiting for readlink to stop, but non readlink to be read and the reload. see `WaitForPauseReadlink` for more information. it seems like it gets stuck if you try to add `PauseReadlink` in case it was not triggered during waiting for it

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -823,6 +823,11 @@ static void cvmfs_getattr(fuse_req_t req, fuse_ino_t ino,
  * Reads a symlink from the catalog.  Environment variables are expanded.
  */
 static void cvmfs_readlink(fuse_req_t req, fuse_ino_t ino) {
+  while (mount_point_->cache_symlinks()
+         && fuse_remounter_->IsPausedReadlink()) {
+    SafeSleepMs(100);
+  }
+
   HighPrecisionTimer guard_timer(file_system_->hist_fs_readlink());
 
   perf::Inc(file_system_->n_fs_readlink());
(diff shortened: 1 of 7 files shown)


HereThereBeDragons:
@cernvm-bot cloudtest

jblomer:
Thank you! 
As you pointed out, there are still races, although probably smaller than before. Let's think about it a little longer.

HereThereBeDragons:
ah i forgot to add: the race can only happen once. so in theory only a single symlink can be broken, compared to all of them in the current state

HereThereBeDragons:
(we could also think about accessing a hidden "fake" symlink to invoke correctly the symlink lock?)

HereThereBeDragons:
the latest commit is the current version i would add to `devel`

it greatly reduces corruption of `readlink` when symlink caching is enabled. in addition, errors are correct within a couple hundreds of ms.

before corrupted symlinks were only correct on a new catalog update. (so minutes later)

```
Current time: 08/30/24 14:47:48.917392122 UTC
SYMLINK CHANGED:	845996	size 18	symlinktest/3ddd3e
Current time: 08/30/24 14:47:48.918172774 UTC
FILE CONTENT CHANGED:	845997		symlinktest/3ddd3eee

Current time: 08/30/24 14:47:48.918552385 UTC
SYMLINK CHANGED:	845997	size 20	symlinktest/3ddd3eee
```

it might be that these changes are enough that in production this issue is not visible anymore as before this PR the issue was only rarely sawn while the replicator could quickly reproduce it. and now the replicator only triggers it in <10% of the reloads.

PR #3646, opened 2024-07-25
ducc: enabling syncronous gc on webhook

marcoverl:
a delete image message from the harbor webhook triggers the garbage collection of that image


ducc/webhook/registry-listener.go
```diff
@@ -75,70 +75,97 @@ func ExecDucc(msg string, logfile_name string, repository_name string) {
...
-                status, exists := data["status"]
-                if exists && status == "error" {
-                        // Perform some action when "status" is "error"
-                        fmt.Printf("[DUCC conversion n.%d failed for layer %s]\n", nOfE, data["layer"])
-                        repeat = true
-                        break;
-                }
+          image_path := "/cvmfs/" + repository_name + "/" + image
+          image_manifest := "/cvmfs/" + repository_name + "/.metadata/" + image
+          _, rmErr := exec.Command("sudo", "rm", "-rf", image_path, image_manifest).Output()
```
vvolkl: Here we should either add an assert that `image_path` and `image_manifest` are really absolute paths to cvmfs - the consequences of a future bug inadvertently changing those strings in the future could be bad :)

I think it should be possible and easier to use `cvmfs_server ingest -d` instead.

marcoverl: i do agree, but when i tried the ingest -d with this file:

```
ls -l /cvmfs/unpacked.infn.it/harbor.cloud.infn.it/unpacked/almalinux:latest
lrwxrwxrwx. 1 root root 79 Sep 16 13:02 /cvmfs/unpacked.infn.it/harbor.cloud.infn.it/unpacked/almalinux:latest -> ../../.flat/fe/fe9652e6c62e5ff85af09ae7cd2cea7d9d8389baf6977ce90a12867c28ad21be

```
i got:
```
$ sudo cvmfs_server ingest -d harbor.cloud.infn.it/unpacked/almalinux:latest unpacked.infn.it
Using auto tag 'generic-2024-09-25T13:57:30Z'
Processing changes...
[WARNING] 'harbor.cloud.infn.it/unpacked/almalinux' cannot be deleted. Unrecognized file type.
[WARNING] 'latest' cannot be deleted. Unrecognized file type.
Waiting for upload of files before committing...
Committing file catalogs...
Wait for all uploads to finish
Exporting repository manifest
Statistics stored at: /var/spool/cvmfs/unpacked.infn.it/stats.db
Tagging unpacked.infn.it
Flushing file system buffers
Signing new manifest
Remounting newly created repository revision
```
and the file is still there:
```
$ ls -l /cvmfs/unpacked.infn.it/harbor.cloud.infn.it/unpacked/almalinux:latest
lrwxrwxrwx. 1 root root 79 Sep 16 13:02 /cvmfs/unpacked.infn.it/harbor.cloud.infn.it/unpacked/almalinux:latest -> ../../.flat/fe/fe9652e6c62e5ff85af09ae7cd2cea7d9d8389baf6977ce90a12867c28ad21be
```

vvolkl:
Hi @marcoverl, I've moved the deletions in their own functions and added the check there. Please do check and let me know it works, then this would be good to merge

marcoverl:
Ok i will check. Is there no hope to make work ` ingest -d` instead?

vvolkl:
> Ok i will check. Is there no hope to make work ` ingest -d` instead?

Thanks! It can certainly be made to work, but requires a patch, and introduces additional dependency constraints. I don't think it would be much better than this solution.

marcoverl:
Hi @vvolkl, i made some fixes and tested that it works, you can merge it now

PR #3652, opened 2024-08-21
Fix ReadHalfPipe for reads larger than the timeout

jblomer:


test/unittests/t_util.cc
```diff
@@ -569,6 +569,51 @@ TEST_F(T_Util, ReadHalfPipe) {
...
+  MakePipe(fd);
+
+  int size = to_write.length();
+
+  ReadHalfPipeInfo info;
+  info.fd = fd[0];
+  info.length = size;
+  info.dst = buffer_output;
+  pthread_t thread_read;
+  int retval = pthread_create(&thread_read, NULL, MainReadHalfPipe, &info);
```
vvolkl: clang-tidy
```suggestion
 const int retval = pthread_create(&thread_read, NULL, MainReadHalfPipe, &info);
```

test/unittests/t_util.cc
```diff
@@ -569,6 +569,52 @@ TEST_F(T_Util, ReadHalfPipe) {
...
+}
+
+}  // anonymous namespace
+
+TEST_F(T_Util, ReadHalfPipeTimeout) {
+  int fd[2];
+  void *buffer_output = scalloc(20, sizeof(char));
+  MakePipe(fd);
+
+  int size = to_write.length();
```
vvolkl: clang-tidy

```suggestion
  const int size = static_cast<int>(to_write.length());
```

vvolkl:
Unittests may fail due to externals, will check in devel

vvolkl (approved):

PR #3662, opened 2024-09-11
Add support for supervisord in place of systemd

### LINKED ISSUE #3661: cvmfs_server fails in kubernetes containers based on supervisord
I wanted to do some performance tests of cvmfs_server operations under kubernetes, and at least on the cluster I am using (the OSG Tiger cluster) the containers are based on supervisord instead of systemd.  The cvmfs_server add-replica command failed because of the lack of systemd, and it would be nice to be able to use it there.

DrDaveD:
If supervisorctl exists, use that instead of systemctl for managing apache in cvmfs_server.

- Fixes #3661

### DIFF
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -913,8 +913,12 @@ minpidof() {
 }
 
 
+# Return true if systemd is to be used, that is, if neither the service
+# command nor the supervisorctl command has been located.  Note that
+# the service command is only looked for if systemctl is missing, but
+# the supervisorctl command will take precedence over systemctl.
 is_systemd() {
-  [ x"$SERVICE_BIN" = x"false" ]
+  [ x"$SERVICE_BIN" = x"false" ] && [ x"$SUPERVISOR_BIN" = x"false" ]
 }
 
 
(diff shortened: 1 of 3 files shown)


vvolkl:
@cernvm-bot ok to test

vvolkl:
@cernvm-bot all

vvolkl:
@cernvm-bot cloudtest

cvmfs/server/cvmfs_server_util.sh
```diff
@@ -914,7 +914,7 @@ minpidof() {
 
 
 is_systemd() {
```
vvolkl, line 920: Is this function still named correctly?  If I understand correctly this will now  always prefer supervisord if the binary is available. Could you maybe add a comment to clarify this?

DrDaveD: Yes it is still named correctly.  It returns true if systemd is to be used.  It's really not a change in the way it worked before; previously it returned true if the service command was not available, and now it returns true if neither the service command or supervisorctl command are available, defaulting to systemd.  I will add a comment although it seems pretty obvious to me.

DrDaveD: Oh I guess it is subtly different, in that the service command is only searched for if systemctl is missing.

vvolkl (approved):
Thanks Dave!

PR #3669, opened 2024-09-24
Remove Taiwan stratum 1 from config rpm for egi and osg

DrDaveD:
Remove the ASGC Taiwan stratum 1 from the egi and opensciencegrid domains in the cvmfs-config-default rpm.

The list of stratum 1s is normally overriden by the configuration repository so this only affects cases where the configuration repository is not available.

### DIFF
--- a/ci/cvmfs/deb.sh
+++ b/ci/cvmfs/deb.sh
@@ -19,7 +19,7 @@ CVMFS_SOURCE_LOCATION="$1"
 CVMFS_RESULT_LOCATION="$2"
 CVMFS_NIGHTLY_BUILD_NUMBER="${3-0}"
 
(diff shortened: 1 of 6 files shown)


vvolkl:
@cernvm-bot cloudtest

vvolkl (changes requested):
We need a few more changes for a new config-default package, in particular a similar  update to the changelog for debian: `cvmfs/packaging/debian/config-default/changelog` and bumping the version numbers in the ci scripts:  `ci/cvmfs/rpm.sh` and ci/`cvmfs/deb.sh`. I'll go ahead and add those directly in your branch if that's ok.

DrDaveD:
Sure that's OK

vvolkl (approved):

PR #3682, opened 2024-10-24
snapshotter: add continuous benchmarking 

amalthundiyil:


.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,85 @@
...
+          VERSION="1.7.7" 
+          wget https://github.com/containerd/nerdctl/releases/download/v$VERSION/nerdctl-full-$VERSION-linux-amd64.tar.gz
+          sudo tar Cxzvvf /usr/local nerdctl-full-$VERSION-linux-amd64.tar.gz
+          sudo systemctl enable --now containerd
+
+      - name: Install cvmfs
+        run: |
+          sudo apt install lsb-release -y
+          if [ ! -f /tmp/cvmfs-release-latest_all.deb ]; then
+              wget https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest_all.deb -O /tmp/cvmfs-release-latest_all.deb
```
vvolkl:
```suggestion
              wget https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest_all.deb -O /tmp/cvmfs-release-latest_all.deb
```

We'll want to run this for every commit on devel, so I think you have to build cvmfs from source (or run the build as a separate action and move the packages as artifacts). We want to benchmark the client as well as the snapshotter

PR #3684, opened 2024-10-28
Turn CVMFS_CACHE_REFCOUNT on by default

vvolkl:


vvolkl:
@cernvm-bot unittest

cvmfs/cache_posix.h
```diff
@@ -73,7 +73,7 @@ class PosixCacheManager : public CacheManager {
     const std::string &cache_path,
     const bool alien_cache,
     const RenameWorkarounds rename_workaround = kRenameNormal,
-    const bool do_refcount = false);
+    const bool do_refcount = true);
```
HereThereBeDragons, line 76: the following lines need the same change:
 https://github.com/cvmfs/cvmfs/blob/64595763e066911ed28332dfb92bfaf7395e72de/cvmfs/cache_posix.h#L142-L143

HereThereBeDragons (changes requested):
aside from the one missing constructor, lgtm

but: we should add in the doc somewhere clearly that refcounted cache manager so far only exists for posix cache manager

cvmfs/cache_posix.h
vvolkl: Good catch, thanks!

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3687, opened 2024-11-04
client test 098 telemetry: better logging

HereThereBeDragons:
to better understand the flaky failures..

### DIFF
--- a/test/src/098-telemetry/main
+++ b/test/src/098-telemetry/main
@@ -59,11 +59,18 @@ CVMFS_INFLUX_METRIC_NAME=cvmfs_client_counters"
                       | awk '{split($0,a,"md5_path_cache.n_miss="); split(a[2],b,","); print b[1]}')
 
   # static value: 1 unique value in absolute and delta value must be 0
+  echo "Check results"
+  echo "Static value"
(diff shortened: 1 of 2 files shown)


vvolkl (approved):

test/src/098-telemetry/main
```diff
@@ -59,11 +59,18 @@ CVMFS_INFLUX_METRIC_NAME=cvmfs_client_counters"
                       | awk '{split($0,a,"md5_path_cache.n_miss="); split(a[2],b,","); print b[1]}')
 
   # static value: 1 unique value in absolute and delta value must be 0
+  echo "Check results"
+  echo "Static value"
+  echo "   1 unique value in absolute"
```
vvolkl, line 64:
```suggestion
  echo "expect 1 unique value in absolute, got $count_static_absolute"
```
Small suggestion to add the actual counts to the log  while we're adding debugging information, same below

vvolkl: Maybe it's even better to `cat $CVMFS_TEST_098_TMPFILE` on error

vvolkl: Ah missed that below, sorry!

PR #3693, opened 2024-11-11
mkfs: do not overwrite existing manifest in s3 bucket

### LINKED ISSUE #3691: `cvmfs_server mkfs`- Fail before overwriting manifest on s3
`cvmfs_server mkfs` with a local storage backend will fail when the repository exists already, but will overrwrite the manifest on an s3 backend without warning. We should also fail here, possibly adding a `--force` option to go ahead regardless.

vvolkl:
When using mkfs+s3 with a bucket already containing a repository, we will just overwrite the existing manifest, but it's very unlikely that's what the user actually wants to do. For local repositories on local storage this already fails due to the existence of the repository configuration on the local disk. This patch adds a check in mkfs. If one actually does want to overwrite the manifest, it's now necessary to delete the old one manually from the bucket.

Fixes #3691

### DIFF
--- a/cvmfs/server/cvmfs_server_mkfs.sh
+++ b/cvmfs/server/cvmfs_server_mkfs.sh
@@ -205,7 +205,9 @@ cvmfs_server_mkfs() {
   # sanity checks
   check_repository_existence $name  && die "The repository $name already exists"
   is_empty_repository_from_url $stratum0 ||
-      die "A manifest already exists at this url. Did you mean to use cvmfs_server import?"
+      die "Error: A manifest already exists at this url: $stratum0/.cvmfspublished .\n 
+          Delete it manually and retry to force the creation of a new repository in this non-empty
+          storage. \n\n If you meant to setup a new stratum0 for this existing repository, use cvmfs_server import."
   check_upstream_validity $upstream
   if [ $unionfs = "aufs" ]; then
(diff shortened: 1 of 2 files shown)


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot unittest

cvmfs/server/cvmfs_server_mkfs.sh
```diff
@@ -204,6 +204,8 @@ cvmfs_server_mkfs() {
 
   # sanity checks
   check_repository_existence $name  && die "The repository $name already exists"
+  is_empty_repository_from_url $stratum0 ||
+      die "A manifest already exists at this url. Did you mean to use cvmfs_server import?"
```
jblomer: We may want to have a `force` flag or give instructions on how to really force making a new repo in the same space (manually remove the manifest & whitelist?)

jblomer (approved):
LGTM, one point for consideration

cvmfs/server/cvmfs_server_mkfs.sh
vvolkl: Yes, I think this usecase is too rare to warrant a force flag, but I can add a note regarding instructions (just deleting the manifest should be enough)

PR #3699, opened 2024-11-18
Fix detection of supervisord in case systemd is not installed

maxnoe:


DrDaveD (changes requested):
No, this breaks cases where both supervisord and systemctl are installed.  It needs to work with supervisord whether or not systemctl is installed.

DrDaveD:
Refer also to followup discussion in #3661

maxnoe:
Right, sorry, supervisor needs to take precedence over systemd if there

PR #3707, opened 2024-11-25
Run post-install reload in a unit

### LINKED ISSUE #3654: Avoid hanging the rpm transaction with a hung reload.
Currently the  rpm contains amongst other thing:

```
%post
if [ -d /var/run/cvmfs ]; then
  /usr/bin/cvmfs_config reload
fi
```
If this hangs for what ever reason you end up with a stuck for ever rpm transaction.
If you then kill the reload you end up with a dirty transaction.

Generally advice is never hang or fail an rpm transaction on a script-let as it leaves you in a worse mess afterwards
than you would want.

Idea 1: Use a %posttrans

https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/

```
%posttrans
if [ -d /var/run/cvmfs ]; then
  /usr/bin/cvmfs_config reload
fi
```  
These run after the rpm transaction is complete obviously.
Example usage  - https://gitlab.cern.ch/linuxsupport/rpms/cern-ca-certs/-/blob/master/CERN-CA-certs.spec?ref_type=heads#L51

Idea 2: Run it async via a unit

Add a static simple unit file `/lib/systemd/system/cvmfs-reload.service` that will run async

```
[Unit]
Description=Reload any cvmfs instances.
ConditionPathExists=/var/run/cvmfs

[Service]
Type=simple
TimeoutSec=1 hour
ExecStart=/usr/bin/cvmfs_config reload
```
Better I would say to have a TimeoutSec as then it would enter a failed state rather than "run" forever.

I prefer the unit to the %posttrans as it gets it well out of rpm transaction and the log and return code will be preserved in far more useful place.

vvolkl:
Fixes #3654

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -359,6 +359,13 @@ if (BUILD_CVMFS)
         PERMISSIONS   OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ
     )
   endif ()
+  # Systemd service unit for reloads after package upgrades
+  install (
+    FILES pkg/cvmfs-reload.service
+    DESTINATION /usr/lib/systemd/system
+    PERMISSIONS   OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ
+  )
+
 endif (BUILD_CVMFS)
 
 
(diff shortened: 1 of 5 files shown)


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

Issue #3676, opened 2024-10-12
Ubuntu24.10: packaging/debian/cernvm.gpg uses untrusted public key algorithm dsa1024


wdconinc:
With the instructions at https://cernvm.cern.ch/fs/, installation on the [recently released](https://canonical.com/blog/canonical-releases-ubuntu-24-10-oracular-oriole) Ubuntu 24.10 fails since it does not accept dsa1024-signed repositories anymore ([anecdotally](https://security.stackexchange.com/questions/255550/is-dsa-1024-safe-to-sign-package-repositories-what-are-the-reasons), a collision for DSA1024 could be obtained for US$45k in 2021).

I understand that Ubuntu 24.10 is not supported; nevertheless Ubuntu users will be prompted to upgrade.

Reproducer in container:
```console
docker run --rm -it ubuntu:24.10
apt-get -q update
apt-get install -qy sudo wget
sudo apt-get install lsb-release
wget https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest_all.deb
sudo dpkg -i cvmfs-release-latest_all.deb
rm -f cvmfs-release-latest_all.deb
sudo apt-get update
```

Output:
```
Preparing to unpack cvmfs-release-latest_all.deb ...
Unpacking cvmfs-release (4.3-1) ...
Setting up cvmfs-release (4.3-1) ...
Warning: this distribution is not supported. Using Ubuntu 22.04 packages as fallback.
Hit:1 http://archive.ubuntu.com/ubuntu oracular InRelease
Hit:2 http://security.ubuntu.com/ubuntu oracular-security InRelease
Ign:3 http://cvmrepo.s3.cern.ch/cvmrepo/apt jammy-prod InRelease
Hit:4 http://archive.ubuntu.com/ubuntu oracular-updates InRelease
Hit:5 http://archive.ubuntu.com/ubuntu oracular-backports InRelease
Get:6 http://cvmrepo.s3.cern.ch/cvmrepo/apt jammy-prod Release [2316 B]
Get:7 http://cvmrepo.s3.cern.ch/cvmrepo/apt jammy-prod Release.gpg [198 B]
Ign:7 http://cvmrepo.s3.cern.ch/cvmrepo/apt jammy-prod Release.gpg
Reading package lists... Done
[...]

PR #3709, opened 2024-11-27
packaging: add stronger signing key for apt

vvolkl:


vvolkl:
Issue: https://github.com/cvmfs/cvmfs/issues/3676

Issue #3694, opened 2024-11-11
Add curl debug mode when downloads are retried


HereThereBeDragons:
Follow up to #3616

DrDaveD:
Another possibility which we discussed in today's coordination meeting was to leave curl debug mode enabled but normally do not show the output anywhere, just save it in a buffer and only use it if there is an error.  That may not have a significant performance impact.

vvolkl:
I think this may be better implemented while doing a larger overhaul and rethink of the debug logging.

PR #3718, opened 2024-12-04
Turn curl debug mode on when retrying a download

HereThereBeDragons:
Issue #3694

Turn on curl debug mode and logging it to syslog when retrying a download.
if debug mode is on it is only written to the debuglog.

i verified it manually that it works by hard coding to always go in the retry branch.
i did not find any unittest or integration test that covers this case.


vvolkl:
In the current state this will be too much syslog volume. Closing, maybe revisit in the future.

PR #3720, opened 2024-12-04
externals: fix  builds on arm/gcc14+

vvolkl:


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3726, opened 2024-12-06
Add -q option to fsck to check for zero-byte cache files quickly

vvolkl:


cvmfs/cvmfs_config
```diff
@@ -1294,9 +1294,19 @@ cvmfs_config_fsck() {
       continue
     fi
 
-    echo "CernVM-FS fsck on $dir"
-    cvmfs_fsck $@ $dir
-    retval=$(($? + $retval))
+    if [ "x$1" = "x-q" ]; then
+      # this hardcodes the hash for the (legitimately) empty file - if anything about compression or hashing changes,
+      # this needs to be updated
+      for zeroed_file in $(find $dir -empty -wholename "$dir/[a-z0-9][a-z0-9]/*" ! -name ec3d88b62ebf526e4e5a4ff6162a3aa48a6b78); do
```
jblomer, line 1300: @vvolkl Sorry for the late comment, I just noticed that the list of allowed hashes should also cover for the corresponding RMD-160 and SHAKE-128 versions.

vvolkl: Good catch! See https://github.com/cvmfs/cvmfs/pull/3739

PR #3728, opened 2024-12-11
Hotfix for cache corruption alarms

vvolkl:
This fixes a bug introduced in #3156 that can lead to empty certificate files being left in the cache, triggering fsck alarms.

Has been deployed and confirmed to fix the issue on QA nodes. Followups should be a regression test and further investigation of all code paths.

### DIFF
--- a/cvmfs/catalog_mgr_client.cc
+++ b/cvmfs/catalog_mgr_client.cc
@@ -317,12 +317,14 @@ LoadReturn ClientCatalogManager::LoadCatalogByHash(
           CacheManager::Label label;
           label.path = repo_name_;
           label.flags |= CacheManager::kLabelCertificate;
+          if (ctlg_context->manifest_ensemble()->cert_size > 0) {
           fetcher_->cache_mgr()->CommitFromMem(
                 CacheManager::LabeledObject(ctlg_context->manifest_ensemble()->
                                                         manifest->certificate(),
                                             label),
                                   ctlg_context->manifest_ensemble()->cert_buf,
                                   ctlg_context->manifest_ensemble()->cert_size);
+          }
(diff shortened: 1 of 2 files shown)


jblomer (commented):
It's plausible that it fixes the issue but I think it's not quite ideal because now, whenever we look if there is a new remote catalog, we will not only load the manifest but also the certificate and the whitelist.

I think I'd prefer to fix it when we try to write the certificate to the cache. I.e. don't write it if there is no certificate available in the ensemble.

vvolkl:
Yes indeed - I'll improve the fix. But it's good to know that this actually fixes the issue.

cvmfs/catalog_mgr_client.cc
```diff
@@ -317,12 +317,14 @@ LoadReturn ClientCatalogManager::LoadCatalogByHash(
...
           label.path = repo_name_;
           label.flags |= CacheManager::kLabelCertificate;
+          if (ctlg_context->manifest_ensemble()->cert_size > 0) {
           fetcher_->cache_mgr()->CommitFromMem(
                 CacheManager::LabeledObject(ctlg_context->manifest_ensemble()->
                                                         manifest->certificate(),
                                             label),
                                   ctlg_context->manifest_ensemble()->cert_buf,
                                   ctlg_context->manifest_ensemble()->cert_size);
+          }
```
jblomer: Whitespace issue?

jblomer (approved):
LGTM (but see style comment)

cvmfs/catalog_mgr_client.cc
HereThereBeDragons: agree, fetcher needs to be indented
otherwise lgtm to fix the issue

HereThereBeDragons (approved):

Issue #3666, opened 2024-09-19
Extending cvmfs-info header would be very useful


DrDaveD:
The frontier-squid package by default shows the contents of any `cvmfs-info` header in its log.  It would be very useful to extend that header so requests could be traced to a user job.  I looked at CVMFS_HTTP_TRACING and CVMFS_HTTP_TRACING_HEADERS but it doesn't fit well for squid to use a different header for each piece of information.  It would be especially useful to be able to include selected environment variables from the application process where job managers can set information identifying a job.

I propose a configuration parameter CVMFS_INFO_HEADER, corresponding to the existing toggle CVMFS_SEND_INFO_HEADER which enables sending the access path in the `cvmfs-info` header.  In order to avoid problems with people setting config parameters with double quotes, I would use a `%{}` syntax instead of `${}`.  I propose the following substitutions:
- `%{path}` - the path accessed by the application
- `%{pid}` - the application process id
- `%{uid}` - the application user id
- `%{gid}` - the application group id
- `%{env:VARNAME}` - a sanitized value of the environment variable VARNAME from the application process environment (from `/proc/%{pid}/environ`.  I would like it to be preceded by a blank if there is a value, or be completely empty if there is no value.  That way more than one environment variable could be chosen from without introducing extra blanks if they're missing.

If CVMFS_INFO_HEADER is set, CVMFS_SEND_INFO_HEADER will default to true.  On the other hand if CVMFS_SEND_INFO_HEADER is set true, CVMFS_INFO_HEADER will default to `%{path}`.

DrDaveD:
In order for the cvmfs fuse process to be able to read `/proc/%{pid}/environ` of any process, it will need to retain the ability to elevate its permissions to have the `CAP_DAC_READ_SEARCH` capability.

DrDaveD:
Unfortunately I found that, although necessary, `CAP_DAC_READ_SEARCH` is insufficient to read `/proc/%{pid}/environ` of any process; `CAP_SYS_PTRACE` is also required.  I consider `CAP_SYS_PTRACE` to be much more powerful, almost as powerful as root.

labkode:
[...]

PR #3730, opened 2024-12-13
Reduce client capabilities to minimum

DrDaveD:
This is an initial PR setting up for a PR fixing #3666.  It reduces the capabilities in all the client processes: the main fuse process, the cache manager process, and the watchdog/monitor processes of each.  For now this leaves CAP_DAC_READ_SEARCH capability available to the main fuse process, but in the followup PR that will be made available only when needed.

In order to avoid the fuse process to need CAP_SYS_ADMIN, this depends on the watchdog to do the unmount also in the case of the fuse process exiting on its own for some reason, for example with a kill signal.  In the normal case it's the automounter that does the unmount.

In summary, the capabilities permitted by the 4 processes after this PR are (when started as root):

1. fuse process - CAP_DAC_READ_SEARCH
2. fuse watchdog - CAP_SYS_ADMIN and CAP_DAC_READ_SEARCH.  CAP_DAC_READ_SEARCH is inheritable & ambient so only it is passed to gdb when it does a stack trace.  That capability is only needed because the fuse process has it and ptrace requires the attaching process to have at least the capabilities of the attached process.
3. cache manager - none
4. cache manager watchdog - none

The fuse process and cache manager process have their PR_SET_DUMPABLE flag enabled so they can be ptraced, but the RLIMIT_CORE set to zero so they still don't dump core.  They didn't dump core before because PR_SET_DUMPABLE was disabled because they started privileged.

For some reason the github diff does't show that `swissknife_capabilities.{h|cc}` were first renamed to `capabilities.{h|cc}`, although the git diff showed that.


DrDaveD:
@vvolkl The ci checks all pass now, please run the cloud tests.

vvolkl:
@cernvm-bot cloudtest

DrDaveD:
The Libcvmfs test (602-libcvmfs) is failing due to a lack of -lcap.  It passed on the CI/Docker test.  Is there another place where the packages installed on the cloudtest machines is defined?

DrDaveD:
One of the metalink tests failed on one machine too, which I have addressed in #3733.

jblomer:
Sorry, I realized too late that #3735 is based on this one. So several comments over there apply to this PR.

DrDaveD:
The cpp-linter errors do not seem to be relevant to the changes made in this PR, and I don't know what to do about them. There were some additional errors both relevant and not relevant until I disabled the `misc-include-cleaner` test.

DrDaveD:
@cernvm-bot cloudtest

DrDaveD:
@vvolkl The docker & mac github actions succeeded now and I still have this issue from above:

> The Libcvmfs test (602-libcvmfs) is failing due to a lack of -lcap. It passed on the CI/Docker test. Is there another place where the packages installed on the cloudtest machines is defined?

This version doesn't address yet Jakob's suggestion of starting the watchdog once at startup and not at reload time in order to reduce the privileges needed to be retained.

DrDaveD:
Closing & reopening to re-run CI checks

DrDaveD:
This has now been rebased on #3796 so that has to be merged first.

DrDaveD:
#3796 is now merged so this is again rebased on devel and ready for review @vvolkl @jblomer.

DrDaveD:
@cernvm-bot cloudtest

DrDaveD:
@cernvm-bot cloudtest

DrDaveD:
@vvolkl @jblomer Please review

DrDaveD:
Rebased on latest devel branch.

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+ */
+
+
+#include <errno.h>
+
+#include <cassert>
+
+#ifdef __APPLE__
+#include <unistd.h>
+#endif
```
jblomer, line 11: Move up (first the system includes)

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+bool ClearPermittedCapabilities(int,
+                                const cap_value_t *,
+                                int,
+                                const cap_value_t *) {
+  return true;
+}
+
+namespace {
+
+bool obtainCapability(const cap_value_t, const char *) {
```
jblomer: Style:
```suggestion
bool ObtainCapability(const cap_value_t, const char *) {
```

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+  // there are no individual capabilities on OSX so switch back to root
+  return (SwitchCredentials(0, getgid(), true));
+}
+
+bool checkCapabilityPermitted(const cap_value_t) {
```
jblomer: Style:
```suggestion
bool CheckCapabilityPermitted(const cap_value_t) {
```

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+  return (getuid() == 0);
+}
+
+} // namespace
+
+#else
+
+#include <sys/prctl.h>
```
jblomer: Move up to the preamble with includes

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+// Clear all CAP_PERMITTED capabilities except those requested.
+// This function requires being run with CAP_SETPCAP capability permitted.
+// If the real uid & gid do not match the effective uid & gid, it also
+// requires CAP_SETUID and CAP_SETGID capabilities to be permitted and
+// ends up switching the real uid & gid to match the incoming effective
+// uid & gid.  Beware that switching the uid is not thread-safe; it is
+// process-wide and clears all capabilities from threads that do not
+// have keepcaps enabled.
+// Capabilities listed in inheritcaps will be made inheritable.
+
```
jblomer:
```suggestion
```

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+// Clear all CAP_PERMITTED capabilities except those requested.
+// This function requires being run with CAP_SETPCAP capability permitted.
+// If the real uid & gid do not match the effective uid & gid, it also
+// requires CAP_SETUID and CAP_SETGID capabilities to be permitted and
+// ends up switching the real uid & gid to match the incoming effective
+// uid & gid.  Beware that switching the uid is not thread-safe; it is
+// process-wide and clears all capabilities from threads that do not
+// have keepcaps enabled.
+// Capabilities listed in inheritcaps will be made inheritable.
```
jblomer: Doxygen syntax

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+    return true;
+  }
+  
+  uid = geteuid();
+  gid = getegid();
+  if ((uid != getuid()) || (gid != getgid())) {
+    // Only do setuid & setgid when necessary because it is a process-wide
+    // setting, not a per-thread setting.
+    if (!ObtainSetuidgidCapabilities()) {
+      LogCvmfs(kLogCvmfs, kLogStderr | kLogDebug,
```
jblomer: Since this function can be executed after `Deamonize()`, we should not log to stderr but only to syslog.
```suggestion
      LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
```

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+// This function requires being run with CAP_SETPCAP capability permitted.
+// If the real uid & gid do not match the effective uid & gid, it also
+// requires CAP_SETUID and CAP_SETGID capabilities to be permitted and
+// ends up switching the real uid & gid to match the incoming effective
+// uid & gid.  Beware that switching the uid is not thread-safe; it is
+// process-wide and clears all capabilities from threads that do not
+// have keepcaps enabled.
+// Capabilities listed in inheritcaps will be made inheritable.
+
+bool ClearPermittedCapabilities(int nreservecaps,
```
jblomer: It seems that the mix of error return codes and asserts is inconsistent. We call this always wrapped in `assert()`, so I'd suggest not to use assert inside this function but to log errors and return false.

cvmfs/util/capabilities.h
```diff
@@ -0,0 +1,42 @@
...
+CVMFS_EXPORT bool ObtainSysAdminCapability();
+CVMFS_EXPORT bool ObtainSysPtraceCapability();
+CVMFS_EXPORT bool ObtainSetuidgidCapabilities();
+CVMFS_EXPORT bool ObtainSetpcapCapability();
+CVMFS_EXPORT bool SetuidCapabilityPermitted();
+CVMFS_EXPORT bool SetpcapCapabilityPermitted();
+CVMFS_EXPORT bool ClearPermittedCapabilities(int nreservecaps,
+                                             const cap_value_t *reservecaps,
+                                             int ninheritcaps,
+                                             const cap_value_t *inheritcaps);
```
jblomer: Use a `const std::vector<cap_value_t> &` in the function signature. We need the int+pointer pair for `cap_set_flag()`, but that should be taken care of by the implementation.

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+      assert(retval == 0);
+    }
+  }
+
+  return true;
+}
+
+namespace {
+
+bool obtainCapability(const cap_value_t cap, const char *capname) {
```
jblomer: Style:
```suggestion
bool ObtainCapability(const cap_value_t cap, const char *capname) {
```

cvmfs/util/capabilities.cc
```diff
@@ -0,0 +1,268 @@
...
+             "Cannot reset capabilities for current process "
+             "(errno: %d)",
+             errno);
+    return false;
+  }
+
+  return true;
+}
+
+bool checkCapabilityPermitted(const cap_value_t cap) {
```
jblomer:
```suggestion
bool CheckCapabilityPermitted(const cap_value_t cap) {
```

cvmfs/util/posix.cc
```diff
@@ -1515,6 +1521,18 @@ void GetLimitNoFile(unsigned *soft_limit, unsigned *hard_limit) {
 }
 
 
+/**
+ * Sets soft and hard limit for maximum core size
+ */
+int SetLimitCore(unsigned limit_core) {
```
jblomer: I think we only check for success/failure when we use this, so change to boolean return type.

cvmfs/util/posix.cc
```diff
@@ -802,11 +804,15 @@ bool SwitchCredentials(const uid_t uid, const gid_t gid, const bool temporarily,
...
-    // If effective uid is not root, we must first gain root access back
-    if ((getuid() == 0) && (getuid() != geteuid())) {
-      retval = SwitchCredentials(0, getgid(), true);
+    // If effective uid is not root, we probably need setuid+setgid
+    // capabilities first
+    if (geteuid() != 0) {
+      retval = ObtainSetuidgidCapabilities();
       if (!retval)
-        return false;
+        LogCvmfs(kLogCvmfs, kLogDebug,
```
jblomer, line 812: I think we also need for `avoid_mutexes` before logging.

cvmfs/auto_umount.h
```diff
@@ -10,7 +10,7 @@
 namespace auto_umount {
 
 void SetMountpoint(const std::string &mountpoint);
-void UmountOnCrash();
+void UmountOnExit(const bool crashed);
```
jblomer, line 13:
```suggestion
void UmountOnExit(bool crashed);
```

cvmfs/quota_posix.cc
```diff
@@ -1198,6 +1199,15 @@ int PosixQuotaManager::MainCacheManager(int argc, char **argv) {
   assert(watchdog.IsValid());
   watchdog->Spawn("./stacktrace.cachemgr");
 
+  if ((geteuid() != 0) && SetuidCapabilityPermitted()) {
+    // Permanently drop credentials
+    assert(ClearPermittedCapabilities(0, 0, 0, 0));
+    // Leave this process ptraceable
+    assert(platform_set_dumpable());
+    // but without core dumps
+    assert(SetLimitCore(0) == 0);
```
jblomer: Just curious: why are we disabling core dumps?

cvmfs/loader.cc
```diff
@@ -722,10 +724,27 @@ int FuseMain(int argc, char *argv[]) {
...
+  int origuid = -1;
+  int origgid = -1;
   if (simple_options_parsing_) {
     options_manager = new SimpleOptionsParser(
         new DefaultOptionsTemplateManager(*repository_name_));
   } else {
+    if ((uid_ != 0) || (gid_ != 0)) {
+      // temporarily switch to requested uid/gid while running bash parser
+      origuid = (int) getuid();
+      origgid = (int) getgid();
```
jblomer: Shouldn't we check `geteuid()`/`getegid()`?

cvmfs/loader.cc
```diff
@@ -722,10 +724,27 @@ int FuseMain(int argc, char *argv[]) {
 
   string parameter;
   OptionsManager *options_manager;
+  int origuid = -1;
```
jblomer: That probably works in practice but I think I'd prefer using `uid_t` and `gid_t` types and another marker for whether we need to switch back.

cvmfs/loader.cc
```diff
@@ -902,12 +929,16 @@ int FuseMain(int argc, char *argv[]) {
   }
 
 
-  int fd_mountinfo = -1;  // needs to be declared before start using goto
+  // these need to be declared before start using goto
+  int fd_mountinfo = -1;
+  const bool delegatedunmount = (!suid_mode_ && !disable_watchdog_);
```
jblomer:
```suggestion
  const bool delegated_unmount = (!suid_mode_ && !disable_watchdog_);
```

cvmfs/loader.cc
```diff
@@ -921,7 +952,9 @@ int FuseMain(int argc, char *argv[]) {
   }
   if (disable_watchdog_) {
     LogCvmfs(kLogCvmfs, kLogDebug, "No watchdog, enabling core files");
-    prctl(PR_SET_DUMPABLE, 1, 0, 0, 0);
+    if (!platform_set_dumpable()) {
+      LogCvmfs(kLogCvmfs, kLogDebug, "Failed to set process dumpable");
```
jblomer:
```suggestion
      LogCvmfs(kLogCvmfs, kLogDebug | kSyslogWarn, "Failed to set process dumpable");
```

cvmfs/loader.cc
```diff
@@ -1204,9 +1253,25 @@ int FuseMain(int argc, char *argv[]) {
...
+  // after it starts things that need more privileges
   cvmfs_exports_->fnSpawn();
+
   loader_talk::Spawn();
 
+  if (delegatedunmount) {
+    // Unmounting in this case might be delegated to the watchdog process.
+    // Allow ptracing by the watchdog.
+    if (!platform_set_dumpable()) {
+      LogCvmfs(kLogCvmfs, kLogDebug, "Failed to set process dumpable");
```
jblomer:
```suggestion
      LogCvmfs(kLogCvmfs, kLogDebug | kSyslogWarn, "Failed to set process dumpable");
```

cvmfs/loader.cc
```diff
@@ -1239,15 +1304,58 @@ int FuseMain(int argc, char *argv[]) {
...
+      // This can happen if the cvmfs version was downgraded.  The
+      // watchdog won't know how to do a clean unmount, so the best we
+      // can do is to force a crash so it will do a crash cleanup.
+      LogCvmfs(kLogCvmfs, kLogSyslog,
+               "Watchdog too old for clean error unmount on %s, forcing crash",
+               mount_point_->c_str());
+      assert(false);
+    }
+  } else if (cvmfs_exports_->version > 1) {
+    // This tells the watchdog to not unmount.
```
jblomer, line 1338: Maybe: 

```suggestion
  } else {
     assert(cvmfs_exports_->version > 1);
    // This tells the watchdog to not unmount.
```

cvmfs/loader.cc
```diff
@@ -1239,15 +1304,58 @@ int FuseMain(int argc, char *argv[]) {
...
 
   loader_talk::Fini();
   cvmfs_exports_->fnFini();
 
-  // Unmount
+  if (!delegatedunmount) {
+    // Restore privileges for unmount even if we don't need to do it
+    // ourselves, because the fuse library might need it.
+    if (!ObtainSysAdminCapability()) {
+      LogCvmfs(kLogCvmfs, kLogDebug,
```
jblomer:
```suggestion
      LogCvmfs(kLogCvmfs, kLogDebug | kLogSyslogWarn,
```

cvmfs/loader.cc
```diff
@@ -902,12 +929,16 @@ int FuseMain(int argc, char *argv[]) {
   }
 
 
-  int fd_mountinfo = -1;  // needs to be declared before start using goto
+  // these need to be declared before start using goto
+  int fd_mountinfo = -1;
+  const bool delegatedunmount = (!suid_mode_ && !disable_watchdog_);
+  bool dounmount = false;
```
jblomer, line 934: It may help to comment here what will be the conditions for `dounmount` to be set.

cvmfs/monitor.h
```diff
@@ -18,6 +18,33 @@
 #include "util/pointer.h"
 #include "util/single_copy.h"
 
+
+class Watchdog;
```
jblomer: I don't think that the forward declaration is necessary.

cvmfs/monitor.h
```diff
@@ -18,6 +18,33 @@
...
+
+/**
+ * Information for the FUSE module to communicate with the watchdog process
+ * that needs to be preserved through reloads.
+ */
+
```
jblomer:
```suggestion
```

cvmfs/monitor.h
```diff
@@ -35,12 +62,15 @@ class Watchdog : SingleCopy {
...
    */
-  typedef void (*FnOnCrash)(void);
+  typedef void (*FnOnExit)(const bool crashed);
 
-  static Watchdog *Create(FnOnCrash on_crash);
+  static Watchdog *Create(FnOnExit on_exit, WatchdogState *saved_state = 0);
   static pid_t GetPid();
   ~Watchdog();
   void Spawn(const std::string &crash_dump_path);
+  void ClearOnExitFn() { on_exit_ = 0; }
```
jblomer:
```suggestion
  void ClearOnExitFn() { on_exit_ = NULL; }
```

cvmfs/monitor.cc
```diff
@@ -64,10 +65,13 @@ int Watchdog::g_suppressed_signals[] = {
 int Watchdog::g_crash_signals[] = {SIGQUIT, SIGILL, SIGABRT, SIGFPE,
                                    SIGSEGV, SIGBUS, SIGPIPE, SIGXFSZ};
 
-Watchdog *Watchdog::Create(FnOnCrash on_crash) {
+Watchdog *Watchdog::Create(FnOnExit on_exit, WatchdogState *saved_state) {
   assert(instance_ == NULL);
-  instance_ = new Watchdog(on_crash);
-  instance_->Fork();
+  instance_ = new Watchdog(on_exit);
+  if (saved_state != 0)
```
jblomer:
```suggestion
  if (saved_state != NULL)
```

cvmfs/monitor.h
```diff
@@ -35,12 +62,15 @@ class Watchdog : SingleCopy {
   /**
    * Crash cleanup handler signature.
    */
-  typedef void (*FnOnCrash)(void);
+  typedef void (*FnOnExit)(const bool crashed);
```
jblomer, line 62: Let's use an enum as an argument instead of a bool, so that we know what it means when calling it. Alternatively, always call with comment, like `on_exit_(true /* crashed */)`.

cvmfs/cvmfs.cc
```diff
@@ -234,14 +238,35 @@ void GetReloadStatus(bool *drainout_mode, bool *maintenance_mode) {
...
 #ifndef __TEST_CVMFS_MOCKFUSE  // will be mocked in tests
-static bool UseWatchdog() {
-  if (loader_exports_ == NULL || loader_exports_->version < 2) {
-    return true;  // spawn watchdog by default
+// returns whether or not to start a watchdog
+static bool ShouldStartWatchdog() {
+  if (loader_exports_ == NULL) {
+    // Very old loader?
+    return true;
+  }
```
jblomer: Perhaps just `assert(loader_exports_);`

cvmfs/cvmfs.cc
```diff
@@ -234,14 +238,35 @@ void GetReloadStatus(bool *drainout_mode, bool *maintenance_mode) {
...
+    // This is the initial loader run, not a reload
+    return !loader_exports_->disable_watchdog;
+  }
+
+  // This is a reload
+
+  if ((loader_exports_->version < 6) && !loader_exports_->disable_watchdog) {
+    // This is an older loader so need to start a watchdog
+    // Set the version to 6 so it will only start on the first reload
+    const_cast<loader::LoaderExports*>(loader_exports_)->version = 6;
```
jblomer: Are we sure that this works? E.g., the old loader will not call `fnClearExit()`. We should test this case (or just leave the version at 5, for the time being).

jblomer (commented):
Many thanks, that's a substantial improvement, I think!

Some minor comments inline.

cvmfs/quota_posix.cc
DrDaveD: Previously we weren't setting the dumpable flag, so I didn't want to add more situations where core dumps happen.  I believe the default Linux configuration does not graciously manage disk space when a lot of core dumps happen.

cvmfs/util/capabilities.cc
DrDaveD: It's not always wrapped in assert(); in publish/repository_env.cc it throws an exception.

The thing is that there are a lot of little steps that should never fail, and there are other major steps that might possibly fail without being considered fatal.  So I think having a mix of error returns and asserts is justified.

While working on this I actually did turn all those asserts into messages, but I decided it was worse than leaving the mix.  Also, I don't yet see a need for this function to have an "avoid_mutexes" option but if that becomes necessary it's going to be even more unwieldy to have all those messages.

cvmfs/util/posix.cc
DrDaveD: Oh, right.  And ObtainsetuidgidCapabilities() also needs to support that option.

cvmfs/auto_umount.h
DrDaveD: Why remove the const?

cvmfs/cvmfs.cc
DrDaveD: I did test for it.  It's OK that `fnClearExit()` is not called, because the old loader retains its capabilities and can do the unmount itself.  If I left it at version 5, I would also need to change the code to detect version 5 at shutdown time and stop the watchdog like the old code did.  Maybe that would be better.  I went ahead and made that change.

cvmfs/loader.cc
DrDaveD: There's a very extensive comment about that later (beginning "Decide whether we will need to do our own unmount") so I don't think it's good to duplicate that.

cvmfs/loader.cc
DrDaveD: No, we don't want it to crash in that situation.  It just doesn't need to clear the exit function.  This is a legitimate downgrade situation when there's nothing that needs to be unmounted.

Oh, maybe you were thinking that this `else` was to the `if (cvmfs_exports_->version == 1)` and not the `if (dounmount && delegated_unmount)`?

cvmfs/loader.cc
DrDaveD: Yes, good catch.

cvmfs/loader.cc
DrDaveD: Actually there is no `kSyslogWarn`; I put in `kLogWarning`.

DrDaveD (commented):
All the suggestions that have no reply comment were applied.

cvmfs/auto_umount.h
jblomer: I'm not sure what we gain by adding the `const`. I see the point for const references but not so much for a boolean argument. Anyway, it's minor and up to you.

cvmfs/loader.cc
jblomer: Right, got confused in the nested `if`s

jblomer (commented):
Thank you! LGTM.

Approval and merge should be done by @vvolkl

vvolkl (approved):
Looks good to me, very big patch but a nice improvement!

I pushed a small change to the test to make it check also the effective capabilities. @DrDaveD  If you agree with that I'm happy to merge and prepare a 2.14.0~pre1 rerelease version for testing.

DrDaveD:
I am fine with the change and with merging.  However, please include my other PRs in 2.14.0-pre1 too, especially #3735.

PR #3735, opened 2024-12-24
Add CVMFS_INFO_HEADER

### LINKED ISSUE #3666: Extending cvmfs-info header would be very useful

DrDaveD:
This extends the `cvmfs-info` header when a template passed in `CVMFS_INFO_HEADER` is set, as defined in #3666.

This PR depends on #3730 which must be merged first.

- Fixes #3666

### DIFF
--- a/cvmfs/loader.h
+++ b/cvmfs/loader.h
@@ -210,7 +210,8 @@ struct LoaderExports {
 
 /**
  * This contains the public interface of the cvmfs fuse module.
- * Whenever something changes, change the version number.
+ * Whenever something changes, change the version number and add only
+ * to the end of the struct.
  * A global CvmfsExports struct is looked up by the loader via dlsym.
  * Since the cvmfs fuse module gets replaced each time there's an
  * upgrade or downgrade, the only way to get a cvmfs module older than
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -2207,12 +2207,29 @@ void MountPoint::SetupHttpTuning() {
       && options_mgr_->IsOn(optarg)) {
     download_mgr_->EnableRedirects();
   }
-  if (options_mgr_->GetValue("CVMFS_SEND_INFO_HEADER", &optarg)
-      && options_mgr_->IsOn(optarg)) {
+  if (options_mgr_->GetValue("CVMFS_INFO_HEADER", &optarg) && (optarg != ""))
+  {
     download_mgr_->EnableInfoHeader();
+    download_mgr_->SetInfoHeaderTemplate(optarg);
+  } else if (options_mgr_->GetValue("CVMFS_SEND_INFO_HEADER", &optarg)
+      && options_mgr_->IsOn(optarg))
+  {
+    download_mgr_->EnableInfoHeader();
+    download_mgr_->SetInfoHeaderTemplate("%{path}");
   }
 }
 
+/*
+ * Check whether permission is needed to read from user process environment.
+ */
+bool MountPoint::NeedsReadEnviron(OptionsManager *omgr) {
+  // This is a class (static) method because it is used early, before 
+  // all the above initialization is done, so can't rely on mountpoint
+  // object data.
+  string info_header;
+  omgr->GetValue("CVMFS_INFO_HEADER", &info_header);
+  return (info_header.find("%{env:") != std::string::npos);
+}
 
 void MountPoint::SetupInodeAnnotation() {
   string optarg;
--- a/cvmfs/mountpoint.h
+++ b/cvmfs/mountpoint.h
@@ -502,6 +502,9 @@ class MountPoint : SingleCopy, public BootFactory {
                             OptionsManager *options_mgr = NULL);
   ~MountPoint();
 
+  // Check whether permission is needed to read from user process environment
+  static bool NeedsReadEnviron(OptionsManager *omgr);
+
   unsigned GetMaxTtlMn();
   unsigned GetEffectiveTtlSec();
   void SetMaxTtlMn(unsigned value_minutes);
--- a/cvmfs/network/jobinfo.cc
+++ b/cvmfs/network/jobinfo.cc
@@ -137,37 +137,49 @@ std::string EscapeHeader(const std::string &header) {
  * Return the filled-in template of CVMFS_INFO_HEADER
  */
 std::string JobInfo::GetInfoHeaderContents(const std::string &templ) {
+  enum ParseMode {
+    kParseModeDefault,       // reading literal characters
+    kParseModeAfterPercent,  // just saw '%'
+    kParseModeInKey,         // inside '%{...}'
+  };
+
+  enum MatchMode {
+    kMatchModeSkipping,    // skipping to next '\0' separator
+    kMatchModeMatching,    // comparing chars against variable name
+    kMatchModeCollecting,  // name matched, collecting value
+  };
+
   std::string answer = "";
   std::string key;
-  int parsemode = 0;  // 0 default, 1 after '%', 2 after '%{'
-  int envsize = -1;
-  char *envbuf = NULL;
+  ParseMode parsemode = kParseModeDefault;
+  std::vector<char> envbuf;
+  bool env_attempted = false;
 
   for (std::string::const_iterator i = templ.begin(); i != templ.end(); i++) {
     const char c = *i;
     switch (parsemode) {
-    case 0:
+    case kParseModeDefault:
       if (c == '%') {
-        parsemode = 1;
+        parsemode = kParseModeAfterPercent;
         key = "";
       } else {
         answer += c;
       }
       break;
-    case 1:
+    case kParseModeAfterPercent:
       if (c == '{') {
-        parsemode = 2;
+        parsemode = kParseModeInKey;
       } else {
-        parsemode = 0;
+        parsemode = kParseModeDefault;
         answer += '%';
         answer += c;
       }
       break;
-    case 2:
+    case kParseModeInKey:
       if (c != '}') {
         key += c;
       } else {
-        parsemode = 0;
+        parsemode = kParseModeDefault;
         if (key == "path") {
           if (path_info_ != NULL) {
             answer += EscapeHeader(*path_info_);
@@ -193,9 +205,9 @@ std::string JobInfo::GetInfoHeaderContents(const std::string &templ) {
             answer += "-";
           }
         } else if (key.substr(0, 4) == "env:") {
+          if (!env_attempted) {
+            env_attempted = true;
 #ifndef __APPLE__
-          if (envsize == -1) {
-            envsize = 0;
             if (pid_ != static_cast<pid_t>(-1)) {
               ObtainDacReadSearchCapability();
               ObtainSysPtraceCapability();
@@ -214,17 +226,13 @@ std::string JobInfo::GetInfoHeaderContents(const std::string &templ) {
                 }
                 if ((n >= 0) && (size > 0)) {
                   if (lseek(fd, 0, SEEK_SET) >= 0) {
-                    envbuf = static_cast<char *>(malloc(size));
-                    if (envbuf != NULL) {
-                      if (read(fd, envbuf, size) > 0) {
-                        envsize = size;
-                        LogCvmfs(kLogDownload, kLogDebug,
-                          "(job id %" PRId64 ") read %d bytes from %s",
-                          id_, size, fname.c_str());
-                      } else {
-                        free(envbuf);
-                        envbuf = NULL;
-                      }
+                    envbuf.resize(size);
+                    if (read(fd, envbuf.data(), size) > 0) {
+                      LogCvmfs(kLogDownload, kLogDebug,
+                        "(job id %" PRId64 ") read %d bytes from %s",
+                        id_, size, fname.c_str());
+                    } else {
+                      envbuf.clear();
                     }
                   }
                 }
@@ -237,24 +245,24 @@ std::string JobInfo::GetInfoHeaderContents(const std::string &templ) {
               DropSysPtraceCapability();
               DropDacReadSearchCapability();
             }
-          }
 #endif
-          if (envbuf != NULL) {
+          }
+          if (!envbuf.empty()) {
             const char * const var = key.c_str() + 4; // everything after "env:"
             const char *varp = var;
             std::string val = "";
-            int matchmode = 1;
-            const char * const endp = envbuf + envsize;
-            for (const char *p = envbuf; p < endp; p++) {
-              switch(matchmode) {
-              case 0:
+            MatchMode matchmode = kMatchModeMatching;
+            const char * const endp = envbuf.data() + envbuf.size();
+            for (const char *p = envbuf.data(); p < endp; p++) {
+              switch (matchmode) {
+              case kMatchModeSkipping:
                 // skipping to next null character
                 if (*p == '\0') {
                   varp = var;
-                  matchmode = 1;
+                  matchmode = kMatchModeMatching;
                 }
                 break;
-              case 1:
+              case kMatchModeMatching:
                 // matching variable name
                 if (*p == '\0') {
                   // premature end without an '='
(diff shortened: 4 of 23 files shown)


DrDaveD:
Closing & reopening to re-run clang-tidy & unittest, which failed but show no failures

vvolkl:
I've fixed the clang-tidy workflow in  #3736, let's try with that once it's merged.

test/unittests/main.cc
```diff
@@ -11,7 +11,7 @@
 #include "monitor.h"
 
 int main(int argc, char **argv) {
-  Watchdog *watchdog = Watchdog::Create(NULL);
+  Watchdog *watchdog = Watchdog::Create(NULL, null);
```
vvolkl:
```suggestion
  Watchdog *watchdog = Watchdog::Create(NULL, NULL);
```

cvmfs/network/jobinfo.cc
```diff
@@ -80,4 +88,211 @@ void JobInfo::Init() {
...
+                             (input % 16) + ((input % 16 <= 9) ? '0' : 'A'-10));
+  return true;
+}
+
+
+static std::string EscapeHeader(const std::string &header)
+{
+  std::string escaped = "";
+  char escaped_char[3];
+  for (char c : header) {
```
vvolkl:
```
error: range-based 'for' loops are not allowed in C++98 mode
   for (char c : header) {
                 ^
```
We'll have to at least deprecate centos7 for that, maybe work around it for now

cvmfs/capabilities.h
jblomer, line 1: Move to libcvmfs_util?

cvmfs/util/posix.h
```diff
@@ -157,6 +157,7 @@ CVMFS_EXPORT std::string GetArch();
 
 CVMFS_EXPORT int SetLimitNoFile(unsigned limit_nofile);
 CVMFS_EXPORT void GetLimitNoFile(unsigned *soft_limit, unsigned *hard_limit);
+CVMFS_EXPORT int SetLimitCore(unsigned limit_core);
```
jblomer: We should try adding a unit test for this.

cvmfs/capabilities.h
jblomer: I think we should add unit tests for adding/dropping capabilities. Perhaps in a separate executable with suid flag.

cvmfs/mountpoint.h
```diff
@@ -22,6 +22,7 @@
 #include "gtest/gtest_prod.h"
 #include "loader.h"
 #include "magic_xattr.h"
+#include "network/download.h"
```
jblomer: This change seems unnecessary.

cvmfs/mountpoint.h
```diff
@@ -493,6 +491,8 @@ class MountPoint : SingleCopy, public BootFactory {
                             OptionsManager *options_mgr = NULL);
   ~MountPoint();
 
+  static bool needs_read_environ(OptionsManager *omgr);
```
jblomer:
```suggestion
  static bool NeedsReadEnviron(OptionsManager *omgr);
```

cvmfs/mountpoint.cc
```diff
@@ -2244,13 +2244,30 @@ void MountPoint::SetupHttpTuning() {
...
   {
     download_mgr_->EnableInfoHeader();
+    download_mgr_->SetInfoHeaderTemplate("%{path}");
   }
 }
 
+/*
+ * Check whether permission is needed to read from user process environment.
+ */
+
```
```suggestion
```

cvmfs/mountpoint.h
jblomer: Please add documentation

cvmfs/loader.h
```diff
@@ -223,6 +223,7 @@ struct CvmfsExports {
     fnSaveState = NULL;
     fnRestoreState = NULL;
     fnFreeSavedState = NULL;
+    fnNeedsReadEnviron = NULL;
```
jblomer: I think we should change the version number to make sure old loaders work with newer fuse modules (and test that)

cvmfs/network/jobinfo.h
```diff
@@ -5,6 +5,7 @@
 #ifndef CVMFS_NETWORK_JOBINFO_H_
 #define CVMFS_NETWORK_JOBINFO_H_
 
+#include <cstdio>
```
jblomer: Seems unnecessary

cvmfs/network/jobinfo.cc
```diff
@@ -80,4 +88,211 @@ void JobInfo::Init() {
...
+  }
+
+  return escaped;
+}
+
+
+/*
+ * Return the filled-in template of CVMFS_INFO_HEADER
+ */
+
```
jblomer, line 151:
```suggestion
```

cvmfs/publish/cmd_transaction.cc
```diff
@@ -72,8 +72,7 @@ int CmdTransaction::Main(const Options &options) {
     settings->GetTransaction()->SetTemplate(tokens[0], tokens[1]);
   }
 
-  if (!SwitchCredentials(settings->owner_uid(), settings->owner_gid(),
-                         false /* temporarily */))
+  if (!SwitchCredentials(settings->owner_uid(), settings->owner_gid(), false))
```
jblomer: I'd prefer to keep the inline documentation of the bool argument

cvmfs/publish/cmd_commit.cc
```diff
@@ -49,8 +49,7 @@ int CmdCommit::Main(const Options &options) {
     throw;
   }
 
-  if (!SwitchCredentials(settings->owner_uid(), settings->owner_gid(),
-                         false /* temporarily */))
+  if (!SwitchCredentials(settings->owner_uid(), settings->owner_gid(), false))
```
jblomer: I'd prefer to keep the inline documentation of the bool argument

cvmfs/publish/cmd_abort.cc
```diff
@@ -61,8 +61,7 @@ int CmdAbort::Main(const Options &options) {
     throw;
   }
 
-  if (!SwitchCredentials(settings->owner_uid(), settings->owner_gid(),
-                         false /* temporarily */))
+  if (!SwitchCredentials(settings->owner_uid(), settings->owner_gid(), false))
```
jblomer: I'd prefer to keep the inline documentation of the bool argument

cvmfs/network/jobinfo.h
```diff
@@ -185,13 +188,14 @@ class JobInfo {
...
   const shash::Any *expected_hash() const { return expected_hash_; }
-  const std::string *extra_info() const { return extra_info_; }
+  const std::string *path_info() const { return path_info_; }
 
   off_t range_offset() const { return range_offset_; }
   off_t range_size() const { return range_size_; }
 
   CURL *curl_handle() const { return curl_handle_; }
   curl_slist *headers() const { return headers_; }
+  const std::string info_header_contents(const std::string templ);
```
jblomer: That's more than just a getter, and shouldn't be `const`. Also, I think the parameter should be passed by const reference. So I'd suggest
```suggestion
  std::string GetInfoHeaderContents(const std::string &templ);
```

cvmfs/network/jobinfo.cc
```diff
@@ -80,4 +88,211 @@ void JobInfo::Init() {
...
+const std::string JobInfo::info_header_contents(const std::string templ) {
+  std::string answer = "";
+  std::string key;
+  int parsemode = 0;  // 0 default, 1 after '%', 2 after '%{'
+  int envsize = -1;
+  char *envbuf = NULL;
+
+  for (char c : templ) {
```
jblomer: Not C++03 compliant

cvmfs/network/jobinfo.cc
```diff
@@ -80,4 +88,211 @@ void JobInfo::Init() {
...
+
+  return escaped;
+}
+
+
+/*
+ * Return the filled-in template of CVMFS_INFO_HEADER
+ */
+
+const std::string JobInfo::info_header_contents(const std::string templ) {
```
jblomer: Should be unit tested.

cvmfs/network/jobinfo.cc
```diff
@@ -80,4 +88,211 @@ void JobInfo::Init() {
...
+      } else {
+        answer += c;
+      }
+      break;
+    case 1:
+      if (c == '{') {
+        parsemode = 2;
+      } else {
+        answer += '%';
+        answer += c;
```
jblomer, line 175: Shouldn't this reset the parsemode?

cvmfs/network/jobinfo.cc
```diff
@@ -80,4 +88,211 @@ void JobInfo::Init() {
...
+            answer += StringifyInt(gid_);
+          } else {
+            answer += "-";
+          }
+        } else if (key.substr(0, 4) == "env:") {
+          if (envsize == -1) {
+            envsize = 0;
+            if (pid_ != (pid_t) -1) {
+              ObtainDacReadSearchCapability();
+              ObtainSysPtraceCapability();
```
jblomer, line 213: We would do this for each request... how heavy is this? Should we warn users when using `env:` in the template?

jblomer: Also, I think the regular and the external download manager may step on each others feet obtaining and dropping capabilities. Perhaps we need an helper process with elevated capabilities to read the environment, similar to the authorization helper.

cvmfs/util/posix.h
DrDaveD: There aren't other unit tests for posix.cc functions so it doesn't seem like much of a priority to  me.

cvmfs/capabilities.h
DrDaveD: > Move to libcvmfs_util?

I tried first putting `capabilities.[ch]` in the util subdirectory but ran into compilation snags and moved it back up to the main level where it was. I don't remember exactly what the snags were.  Perhaps since then I have learned more about the cmake setup and should try it again.

cvmfs/publish/cmd_transaction.cc
DrDaveD: Done in #3730

cvmfs/publish/cmd_commit.cc
DrDaveD: Done in #3730

cvmfs/network/jobinfo.cc
DrDaveD: > We would do this for each request... how heavy is this? Should we warn users when using `env:` in the template?

Obtaining & dropping capabilities are a trivial amount of time.  The longer time is reading & parsing the environment, but it is all in memory and I assume it is negligible compared to the time of doing a download.  This doesn't happen when reading from cache.

DrDaveD: > Also, I think the regular and the external download manager may step on each others feet obtaining and dropping capabilities. Perhaps we need an helper process with elevated capabilities to read the environment, similar to the authorization helper.

Maybe, but that's a very heavyweight solution.  Perhaps there should be a lock, or perhaps it should obtain capabilities and not drop them.  On the other hand keeping these dangerous capabilities out of the fuse process is attractive for security reasons.  Maybe it could be delegated to the watchdog process so we don't need yet another process.

I thought about keeping a cache of recent cvmfs-info strings based on PID (not including the `%{path}` substitution because that changes) but wasn't sure about the implications for locking.  So I decided to leave that for a later optimization if needed.

cvmfs/capabilities.h
DrDaveD: Ok this time I succeeded in getting it put into libcvmfs_util, updated in #3730.

DrDaveD: > I think we should add unit tests for adding/dropping capabilities. Perhaps in a separate executable with suid flag.

I wouldn't know how to do that.  To me it seems a lot more straightforward to add a numbered test written in bash that verifies that capabilities are dropped when they should be and that stack dumps & unmounts work.  How does that sound?

cvmfs/network/jobinfo.cc
DrDaveD: > Shouldn't this reset the parsemode?

Yes, good catch.

cvmfs/loader.h
DrDaveD: > I think we should change the version number to make sure old loaders work with newer fuse modules (and test that)

I tried to do that, please check to see if it looks like I did it correctly.  I think I probably did something wrong, because I can smoothly upgrade from 2.11.5 to 2.12.3, but if I upgrade from 2.12.3 to my test version the reload hangs and I have to abort the fuse connection.

DrDaveD: I fixed the reload problem.  The fix was to move the new field to the end of the struct so it doesn't move the `cvmfs_operations` substructure.

It turns out that the new field is only used during initial load, so it is always present when it is accessed and there's no need to check for a version number change.

Downgrading from the new loader to an older cvmfs does cause a failure to unmount (and a failure to get a stack trace when the condition calls for it) if the fuse process dies with any kind of signal.  I'm not sure if that's anything to worry about though since it should be a very rare condition.

cvmfs/capabilities.h
DrDaveD: I added test 106 here for that.

cvmfs/network/jobinfo.cc
DrDaveD: I added test 107 which tests all the functionality in this function.  There doesn't need to be a separate unittest does there?

cvmfs/network/jobinfo.cc
DrDaveD: > On the other hand keeping these dangerous capabilities out of the fuse process is attractive for security reasons. Maybe it could be delegated to the watchdog process so we don't need yet another process.

Oh but there's a problem.  The loader is part of the main fuse process.  After it forks the watchdog it drops it privileges.  Then when it is time to upgrade and/or reload, it stops the watchdog and restarts the watchdog after the reload.  It won't have the privileges to pass to the new watchdog.

I don't know what to do about that.

DrDaveD: > I don't know what to do about that.

Capabilities are per-thread, and it looks like only the loader_talk thread needs to retain capabilities for reloading.  I'm trying to get that to work.  If that works I should probably have that thread also handle all the env lookups.

DrDaveD: It does work and I have updated #3730 with that strategy, not yet merged with this one.

cvmfs/network/jobinfo.cc
```diff
@@ -81,4 +88,218 @@ void JobInfo::Init() {
...
+                // read it to find out the size
+                ssize_t n;
+                int size = 0;
+                char buf[BUFSIZ];
+                while ((n = read(fd, buf, BUFSIZ)) > 0) {
+                  size += static_cast<int>(n);
+                }
+                if ((n >= 0) && (size > 0)) {
+                  if (lseek(fd, 0, SEEK_SET) >= 0) {
+                    envbuf = static_cast<char *>(malloc(size));
```
vvolkl: Since you now opened the gates to C++11, this could be a unique_ptr

cvmfs/network/jobinfo.cc
```diff
@@ -81,4 +88,218 @@ void JobInfo::Init() {
...
+
+} // namespace
+
+/*
+ * Return the filled-in template of CVMFS_INFO_HEADER
+ */
+std::string JobInfo::GetInfoHeaderContents(const std::string &templ) {
+  std::string answer = "";
+  std::string key;
+  int parsemode = 0;  // 0 default, 1 after '%', 2 after '%{'
```
vvolkl: Could be an enum

vvolkl (approved):
Some code quality improvements, and a fix for the capabilities already added to the PR.

vvolkl:
I'll move the last commit to its own PR, as it's not directly related.

PR #3739, opened 2025-01-05
fsck -q: correctly handle other hash algs 

vvolkl:


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest full

Issue #3744, opened 2025-01-09
Replace cpp-lint with clang-format


HereThereBeDragons:

HereThereBeDragons:
The two PRs #3745 and #3746 offer to possible versions of clang-format as close as possible to the old cpp-lint style.

I would prefer #3746 as i think the alignment of multiple assignment operands is superior. 
This alignment is only performed of consecutive assignments. No alignment across empty lines or comments are done.
See: https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html#alignconsecutiveassignments

HereThereBeDragons:
There is some github action that might work? 
https://github.com/purduesigbots/clang-format-action/tree/main

However building ubuntu:latest comes with clang-format-18, while the latest stable one is version 19.
I had to adjust the Dockerfile a bit to get clang-format-19:
``` 
 # Jan 2025: ubuntu:latest default clang-format is version 18, not version 19
 # so manually download version 19
 RUN wget https://apt.llvm.org/llvm.sh && \
   chmod +x llvm.sh && \
   ./llvm.sh 19
 
 RUN apt-get update && \
     apt-get install -y clang-format-19
 
 # make symlink so that normal clang-format command works
 RUN cd /bin && ln -n clang-format-19 /bin/clang-format
 RUN cd /bin && ln -n git-clang-format-19 /bin/git-clang-format
```

the entrypoint i also had to change a bit as for the local docker run to be able to mount and run `git-clang-format` on a given host directory

vvolkl:
We can add clang-format here: https://github.com/cvmfs/cvmfs/blob/devel/.github/workflows/linter.yaml

vvolkl:
That uses clang 18 as well, but I'm OK with updating to 19

HereThereBeDragons:
the final implementation of clang format is  in PR #3752 and got merged

PR #3746, opened 2025-01-09
Example of using clang-format-19: With alignment of assignments

HereThereBeDragons:
This is an example how the created `.clang-format` using clang-format v19 would look like.
The this version also aligns assignment operators.

For more read issue #3744

- this would be my preferred version -


HereThereBeDragons:
replaced by #3748

PR #3750, opened 2025-01-14
Revert CVMFS_PATCH_LEVEL for check_cvmfs.sh

vvolkl:
The CVMFS_PATCH_LEVEL is largely unused and has been set to 0 for quite a while, see
https://github.com/cvmfs/cvmfs/commit/4f54395bbc24bd989a42c7e642bf9de353a86903

In a recent cleanup I've set this to the actual patch version of the package, however this breaks the cvmfs_check.sh script. 

CVMFS_PATCH_LEVEL is used in two places:
* `cvmfs_talk version patchlevel` returns  CVMFS_PATCH_LEVEL (usually "0")
* `attr -g version /mountpoint` returns CVMFS_VERSION.CVMFS_PATCH_LEVEL (so "2.11.5.0" for example)

As the patch level is mostly unused it could be removed, but I would tend to fix it first and deprecate it slowly as it is user facing via the two commands above (or keep it, if a usecase can be found).


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3752, opened 2025-01-17
clang-format for v18

HereThereBeDragons:
already ready for v19, but commented out

uses google cpp style as base and then changed it to fit better our old cpplint style.
comments included to show which parameters deviate from which default value

only missing cpplint feature is to mark if a function parameter that is a reference:
- with `*param` pointer instead of `&param` to show that the values within are modified
- with `const &param` that it is not modified
maybe something clang-tidy can do?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,289 @@
 ---
-Language: Cpp
-BasedOnStyle: Google
+Language:        Cpp
+# BasedOnStyle:  Google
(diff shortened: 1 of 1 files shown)


vvolkl (approved):
Thanks!

HereThereBeDragons:
just for better issue linking. the relevant issue is #3744

PR #3754, opened 2025-01-21
Make telemetry aggregator delta calculation more reliable

HereThereBeDragons:
integration test `Telemetry Aggregator: Influx (test number 098)` is sometimes flaky due to wrong delta values.

delta values are calculated each time based on the raw values and as such it is only a single occurence that goes wrong.

it does not occur so often but it seems like those faulty tests have a longer than normal run time.
in the single output i got the 1st print of counters is missing some counters listed in the 2nd

so lets see if this `try/catch` resolves the issue...


HereThereBeDragons:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3760, opened 2025-01-28
Applied clang-tidy rules to cvmfs.cc and included headers, and removed misc-include-cleaner from .clang-tidy

Siempie30:
All clang-tidy rules (specified in the .clang-tidy file) have been applied to cvmfs.cc. All suggestions for the headers included in cvmfs.cc are also applied.
Logging feature are marked with NOLINT, because there is no solution yet for avoiding directly including the internal logging header.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -12,6 +12,7 @@ Checks: -*
   ,google-*
   ,-google-readability-avoid-underscore-in-googletest-name
   ,misc-*
+  ,-misc-include-cleaner
   ,-misc-static-assert
   ,-misc-no-recursion
   ,-misc-non-private-member-variables-in-classes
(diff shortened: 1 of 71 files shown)


vvolkl (changes requested):
Thanks a lot! I think we need to break up the different types of fixes though, especially for cvmfs.cc. Some, like adding missing `consts` are straightforward, but others like the removal of static require a static look.

For misc-include-cleaner, my thinking was actually to remove it from .clang-tidy, and do one separate run of clang-tidy with the inline NOLINTS in front of logging statements, not add them permanently to the source.

vvolkl:
Thanks for this PR. Unfortunately  fixing all checks in one PR turned out to be impractical even for a limited fileset, so I'm closing.

Issue #3531, opened 2024-03-14
cvmfs_receiver performance: curl_global_init


mharvey-jt:
`curl_global_init()` is always called with `CURL_GLOBAL_ALL`
https://github.com/cvmfs/cvmfs/blob/2333588f3d2afd54e3b145336740539bdf053f9c/cvmfs/network/download.cc#L1783

This causes ssl libraries to be initialized, and can take some time.

If SSL is not in use, calling with `CURL_GLOBAL_NOTHING` is faster.

Siempie30:
Hi, thanks for the suggestion.
After investigating, it turns out that switching to `CURL_GLOBAL_NOTHING` does not improve performance, because the `CURL_GLOBAL_SSL` flag (which is used in the `CURL_GLOBAL_ALL` flag) serves no meaning. [This is the case since libcurl 7.57.0](https://curl.se/libcurl/c/curl_global_init.html), CVMFS currently uses 7.86.0.

Initialization of the SSL libraries can be prevented by adding the `--without-ssl` option to the libcurl configure script, this is currently set to `--with-openssl`:
https://github.com/cvmfs/cvmfs/blob/8444c2ccf2c2eb91ac572b62a5cadd3ae05e26b5/externals/libcurl/src/configureHook.sh#L3

Because SSL is still used for some CVMFS features (such as S3), it might be a good solution to add a CMake option to disable SSL, in case the SSL features are not used.

vvolkl:
We just discussed this - consensus was to add a CMake option.

PR #3761, opened 2025-01-30
Added a cmake option to allow for building libcurl without SSL support.

Siempie30:
This option was requested in issue #3531. Building curl without SSL support improves the speed of the curl initialization in https://github.com/cvmfs/cvmfs/blob/8444c2ccf2c2eb91ac572b62a5cadd3ae05e26b5/cvmfs/network/download.cc#L1949
However, it cannot be set as a standard option, because some features (such as S3) require SSL support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,6 +206,9 @@ if (BUILTIN_EXTERNALS)
     if (BUILD_SNAPSHOTTER)
       set(ENV{BUILD_SNAPSHOTTER} "true")
     endif()
+    if (BUILD_CURL_WITHOUT_SSL)
+      set(ENV{BUILD_CURL_NO_SSL} "true")
+    endif()
 
     message (STATUS "running bootstrap.sh ...")
     execute_process (
@@ -230,6 +233,7 @@ if (BUILTIN_EXTERNALS)
     set(ENV{BUILD_GATEWAY} "")
     set(ENV{BUILD_DUCC} "")
     set(ENV{BUILD_SNAPSHOTTER} "")
+    set(ENV{BUILD_CURL_WITHOUT_SSL} "")
   endif (EXISTS "${CMAKE_SOURCE_DIR}/bootstrap.sh")
 
   # In the case of built-in external libraries, we need to set CMAKE_PREFIX_PATH to
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,9 +206,11 @@ if (BUILTIN_EXTERNALS)
     if (BUILD_SNAPSHOTTER)
       set(ENV{BUILD_SNAPSHOTTER} "true")
     endif()
-    if (BUILD_CURL_WITHOUT_SSL)
-      set(ENV{BUILD_CURL_NO_SSL} "true")
+    if (NO_SSL_SUPPORT_EXPERIMENTAL)
+      message(WARNING "SSL support for CURL has been disabled. This is an experimental feature, it might break CVMFS. Only use if you are absolutely certain this is correct.")
+      set(ENV{NO_SSL_SUPPORT_EXPERIMENTAL} "true")
     endif()
+    execute_process(COMMAND sed -i /ssl_support_disabled=/c\\\ \ ssl_support_disabled=${NO_SSL_SUPPORT_EXPERIMENTAL} ${CMAKE_SOURCE_DIR}/cvmfs/cvmfs_config)
 
     message (STATUS "running bootstrap.sh ...")
     execute_process (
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -206,6 +206,9 @@ if (BUILTIN_EXTERNALS)
     if (BUILD_SNAPSHOTTER)
       set(ENV{BUILD_SNAPSHOTTER} "true")
     endif()
+    if (BUILD_CURL_WITHOUT_SSL)
```
vvolkl: The BUILD_* arguments generally include/exclude parts of CVMFS in the build. I would rename it to something like NO_SSL_SUPPORT_EXPERIMENTAL to make sure that users know what they are doing when enabling this.

CMakeLists.txt
```diff
@@ -206,6 +206,9 @@ if (BUILTIN_EXTERNALS)
...
+      set(ENV{BUILD_CURL_NO_SSL} "true")
```
vvolkl: Add a warning printout that says that parts of CVMFS might not work without.

externals/libcurl/src/configureHook.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/sh
 
 curl_ssl_config="--with-openssl"
+if [ x$BUILD_CURL_WITHOUT_SSL = x"true" ]; then
+  curl_ssl_config="--without-ssl"
+fi
+
 FIX_COMP=""
 LIBS=""
 if [ x"$(uname)" = x"Darwin" ]; then
```
vvolkl, line 7: Might as well add the same switch for macos

vvolkl (changes requested):
Looks good, thank you! One thing we might want is to add a warning in `cvmfs_config chkconfig` that SSL support has been disabled.

mharvey-jt:
With this set, libcurl is built without ssl support,  but the OpenSSL libraries are still linked against in the cmake rules for cvmfs2, libcvmfs_fuse3.so, etc.
Simply commenting those out for libcvmfs_fuse3.so produces a library that is missing X509_* and BIO_* symbols (signature.cc) which I think could be satisfied if the vendored libcrypto were linked in.

mharvey-jt:
As an aside, we don't actually care about manifest signatures and have that disabled via a runtime feature flag.

cvmfs/cvmfs_config
```diff
@@ -912,6 +912,10 @@ cvmfs_showconfig() {
       [ $short_output -eq 0 ] && echo "${var}="
     fi
   done <<< "$all_vars"
+
+  # Check if CVMFS was built without SSL support
+  ssl_support_disabled=YES
+  echo "SSL support disabled=${ssl_support_disabled}"
```
vvolkl: Better to move this to `chksetup` as a warning

cvmfs/cvmfs_config
```diff
@@ -839,6 +839,13 @@ cvmfs_chksetup() {
     fi
   done
 
+  # Check if CVMFS was built without SSL support (the following line is automatically substituded by CMake)
+  ssl_support_disabled=
+  if [[ "x$ssl_support_disabled" = "xtrue" ]]; then
+    echo "Warning: SSL support is disabled. This is experimental and might cause problems"
+    num_warnings=$(($num_warnings+1))
+  fi
+
```
vvolkl, line 848: On second thought, let's not do it like this. Just doing this based on configure flags is a bit brittle. Plus,  like this it will leave a changed cvmfs_config in the git repository after running cmake, we should at least make a copy in the build directory. But let's remove it altogether, and see if we can figure out a runtime test based on ldd.

vvolkl (changes requested):
The library linked in when NO_SSL_... is set should be `${Libcrypto_LIBRARIES}` instead of  `cvmfs_crypto`. But even then this isn't working:
```
CernVM-FS: running in debug mode
CernVM-FS: loading Fuse module... failed to load cvmfs library, tried: './libcvmfs_fuse3_debug.so' '/usr/lib/libcvmfs_fuse3_debug.so' '/usr/lib64/libcvmfs_fuse3_debug.so'
./libcvmfs_fuse3_debug.so: cannot open shared object file: No such file or directory
/usr/lib/libcvmfs_fuse3_debug.so: undefined symbol: SSL_CTX_set_keylog_callback
/usr/lib64/libcvmfs_fuse3_debug.so: cannot open shared object file: No such file or directory
```
We might need to implement the runtime switch to disable manifest signatures Matt was mentioning, to be discussed.

vvolkl:
Closing for now. This requires the signatures to be disabled as per Matt's comment, and we first need to add a cmake flag for that .

PR #3767, opened 2025-02-13
Gateway nested catalog recursion test

Siempie30:
When a nested catalog mountpoint is added or deleted, the gateway does not need to recurse into that directory (unlike when it is *not* a nested catalog mountpoint). These two tests verify that the commit processor does not unnecessarily recurse into a nested catalog mountpoint directory, when this directory is added or deleted, or when a nested catalog is added to or removed from an existing directory.


vvolkl:
@cernvm-bot cloudtest

Issue #3764, opened 2025-02-03
Make cvmfs_config bugreport resilient against hanging processes and improve reporting


HereThereBeDragons:
When cvmfs hangs, creating a bugreport can file as when trying to retrieve information from cvmfs those processes also hang.

It would be good if the subcommands are wrapped by a timeout and killed if they take too long. 
So that in the end you get the bugreport with as much as information as possible.

In addition, more info should be retrieved to allow easier debugging.
E.g.  for #3751 it would be useful to have the content of the locking files used for reloading cvmfs ( ` /var/run/cvmfs/` )

PR #3768, opened 2025-02-17
Resilient bugreport: add timeout to commands + write more stuff

HereThereBeDragons:
Issue #3764 

- subcommands in bugreport now timeout after default 90sec (can be changed by running `cvmfs_config bugreport <timeout_in_sec>`
- add all files listed in `/var/run/cvmfs` (if available will include guard_file used for reload etc)

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -151,7 +151,7 @@ cvmfs_config_usage() {
   echo "  umount"
   echo "  wipecache"
   echo "  killall"
-  echo "  bugreport"
+  echo "  bugreport <timeout_in_sec_per_command (default = 120)>"
 }
 
 has_selinux() {
@@ -1454,6 +1454,13 @@ cvmfs_bugreport() {
   mkdir bugreport
   cd bugreport
 
+  local timeout_sec
+  if [ -z "$1" ]; then
+    timeout_sec=120  # 2min per command until it is killed
+  else
+    timeout_sec=$1
+  fi
+
   dist_default="/etc/cvmfs/default.d/*.conf"
   for file in /etc/cvmfs/default.conf $dist_default /etc/cvmfs/default.local
   do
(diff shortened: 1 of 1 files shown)


vvolkl:
Thanks!  Did you check how `timeout` behaves if `cmd` does not respond to a SIGKILL? That's a possibility for df for example, when the cvmfs process is in D state.

DrDaveD:
In my experience I have always had to kill the parent process of the stuck command process, because the stuck command itself becomes unkillable even with -9.  You might need to use something like
```
timeout -s 9 ${timeout_sec} sh -c "$cmd &
                                   wait"
```
With that if $cmd is unkillable it will still continue by just killing the wait command.

cvmfs/cvmfs_config
```diff
@@ -1485,35 +1496,38 @@ cvmfs_bugreport() {
 
   local commands
 
+  cvmfs_cmds=('cvmfs2 --version' 'cvmfs_config probe' 'cvmfs_config status' \
+              'cvmfs_config showconfig' 'cvmfs_config chksetup' \
+              'cvmfs_config stat -v' 'cvmfs_talk internal affairs' \
+              "find ${CVMFS_CACHE_BASE} -maxdepth 1 -exec ls -lah {} ;")
```
vvolkl:
```suggestion
              "find ${CVMFS_CACHE_BASE} -maxdepth 1 -exec ls -lah \{\} \;")
```

cvmfs/cvmfs_config
```diff
@@ -1466,6 +1473,10 @@ cvmfs_bugreport() {
   mkdir etc
   cp -r /etc/cvmfs etc/
 
+  echo "Gathering /var/run/cvmfs"
+  mkdir -p var/run/cvmfs
+  cp -r /var/run/cvmfs var/run/cvmfs
```
vvolkl:
```
tar: bugreport/var/run/cvmfs/cvmfs/cvmfs.cvmfs-config.cern.ch: socket ignored
```
Let's silence this  - maybe with a `find -type f -exec cp ...`

DrDaveD: `find` with `-exec` can be really slow since it needs to start a program for every file.  I think it would be better to redirect stderr to /dev/null.

vvolkl: Yes, that works too. Although I expect there to be 1-3 files per repo, so performance isn't a big concern

HereThereBeDragons: sadly not a single line replacement to just surpress the warning

vvolkl (approved):

PR #3770, opened 2025-02-18
Removed performance-enum-size from clang-tidy

Siempie30:
Adding typed enums, which is suggested by the performance-enum-size check, is only supported since C++11, which is not yet supported by all the platforms that use CVMFS.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -19,6 +19,7 @@ Checks: -*
   ,-performance-enum-size
   ,-performance-inefficient-string-concatenation
   ,portability-*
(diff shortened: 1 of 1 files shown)


.clang-tidy
```diff
@@ -16,6 +16,7 @@ Checks: -*
   ,-misc-no-recursion
   ,-misc-non-private-member-variables-in-classes
   ,performance-*
+  ,-performance-enum-size
```
vvolkl, line 19: Maybe add a comment saying that this can be re-enabled with `--std=c++11`

vvolkl (approved):

PR #3771, opened 2025-02-20
Fix for `cvmfs_server transaction -t`: update repository state  after waiting for lease

vvolkl:
This is an important fix relevant when using the `cvmfs_server transaction -t`  option to wait for a lease on a publisher connected to a gateway. Without this patch, the transaction that is waited on can be lost (to be precise, it is reverted by the following transaction due to the wrong choice of the base of the catalog diff)

This is a rather narrow fix that updates the current readonly mount with the latest revision. We do this only when we are waiting for a lease to avoid the overhead of the manifest download and remount - to do that is really not required when we change a subrepository that isn't currently in a transaction by another publisher.

However, there might still be an unlikely race condition when the start of the transaction happens before an ongoing publication is concluded, but the check for the lease happens after. To be investigated.

A test reproducing this issue is included, but has yet to be added to the CI - needs a good way to build an image with the current cvmfs installed.

This issue was first reported in https://its.cern.ch/jira/projects/SPI/issues/SPI-2652

### DIFF
--- a/cvmfs/publish/repository.h
+++ b/cvmfs/publish/repository.h
@@ -359,7 +359,7 @@ class __attribute__((visibility("default"))) Publisher : public Repository {
   void CheckTagName(const std::string &name);
 
   void TransactionRetry();
-  void TransactionImpl();
+  void TransactionImpl(bool waiting_on_lease=false);
 
   SettingsPublisher settings_;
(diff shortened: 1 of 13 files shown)


vvolkl:
@cernvm-bot cloudtest

Issue #3773, opened 2025-02-25
Fedora 41 package error


Zeldhoron:
When trying to install cvmfs on Fedora 41 I run into the following error:
`nothing provides redhat-release >= 5 needed by cvmfs-release-4-1.noarch`

I tried following the instructions at https://cernvm.cern.ch/fs/ and https://www.eessi.io/docs/getting_access/native_installation/.
They required me to run the following commands (respectively):
```
# CernVM website:
sudo yum install https://cvmrepo.web.cern.ch/cvmrepo/yum/cvmfs-release-latest.noarch.rpm
# EESSI website:
sudo yum install -y https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest.noarch.rpm
```
Before running the following:
`sudo yum install -y cvmfs`
Which then resulted in the error I got.

vvolkl:
Thanks for the report. We actually need to build separate release packages for fedora - I can update this fairly quickly.

vvolkl:
This should now be fixed with version 5-1 of the cvmfs-release package. The `latest` links have been updated to point to that version, so no update of the instructions is necessary. It wasn't necessary after all to build separate packages.

If you still see an issue, feel free to re-open.

PR #3774, opened 2025-02-25
[rpm] cvmfs-release: weaken dependencies on redhat-release et al.

vvolkl:
This allows one rpm to be installed on most rpm-based distros. The conditional boolean is a feature of `rpm` > 4.14, that's why I only recommend to install the various system-release packages. They are very likely installed anyway.


vvolkl:
Fixes #3773

PR #3779, opened 2025-03-03
Apply %pretrans transaction check in most subpackages

### LINKED ISSUE #3778: Upgrade of cvmfs+cvmfs_server while transaction open leaves unusable state
Doing a `dnf upgrade cvmfs` on a system that is used for publishing repositories while a transaction is open does succeed in upgrading cvmfs and cvmfs-libs but fails to upgrade cvmfs-server.  The upgrade error message is
```
Running transaction
  Running scriptlet: cvmfs-server-2.12.6-1.osg23.el8.x86_64                 1/1
     Found open CernVM-FS repository transactions.
     Please abort or publish them before updating CernVM-FS.
error: %pretrans(cvmfs-server-2.12.6-1.osg23.el8.x86_64) scriptlet failed, exit status 1

Error in PRETRANS scriptlet in rpm package cvmfs-server
```
Because cvmfs-libs is shared between cvmfs and cvmfs-server, that leaves the system in an inconsistent state where `cvmfs_server abort` fails with
```
/usr/bin/cvmfs_publish: error while loading shared libraries: libcvmfs_crypto.so.2.11.5: cannot open shared object file: No such file or directory
```

I don't have any ideas right now on how to prevent this situation from happening, but to clean up you can do
```
dnf downgrade cvmfs
cvmfs_server abort
dnf install cvmfs
```

DrDaveD:
This turns the transaction check that was only in `%pretrans server` into a macro and applied in `%pretrans` scriptlets for all the subpackages that need to have their version stay in sync.

- Fixes #3778

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -70,7 +70,7 @@
 Summary: CernVM File System
 Name: cvmfs
 Version: 2.12.7
-Release: 1%{?dist}
+Release: 2%{?dist}
 URL: https://cernvm.cern.ch/fs/
 Source0: https://ecsft.cern.ch/dist/cvmfs/%{name}-%{version}/%{name}-%{version}.tar.gz
 %if 0%{?selinux_cvmfs}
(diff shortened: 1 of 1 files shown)


DrDaveD:
@cernvm-bot  cloudtest

vvolkl (approved):
LGTM. Does it need to be a patch release?

DrDaveD:
> LGTM. Does it need to be a patch release?

A revision increase is sufficient because it only modifies the package.  It should be released though.

PR #3782, opened 2025-03-13
Tidy Fixes for #3562

vvolkl:


vvolkl:
There's still a few things to fix here. I'll close this and make separate PRs that fix different tidy checks, I think that's the only way to be able to review those.

PR #3783, opened 2025-03-13
wipecache: make sleep only required for mac

HereThereBeDragons:


cvmfs/cvmfs_config
```diff
@@ -1213,7 +1213,9 @@ cvmfs_reload() {
           rm -rf "$dir" && \
             mkdir -p "$dir" && \
             chown $CVMFS_USER "$dir"
-            sleep 15
+            if [ "$sys_arch" = "Darwin" ]; then
```
vvolkl, line 1216:
```suggestion
            if [ x"$sys_arch" = x"Darwin" ]; then
```

vvolkl:
Thanks!

cvmfs/cvmfs_config
vvolkl: Well, I guess that's actually not needed if properly quoted.

vvolkl (approved):

Issue #3787, opened 2025-03-14
cmake: fuse3 detection does not correctly check for minor version


HereThereBeDragons:
The cmake `try_compile()`  to check if fuse3 is version >= 3.12 to enable `fuse_session_loop_mt` is incorrect and always fails.

PR #3788, opened 2025-03-14
fix cmake fuse3 detection + allow for manual overwrite of fuse3 lib

HereThereBeDragons:
Issue #3787 

- `try_compile` would always fail because of missing flags. as such HAS_FUSE_LOOP_CONFIG would never be enabled
- even with this fix `HAS_FUSE_LOOP_CONFIG` was not correctly propagated to all cvmfs_fuse3 libraries (only the stubs had it correctly included)

for easier custom fuse3 library:
- make sure that if `FUSE3_INCLUDE_DIR` and `FUSE3_LIBRARY` are only set by the system if not already set.
- before `FUSE3_LIBRARY` was always overwritten by the automatic detection.. still surprised that in most cases it found the right library when running a custom version

Needs at least cmake version 3.20 because of `cmake_path`  -- which is the default from centos 8 stream on. 
as we drop the cc7 support i guess that is ok? (cc7 has cmake 2.8.xx as default)


vvolkl:
> Needs at least cmake version 3.20 because of cmake_path -- which is the default from centos 8 stream on.
as we drop the cc7 support i guess that is ok? (cc7 has cmake 2.8.xx as default)

Let's wait a bit with dropping compatibility for cc7. Can you find another solution and leave the cmake_path line as a comment?

vvolkl (changes requested):
See comment above

PR #3791, opened 2025-03-20
Install apparmor profile for fusermount3/cvmfs

vvolkl:
Newer versions of apparmor come with a fusermount3 profile that prohibits mounts outside of certain directories like /mnt. This means that `/cvmfs/*` cannot be mounted on Ubuntu 24.10 and newer.

This patch installs an extension to the apparmor profile that fixes this issue. I've tried to upstream the change: https://gitlab.com/apparmor/apparmor/-/merge_requests/1587

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -365,6 +365,13 @@ if (BUILD_CVMFS)
     DESTINATION /usr/lib/systemd/system
     PERMISSIONS   OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ
   )
+  # Apparmor fusermount3 profile extension
+  install (
+    FILES pkg/apparmor.local.fusermount3
+    DESTINATION /etc/apparmor.d/local
+    RENAME fusermount3
+    PERMISSIONS   OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ
+  )
 
 endif (BUILD_CVMFS)
 
--- a/cvmfs/pkg/apparmor.local.fusermount3
+++ b/cvmfs/pkg/apparmor.local.fusermount3
@@ -1,0 +1,2 @@
+# Added by cvmfs to allow mounts on /cvmfs
+mount fstype=@{fuse_types} options=(nosuid,nodev,ro) -> /cvmfs/**/, 
--- a/packaging/debian/cvmfs/conffiles
+++ b/packaging/debian/cvmfs/conffiles
@@ -1,0 +1,1 @@
+/etc/apparmor.d/local/fusermount3
--- a/packaging/debian/cvmfs/cvmfs.install.in
+++ b/packaging/debian/cvmfs/cvmfs.install.in
@@ -24,3 +24,4 @@ etc/cvmfs/default.conf
 etc/cvmfs/default.d/README
 etc/bash_completion.d/cvmfs
 usr/lib/systemd/system/cvmfs-reload.service
+etc/apparmor.d/local/fusermount3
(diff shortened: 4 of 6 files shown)


vvolkl:
@cernvm-bot cloudtest

DrDaveD:
Hang on.  Why is fusermount3 being used at all?  When a mount is run as root or in a root-mapped unprivileged user+mount namespace, fusermount3 should not be used.  Under what circumstance is fusermount3 invoked?

DrDaveD:
fusermount3 is used with cvmfsexec "mode 1", but then it gets used by an unprivileged user to mount somewhere they have write access to, not /cvmfs.

vvolkl:
That's an interesting question. It could be that the fusermount3 profile applies for fuse mounts in general (just by denying mount operations outside of the explicitly allowed ones). Are you sure that fusermount3 is never invoked by the mount command behind the scenes?

DrDaveD:
fusermount's only purpose is as an enhanced-capability (setuid root) program for unprivileged users.  It is not used by the fuse library when a fuse program is started as root nor in a root-mapped user namespace, I am sure.  The primary cvmfsexec mode (mode 3, using the cvmfsexec script) depends on that behavior.  apptainer does not allow setuid-root programs, yet cvmfsexec works within it, even in containers that do not have fusermount or fusermount3 installed.

DrDaveD:
The thing about Ubuntu 24.04 is that it does not allow unprivileged user namespaces by default, except for programs that are listed in apparmor.   That's going to interfere with cvmfsexec unless it is turned off.

It sounds like Ubuntu 24.10 has a new restriction on fusermount/fusermount3 but I can't think of a scenario where that could be used to mount to /cvmfs.  Does it apply to all FUSE mounts, not just those using fusermount/fusermount3?

DrDaveD:
I just installed a fresh ubuntu 24.10 VM with vagrant (box alvistack/ubuntu-24.10) and then installed cvmfs using the standard apt instructions.  It installed the Ubuntu 24.04 binaries.  I was able to do `sudo mount -t cvmfs cvmfs-config.cern.ch /cvmfs/cvmfs-config.cern.ch/` and it mounted fine without needing any apparmor modifications.

On the other hand, it did install the fuse3 package including fusermount3.  When I renamed that binary to something else, the mount command complained about it being missing, so I was wrong about it not being used by the standard mount command.  I think that must be because of something that mount.cvmfs is doing which I don't yet understand.  cvmfsexec bypasses that and directly invokes the cvmfs2 program.

DrDaveD:
Oh. it's because of the uid=/gid= options that mount.cvmfs passes to the cvmfs2 program, causing it switch the user id before it invokes the fuse library.  

That invocation of fusermount3 can be avoided (with fuse3 only, not fuse2) by doing the mount command first as root and passing in a string listing /dev/fd/N where N is an open file descriptor to the mountpoint, in place of the mountpoint parameter.  I think that would be a better fix than messing with apparmor.  Probably with that we would be able to drop the dependency on the fuse3 package completely, using only fuse3-libs.

One thing that's still not clear to me is why fusermount3 allows the unprivileged cvmfs user to mount to a mountpoint owned by root.  There's a `grab_mountpoint` option also passed in from mount.cvmfs which tells cvmfs2 to chown the mountpoint to the give uid, but the comment says that's needed for autofs. and if I do a manual mount and leave the option off it still works.  Oh, I found out with strace that it's because the fuse library is invoked only with the effective user id as cvmfs; the real uid is still root.  The [fusermount code](https://github.com/libfuse/libfuse/blob/fuse-3.3.0/util/fusermount.c#L932) accepts any target if the real user id is root.

vvolkl:
Reverting this PR as it is no longer needed following the better fix by Dave.

PR #3792, opened 2025-03-20
Fix  `ingest -d` with `:` in filename

vvolkl:
Not that it's a good idea to use `:` in filenames, but we unfortunately do this ourselves for symlinks to container images. This fixes ingest to be able to delete paths with colons in them. The new delimiter used is `///` because it's one of the few characters that's not allowed in filenames, and we can safely sanitize the pathname before joining it with the delimiter.


vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3793, opened 2025-03-20
Fedora 41 support

andrey-utkin:
Trivial patches fixing issues encountered while doing `cmake -DBUILD_UNITTESTS=yes ../ && make` on Fedora 41.

### DIFF
--- a/cvmfs/swissknife_lease_json.h
+++ b/cvmfs/swissknife_lease_json.h
@@ -7,6 +7,7 @@
 
 #include "swissknife_lease_curl.h"
 
+#include <cstdint>
 #include <string>
 
 enum LeaseReply {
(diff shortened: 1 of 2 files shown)


vvolkl:
Thanks, but there shouldn't be a problem with Fedora 41 (we're building packages for it). I have to double-check if openssl engine can be safely removed, but I don't think so. Just install

```
dnf -y install openssl-devel-engine
```

vvolkl:
@cernvm-bot cloudtest

cvmfs/swissknife_lease_json.h
```diff
@@ -7,6 +7,7 @@
 
 #include "swissknife_lease_curl.h"
 
+#include <cstdint>
```
vvolkl:
```suggestion
```
Let's keep this out of this PR - That won't work on centos7, and we may want to do one more release for that platform.

vvolkl (commented):
Seems like the openssl/engine header is indeed not needed.

vvolkl:
@cernvm-bot cloudiest

vvolkl:
@cernvm-bot  cloudtest

PR #3805, opened 2025-04-04
Improve args docs of ducc commands

maxnoe:
- Use `<>` for required args 
- Add missing required args to some usage messages


vvolkl (approved):
Great, thanks a lot!

Issue #3780, opened 2025-03-04
cvmfs_ducc: add option for putting images into a subdirectory


maxnoe:
For us, it would be preferable I think to be able to use a single cvmfs repository but then the structure that ducc creates seems somewhat confusing to users.

We would prefer the option to use a subdirectory for the output of ducc.

Currently it seems not such option exists.

I am somewhat proficient in go and could give it a try (or go... if you like bad puns).

Looking through the code, it would probably mean

- add a subdirectory option to the wish struct
- add command line options for the scripts
- extend the wishlisht format to include it

vvolkl:
Sounds good. Instead of a dedicated subdirectory option, you could also consider to allow an extended repository name, like `cvmfs_ducc convert-single-image unpacked.repo.org/path/to/subdir`. That somewhat mirrors the cvmfs_transaction command when publishing to a gateway.

maxnoe:
> like cvmfs_ducc convert-single-image unpacked.repo.org/path/to/subdir

That's actually what I tried. So It seems to be an intuitive option, so maybe we start with that

maxnoe:
@vvolkl How would you handle changing the verification steps like `RepositoryExists` for this change?

Should we split-off the sub directories before passing? Or inside the function?

Can we assumed that a cvmfs repository is always `/cvmfs/<filesystem name>` or do we need to support different mount paths?

I.e. would it be correct to have a function:
```
func ExtractRepo(path string) string {
	elems := strings.Split(path, "/")
	return strings.Join(elems[:3], "/")
}
```
?

maxnoe:
Nevermind, `repo` is just the name, without the `/cvmfs` prefix, so we can just remove the part after `/`.

PR #3806, opened 2025-04-04
Allow subdirectories for ducc

maxnoe:
I start to tackle #3780 

How is this tested? Can I add a test case somehow for using ducc with a subdirectory?

### DIFF
--- a/ducc/cmd/convert_single_image.go
+++ b/ducc/cmd/convert_single_image.go
@@ -58,8 +58,8 @@ var convertSingleImageCmd = &cobra.Command{
 		}
 
 		if !cvmfs.RepositoryExists(cvmfsRepo) {
-			l.Log().Error("The repository does not seem to exists.")
-			return fmt.Errorf("The repository does not seem to exists.")
+			l.Log().Errorf("The repository %s does not seem to exist.", cvmfsRepo)
+			return fmt.Errorf("The repository %s does not seem to exist.", cvmfsRepo)
 		}
 
 		input, err := lib.ParseImage(inputImage)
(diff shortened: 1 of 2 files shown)


vvolkl:
Yes, the unpacker is a bit undertested, but I've started adding tests with #3829 . Unfortunately the way I'm mocking the cvmfs commands and repository directory might interfere with testing this functionality, so I'll look to add a full integration test with a proper cvmfs repository as well.

But having looked  at the code base a bit more, I found several circumstances where the repository name is used.
* It's used in a sanity check that checks if the string is contained  in cvmfs_server list https://github.com/cvmfs/cvmfs/blob/devel/ducc/cvmfs/raw.go#L131 The subdirectory needs to be trimmed here
* It's used as an argument for other cvmfs_server commands, like cvmfs_server transaction, cvmfs_server publish and ingest. At least cvmfs_server transaction works with the "extended reponame" - the idea is if the repository is connected to a gateway, it will get a lease only for the subdirectory, otherwise it will ignore it. I'm not sure about publish and ingest, needs to be checked. I think the template transactions still need to be prefixed with subdir
* It's used to find the upperdir for the overlay of cvmfs_server: https://github.com/cvmfs/cvmfs/blob/devel/ducc/cvmfs/cvmfs.go#L515 Here the subdir needs to be trimmed


All in all I think it makes sense to keep reponame and subdir for unpacking separate. I think it'll help to add a config package like this: https://github.com/cvmfs/cvmfs/tree/rest_api3/ducc/config I'll cherry-pick that from this bigger changeset

maxnoe:
> All in all I think it makes sense to keep reponame and subdir for unpacking separate. 

So instead of appending to the repo string, we add a new configuration value?

vvolkl:
Superseded by #4076

PR #3809, opened 2025-04-15
Fix unittests linkage on Gentoo

andrey-utkin:
It gave to the linkage command the protobuf lib path only, not the pkgconfig-provided list of flags to supply.

pkg-config metadata of protobuf package on Gentoo system indicates that linker flags for transitive deps, specifically libabsl* are necessary.

------------------

Reviewed by Gentoo package maintainer https://github.com/gentoo/gentoo/pull/41411#discussion_r2044480304

### DIFF
--- a/cmake/Modules/FindProtobuf.cmake
+++ b/cmake/Modules/FindProtobuf.cmake
@@ -214,28 +214,43 @@ endif()
 #    if the debug one is specified also include debug/optimized keywords
 #    in *_LIBRARIES variable
 function(_protobuf_find_libraries name filename)
-   find_library(${name}_LIBRARY
-       NAMES ${filename}
-       PATHS ${PROTOBUF_SRC_ROOT_FOLDER}/vsprojects/${_PROTOBUF_ARCH_DIR}Release)
-   mark_as_advanced(${name}_LIBRARY)
+  find_package(PkgConfig)
+  if (PKG_CONFIG_FOUND)
+    pkg_check_modules(${name} IMPORTED_TARGET GLOBAL ${filename})
+    if (${${name}_FOUND})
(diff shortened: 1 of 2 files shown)


andrey-utkin:
Please don't merge yet:
I noticed that fix-unittests-linkage.patch breaks mounting.

```
gentoo ~ # cvmfs2 -d atlas.cern.ch /cvmfs/atlas.cern.ch
Debug: using library /usr/lib64/libcvmfs_fuse3_stub.so
CernVM-FS: running in debug mode
CernVM-FS: loading Fuse module... failed to load cvmfs library, tried: './libcvmfs_fuse3_debug.so' '/usr/lib/libcvmfs_fuse3_debug.so' '/usr/lib64/libcvmfs_fuse3_debug.so'
./libcvmfs_fuse3_debug.so: cannot open shared object file: No such file or directory
/usr/lib/libcvmfs_fuse3_debug.so: cannot open shared object file: No such file or directory
/usr/lib64/libcvmfs_fuse3_debug.so: undefined symbol: _ZN6google8protobuf8internal28init_protobuf_defaults_stateE
```

vvolkl:
It looks like the error during mounting is due to a `${PROTOBUF_LITE_LIBRARY}` in `cvmfs/CMakeLists.txt` that should also be changed to `${PROTOBUF_LITE_LIBRARIES}`. The protobuf build system changed quite a bit on devel in the meantime, but I think the issue with the unittest linkage this PR tries to adress is still there. The fix should now be to prefer the upstream Protobuf module `find_package(Protobuf CONFIG QUIET)` that correctly links absl instead of the vendored one in `externals/protobuf/CMakeLists.txt` that I added to better deal with some system packages on RHEL. I'll open another PR with the fix I mentioned.

PR #3815, opened 2025-04-24
Fix lock between gc and check

DrDaveD:
This moves acquiring the gc lock earlier in the check function, so it covers all the check steps and not just one.  This has been broken ever since the lock was implemented in #2850, and we have experienced numerous check failures because of it.

### DIFF
--- a/cvmfs/server/cvmfs_server_check.sh
+++ b/cvmfs/server/cvmfs_server_check.sh
@@ -40,6 +40,14 @@ __do_check() {
     url=$CVMFS_STRATUM0
   fi
 
+  if [ "x$tag" = "x" ] && is_garbage_collectable $name; then
+    # acquire gc lock
+    # waits for gc on the same repository to finish
+    # and prevents a gc from starting
+    acquire_gc_lock $name check || die "Failed to acquire gc lock for $name"
+    trap "release_gc_lock $name" EXIT HUP INT TERM
+  fi
+
   # do it!
   if [ $check_integrity -ne 0 ]; then
     if ! is_local_upstream $upstream; then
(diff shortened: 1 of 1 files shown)


DrDaveD:
Closing & reopening to retry CI checks

vvolkl:
I rebuilt our build images, that caused a few strange issues, which should be fixed with https://github.com/cvmfs/cvmfs/pull/3816  . Btw, you can also retrigger gh actions checks from the "Checks" tab with the "Rerun all workflows" button

vvolkl:
@cernvm-bot unittest

vvolkl (changes requested):
Ah, please use devel as the target for PRs. I prefer to backport patches to cherry-picking from older branches.


In particular for this PR, we should be good to release 2.13 next week instead of making another patch release.

vvolkl:
The CI runs need some fixes from devel.

DrDaveD:
@cernvm-bot cloudtests

DrDaveD:
@cernvm-bot cloudtest

DrDaveD:
@cernvm-bot cloudtest

DrDaveD:
> Btw, you can also retrigger gh actions checks from the "Checks" tab with the "Rerun all workflows" button

It doesn't give me that option.  I must not have a high enough permission level.

vvolkl (approved):
LGTM. Do you still know what were the errors you were getting? I'm thinking if it's possible to add a regression test.

vvolkl:
@cernvm-bot cloudtest

DrDaveD:
> Do you still know what were the errors you were getting? I'm thinking if it's possible to add a regression test.

A regression test would be very challenging.  The check ran into missing files that the garbage collection removed.  A regression test would have to create tons of files and remove them in gc while doing checks on the same repository in parallel, and even then it probably wouldn't fail reliably.

DrDaveD:
I'm not sure why this was on the 2.14 milestone, because it was included in 2.13.0

PR #3816, opened 2025-04-25
ci: fix sudo in build image

vvolkl:
Rebuilding  the build images caused sudo to fail, with an error reminiscent of https://github.com/apptainer/apptainer/issues/2756 (although gh uses docker):
```
 [100%] Built target auto.cvmfs
sudo: PAM account management error: Authentication service cannot retrieve authentication info

sudo: a password is required
```
The chmod fixes this issue, although it's not yet clear to me why.


vvolkl:
@cernvm-bot unittest

vvolkl:
@cernvm-bot unittest

PR #3821, opened 2025-04-28
Re-enable bootstrapping of go

vvolkl:
Only if no recent installation found. Currently only for x86_64 and aarch64. ~Will need developers to remove golang from `externals_install/.bootstrapDone` for it to do anything.~ Renames the external to golang_rev2 so simply rerunning cmake should be enough.


vvolkl:
@cernvm-bot unittest

PR #3823, opened 2025-05-01
Fix gateway, swissknife stuff

andrey-utkin:


vvolkl:
@cernvm-bot cloudtest

cvmfs/swissknife_lease_curl.cc
```diff
@@ -39,9 +39,9 @@ size_t RecvCB(void* buffer, size_t size, size_t nmemb, void* userp) {
     return 0;
   }
 
-  my_buffer->data = static_cast<char*>(buffer);
+  my_buffer->data = my_buffer->data + std::string((const char*)buffer, nmemb);
```
vvolkl:
```suggestion
  my_buffer->data = my_buffer->data + std::string(static_cast<char*>(buffer), nmemb);
```
code linting

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl (approved):
That's a good catch! The original implementation may have assumed that the reply is short enough to only require one callback, but that seems clearly incorrect.

I just saw that we have further callback methods with the same issue, will have to check them all.

PR #3827, opened 2025-05-05
Containerized tests multistage dockerfile improvements

Siempie30:
Common container tests now use a multistage dockerfile. It copies the repository from the host and uses this for the build stage. The resulting rpm files are copied to a second stage, where they are installed. This multistage setup ensures the large amount of build files do not end up in the resulting container. Because the docker `COPY` command only works inside of the build context, the build context is changed to the repository root. The concurrent publishing tests require the same container setup, which is why this now also use this common dockerfile (instead of its own).


vvolkl (commented):
Great, thanks! Can you also update the Dockerfile in test/common/container (and deduplicate the Dockerfiles if possible)?

vvolkl (approved):
Added a few dependencies used for cloud tests ( we could also use the *_setup.sh script from test/cloudtesting/platforms, but I thought it was good to keep it minimal ).

Also removed a dependency by replacing wget with curl.

PR #3831, opened 2025-05-10
Decorate cvmfs_config killall with extra options -r(reset fuse) / -s(stuck fuse reset)

christge-sft:
[Issue 3807](https://github.com/issues/assigned?issue=cvmfs%7Ccvmfs%7C3807)

A blocked fuse connection could bring a cvmfs2 process in D-state (unresponsive and not kill-able).
Invoking `cvmfs_config killall -r/-s` aborts the unresponsive fuse connection (`-s`) or every cvmfs fuse connection (`-r`), allowing the cvmfs2 process to become responsive again.

For testing, we can not replicate the exact scenario of freezing a fuse connection, hence we bring the cvmfs2 process in D state by  creating and freezing another fs in `/tmp/freezeme/` and instructing cvmfs2 to interact with this frozen fs.

**NOTE**: The solution is not tested with mount points outside `/cvmfs`.

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -99,6 +99,8 @@ var_list="$parm_list $switch_list"
 
 cvmfs_fs_bundle="/Library/Filesystems/cvmfs.fs/Contents/Resources" # Mac OS X
 
+PROCNAME=cvmfs2
+MOUNTPOINT=/cvmfs
 IS_KILLALL_STUCK=FALSE
 IS_KILLALL_FORCE=FALSE
 
@@ -106,7 +108,11 @@ IS_KILLALL_FORCE=FALSE
 # -f(orce): abort all cvmfs associated fuse devices
 # -s(tuck): abort all cvmfs associated fuse devices in waiting state
 parse_killall_opts(){
-  while [[ "${1:-}" == -* ]]; do
+  for arg in "$@"; do
+    [[ "$arg" == "killall"  ]] && shift && break
+    shift
+  done
+  while [[ $# -gt 0  ]]; do
     case "$1" in
       "-f")
         IS_KILLALL_FORCE=true
@@ -119,9 +125,9 @@ parse_killall_opts(){
         ;;
       *)
         echo "Ignoring unknown option $1"
+        break
         ;;
     esac
-    [ $# -gt 0 ] || break
     shift
   done
 }
@@ -157,7 +163,7 @@ get_cvmfs_devices()
       device_id=$(echo "$line" | cut -d' ' -f 3)
       current_mountpoint=$(echo "$line" | cut -d' ' -f 5)
 
-      if [[ "$current_mountpoint" == "$MOUNTPOINT"/*  ]]; then
+      if [[ "$current_mountpoint" == "$MOUNTPOINT/"*  ]]; then
         if ! grep -q "$device_id" <<< "$cvmfs_devices"; then
           cvmfs_devices=$cvmfs_devices" $device_id"
         fi
@@ -165,6 +171,27 @@ get_cvmfs_devices()
     done < /proc/$pid/mountinfo
   done
   echo $cvmfs_devices
+}
+
+get_devices_map()
+{
+  local devices_map=""
+  for pid in $(pidof "$PROCNAME")
+  do
+    while read -r line; do
+      # Get the interesting fields from every line of mountinfo;
+      # those are the mountpoint and the device id
+      device_id=$(echo "$line" | cut -d' ' -f 3)
+      current_mountpoint=$(echo "$line" | cut -d' ' -f 5)
+
+      if [[ "$current_mountpoint" == "$MOUNTPOINT/"*  ]]; then
+        if ! grep -q "$device_id" <<< "$devices_map"; then
+          devices_map="$current_mountpoint>$device_id "$devices_map
+        fi
+      fi
+    done < /proc/$pid/mountinfo
+  done
+  echo $devices_map
 }
 
 abort_all_devices()
@@ -180,14 +207,18 @@ abort_all_devices()
 abort_stuck_devices()
 {
   echo "Aborting stuck cvmfs fuse devices.."
-  local devices="$@"
-  for device in $devices
+  local device_map="$@"
+  for map_entry in $device_map
   do
+    local mountpoint=$(echo $map_entry | cut -d'>' -f 1)
+    local device=$(echo $map_entry | cut -d'>' -f 2)
+
     fuse_dev=$(get_minor_device_id $device)
     local fuse_dev_path="/sys/fs/fuse/connections/$fuse_dev"
     # Abort waiting devices only
     if [ "$(cat $fuse_dev_path/waiting)" != "0" ] ;
     then
+      echo "Aborting $mountpoint"
       abort_fuse_device $fuse_dev
     fi
   done
@@ -1482,11 +1513,14 @@ cvmfs_killall() {
   local pid=$$
 
   parse_killall_opts "$@"
-  local cvmfs_devices=$(get_cvmfs_devices)
+
+
   if [ ${IS_KILLALL_FORCE} = true ] ; then
+    local cvmfs_devices=$(get_cvmfs_devices)
     abort_all_devices $cvmfs_devices
   elif [ ${IS_KILLALL_STUCK} = true ] ; then
-    abort_stuck_devices $cvmfs_devices
+    local devices_map=$(get_devices_map)
+    abort_stuck_devices  $devices_map
   fi
 
   echo -n "Terminating cvmfs_config processes... "
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -99,11 +99,6 @@ var_list="$parm_list $switch_list"
 
 cvmfs_fs_bundle="/Library/Filesystems/cvmfs.fs/Contents/Resources" # Mac OS X
 
-PROCNAME=cvmfs2
-MOUNTPOINT=/cvmfs
-IS_KILLALL_STUCK=FALSE
-IS_KILLALL_FORCE=FALSE
-
 # Check if killall has extra arguments
 # -f(orce): abort all cvmfs associated fuse devices
 # -s(tuck): abort all cvmfs associated fuse devices in waiting state
@@ -114,11 +109,11 @@ parse_killall_opts(){
   done
   while [[ $# -gt 0  ]]; do
     case "$1" in
-      "-f")
-        IS_KILLALL_FORCE=true
+      "-r")
+        abort_all_devices $(get_cvmfs_devices)
         ;;
       "-s")
-        IS_KILLALL_STUCK=true
+        abort_stuck_devices $(get_devices_map)
         ;;
       "--help")
         cvmfs_config_usage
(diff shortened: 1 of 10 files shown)


cvmfs/talk.cc
```diff
@@ -426,6 +426,8 @@ void *TalkManager::MainResponder(void *data) {
         mount_point->download_mgr()->SetDnsServer(host);
         talk_mgr->Answer(con_fd, "OK\n");
       }
+    } else if (line == "__testing_freeze_cvmfs") {
+      mkdir("/tmp/freezeme/dir",0700);
```
vvolkl: Better to use mkdirtemp instead of the hardcoded directory name.

The pathname could be passed in via argument to cvmfs_talk __testing_freeze_cvmfs

test/src/066-killall/main
```diff
@@ -3,9 +3,134 @@
 cvmfs_test_name="Killing all cvmfs2 instances"
 cvmfs_test_suites="quick"
 
+SRC="$(mktemp --suffix=FROZENFS)"
+DEST=/tmp/freezeme
+MOUNTPOINT=/cvmfs
+
+fs_setup(){
```
vvolkl, line 14: Needs a comment explaining what is happening. I'd also  link to the blogpost.

Maybe we can start adding a "docstring" to the top that explains what the test is trying to do.

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,100 @@ var_list="$parm_list $switch_list"
...
+        ;;
+    esac
+    [ $# -gt 0 ] || break
+    shift
+  done
+}
+
+# Given a device id in the form <Major>:<minor>
+# we need only the <minor> part to identify the fuse device
+get_fuse_device()
```
vvolkl: The naming of this function is not very precise. I'd propose something like `get_fuse_device_id_minor`

test/src/066-killall/main
```diff
@@ -3,9 +3,134 @@
...
+  logdebug "\tbefore killall $arg FUSE devices: $fuse_devices"
+
+  logdebug "Killing cvmfs using cvmfs_config killall $arg"
+  sudo cvmfs_config killall "$arg" &
+
+  unfreeze
+}
+
+logdebug(){
+  echo -e "\e[35m$1\e[0m"
```
vvolkl: Nice idea, but the output is often logged to log files or other sinks without support for colored output, where the colour codes make it less readable.

We can think about adding a solution at the level of the run.sh script which adds these only when we log to a terminal that supports them.

test/src/066-killall/main
```diff
@@ -3,9 +3,134 @@
...
+  freeze
+
+  logdebug "Mounting repositories and reloading.."
+  cvmfs_mount $repo || return 40
+  sudo cvmfs_config reload || return 41
+
+  logdebug "Freezing cvmfs.."
+  sudo cvmfs_talk -i $repo __testing_freeze_cvmfs &
+
+  local d_pid=$(get_D_state_pid)
```
vvolkl: Checking the process state  will be interesting to follow up in the future, but I don't think it's strictly necessary for the test. you could just do an `ls $repo` save the PID of that and then check that it is killed after the connection abort

vvolkl (changes requested):
Very nice! Using `mountinfo` for the device id was a great idea. A few small comments inline.

cvmfs/cvmfs_config
christge-sft: I decided to give this name because here `/sys/fs/fuse/connections/<device id>/` is the minor part of a the full device id (\<major\>:\<minor\>).

If this is not clear, I propose to rename the function to something more generic like `get_minor_device_id` since, if needed, it could be used for any device (even not fuse) to get its minor id.

test/src/066-killall/main
christge-sft: Right, I used that only because it helped me to jump directly to the killall-relevant parts! I see these options:

1. replace `logdebug` with `echo` everywhere
2. use `echo "$1"` as the production version of `logdebug`, allowing an easier upgrade to a more syntactically rich version
3. look for an environment variable like `ENABLE_COLORED_DEBUG_LOGGING` to log with the colored version and go with the vanilla version if not set.

I think one of the three can easily be included in this PR.

cvmfs/cvmfs_config
vvolkl: That's a good name I think!

test/src/066-killall/main
vvolkl: Yes, those are all good options. We can discuss the best solution for the tests in general

cvmfs/cvmfs_config
```diff
@@ -150,7 +244,7 @@ cvmfs_config_usage() {
   echo "  reload [-c | <repository>]"
   echo "  umount"
   echo "  wipecache"
-  echo "  killall"
+  echo "  killall [-f(orce) ] [-s(tuck)]"
```
DrDaveD: I think the name `-f` isn't a sufficient description of what it's doing.  How about `[-a(llfuse)]`.

DrDaveD (changes requested):

cvmfs/cvmfs_config
christge-sft: @vvolkl what's your take on that?

If we want to be super precise, I think it should be `[ -ff(fuse first)] [-sff(stuck fuse first)]`

christge-sft: Also, @DrDaveD what do you think about the alternatvies? (I just wanted to make sure that Valentin is also in the loop, that's why I tagged him)

DrDaveD: I don't think we should use multi-character options with a single hyphen, that's non-standard and not the case for most cvmfs options.  However now I see that "allfuse" is bad too because I misunderstood and thought it applied all fuse on the host, not just cvmfs.  Maybe it should be just `-a(ll)` and `-s(tuck)`.

vvolkl: Let's stick to the original naming - I think Dave assumed that all fuse connections are going to be aborted, while we only abort those with a mountpoint in /cvmfs

DrDaveD: Or since the command name already includes "all" how about `-r(esetfuse)` and `-s(tuckfusereset)`.

vvolkl: That's good too!

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,131 @@ var_list="$parm_list $switch_list"
 
 cvmfs_fs_bundle="/Library/Filesystems/cvmfs.fs/Contents/Resources" # Mac OS X
 
+PROCNAME=cvmfs2
+MOUNTPOINT=/cvmfs
+IS_KILLALL_STUCK=FALSE
+IS_KILLALL_FORCE=FALSE
```
vvolkl:
```suggestion
```
I think it'd be better to avoid these global variables. PROCNAME and MOUNTPOINT are essentially const, and can just be hardcoded where they are used, that makes the code more readable.

To avoid  `IS_KILLALL*` you could call `abort_*devices` directly from `parse_killall_args`.

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,131 @@ var_list="$parm_list $switch_list"
...
+# -s(tuck): abort all cvmfs associated fuse devices in waiting state
+parse_killall_opts(){
+  for arg in "$@"; do
+    [[ "$arg" == "killall"  ]] && shift && break
+    shift
+  done
+  while [[ $# -gt 0  ]]; do
+    case "$1" in
+      "-ff")
+        IS_KILLALL_FORCE=true
```
vvolkl:
```suggestion
        abort_all_devices $(get_cvmfs_devices)
```

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,131 @@ var_list="$parm_list $switch_list"
...
+        ;;
+      "-sff")
+        IS_KILLALL_STUCK=true
```
vvolkl:
```suggestion
        abort_stuck_devices $(get_cvmfs_devices)
```

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,131 @@ var_list="$parm_list $switch_list"
...
+  local cvmfs_devices=""
+  for pid in $(pidof "$PROCNAME")
+  do
+    while read -r line; do
+      # Get the interesting fields from every line of mountinfo;
+      # those are the mountpoint and the device id
+      device_id=$(echo "$line" | cut -d' ' -f 3)
+      current_mountpoint=$(echo "$line" | cut -d' ' -f 5)
+
+      if [[ "$current_mountpoint" == "$MOUNTPOINT/"*  ]]; then
```
vvolkl:
```suggestion
      if [[ "$current_mountpoint" == "/cvmfs/"*  ]]; then
```

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,131 @@ var_list="$parm_list $switch_list"
...
+  local devices_map=""
+  for pid in $(pidof "$PROCNAME")
+  do
+    while read -r line; do
+      # Get the interesting fields from every line of mountinfo;
+      # those are the mountpoint and the device id
+      device_id=$(echo "$line" | cut -d' ' -f 3)
+      current_mountpoint=$(echo "$line" | cut -d' ' -f 5)
+
+      if [[ "$current_mountpoint" == "$MOUNTPOINT/"*  ]]; then
```
vvolkl:
```suggestion
      if [[ "$current_mountpoint" == "/cvmfs/"*  ]]; then
```

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,131 @@ var_list="$parm_list $switch_list"
...
+      fi
+    done < /proc/$pid/mountinfo
+  done
+  echo $cvmfs_devices
+}
+
+get_devices_map()
+{
+  local devices_map=""
+  for pid in $(pidof "$PROCNAME")
```
vvolkl:
```suggestion
  for pid in $(pidof "cvmfs2")
```

cvmfs/cvmfs_config
```diff
@@ -99,6 +99,131 @@ var_list="$parm_list $switch_list"
...
+    echo "OK"
+  fi
+}
+
+# Collect every device associated to a cvmfs mount point,
+# based on /proc/<cvmfs2_pid>/mountinfo
+get_cvmfs_devices()
+{
+  local cvmfs_devices=""
+  for pid in $(pidof "$PROCNAME")
```
vvolkl:
```suggestion
  for pid in $(pidof "cvmfs2")
```

vvolkl (changes requested):

vvolkl (approved):

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

vvolkl:
@cernvm-bot cloudtest

PR #3833, opened 2025-05-15
Add  manpages for some cvmfs commands

vvolkl:
Adds support for:
* `man cvmfs2`
* `man cvmfs_config`
* `man cvmfs_server`
* `man cvmfs_swissknife`

This is mainly for compliance with packaging guidelines of linux distro repositories. The man pages don't add much more information than what is currently available in the `--help` output, and use `help2man` in order to keep the information in sync with the codebase.

The `--help` output needs to be improved in a future patch - in particular cvmfs_server and swissknife should show the usage for individual subcommands and not just one long list with all of them.

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -286,7 +286,7 @@ cvmfs_config_usage() {
   echo ""
   echo "  killall [-r(reset fuse) ] [-s(reset stuck fuse)]"
   echo ""
-  echo "  bugreport <timeout_in_sec_per_command (default = 90)>"
+  echo "  bugreport [timeout_in_sec_per_command (default = 90)]"
 }
 
 has_selinux() {
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -257,22 +257,35 @@ compare_versions() {
 cvmfs_config_usage() {
   echo "Common configuration tasks for CernVM-FS"
   echo "Usage: $0 <command>"
-  echo "Commands are"
+  echo "Commands:"
   echo "  setup [nouser] [nocfgmod] [nostart] [noautofs]"
+  echo ""
   if is_wsl2; then
     echo "  wsl2_start"
+    echo ""
   fi
   echo "  chksetup"
+  echo ""
   echo "  showconfig [-s(hort output, only non-empty parameters)] [<repository>]"
+  echo ""
   echo "  stat [-v | <repository>]"
+  echo ""
   echo "  status"
+  echo ""
   echo "  probe [<repository list>]"
+  echo ""
   echo "  fsck [-q(uick check for zero byte files only) | <cvmfs_fsck options>]"
+  echo ""
   echo "  fuser <repository>"
+  echo ""
   echo "  reload [-c | <repository>]"
+  echo ""
   echo "  umount"
+  echo ""
   echo "  wipecache"
+  echo ""
   echo "  killall [-r(reset fuse) ] [-s(reset stuck fuse)]"
+  echo ""
   echo "  bugreport <timeout_in_sec_per_command (default = 90)>"
 }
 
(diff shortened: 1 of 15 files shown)


doc/manpages/cvmfs.md
```diff
@@ -0,0 +1,9 @@
+# The CernVM Filesystem
+
+## Description
+
+The CernVM filesystem (CVMFS) is a global, readonly distributed file system optimized for software delivery - it lets you stream software from a remote server to your local machine on demand! 
+
+CVMFS is a filesystem in userspace that presents files on a remote server as if they were already on a local filesystem, and downloads them on demand. Unlike sshfs and similar tools, CernVM-FS needs the files on the server to be in a special format, the "CVMFS repository", which includes file metadata. The cvmfs_server command can be used to author and add files to CVMFS repositories.
+
+This lets
```
DrDaveD: Why does this end so abruptly?

DrDaveD (changes requested):
Could you please attach the generated man pages to a comment so I can see what they look like?

doc/manpages/cvmfs.md
vvolkl: Right, that file was included by accident. I was starting to write a manpage for cvmfs in section 5, but then ran out of time.

vvolkl:
I uploaded the generated manpages here: https://cernbox.cern.ch/s/kadkaP5f2yzqTOQ

View with `man cvmfs2.1` etc.

cvmfs/cvmfs_config
```diff
@@ -135,22 +135,35 @@ compare_versions() {
...
+  echo ""
   echo "  reload [-c | <repository>]"
+  echo ""
   echo "  umount"
+  echo ""
   echo "  wipecache"
+  echo ""
   echo "  killall"
+  echo ""
   echo "  bugreport <timeout_in_sec_per_command (default = 90)>"
```
DrDaveD: Put square brackets around the timeout since it is optional.

DrDaveD (changes requested):
It would be better if line breaks could be forced in the cvmfs_server page wherever they are in the script help.  It's pretty ugly as it is.

Reducing the `--------------` from the cvmfs_swissknife page to just `--` or something like that would also be good.  Also remove the version number from there if you can.

vvolkl:
> It would be better if line breaks could be forced in the cvmfs_server page wherever they are in the script help. It's pretty ugly as it is.

Yes, as I said, writing them properly will be a larger effort. I think the way to go is to add detailed usage information to the subcommands , so for example 

cvmfs_server mkfs --help

and only print a line for each subcommand on

cvmfs_server --help

That makes more sense to me than fiddling with the formatting, but I won't do it in this PR.

> 
> Reducing the `--------------` from the cvmfs_swissknife page to just `--` or something like that would also be good. Also remove the version number from there if you can.

Done

DrDaveD:
>>    It would be better if line breaks could be forced in the cvmfs_server page wherever they are in the script help. It's pretty ugly as it is.

> Yes, as I said, writing them properly will be a larger effort

Can't a filter be applied to insert a man page break (`.br`) or an extra blank line at every input line break?

vvolkl:
> Can't a filter be applied to insert a man page break (`.br`) or an extra blank line at every input line break?

I don't see any way to do this with [help2man](https://www.gnu.org/software/help2man/), which just allows you to include additional sections from a file. That leaves postprocessing of the troff file, and I think the time and effort would be better invested in adding proper usage strings for the individual subcommands.

DrDaveD:
You can process the input to help2man.  This inserts an extra blank line:
```
help2man "bash -c 'cvmfs_server|sed "s/$/\\\\n/"'"
```
It could be made less ugly by putting the pipleine in a script instead.

vvolkl:
> "bash -c 'cvmfs_server|sed "s/$/\\\\n/"'"

Ah nice, that's actually very simple and a definitive improvement. Added it to the PR, although I think it doesn't help that much. The manpages for server and swissknife are still impractical because too long  (600 and 1600 lines ...)

PR #3835, opened 2025-05-19
ci: Fix permissions in 81 and timeouts on find

vvolkl:
In a new cloudtesting setup that avoids the test/cloudtesting/platforms/*.sh scripts, there were several timeouts that seemed to come from `find` traversing more of the repository than intended.

The command find $repo | head -n 1 is indeed not guaranteed to stop the find after one file, and depends on line buffering of the shell.

This uses subdirectories of the repositories to find test files, to limit find and fix the tests. The downside is that those subdirectories could be deleted and the test logic needs to be updated, but that should be rare and not a problem.

### DIFF
--- a/test/src/081-shrinkwrap/main
+++ b/test/src/081-shrinkwrap/main
@@ -7,8 +7,8 @@ CVMFS_TEST081_REPO=
 
 cleanup() {
     echo "*** [Cleanup]"
-    sudo rm -f /tmp/cvmfs-$CVMFS_TEST081_REPO.trace.log 2> /dev/null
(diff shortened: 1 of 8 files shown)


vvolkl:
Example of this issue: https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsPRBuilderTest2/541/

vvolkl:
The old cloudtest does not have this problem: https://lcgapp-services.cern.ch/cvmfs-jenkins/job/CvmfsCloudTesting/2478/CVMFS_PLATFORM=el8,CVMFS_PLATFORM_CONFIG=x86_64,label=trampoline/artifact/centos8-x86_64_test_client.log

PR #3837, opened 2025-05-20
[server] add a logrotate config for the stats.db

### LINKED ISSUE #2845: Implement rotation of stats.db
The `/var/spool/cvmfs/<REPO>/stats.db` file should be rotated based on creation time and size, to avoid it growing unbounded.

The rotation thresholds should be configurable:
* `CVMFS_STATSDB_ROTATION_MAX_DAYS` - default 30
* `CVMFS_STATSDB_ROTATION_MAX_SIZE_BYTES` - default 5 * 1024 * 1024 (5 MB)

vvolkl:
Fixes #2845

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -379,6 +379,14 @@ if (BUILD_CVMFS)
   )
   endif()
 
+  # Apparmor fusermount3 profile extension
+  install (
+    FILES pkg/logrotate.d/cvmfs-statsdb
+    DESTINATION /etc/logrotate.d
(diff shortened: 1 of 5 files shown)


cvmfs/CMakeLists.txt
```diff
@@ -377,6 +377,13 @@ if (BUILD_CVMFS)
     RENAME fusermount3
     PERMISSIONS   OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ
   )
+
+  # Apparmor fusermount3 profile extension
```
vvolkl:
```suggestion
```

PR #3839, opened 2025-05-20
install default logrotate config for /var/log/cvmfs

vvolkl:
At a recent tutorial, we saw that there is some friction for new users in the way cvmfs_server snapshot errors when no logrotate config is present. There is the possibility to turn it into a warning, but having a sensible default config that can be configured seems better.


DrDaveD (changes requested):
I don't think it's a good idea to add a logrotate.d file that could clash with one installed by the admin.  I think it would be better to install a default one with the `cvmfs_server snapshot` command if one doesn't exist.

vvolkl:
So I think `%config(noreplace) %{_sysconfdir}/logrotate.d/cvmfs-logs` behaves correctly  in that it allows admins to delete or change it, and won't overwrite an existing file. I'm not sure what happens if an admin has an existing logrotate for /var/log/cvmfs installed under a different name, they might indeed clash. Maybe I should keep the name /etc/lograte.d/cvmfs that was suggested previously.

Installing it in cvmfs_server snapshot will suffer from the same issue though, or am I missing something?

vvolkl:
I've confirmed that this won't overwrite any existing configuration - it will be installed as /etc/logrotate.d/cvmfs.rpmnew and skipped by logrotate. 

I'll go ahead and merge, I'll add this change prominently to the release notes to make admins aware.

PR #3840, opened 2025-05-21
Minio S3 bucket for containerized testing

Siempie30:
This adds two containers to the docker compose file for concurrent publishing tests.
1. The `cvmfs-s3` container runs the minio server docker image, and acts as the S3 server.
2. The `cvmfs-s3-creator` container runs the minio client docker image, and is used to create a bucket on `cvmfs-s3`.

The gateway and publisher containers depend on the `cvmfs-s3-creator` having completed succesfully, so that it is guaranteed that the bucket is up and running before any tests are set up or started.
The `cvmfs-s3-creator` also depends on the minio server being up and healthy, in order to prevent attempting to create a bucket before the server is up.


vvolkl (approved):
Very nice, thanks!

Issue #2698, opened 2021-03-22
cvmfs rpm creation: use rpmlint to check cvmfs-universal.spec


skluth:
the script cvmfs/ci/makerpm.sh does the local rpm build. I would suggest to check at this point the validity of the cvmfs-universal.spec with the rpmlint command and fail the rpm-build in ci if rpmlint has complaints. This connects with my PR for updates to cvmfs-universal.spec for a successful build on OBS. Checking spec file consistency in ci keeps it clean and increases chances of a successful OBS/COPR builds

PR #3844, opened 2025-05-23
[rpm] add rpmlint check and small fixes to silence warnings

vvolkl:
`%check` sections  are needed to silence warnings, This macro is available as of rpm version 4.2, the workaround 
```
%check || :
```

to not introduce errors in earlier versions is taken from [here](http://ftp.rpm.org/max-rpm/s1-rpm-inside-scripts.html#FTN.AEN8053)

Upon suggestion by @traylenator I'm adding actual unittests to the check in cvmfs-universal.spec and ignoring the warning for the other spec files which don't have much to test.

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -507,8 +507,7 @@ EOF
 mkdir -p $RPM_BUILD_ROOT/var/lib/cvmfs-gateway
 %endif
 
-%clean
-rm -rf $RPM_BUILD_ROOT
+%check || :
 
 %post
 if [ $1 -eq 1 ]; then
@@ -764,7 +763,7 @@ systemctl daemon-reload
 %changelog
 # - When using fuse3, require at least version 3.3.0 (for premounting).
 * Mon Mar 3 2025 Dave Dykstra <dwd@cern.ch>> - 2.12.7-2
-- Apply %pretrans transaction check to all subpackages that need to
+- Apply pretrans transaction check to all subpackages that need to
   have their version stay in sync
 * Mon Dec 2 2024 Valentin Volkl <vavolkl@cern.ch> - 2.12
 - Add cvmfs-snapshotter package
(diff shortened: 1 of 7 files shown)


vvolkl:
Fixes #2698

packaging/rpm/cvmfs-universal.spec
```diff
@@ -507,8 +507,8 @@ EOF
 mkdir -p $RPM_BUILD_ROOT/var/lib/cvmfs-gateway
 %endif
 
-%clean
-rm -rf $RPM_BUILD_ROOT
+%check || :
+make check
```
traylenator: You might want to set an environment here if you have a suitable one so you used what is going to be installed rather than what is in the build tree where you are actually executing.

Taking a python example as they have better macros, you see it actually uses the `%buildroot` path as the python lib. 
```
rpm -E '%{pytest}'
\
  CFLAGS="${CFLAGS:-${RPM_OPT_FLAGS}}" LDFLAGS="${LDFLAGS:-${RPM_LD_FLAGS}}"\
  PATH="/home/straylen/rpmbuild/BUILDROOT/%{NAME}-%{VERSION}-%{RELEASE}.x86_64/usr/bin:$PATH"\
  PYTHONPATH="${PYTHONPATH:-/home/straylen/rpmbuild/BUILDROOT/%{NAME}-%{VERSION}-%{RELEASE}.x86_64/usr/lib64/python3.9/site-packages:/home/straylen/rpmbuild/BUILDROOT/%{NAME}-%{VERSION}-%{RELEASE}.x86_64/usr/lib/python3.9/site-packages}"\
  PYTHONDONTWRITEBYTECODE=1\
  \
  PYTEST_XDIST_AUTO_NUM_WORKERS=1 /usr/bin/pytest
```

vvolkl: Hmm, I think our unittests are completely statically linked, so won't check the installation at all. So questionable if we want to run them with rpmbuild actually. It won't hurt, but I'll take a look if I can change that.

DrDaveD (approved):

vvolkl:
Tests are failing because CreateUserNamespace et al. don't work in a docker container (I'm assuming rootless containers do work because the regular unittest CI job runs in podman). I think I'll just remove the %check for now and come back to it once the unittests run reliably on docker.

PR #3850, opened 2025-05-30
[ci] Add github actions workflow on ubuntu

vvolkl:
This is intentionally not using ci-scripts, in order to streamline the test infrastructure a bit. Also adds an experimental setup for a local squid proxy that gets preloaded from gh actions.


vvolkl:
@cernvm-bot unittest

PR #3872, opened 2025-06-16
Have premount check for -oro instead of !-orw to determine readonly

### LINKED ISSUE #3868: `futimes` error following fuse premount change introduced in 2.13
As originally reported by @tmadlener in SPI, the following fails with the docker runtime (although it works with podman-docker):

```
ubuntu@cvm-test-ubuntu25:~$ docker run -it --pull always  -e SHELL -u root --cpus=10 --security-opt seccomp:unconfined  -v /cvmfs:/cvmfs:shared --network host  ghcr.io/aidasoft/el9:latest bash
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Trying to pull ghcr.io/aidasoft/el9:latest...
Getting image source signatures
Copying blob d680b6993382 skipped: already exists  
Copying blob 356250a30fee skipped: already exists  
Copying blob 839eedb24528 skipped: already exists  
Copying blob b3403c049af6 skipped: already exists  
Copying blob 52c797bb34c4 skipped: already exists  
Copying config e32b153a9e done   | 
Writing manifest to image destination
[root@cvm-test-ubuntu25 /]#  source /cvmfs/sft.cern.ch/lcg/views/dev4/latest/x86_64-el9-gcc13-opt/setup.sh
[root@cvm-test-ubuntu25 /]# julia -e 'import Pkg'
ERROR: SystemError: futimes: Function not implemented
Stacktrace:
  [1] systemerror(p::Symbol, errno::Int32; extrainfo::Nothing)
    @ Base ./error.jl:176
  [2] systemerror
    @ ./error.jl:175 [inlined]
  [3] touch
    @ ./filesystem.jl:361 [inlined]
  [4] touch(path::String)
    @ Base.Filesystem ./file.jl:460
  [5] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt128, stalecheck::Bool; reasons::Dict{String, Int64}, DEPOT_PATH::Vector{String})
    @ Base ./loading.jl:2017
  [6] _require(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:2527
  [7] __require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:2388
  [8] #invoke_in_world#3
    @ ./essentials.jl:1089 [inlined]
  [9] invoke_in_world
    @ ./essentials.jl:1086 [inlined]
 [10] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:2375
 [11] macro expansion
    @ ./loading.jl:2314 [inlined]
 [12] macro expansion
    @ ./lock.jl:273 [inlined]
 [13] __require(into::Module, mod::Symbol)
    @ Base ./loading.jl:2271
 [14] #invoke_in_world#3
    @ ./essentials.jl:1089 [inlined]
 [15] invoke_in_world
    @ ./essentials.jl:1086 [inlined]
 [16] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:2260

```

A simpler reproducer is running 

```
#include <sys/time.h>
#include <fcntl.h>
#include <unistd.h>
#include <stdio.h>
#include <errno.h>int main(){
  // Open (or create) the file and get a valid fd
  int fd = open("/cvmfs/sft.cern.ch/README.md", O_WRONLY , 0666);
  if (fd < 0) { perror("open"); return 1; }  struct timeval tv[2] = {{0,0},{0,0}};
  int res = futimes(fd, tv);
  if (res < 0) {
    perror("futimes");
    return errno;
  }  printf("futimes OK\n");
  return 0;
}
```

which, in cvmfs-2.13.0 will fail with 
```
futimes: Function not implemented
```
instead of 
```
open: Read-only file system
```

I've bisected to problem and confirmed this was introduced by https://github.com/cvmfs/cvmfs/pull/3796

DrDaveD:
When doing premount look for the "-oro" fuse option instead of a lack of "-orw".  Apparently both are being included.  This was causing the mount to be done as read-write instead of read-only, and on Ubuntu 24 inside a docker container it ended up with a different error from the `futimes()` system call that julia was apparently not checking for.

- Fixes #3868

### DIFF
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -926,7 +926,7 @@ int FuseMain(int argc, char *argv[]) {
                                      ",default_permissions" : "",
       MatchFuseOption(mount_options, "allow_other") ? ",allow_other" : "");
     unsigned long flags = MS_NOSUID | MS_NODEV | MS_RELATIME;
-    if (!MatchFuseOption(mount_options, "rw")) {
+    if (MatchFuseOption(mount_options, "ro")) {
       flags |= MS_RDONLY;
     }
     if (mount("cvmfs2", mount_point_->c_str(), "fuse", flags, opts) == -1) {
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -926,6 +926,9 @@ int FuseMain(int argc, char *argv[]) {
                                      ",default_permissions" : "",
       MatchFuseOption(mount_options, "allow_other") ? ",allow_other" : "");
     unsigned long flags = MS_NOSUID | MS_NODEV | MS_RELATIME;
+    // Note that during the handling of the `CVMFS_MOUNT_RW` option, we ensure that
+    // at least one of `rw` or `ro` is part of the mount option string (we won't have both unset).
+    // If both `rw` and `ro` are set, the read-only option takes precedence.
     if (MatchFuseOption(mount_options, "ro")) {
       flags |= MS_RDONLY;
     }


cvmfs/loader.cc
```diff
@@ -926,7 +926,7 @@ int FuseMain(int argc, char *argv[]) {
                                      ",default_permissions" : "",
       MatchFuseOption(mount_options, "allow_other") ? ",allow_other" : "");
     unsigned long flags = MS_NOSUID | MS_NODEV | MS_RELATIME;
-    if (!MatchFuseOption(mount_options, "rw")) {
+    if (MatchFuseOption(mount_options, "ro")) {
```
jblomer, line 932: For clarity, I suggest adding a comment like

```
Note that during the handling of the `CVMFS_MOUNT_RW` option, we ensure that at least one of `rw` or `ro` is part of the mount option string (we won't have both unset). If both `rw` and `ro` are set, the read-only option takes precedence.
```

vvolkl (approved):
Confirmed that this fixes the originally reported issue. Will have to think of the best way to add a regression test

PR #3875, opened 2025-06-17
Add a simple CVMFS Client prometheus exporter

jcpunk:
The existing exporters I've found have a number of python dependencies which makes support more complex across Mac and Linux systems.

This exporter uses bash and POSIX tools to run in a fairly minimal environment.

If you run it "as is" you can dump the output to a file for inclusion by node_exporter. Alternatively, sample systemd unit files are included for setting this script up as a simple socket service. The selected port is already registered with Prometheus for CVMFS exporters.

In my perfect world the script and unit files would be packaged along side the client for both `deb` and `rpm` systems.


vvolkl:
This is great, looks very useful! I think it'd be best to keep add-ons like these in its own repository, similar to what we do for the collectd plugin. That doesn't  mean we cannot package and include it in the main package repo, if that is more convenient than the cvmfs-contrib repo.

I've just sent you an invitation for cvmfs-contrib/cvmfs-prometheus, feel free to push this version there. I can also help with adding packaging scripts.

jcpunk:
Pushed over to : https://github.com/cvmfs-contrib/prometheus-cvmfs/commit/707db1583812d97b90cf0bb56dc4f63240484b5f

jcpunk:
So long as it is packaged up somewhere I can find it, I can probably work from there.

PR #3879, opened 2025-06-21
Allow builds without libfuse2

### LINKED ISSUE #3866: Consider removing `Depends` in Debian packaging
As a followup to #3845:

I just realized that the [Debian packaging for apptainer](https://github.com/apptainer/apptainer/blob/main/dist/debian/control) doesn't specify any `Depends` at all, it only uses `Build-Depends` including `libfuse3-dev`.    Then it lets the build system figure out the corresponding `Depends`.  That seems to me to be superior to asking for either of the two install time dependencies, because for a particular built .deb package in fact only one of libfuse3-3 or libfuse3-4 will work, not either.

vvolkl:
Since hotpatching  requirements can make it difficult to remove libfuse2, we decided to build cvmfs without libfuse2 support for the platforms that will be added in cvmfs 2.13.1 (alma10, debian13 , fedora42, ubuntu25.04). This is a first step towards dropping libfuse2 which will remove quite a bit of complexity in the code.

The cvmfs-fuse3 package is kept separate for now, to minimize disruptions.

This PR:

* introduces a `-DBUILD_LIBFUSE2=no` cmake option that allows to not build the relevant libraries
* changes libfuse3 from an optional to a required dependency in cmake
* makes cvmfs-fuse3 a required dependency of cvmfs
* updates rpm packaging to drop libfuse2 for fedora >= 42 and rhel >=10
* updates debian packaging to drop libfuse2 for debian >= 13 and ubuntu >= 25.04
  * Fixes #3866
  * Cleans up the control file, removing comments regarding multi-arch support, and improving the formatting
  * In order not to do too much preprocessing, the control file is the same for older and newer platform and now recommends libfuse2 instead of requiring it.


packaging/debian/cvmfs/control
```diff
@@ -1,85 +1,133 @@
...
-Build-Depends: libfuse3-dev (>= 3.3.0), debhelper (>= 9), autotools-dev, cmake, cpio, libcap-dev, libssl-dev, libfuse-dev, pkg-config, libattr1-dev, patch, python3-dev, python3-setuptools, unzip, uuid-dev, valgrind, libz-dev
+Build-Depends:
+ libfuse3-dev (>= 3.3.0),
+ debhelper (>= 9),
+ autotools-dev,
+ cmake,
+ cpio,
+ libcap-dev,
+ libssl-dev,
+ libfuse-dev,
```
DrDaveD: Doesn't this continue to require libfuse2 for all debian based platforms?  I guess if it's only installed on the build machine and not used, it's OK as long as the package continues to be available.  Otherwise at some point the control file will probably need to become a template that is edited depending on the platform.  Perhaps a comment should be added to that effect.

packaging/debian/cvmfs/control
```diff
@@ -1,85 +1,133 @@
...
+ gdb,
+ uuid-dev,
+ uuid,
+ adduser,
+ cvmfs-libs (= ${binary:Version}),
+ ${misc:Depends},
+ ${shlibs:Depends}
+Recommends:
+ autofs (>= 5.1.2)
+Suggests:
```
DrDaveD: I don't think this should be Suggests.  I'm quite sure that libfuse2 should be automatically figured out as a Depends so listing it here won't help.  I'm not sure however if fuse will be automatically figured out though because it is a runtime dependency.  Assuming that's the case, it ought to be listed as a Depends, not Suggests.  Maybe this is an argument for turning the control file into a template where this fuse Depends and the libfuse-dev Build-Depends get removed when not building with fuse2 support.

DrDaveD (changes requested):

packaging/debian/cvmfs/control
vvolkl: Good catch, that was definitely an oversight. Agreed, thinking about it again I think a template is better. Implemented a simple replacement now.

packaging/debian/cvmfs/control
vvolkl: Agreed. Tried removing fuse completely, the tests should tell if that's possible.

DrDaveD: I don't see how the build system would be able to detect the need for the fusermount in the fuse package at runtime, so I'm pretty sure it's going to be needed in Depends when fuse2 is supported.

vvolkl: Right, with fuse2 we still need fusermount.

PR #3884, opened 2025-06-23
Further cleanup of the repo root

vvolkl:
Move rarely used folders to add-ons, and scripts to relevant subdirectories


jblomer (approved):

vvolkl:
The blacklist and offline mode integration test 047, 078 have been flaky recently, to be investigated.

PR #3890, opened 2025-06-26
Update release package with new signing key for Alma10

vvolkl:
Like Debian-based platforms started doing a while ago, Alma 10 doesn't accept 1024 bit signing keys anymore.  For Debian we just signed the packages twice with the old and a new key, but that does not seem to be possible with rpms. 

So I update the release package to use the same new signing key already used for debian exclusively from Almalinux 10 on.


packaging/rpm/cvmfs-release.spec
```diff
@@ -13,6 +13,7 @@ URL:            http://cvmrepo.s3.cern.ch/cvmrepo/yum
 Source0:        http://cvmrepo.s3.cern.ch/cvmrepo/yum/RPM-GPG-KEY-CernVM
 Source1:        BSD
 Source2:        cernvm.repo
+Source3:        RPM-GPG-KEY-CernVM-2048
```
DrDaveD, line 16: Isn't it better to show the full URL?  In rpms I think that's mostly for documentation so maybe it doesn't matter.

packaging/rpm/cvmfs-release.spec
```diff
@@ -70,10 +73,20 @@ fi
 if  [[ "$ID" == "fedora" ]]; then
 sed -i 's/EL/fedora/g' /etc/yum.repos.d/cernvm.repo
 fi
+if  [[ "$ID" == "almalinux" ]]; then
```
DrDaveD, line 76: What about rocky?  I think it's best to check for `[[ " $ID_LIKE " == *" rhel "* ]]`.

vvolkl: Good catch, see #3891

PR #3891, opened 2025-06-27
cvmfs-release: fix postinst for RHEL-clones that aren't Alma

vvolkl:
Thanks to Dave for pointing this out.


vvolkl:
@cernvm-bot unittest

PR #3895, opened 2025-06-30
Change gc -a to only do repos where gc was run on the stratum0

### LINKED ISSUE #3802: Make gc -a on stratum 1 only run if stratum 0 has
I just noticed that there are a lot more repositories marked as garbage-collectable than I thought; over 60% of them are on the FNAL stratum 1.

This has a cost on stratum 1s because they regularly (daily or weekly) run gc on every garbage-collectable repository.  However, I believe a lot of the repositories only run gc very rarely on the stratum 0 because they're not the type of repository that deletes much.

Perhaps `cvmfs_server gc -a` should only run if the `last_gc` field in .cvmfs_status.json on the stratum 0 is newer than the corresponding field on the stratum 1.   There could be a `gc -A` which behaves as the current one does and run on all garbage collectable repositories.

DrDaveD:
- Fixes #3802

### DIFF
--- a/cvmfs/server/cvmfs_server_apache.sh
+++ b/cvmfs/server/cvmfs_server_apache.sh
@@ -69,14 +69,23 @@ wait_for_apache() {
 }
 
 
+get_url() {
+  local url="$1"
+  local timeout="$2"
+  local curlopts="$3"
+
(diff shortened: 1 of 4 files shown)


vvolkl (approved):
LGTM, thanks!

PR #3900, opened 2025-07-04
Introduce an open file aware cache cleanup policy

christge-sft:
Upgrade cache cleanup by adopting an open file aware LRU policy.

A Posix Cache Manager holds the information of refcounting, which is passed to a Posix Quota Manager.
During cache cleanup, if a file that would otherwise be selected for cleanup is open will be skipped, since it can not be removed from the cache without breaking the process that holds it open.

There are unit and integration tests for the features.

### DIFF
--- a/cvmfs/fd_refcount_mgr.cc
+++ b/cvmfs/fd_refcount_mgr.cc
@@ -85,12 +85,6 @@ FdRefcountMgr *FdRefcountMgr::Clone() {
   return clone;
 }
 
-bool FdRefcountMgr::Contains(const shash::Any& hash){
-  int result=-1;
-  const MutexLockGuard lock_guard(lock_cache_refcount_);
-  return map_fd_.Lookup(hash,&result);
-}
-
 SmallHashDynamic<shash::Any, int> *FdRefcountMgr::GetFdMapPtr() {
   return &map_fd_;
 }
--- a/cvmfs/fd_refcount_mgr.cc
+++ b/cvmfs/fd_refcount_mgr.cc
@@ -85,6 +85,12 @@ FdRefcountMgr *FdRefcountMgr::Clone() {
   return clone;
 }
 
+bool FdRefcountMgr::Contains(const shash::Any& hash){
+  int result=-1;
+  const MutexLockGuard lock_guard(lock_cache_refcount_);
+  return map_fd_.Lookup(hash,&result);
+}
+
 SmallHashDynamic<shash::Any, int> *FdRefcountMgr::GetFdMapPtr() {
   return &map_fd_;
 }
--- a/cvmfs/fd_refcount_mgr.h
+++ b/cvmfs/fd_refcount_mgr.h
@@ -23,7 +23,6 @@ static inline uint32_t hasher_int(const int &key) {
 }
 
 class FdRefcountMgr {
-  FRIEND_TEST(T_QuotaManager, CleanupLru);
  public:
   /**
    * Helper class containing the values for the map: fd -> refcount+id
@@ -55,8 +54,6 @@ class FdRefcountMgr {
   int Dup(int fd);
 
   FdRefcountMgr *Clone();
-
-  bool Contains(const shash::Any& );
 
  private:
   /**
--- a/cvmfs/fd_refcount_mgr.h
+++ b/cvmfs/fd_refcount_mgr.h
@@ -23,6 +23,7 @@ static inline uint32_t hasher_int(const int &key) {
 }
 
 class FdRefcountMgr {
+  FRIEND_TEST(T_QuotaManager, CleanupLru);
  public:
   /**
    * Helper class containing the values for the map: fd -> refcount+id
@@ -54,6 +55,8 @@ class FdRefcountMgr {
   int Dup(int fd);
 
   FdRefcountMgr *Clone();
+
+  bool Contains(const shash::Any& );
 
  private:
   /**
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -306,6 +306,13 @@ FileSystem::PosixCacheSettings FileSystem::DeterminePosixCacheSettings(
     settings.do_refcount = false;
   }
 
+  if (options_mgr_->GetValue(MkCacheParm("CVMFS_CACHE_CLEANUP_NONOPENLRU", instance),
+                             &optarg)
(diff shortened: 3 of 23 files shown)


test/unittests/t_smallhash.cc
```diff
@@ -5,28 +5,32 @@
...
+  static uint32_t hasher_int(const int &key) {
+    return MurmurHash2(&key, sizeof(key), 0x07387a4f);
+  }
 
-static uint32_t hasher_md5(const shash::Md5 &key) {
-  // Don't start with the first bytes, because == is using them as well
-  return static_cast<uint32_t>(*(reinterpret_cast<const uint32_t *>(key.digest) + 1));
+  static uint32_t hasher_md5(const shash::Md5 &key) {
+    // Don't start with the first bytes, because == is using them as well
+    return (uint32_t) * (reinterpret_cast<const uint32_t *>(key.digest) + 1);
```
vvolkl: This reverts a fix for clang-format, let's keep the static cast

cvmfs/mountpoint.cc
```diff
@@ -308,6 +308,13 @@ FileSystem::PosixCacheSettings FileSystem::DeterminePosixCacheSettings(
     settings.do_refcount = false;
   }
 
+  if (options_mgr_->GetValue(MkCacheParm("CVMFS_CACHE_OF_AWARE_CLEANUP", instance),
```
vvolkl: The abbreviation OF isn't very clear in this name. What about something like `CVMFS_CACHE_NONOPENLRU`  or `CVMFS_CACHE_CLEANUP_NONOPENLRU`?

test/src/034-cachecleanup/main
```diff
@@ -2,21 +2,132 @@
 
 cvmfs_test_name="Cache cleanup"
```
vvolkl, line 4: Can you add a short "docstring" style comments to the test, summarizing what it is intending to test?

test/src/034-cachecleanup/main
```diff
@@ -2,21 +2,132 @@
...
+
+  sudo cvmfs_config wipecache > /dev/null && echo "Whipping cache.. OK" || $(echo "Wipping cache.. FAILED!" && return 30)
+  sudo cvmfs_talk -i atlas.cern.ch cache size
+
+  # read files one after the other to create an LRU pattern
+  while IFS= read -r -d '' file; do
+    cat $file > /dev/null
+  done < $FILE_DB
+
+  # Keep the first half filse open
```
vvolkl:
```suggestion
  # Keep the first half of files open
```
typo

test/src/034-cachecleanup/main
```diff
@@ -2,21 +2,132 @@
...
+    less $file > /dev/null &
+    OPEN_PIDS="$pids $!"
+  done < $FILE_DB
+
+  local cache_size_before_cleanup=$(get_cache_size_quiet)
+  echo -n "Target: $CLEANUP_TARGET. Cleaning up.. "
+  sudo cvmfs_talk -i atlas.cern.ch cleanup $CLEANUP_TARGET || $(echo "FAILED" && return 31)
+  local cache_size_after_cleanup=$(get_cache_size_quiet)
+
+  #Check that cache size has been reduced at first and then that the goal has been reached
```
vvolkl:
```suggestion
  # Check that cache size has been reduced at first and then that the goal has been reached
```

test/src/034-cachecleanup/main
```diff
@@ -2,21 +2,132 @@
...
+    cat $file > /dev/null
+  done < $FILE_DB
+
+
+  local cache_size_before_cleanup=$(get_cache_size_quiet)
+  echo -n "Target: $CLEANUP_TARGET. Cleaning up.. "
+  sudo cvmfs_talk -i atlas.cern.ch cleanup $CLEANUP_TARGET || $(echo "FAILED" && return 41)
+  local cache_size_after_cleanup=$(get_cache_size_quiet)
+
+  #Check that cache size has been reduced at first and then that the goal has been reached
```
vvolkl:
```suggestion
  # Check that cache size has been reduced at first and then that the goal has been reached
```

vvolkl (changes requested):

cvmfs/fd_refcount_mgr.cc
```diff
@@ -85,6 +85,12 @@ FdRefcountMgr *FdRefcountMgr::Clone() {
   return clone;
 }
 
+bool FdRefcountMgr::Contains(const shash::Any& hash){
+  int result{-1};
```
vvolkl: MacOS doesn't like this declaration:
```
2025-07-04T13:19:01.5144270Z /Users/runner/work/cvmfs/cvmfs/cvmfs/fd_refcount_mgr.cc:89:13: error: expected ';' at end of declaration
2025-07-04T13:19:01.5246470Z   int result{-1};
2025-07-04T13:19:01.5347830Z             ^
2025-07-04T13:19:01.5448900Z      
```

cvmfs/mountpoint.cc
christge-sft: As discussed in person, I am totally in favor of this change. I would go with the last (longer but more self explanatory) version.

Q: I use this "of" abbreviation for local variables related to the feature. Is it an issue there? (they are not part of an interface, they're either function local variables, or private/protected class members).

vvolkl: Not a problem if they're not part of the interface.

vvolkl:
Fixes https://github.com/cvmfs/cvmfs/issues/2792

cvmfs/fd_refcount_mgr.h
```diff
@@ -9,7 +9,7 @@
 
 #include "cache.h"
 #include "smallhash.h"
-
+#include <gtest/gtest_prod.h>
```
vvolkl:
```suggestion
#include "duplex_testing.h"
```
See https://github.com/cvmfs/cvmfs/pull/3949

cvmfs/util/socket.h
```diff
@@ -0,0 +1,320 @@
...
+enum class Command {
+  SendHashes,
+  RecvHashes
+};
+};  // namespace util
+
+template<ProcessType PT>
+class LocalUnixSocket {
+ public:
+  template<ProcessType X = PT,
```
vvolkl, line 46: This class needs some doxygen-style comments explaining  template args and args

Issue #3901, opened 2025-07-07
mkfs produces fstab entries which fail to mount


andrey-utkin:
This exact sequence of commands used to work on Debian 12 at least for the entire duration of March 2025. Now it fails and I wonder how it ever worked before, because the fstab entry it produces doesn't look right, and there's no obvious related recent changes, and Debian isn't known to ship breaking changes into stable releases. Maybe cvmfs just uses a deprecated syntax form and support of it broke or was withdrawn.

See workaround at the bottom.

```
incus launch --vm images:debian/12 cvmfs-server-fstab-issue
incus exec cvmfs-server-fstab-issue -- bash -l <EOF
apt install -y wget && wget http://cvmrepo.s3.cern.ch/cvmrepo/apt/cvmfs-release-latest_all.deb && dpkg -i cvmfs-release-latest_all.deb && apt update && apt install -y cvmfs cvmfs-server && cvmfs_server mkfs -o root my.repo.name
EOF
```

```
...
Warning: Failed to restart apache after enabling necessary modules
Creating Configuration Files... done
Creating CernVM-FS Master Key and Self-Signed Certificate... done
Creating CernVM-FS Server Infrastructure... done
Creating Backend Storage... done
Signing 30 day whitelist with master key... done
Creating Initial Repository... done
Mounting CernVM-FS Storage... (overlayfs) mount: /var/spool/cvmfs/my.repo.name/rdonly: wrong fs type, bad option, bad superblock on cvmfs2#my.repo.name, missing codepage or helper program, or other error.
       dmesg(1) may have more information after failed mount system call.
fail
root@cvmfs-server-fstab-issue:~# cat /etc/fstab
LABEL=rootfs  /         ext4  defaults  0 0
LABEL=UEFI    /boot/efi vfat  defaults  0 0
cvmfs2#my.repo.name /var/spool/cvmfs/my.repo.name/rdonly fuse allow_other,fsname=my.repo.name,config=/etc/cvmfs/repositories.d/my.repo.name/client.conf:/var/spool/cvmfs/my.repo.name/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for my.repo.name
[...]

PR #3796, opened 2025-03-24
Avoid use of fusermount3 by premounting & remove fuse3 dependency

DrDaveD:
- Fixes #3795

PR #3903, opened 2025-07-07
cvmfs_server: fstab: change type from fuse to cvmfs, abandon cvmfs2# prefix in name

andrey-utkin:
See https://github.com/cvmfs/cvmfs/issues/3901

What seemed to work until recently now fails:

	incus launch --vm images:debian/12 cvmfs-server-fstab-issue
	incus exec cvmfs-server-fstab-issue -- bash -l <EOF
	apt install -y wget && wget http://cvmrepo.s3.cern.ch/cvmrepo/apt/cvmfs-release-latest_all.deb && dpkg -i cvmfs-release-latest_all.deb && apt update && apt install -y cvmfs cvmfs-server && cvmfs_server mkfs -o root my.repo.name
	EOF

	...
	Warning: Failed to restart apache after enabling necessary modules
	Creating Configuration Files... done
	Creating CernVM-FS Master Key and Self-Signed Certificate... done
	Creating CernVM-FS Server Infrastructure... done
	Creating Backend Storage... done
	Signing 30 day whitelist with master key... done
	Creating Initial Repository... done
	Mounting CernVM-FS Storage... (overlayfs) mount: /var/spool/cvmfs/my.repo.name/rdonly: wrong fs type, bad option, bad superblock on cvmfs2#my.repo.name, missing codepage or helper program, or other error.
	       dmesg(1) may have more information after failed mount system call.
	fail
	root@cvmfs-server-fstab-issue:~# cat /etc/fstab
	LABEL=rootfs  /         ext4  defaults  0 0
	LABEL=UEFI    /boot/efi vfat  defaults  0 0
	cvmfs2#my.repo.name /var/spool/cvmfs/my.repo.name/rdonly fuse allow_other,fsname=my.repo.name,config=/etc/cvmfs/repositories.d/my.repo.name/client.conf:/var/spool/cvmfs/my.repo.name/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for my.repo.name
	overlay_my.repo.name /cvmfs/my.repo.name overlay upperdir=/var/spool/cvmfs/my.repo.name/scratch/current,lowerdir=/var/spool/cvmfs/my.repo.name/rdonly,workdir=/var/spool/cvmfs/my.repo.name/ofs_workdir,noauto,nodev,ro 0 0 # added by CernVM-FS for my.repo.name

dmesg says:

	fuse: Unknown parameter 'fsname'

`cvmfs2#` prefix in the first field, which is called deprecated by mount(8) and fstab(5).

Third field (with value fuse) gives mount an indication of which helper program to call. Says fuse, with no subtype (i'd expect fuse.cvmfs). cvmfs installs `mount.cvmfs`, so let's use `cvmfs` as type.

### DIFF
--- a/cvmfs/server/cvmfs_server_common.sh
+++ b/cvmfs/server/cvmfs_server_common.sh
@@ -1194,7 +1194,7 @@ setup_and_mount_new_repository() {
   if [ x"$CVMFS_UNION_FS_TYPE" = x"overlayfs" ]; then
     echo -n "(overlayfs) "
     cat >> /etc/fstab << EOF
-cvmfs2#$name $rdonly_dir fuse allow_other,fsname=$name,config=/etc/cvmfs/repositories.d/${name}/client.conf:${CVMFS_SPOOL_DIR}/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for $name
+$name $rdonly_dir cvmfs allow_other,fsname=$name,config=/etc/cvmfs/repositories.d/${name}/client.conf:${CVMFS_SPOOL_DIR}/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for $name
 overlay_$name /cvmfs/$name overlay upperdir=${scratch_dir},lowerdir=${rdonly_dir},workdir=$ofs_workdir,noauto,nodev,ro 0 0 # added by CernVM-FS for $name
 EOF
   else
@@ -1203,7 +1203,7 @@ EOF
       selinux_context=",context=\"system_u:object_r:default_t:s0\""
     fi
     cat >> /etc/fstab << EOF
-cvmfs2#$name $rdonly_dir fuse allow_other,fsname=$name,config=/etc/cvmfs/repositories.d/${name}/client.conf:${CVMFS_SPOOL_DIR}/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for $name
+$name $rdonly_dir cvmfs allow_other,fsname=$name,config=/etc/cvmfs/repositories.d/${name}/client.conf:${CVMFS_SPOOL_DIR}/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for $name
 aufs_$name /cvmfs/$name aufs br=${scratch_dir}=rw:${rdonly_dir}=rr,udba=none,noauto,nodev,ro$selinux_context 0 0 # added by CernVM-FS for $name
 EOF
   fi
--- a/cvmfs/server/cvmfs_server_common.sh
+++ b/cvmfs/server/cvmfs_server_common.sh
@@ -1194,7 +1194,7 @@ setup_and_mount_new_repository() {
   if [ x"$CVMFS_UNION_FS_TYPE" = x"overlayfs" ]; then
     echo -n "(overlayfs) "
     cat >> /etc/fstab << EOF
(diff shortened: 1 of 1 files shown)


vvolkl:
That's a good catch, thanks a lot Andrey! The breaking change is on our side, actually. Following https://github.com/cvmfs/cvmfs/pull/3796, we removed the dependency on `fuse3` but for this type of fstab entry it seems that fusermount is invoked. You can check that `apt install fuse3` fixes the error with the original fstab.

The fstab syntax that is used for the server rdonly mount is indeed outdated. Even in our documentation a different syntax is used: https://cvmfs.readthedocs.io/en/stable/cpt-configure.html#mounting . However with your suggestion, the `CVMFS_HTTP_PROXY not found` error means that the config option is not propagated correctly, because it is set in the client.conf. It should be `fuse.cvmfs2` instead of `cvmfs` for `vfs_type`, for example:

```
my.repo.name /var/spool/cvmfs/my.repo.name/rdonly fuse.cvmfs2 allow_other,fsname=my.repo.name,config=/etc/cvmfs/repositories.d/my.repo.name/client.conf:/var/spool/cvmfs/my.repo.name/client.local,cvmfs_suid,noauto 0 0 # added by CernVM-FS for my.repo.name
```

The fix could either be re-adding fuse3 as a dependency or changing the fstab syntax. I think even for a patch release it should be fine to fix the fstab.

As a side-note, it's a bit strange that we treat mounts for the server different than regular client repositories. For the server, we set fsname=fqrn , while for a regular client it's fsname=cvmfs2:
```
my.repo.name on /var/spool/cvmfs/my.repo.name/rdonly type fuse (ro,nodev,relatime,user_id=0,group_id=0,default_permissions,allow_other)
cvmfs2 on /mnt/test type fuse (ro,nosuid,nodev,relatime,user_id=0,group_id=0,default_permissions,allow_other)
```
But then again, some way of distinguishing these two types of mounts is useful, because we might want to treat them differently in some operations.

andrey-utkin:
On 27 July 2025 13:08:45 EEST, Valentin Volkl ***@***.***> wrote:
>As a side-note, it's a bit strange that we treat mounts for the server different than regular client repositories. For the server, we set fsname=fqrn , while for a regular client it's fsname=cvmfs2

I believe I have seen fstab entries which had both fsname=fqrn and fsname=cvmfs2 on the same options list.

Thanks for fixing the reported issue!

vvolkl:
Looking into this a bit deeper, the difference is whether mount.cvmfs or mount.fuse will be invoked. Maybe it's even better to call mount.cvmfs (by specifying vfs_type=cvmfs), but the problem seems to be that mount.cvmfs tries to mount with the cvmfs user as owner by default, wheras here the mkfs owner should also own the mount. That's why the original version couldn't read the config files

vvolkl:
Ah, but my solution still needs fuse3. As a stopgap solution for 2.13.2, I'll just add fuse3 as a dependency to cvmfs_server. I'll remove it again with a proper fix this for 2.14, but I want to review the handling of fsname and subtype for cvmfs. In particular I think it'd be useful to add the repository name as a subtype to the mount.

DrDaveD:
Superseded by #3943

vvolkl:
No, #3943 is a temporary workaround, it was the most conservative fix just before tagging a release. As Andrey said, the fstab syntax is outdated and should be updated, but in order to avoid mount.fuse, the uid options  in  mount.cvmfs  need to be fixed, I believe. I'll need to look at this in more detail, but let's keep this open.

andrey-utkin:
Is this PR still being considered for merging?

andrey-utkin:
Please consider this PR.

vvolkl:
Superseded by #4203

PR #3904, opened 2025-07-07
Update python shebangs to python3

andrey-utkin:
This was seen in client tests:

	-- Testing Cancel file system request (Mon Jul  7 13:03:04 UTC 2025 / test number 096)
	restarting autofs...
	using systemd
	*** install a desaster cleanup
	*** mount repository
	*** run a silent HTTP server
	Calling /home/sftnight/cvmfs/test/common/mock_services/silent_socket.py TCP 8019
	silent server started with PID 69599
	env: 'python': No such file or directory

### DIFF
--- a/add-ons/tools/download_catalog_graph.py
+++ b/add-ons/tools/download_catalog_graph.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import cvmfs
 import os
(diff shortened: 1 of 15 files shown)


andrey-utkin:
There remains one `#! /usr/bin/env python2`: test/migration_tests/001-hotpatch/tarsum, haven't looked into it.

vvolkl:
The use of `python` stems from the time when we still had to support python2 and 3 concurrently. Now this is no longer needed, but I prefer to get rid of the python test helpers alltogether. Will rewrite them in c or golang.

Issue #3902, opened 2025-07-07
Proposal: embrace bash strict mode


andrey-utkin:
When using cvmfs codebase for development activities - running test suites, scripting particular test or demo scenarios, a recurring source of unpleasant surprises is that the failure of particular steps of shell scripts is ignored. (And that step's output is often unhelpfully hidden - by sending it to syslog which is disabled by default in containers, or by `>/dev/null 2>&1`).
I have already submitted https://github.com/cvmfs/cvmfs/pull/3801 but there is much more scope to apply this [one single concept](http://redsymbol.net/articles/unofficial-bash-strict-mode/).

The last instance for me is cvmfs_server's keys generation
```
[root@c4110963ba1d /]# sudo -u sftnight sudo cvmfs_server mkfs -o sftnight myrepo                                                                                                                                    Creating Configuration Files... done
Creating CernVM-FS Master Key and Self-Signed Certificate...                                                                                                                                          chmod: cannot access '/etc/cvmfs/keys/myrepo..masterkey': No such file or directory                                                                                                                                  chown: cannot access '/etc/cvmfs/keys/myrepo..masterkey': No such file or directory
chmod: cannot access '/etc/cvmfs/keys/myrepo..pub': No such file or directory
chown: cannot access '/etc/cvmfs/keys/myrepo..pub': No such file or directory                                                                                                                                        fail (master key)
```
The reason was that `openssl` was not installed! But it won't show up in any way, because not only [strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) is not used, that step's error message output is silenced:

https://github.com/cvmfs/cvmfs/blob/20d2be5556a410e034042c99047e6e82118b3137/cvmfs/server/cvmfs_server_ssl.sh#L22

This is a problem packages consumers won't hit, but again this is happening during development activities where I try to run it in novel ways and/or environments.

This is a request for comments. I will likely send patches implementing this proposal in future.

DrDaveD:
Generally cvmfs_server already has `set -e` set.  `set -u` would probably be good except it will probably be difficult to verify that everything has been converted.  I suspect there are cases that subtly depend on `set -o pipefail` not being set, and changing the default `$IFS` would probably break a lot of places.

andrey-utkin:
Another example where lack of strictness about success gives surprising results:

[...]

PR #3905, opened 2025-07-09
server: remove code working around absense of jq

andrey-utkin:
cvmfs_server's JSON utility code makes effort to detect absense of jq. This on its own produces a side-effect of jq absense not showing up in error logs.

If jq is absent, there were different code paths to handle the task at hand, sometimes giving a warning ("it's your responsibility to ensure JSON is valid").

Some of such fallback code paths simply cannot count as accomplishing the task:

* update_repo_status() does nothing
* get_json_field() always provides empty result

This seems risky - it is surprising upon discovery, so it may produce surprising results in operation. Hopefully jq is ubiquitous enough to not need any workarounds.

There seems to be no testing for jq-less configuration, because there is a shell syntax error, as can be seen in test 689. I only stumbled upon this because of running tests in a different setting.

	Verifying integrity of test.cern.ch...
	Inspecting log of references
	Inspecting tag database
	[inspecting catalog] 0b9e8b479c3d6a36e0c1f5a7cfac3492a8feeb04 at /
	no problems found
	/usr/bin/cvmfs_server: line 2206: return: : numeric argument required
	*** setup cleanup handler

This change doesn't make it safe to operate cvmfs_server in jq-less configuration, because [cvmfs_server doesn't use strict mode](https://github.com/cvmfs/cvmfs/issues/3902) (at least `set -eo pipefail`), and running it without jq will not abort the execution immediately upon failure to launch jq, which may cause damage. Consumers of packaged cvmfs which properly declare runtime dependency on jq are safe, though. If this change is not acceptable on its own, it can wait until possibly cvmfs_server adopts necessary elements of strict mode.

### DIFF
--- a/cvmfs/server/cvmfs_server_json.sh
+++ b/cvmfs/server/cvmfs_server_json.sh
@@ -146,30 +146,8 @@ get_editor() {
 }
 
 
-has_jq() {
-  which jq > /dev/null 2>&1
-}
-
-check_jq() {
-  local hasjq=1
-  if ! has_jq; then
-    hasjq=0
-    echo 1>&2
-    echo "Warning: Didn't find 'jq' on your system. It is your responsibility" 1>&2
-    echo "         to produce a valid JSON file." 1>&2
-    echo 1>&2
-    read -p "  Press any key to continue..." nirvana
-  fi
-  echo $hasjq
-}
-
-
 validate_json() {
   local json_file="$1"
-
-  if ! which jq > /dev/null 2>&1; then
-    return 0 # no jq -> assume JSON is valid
-  fi
 
   jq '.' $json_file 2>&1
 }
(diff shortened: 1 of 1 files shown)


DrDaveD:
It seems to me that if you remove the possibility of working without jq then there ought to be instead an explicit check for jq missing and failing out with a fatal error.   Alternatively it should become a dependency on the cvmfs-server rpm and deb packages.

vvolkl:
>  Alternatively it should become a dependency on the cvmfs-server rpm and deb packages.

That's already the case:

https://github.com/cvmfs/cvmfs/blob/bde6e587e28cf898d527091572ed01fb97acbb2d/packaging/rpm/cvmfs-universal.spec#L252

I think this is more about running the test suite from a dev build. Even in this case life will be easier when  using the build_packages.sh script to install from locally built  packages.

DrDaveD:
> I think this is more about running the test suite from a dev build.

Ok then it should probably give a clear error when jq is missing, don't you think?

andrey-utkin:
> > I think this is more about running the test suite from a dev build.
> 
> Ok then it should probably give a clear error when jq is missing, don't you think?

Bash will automatically produce clear errors like `cvmfs_server: line 1: jq: command not found` if we don't redirect invocations' stderr.

PR #3913, opened 2025-07-10
Allocate cvmfs_server lock file descriptors dynamically

### LINKED ISSUE #3909: Duplicate cvmfs_swissknife pull processes running from cvmfs_server snapshot -t
I have seen several instances of duplicate `cvmfs_swissknife pull` running on the IHEP stratum 1, which has cvmfs-server-2.13.1 installed.  If I do an `lsof -p `on them I see that one of them has open a deleted lock file, like this:
```
cvmfs_swi 1072402 root    9u   REG                8,5         0 130669577141 /storage/cvmfs/spool_cvmfs/unpacked.cern.ch/is_updating.lock (deleted)
```
This stratum 1 is not using `cvmfs_server snapshot -a`, it is using the [cvmfs-hastratum1 script](https://github.com/cvmfs-contrib/cvmfs-hastratum1/blob/master/pull_and_push#L234) which just uses `cvmfs_server snapshot -t`.

The machine does also do `cvmfs_server gc -a`, gc also allocates the is_updating.lock, and I believe I have only seen this problem on garbage collectable repositories.  However, at least in this latest case gc has not been run on that repository during the duplicate snapshot runs; it is only run on weekends.  On the other hand, once this situation starts happening, it gets perpetuated because if a snapshot with its lock deleted finishes it deletes the lock of the snapshot that isn't yet finished, so the next one that starts has the same overlap problem.  In this case I see that the stratum 1 was completely caught up on that repository within the last 48 hours so, still, it still doesn't seem like it could have been caused by the gc.

Aha, it was caused by the gc.  Here are the /var/log/messages lines from unpacked.cern.ch on the day the gc ran:
```
Jul  6 15:40:02 cvmfs-stratum-one-a cvmfs_server[1915051]: (unpacked.cern.ch) successfully snapshotted from http://cvmfs-stratum-zero.cern.ch:8000/cvmfs/unpacked.cern.ch
Jul  6 15:40:02 cvmfs-stratum-one-a cvmfs_server[1915069]: (unpacked.cern.ch) started garbage collection
Jul  6 15:40:02 cvmfs-stratum-one-a cvmfs_swissknife[1915073]: Parameter CVMFS_STATISTICS_DB was not set in the repository configuration file. Using default value: /var/spool/cvmfs/unpacked.cern.ch/stats.db
Jul  6 15:45:09 cvmfs-stratum-one-a cvmfs_server[2284921]: (unpacked.cern.ch) started snapshotting from http://cvmfs-stratum-zero.cern.ch:8000/cvmfs/unpacked.cern.ch
Jul  7 04:06:40 cvmfs-stratum-one-a cvmfs_server[1623389]: (unpacked.cern.ch) successfully finished garbage collection
Jul  7 04:09:02 cvmfs-stratum-one-a cvmfs_server[1765868]: (unpacked.cern.ch) started snapshotting from http://cvmfs-stratum-zero.cern.ch:8000/cvmfs/unpacked.cern.ch
```
The garbage collection waited for the first snapshot to finish, but then the next snapshot started before the gc was finished and presumably its lock was removed when the gc exited.

Oh, I see the problem.  More than one lock is being acquired by the gc, but they all use the same flock file descriptor 9.

DrDaveD:
Instead of always using fd 9 for locks, open the next available file descriptor and keep track of it.

- Fixes #3909

I thought about adding some unit tests of these functions to [test_cvmfs_server_util.sh](https://github.com/cvmfs/cvmfs/blob/devel/cvmfs/server/test_cvmfs_server_util.sh) but I haven't been able to figure out if it is being invoked anywhere.  Maybe it needs to be added to [run_unittests.sh](https://github.com/cvmfs/cvmfs/blob/devel/ci/run_unittests.sh)?

### DIFF
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -362,23 +362,24 @@ __hc_transition() {
 
 ### Locking functions
 
-declare -A _lock_fds
+_lock_fds=""
+_lock_paths=""
 
 acquire_lock() {
   local path="$1"
   local wait_for_lock="${2:-0}"
   local lock_file="${path}.lock"
+  local lock_fd
   while true; do
-    exec {_lock_fds[$path]}<>${lock_file}
+    exec {lock_fd}<>${lock_file}
     if [ $wait_for_lock -eq 0 ]; then
-      if ! flock -n ${_lock_fds[$path]}; then
+      if ! flock -n ${lock_fd}; then
         # didn't get it, clean up and return failure
-        exec {_lock_fds[$path]}<&-
-        unset _lock_fds[$path]
+        exec {lock_fd}<&-
         return 1
       fi
     else
-      flock ${_lock_fds[$path]}
+      flock ${lock_fd}
     fi
     # now have the lock
     if [ -f $lock_file ]; then
@@ -386,17 +387,49 @@ acquire_lock() {
       break
     fi
     # was removed by former lock holder; close and try again
-    exec {_lock_fds[$path]}<&-
+    exec {lock_fd}<&-
   done
+  # add the fd number and path to the two lists
+  _lock_fds="$_lock_fds $lock_fd"
+  _lock_paths="$_lock_paths $path"
 }
 
 
 release_lock() {
   local path="$1"
   local lock_file="${path}.lock"
+  local lock_fd=-1
+  # Find the index of $path in $_lock_paths and from that find $lock_fd.
+  # Would be much easier with an associative array if we could rely on bash.
+  local index=0
+  local lock_index=0
+  local lock_path
+  local new_paths=""
+  for lock_path in $_lock_paths; do
+    let index+=1
+    if [ "$path" = "$lock_path" ]; then
+      lock_index=$index
+    else
+      new_paths="$new_paths $lock_path"
+    fi
+  done
+  [ "$lock_index" != 0 ] || die "attempt to release $path lock when it was not acquired"
+  index=0
+  local fd
+  local new_fds
+  for fd in $_lock_fds; do
+    let index+=1
+    if [ "$index" = "$lock_index" ]; then
+      lock_fd=$fd
+    else
+      new_fds="$new_fds $fd"
+    fi
+  done
+  [ "$lock_fd" != -1 ] || die "_lock_fds and _lock_paths out of sync while releasing $path lock"
+  _lock_fds="$new_fds"
+  _lock_paths="$new_paths"
   rm -f $lock_file
-  exec {_lock_fds[$path]}<&-
-  unset _lock_fds[$path]
+  exec {lock_fd}<&-
 }
 
 
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -637,12 +637,14 @@ check_overlayfs_version() {
     if compare_versions "$krnl_version" -ge "$required_version" ; then
       # If the mounted filesystem name is long df will split output into two
       #  lines, so use tail -n +2 to skip first line and echo to combine them
-      local scratch_fstype=$(echo $(df -T /var/spool/cvmfs | tail -n +2) | awk {'print $2'})
+      local scratch_line="$(echo $(df -T /var/spool/cvmfs | tail -n +2))"
+      local scratch_fstype=$(echo $scratch_line | awk {'print $2'})
       if [ "x$scratch_fstype" = "xext3" ] || [ "x$scratch_fstype" = "xext4" ] ; then
         return 0
       fi
       if [ "x$scratch_fstype" = "xxfs" ] ; then
-        if [ "x$(xfs_info /var/spool/cvmfs 2>/dev/null | grep ftype=1)" != "x" ] ; then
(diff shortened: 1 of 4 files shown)


DrDaveD:
Closing & reopening to re-run unittest

DrDaveD:
Also after obtaining the lock check if the file has been removed by a release of the former lock holder and if so, close the lock and obtain it again.  See longer rationale in [my issue comment](https://github.com/cvmfs/cvmfs/issues/3909#issuecomment-3063304960).

This could even happen if the lock aborts on a conflict, so rather than duplicating that logic in two functions I decided to combine them into one `acquire_lock()` function.  That case could happen if process 2 opens the file, then process 1 removes and closes the file before process 2 calls `flock -n`.  It's a much less likely race condition than if process 2 waits for the lock to be freed -- in that case the problem was basically inevitable -- but it is possible.

vvolkl:
Adding tests is a very good idea. For `test_cvmfs_server_util.sh` the best way to invoke it may be in a dedicated github actions workflow that runs whenever the server sources are changed. Let me see if I can add this.

cvmfs/server/cvmfs_server_util.sh
```diff
@@ -362,28 +362,41 @@ __hc_transition() {
 
 ### Locking functions
 
+declare -A _lock_fds
```
vvolkl: Won't work on debian-based platforms using dash

vvolkl (changes requested):

cvmfs/server/cvmfs_server_util.sh
DrDaveD: So what alternative works in dash?

Do we have to support dash?  Why doesn't cvmfs_server start with `#!/bin/bash`?

DrDaveD: Oh, I could do it with two variables that are space-separated strings and two loops.  Much less elegant but doable.

vvolkl: Good question, I also fell into this trap  in https://github.com/cvmfs/cvmfs/pull/3792 and implemented it with bash arrays at first. There's the portability argument for using /bin/sh, and apparently dash is more efficient. It can also be an advantage that it has less features, makes it easier to read for someone not versed in bashisms. Let's keep `#!/bin/sh` for now at least.

vvolkl:
> Adding tests is a very good idea. For `test_cvmfs_server_util.sh` the best way to invoke it may be in a dedicated github actions workflow that runs whenever the server sources are changed. Let me see if I can add this.

https://github.com/cvmfs/cvmfs/pull/3916

vvolkl:
The ubuntu test is failing with 
```
Initial commit... /usr/bin/cvmfs_server: 427: exec: {lock_fd}: not found
```

cvmfs/server/cvmfs_server_util.sh
```diff
@@ -362,28 +362,74 @@ __hc_transition() {
...
-
-
-wait_and_acquire_lock() {
-  local path="$1"
-  local lock_file="${path}.lock"
-  exec 9<>${lock_file}
-  flock 9
+  local lock_fd
+  while true; do
+    exec {lock_fd}<>${lock_file}
```
vvolkl: I can see how bash is nice here. But how about 
```bash
# Function to find available file descriptor
find_available_fd() {
    local fd=9
    local max=$(ulimit -n)
    while [ $fd -lt $max ]; do
        if ! { true <&$fd; } 2>/dev/null; then
            echo $fd
            return
        fi
        fd=$((fd + 1))
    done
    echo "No available file descriptor found" >&2
    return 1
}

lock_fd=$(find_available_fd)
exec ${lock_fd}<>"${lock_file}"
```
?

DrDaveD: Sigh.  dash sure is missing a lot of useful things.  The next thing after this it would hit is the "+=" variable assignment syntax.  `exec ${lock_fd}...` doesn't work in bash or dash; it needs an `eval exec`.

According to [stack overflow](https://stackoverflow.com/a/41603891), it would need to try opening both read & write.  For that reason I like the solution that looks in /proc/$$/fd better.

vvolkl:
OK, now we are running into the issue of dash only being able to manipulate single-digit file descriptors:
https://bugs.launchpad.net/ubuntu/+source/dash/+bug/249620
Didn't know about this either.

My proposal for a fix:
```diff
diff --git a/cvmfs/server/cvmfs_server_util.sh b/cvmfs/server/cvmfs_server_util.sh
index c01bce99f..aadfa03e2 100644
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -363,8 +363,14 @@ __hc_transition() {
 # Find an available file descriptor
 find_available_fd()
 {
-  local fd=11
-  local max=$(ulimit -n) 
+  if [ -n "$BASH_VERSION" ]; then
+    local fd=11
+    local max=$(ulimit -n)
+  else
+    # dash only supports single-digit fds
+    local fd=3
+    local max=9
+  fi
   while [ $fd -lt $max ]; do
     if [ ! -e /proc/$$/fd/$fd ]; then
       echo $fd

```

vvolkl (approved):

PR #3915, opened 2025-07-11
Move the no collectable repos message to gc.log

DrDaveD:
All messages for `gc -a` should go to gc.log instead of stdout.  It is typically run from cron, which ends up sending an email if anything goes to stdout.  This is a follow-on to #3895.


vvolkl:
@cernvm-bot unittest

Issue #3919, opened 2025-07-16
excessive geoapi lookups


deesto:
We've noticed an excessive number of DNS lookups (~2M/day, many of which fail) by our Stratum One that no amount of DNS cache could alleviate -- in fact, while debugging, we decided to replace dnsmasq and its hard-coded 10k cache entry limit with unbound, but this did little to help in this case. We noticed odd patterns in the log entries: unresolvable objects (e.g., remote host short names and uuid) are not only  looked up, but iterating through our local DNS search domains and appending them in further resolution attempts, resulting in two lookups and failed results (NXDOMAIN ANSWER) per each of our four local name servers (for a total of eight failed lookups). This is repeated many times: we're currently averaging about six of these failed queries logged per second, which is a lot of useless noise in our logs. Are these lookups are necessary? If so, can this please be optimized a bit? And possibly honed to look up only FQDNs and not short names and otherwise unresolvable records?
```
Jul 15 19:12:19 [stratum_one] unbound: [9909:1] info: subnet operate: query 8ee8660a-87ea-274d-dba0-151a599fab7b.usatlas.bnl.gov. A IN
Jul 15 19:12:19 [stratum_one] unbound: [9909:1] info: validator operate: query 8ee8660a-87ea-274d-dba0-151a599fab7b.usatlas.bnl.gov. A IN
Jul 15 19:12:19 [stratum_one] unbound: [9909:1] info: resolving 8ee8660a-87ea-274d-dba0-151a599fab7b.usatlas.bnl.gov. A IN
Jul 15 19:12:19 [stratum_one] unbound: [9909:1] info: processQueryTargets: 8ee8660a-87ea-274d-dba0-151a599fab7b.usatlas.bnl.gov. A IN
Jul 15 19:12:19 [stratum_one] unbound: [9909:1] info: sending query: 8ee8660a-87ea-274d-dba0-151a599fab7b.usatlas.bnl.gov. A IN
Jul 15 19:12:19 [stratum_one] unbound: [9909:1] info: subnet operate: query 8ee8660a-87ea-274d-dba0-151a599fab7b.usatlas.bnl.gov. AAAA IN
Jul 15 19:12:19 [stratum_one] unbound: [9909:1] info: validator operate: query 8ee8660a-87ea-274d-dba0-151a599fab7b.usatlas.bnl.gov. AAAA IN
[...]

PR #3920, opened 2025-07-16
Optimize DNS lookups by cvmfs_geo.py to ignore short host names

deesto:
Modify geoinfo function to ignore short names and non-FQDNs in DNS lookups, since they will fail anyway

### DIFF
--- a/cvmfs/webapi/cvmfs_geo.py
+++ b/cvmfs/webapi/cvmfs_geo.py
@@ -124,31 +124,32 @@ def addr_geoinfo(now, addr):
 
     return response['location']
 
-# Look up geo info by name.  Try IPv4 first since that DB is
+# Look up geo info by name (dns or ip address).  Try IPv4 first since that DB is
 # better and most servers today are dual stack if they have IPv6.
 # Store results in a cache.  Wsgi is multithreaded so need to lock
 # accesses to the shared cache.
 # Return geo info record or None if none found.
-def name_geoinfo(now, name):
+def name_geoinfo(now, hostname):
     global geo_cache
-    if (len(name) > 256) or not addr_pattern.search(name):
+    if (len(hostname) > 256) or not addr_pattern.search(hostname):
         return None
 
 # ignore short names and non-FQDNs in lookups, since they will fail anyway:
-    if "." not in name:
+    if not '.' in hostname and not ':' in hostname:
+        # not an fqdn or IPv4 or IPv6 address
         return None
         
     global namelookups
     namelock.acquire()
-    if name in geo_cache:
-        (stamp, gir) = geo_cache[name]
+    if hostname in geo_cache:
+        (stamp, gir) = geo_cache[hostname]
         if now <= stamp + geo_cache_secs:
             # still good, use it
             namelock.release()
             return gir
         # update the timestamp so only one thread needs to wait
         #  when a lookup is slow
-        geo_cache[name] = (now, gir)
+        geo_cache[hostname] = (now, gir)
     elif len(geo_cache) >= geo_cache_max_entries:
         # avoid denial of service by removing one arbitrary entry
         #   before we add one
@@ -158,7 +159,7 @@ def name_geoinfo(now, name):
 
     ai = ()
     try:
-        ai = socket.getaddrinfo(name,80,0,0,socket.IPPROTO_TCP)
+        ai = socket.getaddrinfo(hostname,80,0,0,socket.IPPROTO_TCP)
     except:
         pass
     gir = None
@@ -180,11 +181,11 @@ def name_geoinfo(now, name):
             gir = None
 
     namelock.acquire()
-    if gir == None and name in geo_cache:
+    if gir == None and hostname in geo_cache:
         # reuse expired entry
-        gir = geo_cache[name][1]
+        gir = geo_cache[hostname][1]
 
-    geo_cache[name] = (now, gir)
+    geo_cache[hostname] = (now, gir)
     namelock.release()
 
     return gir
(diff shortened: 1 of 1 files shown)


cvmfs/webapi/cvmfs_geo.py
```diff
@@ -134,6 +134,10 @@ def name_geoinfo(now, name):
     if (len(name) > 256) or not addr_pattern.search(name):
         return None
 
+# ignore short names and non-FQDNs in lookups, since they will fail anyway:
+    if "." not in name:
```
DrDaveD: This function can also look up IPv4 or IPv6 addresses so this should allow also for `:` in the name.  Here's the patch I wrote:
```
diff --git a/cvmfs/webapi/cvmfs_geo.py b/cvmfs/webapi/cvmfs_geo.py
index 56ff1a7fc..2325fdd83 100644
--- a/cvmfs/webapi/cvmfs_geo.py
+++ b/cvmfs/webapi/cvmfs_geo.py
@@ -131,6 +131,11 @@ def addr_geoinfo(now, addr):
 # Return geo info record or None if none found.
 def name_geoinfo(now, name):
     global geo_cache
+
+    if not '.' in name and not ':' in name:
+        # not an fqdn or IPv4 or IPv6 address
+        return None
+
     if (len(name) > 256) or not addr_pattern.search(name):
         return None
```

DrDaveD (changes requested):

cvmfs/webapi/cvmfs_geo.py
deesto: As this function returns names, not IP addresses, I wouldn't think it needs a check for ':'.

DrDaveD: The function parameter is called `name` but if CVMFS_HTTP_PROXY in a client is set to an IP address instead of a name then that gets passed in here.  If getaddrinfo is given an IP address it returns that IP address.  The result of getaddrinfo is used to look up geoinfo, which the function returns.
```
$ python3
Python 3.11.2 (main, Aug 26 2024, 07:20:54) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.getaddrinfo("131.225.189.138",80,0,0,socket.IPPROTO_TCP)
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('131.225.189.138', 80))]
>>> socket.getaddrinfo("2620:6a:0:8421::244",80,0,0,socket.IPPROTO_TCP)
[(<AddressFamily.AF_INET6: 10>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('2620:6a:0:8421::244', 80, 0, 0))]
```

DrDaveD: It occurred to me to wonder whether it would be better to have code to treat addresses differently than names to avoid a call to the DNS, but I confirmed just now by strace that getaddrinfo does not contact the DNS at all if it is given an IPv4 or IPv6 address, so there's no need.

vvolkl: OK; I'd propose to rename the `name`  variable to  `host` to make  that a bit clearer.

vvolkl (approved):
Thanks for the patch!

vvolkl:
Fixes #3919

cvmfs/webapi/cvmfs_geo.py
vvolkl: > The function parameter is called name but if CVMFS_HTTP_PROXY in a client is set to an IP address instead of a name then that gets passed in here. If getaddrinfo is given an IP address it returns that IP address. The result of getaddrinfo is used to look up geoinfo, which the function returns.

I think that's incorrect. The geo call is populated here: https://github.com/cvmfs/cvmfs/blob/c6b909323b8dfecae917d73cbcb5599b79fae209/cvmfs/network/download.cc#L2542
which is in turn called with the host chain in CVMFS_SERVER_URL. Plus the hostname is first passed through:
https://github.com/cvmfs/cvmfs/blob/c6b909323b8dfecae917d73cbcb5599b79fae209/cvmfs/network/download.cc#L2525
which would not allow an IP.

I've also just checked that setting IPv6 adresses in the CVMFS_SERVER_URL works, but won't do a geo-api lookup. I think that's actually a bug and we should allow that.

In any case, I think it's not a bad idea to allow ipv6 adresses in this method.

vvolkl: Currently 
```
curl http://cvmfs-stratum-one.cern.ch/cvmfs/cms.cern.ch/api/v1.0/geo/x/2620:6a:0:8421::244,2401:de00:2:33::72,2001:1458:301:cd::100:41d
```
works same as
```
curl http://cvmfs-stratum-one.cern.ch/cvmfs/cms.cern.ch/api/v1.0/geo/x/cvmfs-s1fnal.opensciencegrid.org,cvmfs-stratum-one.ihep.ac.cn,cvmfs-stratum-one.cern.ch
```

DrDaveD: > > The function parameter is called name but if CVMFS_HTTP_PROXY in a client is set to an IP address instead of a name then that gets passed in here. If getaddrinfo is given an IP address it returns that IP address. The result of getaddrinfo is used to look up geoinfo, which the function returns.
> 
> I think that's incorrect. The geo call is populated here:
...

But I have seen IPv4 addresses listed in geo api calls.  I haven't seen IPv6 addresses, but cvmfs always prefers IPv4 when configured so that's not surprising.  I was able to force it with IPv6 by setting CVMFS_HTTP_PROXY and square brackets around an IPv6 address and it resulted in a small problem.  I'll create a separate issue about that.

PR #3923, opened 2025-07-17
test/run.sh: log tests by dirname everywhere

andrey-utkin:
Test "names" are really long description strings, test numbers have duplicates, but the dirname under test/src/ is where the sources are, so dirname is the most important practical test identifier. Log it everywhere.

-------------

This is an admittedly very opinionated PR, so I don't expect it to be accepted as-is. Hopefully this will spark some discussion at least.

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -279,9 +279,9 @@ do
 
   # write some status info to the screen
   TEST_NR="$(basename $t | head -c3)"
-  to_syslog "Test $TEST_NR (${cvmfs_test_name}) started"
-  echo "-- Testing ${cvmfs_test_name} ($(date) / test number $TEST_NR)" >> $logfile
-  echo -n "Testing ${cvmfs_test_name}... "
+  to_syslog "Test $t (${cvmfs_test_name}) started"
+  echo "-- Testing $t ${cvmfs_test_name} ($(date "+%F %T %z"))" >> $logfile
+  echo -n "Testing $t ${cvmfs_test_name}... "
   echo "$cvmfs_test_name"          > ${scratchdir}/name
   echo "$(basename $t | head -c3)" > ${scratchdir}/number
 
@@ -343,7 +343,7 @@ do
   fi
   test_end=$(get_millisecond_epoch)
   test_time_elapsed=$(( ( $test_end - $test_start ) ))
-  echo "execution took $(milliseconds_to_seconds $test_time_elapsed) seconds" >> $logfile
+  echo "Execution of $t took $(milliseconds_to_seconds $test_time_elapsed) seconds" >> $logfile
   echo "$test_time_elapsed" > ${scratchdir}/elapsed
   echo "$RETVAL"            > ${scratchdir}/retval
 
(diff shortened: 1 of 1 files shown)


vvolkl:
This is opening a can of worms for sure :-) I agree that this printout should show the most relevant information in a way that catches the eye. For me that's actually the failing tests, and the return code they failed with, which can be added in a digest at the end. Then I think all of the skipped test could be similarly grouped in the beginning - we can do a first loop to find those. 

It should be noted that these scripts originally intended for the test report to be generated from the JUnit xml, relying on Jenkins to do this. We could still use JUnit, but it should still work without additional dependencies when just invoking run.sh from bash.

One thing that would make sense to me is to generate CTest testcases for all the integration tests. That way running the tests becomes much more standard. However I also appreciate the flexibility that comes from having our own test runner script, so I would still keep run.sh.

So, back to the run.sh output. With this PR this goes from:
```
docker exec -u sftnight -t cvmfs-dev bash -c \
    "cd /home/sftnight/cvmfs/test/common/container && CVMFS_TEST_PROXY=DIRECT bash test.sh"
  shell: /usr/bin/bash -e {0}
running CernVM-FS client test cases...
Limiting test cases to suite(s): quick
Testing Dummy test... OK
Testing Check installation... OK
Testing Probing atlas, lhcb... OK
Testing Nested catalogs with same prefix... Skipped by suite restriction
...
Testing Check changing CVMFS_CACHE_REFCOUNT between reloads... OK
Testing Check that simultaneous mount attempts only mount the repository once... Skipped by exclusion
Testing Browsing with the streaming cache manager... Skipped by exclusion

Tests:    107
Skipped:  26
Passed:   77
Warnings: 4
Failures: 0
```
To:

```
 docker exec -u sftnight -t cvmfs-dev bash -c \
    "cd /home/sftnight/cvmfs/test/common/container && CVMFS_TEST_PROXY=DIRECT bash test.sh"
  shell: /usr/bin/bash -e {0}
running CernVM-FS client test cases...
Limiting test cases to suite(s): quick
Testing src/000-dummy Dummy test... OK
Testing src/001-chksetup Check installation... OK
Testing src/002-probe Probing atlas, lhcb... OK
Testing src/003-nested Nested catalogs with same prefix... Skipped by suite restriction
Testing src/004-davinci Setup Davinci... Skipped by suite restriction
Testing src/005-asetup Setup ATLAS... Skipped by suite restriction
...
Testing src/103-reloadcachemgr Check changing CVMFS_CACHE_REFCOUNT between reloads... OK
Testing src/104-concurrent_mounts Check that simultaneous mount attempts only mount the repository once... Skipped by exclusion
Testing src/105-streaming-cache Browsing with the streaming cache manager... Skipped by exclusion

Tests:    107
Skipped:  26
Passed:   78
Warnings: 3
Failures: 0
```
A very nice improvement for sure! I think the "src/" can be dropped, as github and any ide allows you to easily find the  location of `105-streaming-cache`.  I would also add some padding to make the OKs more visible, and print the time taken to complete the test. So it could look something like (imagine proper alignment):
```
Testing 001-chksetup   Check installation ............................. OK     in    125.3s
Testing 002-probe         Probing atlas, lhcb ........................... OK      in     34.3s
Testing 003-nested        Nested catalogs with same...  ........ Skipped by suite restriction
```

vvolkl:
Good point about the duplicated test numbers btw, this can easily be fixed, I'll just move them to new numbers or add `abc` suffixes to the number.

vvolkl:
Superseded by https://github.com/cvmfs/cvmfs/pull/4151

PR #3933, opened 2025-07-23
Add test for ingestsql

andrey-utkin:
This test exhibits failures on current devel of upstream, but passes on 2.11.2-based internal version.

This test can be run through test/run.sh.
It can run standalone by invoking run_test (systemd not required).


andrey-utkin:
Please review

andrey-utkin:
Please review?

vvolkl:
Could you generate the testbed keys on the fly instead of checking them into the repo? There are unfortunately some less-than-smart security scanners that will spam me otherwise.

vvolkl (changes requested):
> Could you generate the testbed keys on the fly instead of checking them into the repo? There are unfortunately some less-than-smart security scanners that will spam me otherwise.

Also if you could avoid checking in the `opt`, `spool` and `var` subdirectories of the test, I think that would be cleaner. Do you think that's possible?

andrey-utkin:
> > Could you generate the testbed keys on the fly instead of checking them into the repo? There are unfortunately some less-than-smart security scanners that will spam me otherwise.
> 
> Also if you could avoid checking in the `opt`, `spool` and `var` subdirectories of the test, I think that would be cleaner. Do you think that's possible?

I will work on it, thanks for the review.

andrey-utkin:
I have dropped opt/.

Some more trimming should be possible, and I will keep doing it. `etc/` should go but I am yet to test it.

var/ is currently used for storage of repo files, similarly to archived "guinea pig" repo snapshots in use by other tests. I can convert it to a tarball if that would make a positive difference. I understand it's desirable to make the repo constructed from scratch, this should be possible with more work.

andrey-utkin:
The test took 13 minutes last time, so not quite within 5 minutes timeout used for tests tagged "quick". Is there currently any regular CI workflow for longer-running tests?

I can look into marking a subset of test cases as "quick" so that we have some way to be under 5 minutes and qualify for "quick" tag.

vvolkl:
> I have dropped opt/.
> 
> Some more trimming should be possible, and I will keep doing it. `etc/` should go but I am yet to test it.
> 
> var/ is currently used for storage of repo files, similarly to archived "guinea pig" repo snapshots in use by other tests. I can convert it to a tarball if that would make a positive difference. I understand it's desirable to make the repo constructed from scratch, this should be possible with more work.

Thank you, the effort is appreciated. It may not be the most important thing to keep git repositories small, but once some binary blob is checked in it's kind of hard to get rid of. Once all the external dependency tarballs and other similar files are removed I may do one force push to devel to garbage collect the deleted blobs.

If it's easier, I can also create a separate cvmfs/test-data repository and the tests can clone that on the fly. Ultimately, I think generating test files is cleaner though, it allows to parametrize and extend the tests if needed.

> The test took 13 minutes last time, so not quite within 5 minutes timeout used for tests tagged  "quick". Is there currently any regular CI workflow for longer-running tests?

No, there isn't. I ran the non-quick tests manually ahead of tagging 2.14.0, and there a few known failures and just generally tests that are outdated and should be renovated. So I don't want to add one for all non-quick tests just now, but I think it should also be fairly simple and cheap to create yet another workflow for ingestsql. I imagine there will be more tools for the posix tools so they can be added there.

> I can look into marking a subset of test cases as "quick" so that we have some way to be under 5 minutes and qualify for "quick" tag.

I like that, it's typically not bad to have some quick smoke tests in addition to a more complete suite.

andrey-utkin:
So the repo is now reconstructed with much less magic and definitely with much less storage consumed. Git history was reworked to that at no point the excessive files are included, in the interest of storage footprint.

The test is tagged "quick" which results in only first 20 test cases being run.

test/src/900-notification_system/main
```diff
@@ -36,7 +36,7 @@ publish_and_notify() {
 }
 
 run_standalone_tests() {
-    set_up_repository_gateway
+    set_up_repository_gateway test.repo.org
```
vvolkl, line 39: Why not use `$CVMFS_TEST_REPO`?

vvolkl (changes requested):
There's a missing `const` in swissknife_ingestsql.cc, see inline.

You'll also need to enable tests starting with 9 on gh actions (possibly excluding test 900):
```diff
diff --git i/.github/workflows/ci_ubuntu.yaml w/.github/workflows/ci_ubuntu.yaml
index 0d74acf2b..b982032ef 100644
--- i/.github/workflows/ci_ubuntu.yaml
+++ w/.github/workflows/ci_ubuntu.yaml
@@ -107,7 +107,7 @@ jobs:

           cd test

-          export CVMFS_TESTCASES=${{ github.event.inputs.testCases || '"src/0* src/1* src/5* src/6* src/7* src/8*"' }}
+          export CVMFS_TESTCASES=${{ github.event.inputs.testCases || '"src/0* src/1* src/5* src/6* src/7* src/8* src/9*"' }}
           export CVMFS_TEST_SUITES=${{ github.event.inputs.testSuites || 'quick' }}
           CVMFS_TEST_PROXY=${{ steps.squid-proxy.outputs.proxy-url }} \
           CVMFS_TEST_USE_CDN=yes \
diff --git i/test/common/container/test.sh w/test/common/container/test.sh
index efa77775d..d409fe609 100644
--- i/test/common/container/test.sh
+++ w/test/common/container/test.sh
@@ -74,6 +74,7 @@ CVMFS_TEST_UNIONFS=overlayfs                                                  \
                                  src/5*                                       \
                                  src/6*                                       \
                                  src/7*                                       \
+                                src/9*
                               || exit 1

 # Hotpatch/migration coverage runs via test_migration.sh in the standalone
```

test/src/900-notification_system/main
andrey-utkin: I only parametrised the helper function here, and substituted the old hardcoded name in existing tests, to avoid any disruption.
I also didn't want to change the hardcoded repo name in the test contributed, at that moment.

andrey-utkin:
> There's a missing `const` in swissknife_ingestsql.cc, see inline.

I am used to `int ret` being reused, so it's a bit of a surprise suggestion even though there's no other use for `ret` right now. But I'm fine with this and will push this change shortly.

> You'll also need to enable tests starting with 9 on gh actions (possibly excluding test 900):

This test should not be run in ci_ubuntu workflow as is.

This test doesn't work without S3 service and I have stripped the code path deploying minio in place, so this test can only work in ci-s3-integration workflow right now.

vvolkl (changes requested):
The test failure seems to be from a missing s3cmd. Should be installed in the github actions workflow, and  probably good to skip this test with a warning if it's missing.

vvolkl:
Rebased and appended to the last commit.

EDIT: sorry, did we cross pushes? If so, feel free to push again.

vvolkl:
The last test failure seems to indicate an actual bug, which the last two commits should fix. I'll probably move them to another PR if they indeed do.

PR #3944, opened 2025-07-28
`cvmfs_talk metrics prometheus` 

vvolkl:
This adds a talk command that outputs a lot of information in a format that can be ingested by Prometheus.  The motivation is to speed up the script in https://github.com/cvmfs-contrib/prometheus-cvmfs , which was very slow due to the overhead of calling cvmfs_talk over and over for each metric. Having all metrics returned by one command  reduces the runtime of the script from something like 13s to 2.5s second for 20 mounted repos.

A possible future optimization could be to offload some work, like constructing the string, from talk.cc to cvmfs_talk.cc, and only send the raw values in a suitable format over the socket.


vvolkl:
Since this uses /proc, not sure if the command will work on MacOS. I don't think it has to, though.

PR #3947, opened 2025-07-30
Fix updating of last_gc when no gc is run under gc -a without dry run

DrDaveD:
Fix mistake from #3895 which prevents the "last_gc" time from being updated in .cvmfs_status.json when gc is skipped.  The check for dry_run was reversed.  This bug causes cvmfs-servermon to complain if garbage collection is skipped for a time, default 30 days.


DrDaveD:
This didn't actually fix the problem.  Real fix is in #3950

PR #3949, opened 2025-07-31
Make googletest dependency optional

vvolkl:
Dave reported an issue while rebuilding 2.13.2 from the source package using koji. Because that build is done without network access,  the build failed with a missing googletest dependency due to https://github.com/cvmfs/cvmfs/pull/3874

We've added a packaging patch to include googletest in the src package in https://github.com/cvmfs/cvmfs/tree/cvmfs-2.13.2-2 , but there's really no good reason for our unittest library to be a required dependency. This PR makes it so that googletest is only needed when the unittests are actually built, by adding a forwarding header `duplex_testing.h` that is included instead of `gtest_prod.h`.  It also makes the cvmfs rpm spec file skip the unittest package by default. It can still be built with `rpmbuild --define="build_testing 1"`. I don't really get the need to package the unittests though, and might remove this alltogether in the future ( this PR already removes the unittests from debian packaging ). I find that it's more important to have the option to run the unittests during the package build, failing early before even creating the package. Before we can enable this by default, the T_Namespace tests should be looked at or disabled though, because they make trouble when testing in a container.

Just in case, it also allows for the googletest source to be downloaded to the `externals` directory before cmake is invoked. This can serve as a template for the actually required, vendored dependencies that don't have to be checked into the repo with a CMake configured that way.


packaging/rpm/cvmfs-universal.spec
```diff
@@ -31,6 +31,11 @@
   %define build_snapshotter 1
 %endif
 
+# flag to control building of unittests.
+# No tests are built by default, can be overriden 
+# with rpmbuild --define="build_testing 1"
+%{!?build_testing: %define build_testing 0}
```
DrDaveD, line 37: Defining this by default to 0 isn't really necessary since the ifs already allow for it to be undefined, but I guess it's good for documentation.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -771,6 +788,8 @@ systemctl daemon-reload
 
 %changelog
 # - When using fuse3, require at least version 3.3.0 (for premounting).
+* Thu Jul 31 2025 Valentin Volkl <vavolkl@cern.ch>> - 2.14.0
```
DrDaveD: So if this isn't until 2.14.0, for 2.13.3 are you going to directly change what's now a patch in 2.13.2-2?  I don't think it's good to carry the patch forward.

DrDaveD (approved):

packaging/rpm/cvmfs-universal.spec
vvolkl: Yes, I was going to backport this to the 2.13 branch just in case we do another patch release. So maybe it's better to put 2.13.3, but if there is no more patch release before 2.14.0 I find it also confusing.

DrDaveD: There will need to be a 2.13.3 at least for #3947

PR #3950, opened 2025-07-31
Really fix updating of last_gc when no gc is run under gc -a without dry run

DrDaveD:
It turns out that the "fix" in #3947 didn't actually change any behavior.  That's what I get for changing it after testing.  This is what I meant to submit, after having tested it with `$dry_run -ne 1` but thinking it would be slightly better to go with comparing `$dry_run` with `0` instead.


vvolkl (approved):
Looks good this time.

PR #3951, opened 2025-08-01
CMake improvements

vvolkl:
* Switch on creation of  `compile_commands.json` by default. Not expensive, and useful for developers using an LSP.
* Skip checking of system headers
* rename external build and install dirs to include platform information, p.ex. `externals_install.x86_64.ubuntu22.04`. Useful for caching of those dirs outside of the repo root, when doing container builds for example.


vvolkl:
It's probably not wrong to add some logic to keep using legacy `externals_install` directories if they exist. Otherwise a new set of externals will be built in the new directory, but the build directory will keep using the old one, that can be a bit confusing. Will still add a patch

PR #3954, opened 2025-08-04
[rpm] Fix fuse3 libraries on suse

vvolkl:
The fuse3 libraries on SLES have a different package name. It seems like cvmfs using fuse3 on SUSE has never been tried before, because it seems like this has always been broken, and just surfaced when we switched to fuse3 by default recently.

I'll make a 2.13.2-3 build  for SUSE.


DrDaveD (approved):

vvolkl:
Test 621 failed with retval 82 here, maybe the test is flaky. Definitely unrelated to the changes here since that was on ubuntu.

vvolkl:
I've published the new release version in the repos and tested it, seems to work on opensuse/leap 15.5 and 15.6 . Added the zypper commands to the  quickstart docs as well.

PR #3956, opened 2025-08-06
Fix wrong variable names in cvmfs_server_gc, faulty cherry-pick of (#3575)

vvolkl:
https://github.com/cvmfs/cvmfs/pull/3895 renamed `all_collectable` to `all_collected`, but then, when  https://github.com/cvmfs/cvmfs/pull/3575 was cherry-picked back on devel, it added a block using the old variable names.

Should have been caught by tests, so that's a todo. Maybe we also do need the strict mode for bash.

### DIFF
--- a/cvmfs/server/cvmfs_server_gc.sh
+++ b/cvmfs/server/cvmfs_server_gc.sh
@@ -188,7 +188,7 @@ cvmfs_server_gc() {
   # Use /dev/shm for the lock file because it is world-writable and goes
   # away during reboot
   local gc_all_lock=/dev/shm/cvmfs_is_gcing_all
-  if [ $all_collectable -ne 0 ]; then
+  if [ $all_collected -ne 0 ]; then
     if ! acquire_lock $gc_all_lock; then
       to_syslog "skipping starting cvmfs_server gc -a because $gc_all_lock held by active process"
       return 1
@@ -232,7 +232,7 @@ cvmfs_server_gc() {
     fi
   done
 
-  if [ $all_collectable -ne 0 ]; then
+  if [ $all_collected -ne 0 ]; then
     release_lock $gc_all_lock
   fi
 }


DrDaveD:
Since this is already merged it doesn't allow me to add a review but it looks correct.  For the record, [this is the cherry-pick](https://github.com/cvmfs/cvmfs/commit/994f035406047e8f2149b3ce3e9f8e55e2377f73) that caused the problem and it was done outside of a PR so it didn't get reviewed,  It's probably good practice to do cherry-picks in PRs too.  This was cherry-picked into both devel and cvmfs-2.13 branches so it affects both.

How was this problem detected?

vvolkl:
> It's probably good practice to do cherry-picks in PRs too. 

Agreed. I've already cherry-picked this PR into 2.13, but will open PRs for that in the future.

> How was this problem detected?

I was looking at the test logs for the recent test failures, it had this in it:
```
/usr/bin/cvmfs_server: 7407: [: -ne: unexpected operator
/usr/bin/cvmfs_server: 7451: [: -ne: unexpected operator
```
A bit troubling that the tests didn't consistently detect this bug. And while looking at the test log just now I found another similar issue (but only in the test logic: #3957 ). I think Andrey made a great suggestion with enabling bash strict mode, we should look into that.

DrDaveD:
I'm not sure that bash strict mode would have caught this problem because it just always causes the test expression to return false. 

The impact of this bug is that it doesn't prevent more than one `gc -a` from happening at the same time, which is not a very common issue.

vvolkl:
> I'm not sure that bash strict mode would have caught this problem because it just always causes the test expression to return false.

That's not correct. `set -u` causes the program to immediately exit when it encounters an undefined variable:

```bash
cat > test1.sh<< 'EOF'
test1() {
  local gc_collectable1=yes
  if [ "$gc_collected1" = "yes" ]; then 
    echo "success"
  fi  
}
EOF
source test1.sh; test1; echo $?

cat > test2.sh << 'EOF'
set -u
test2() {
  local gc_collectable2=yes
  if [ "$gc_collected2" = "yes" ]; then 
    echo "success"
  fi  
}
EOF
source test2.sh; test2; echo $?
```
gives:
```bash
~$ source test1.sh; test1; echo $?
0

~$ source test2.sh; test2; echo $?
bash: gc_collected2: unbound variable
~$ echo $?
1
```
Clearly would have caught this bug as soon as cvmfs_server gc was invoked.

PR #3962, opened 2025-08-14
Include date and hostname in bugreport tarball filename

vvolkl:
A quality of life improvement for developers, who otherwise need to manually rename those files on download to keep track of them.


DrDaveD (changes requested):
Looks good but while you're at it why don't you also change the name of the `bugreport` path to match the name of the tar file.  You could do that with a tar `--xform` option.

vvolkl:
Good suggestion, easier to rename the directory before it gets tarred though.

PR #3964, opened 2025-08-19
Add cvmfs-posix-tools

andrey-utkin:
These are tools, modelled after popular file utilities, which offer the convenience of direct effect on CVMFS repo. We believe availability of such commands will make CVMFS administration more attractive to potential adopters.

These tools require "cvmfs_swissknife ingestsql" to be available. ingestsql in turn requires cvmfs_gateway.

One tool is named cvmfs_rsync. There is another script in the git repo called cvmfs_rsync. Please don't confuse them.

Some tests are currently failing which we are looking into, hence draft PR.


DrDaveD:
Wow, this is a very large contribution!

I'm not a primary maintainer of cvmfs but have some initial comments.

First, it's not very clear from glancing at the READMEs what these tools add over the standard posix tools.  Maybe that explanation could be added there.  Presumably these are for publishing machines, since cvmfs clients mount read-only, but that's not clear.

Next, there's a name clash with cvmfs_rsync, because the cvmfs-server package already has a /usr/bin/cvmfs_rsync.

Finally, this may be better distributed as a separate package under [cvmfs-contrib](https://cvmfs-contrib.github.io/) instead of integrated in the cvmfs package.

vvolkl:
Yes, thanks indeed for preparing this big contribution. Yes, the name clash with cvmfs_rsync is a bit unfortunate, but since cvmfs_rsync does little more than add a few options to rsync, and Jump Trading has been using these tools for a while,  I think we can replace it .

I would certainly package these tool in a separate package, but I have no problem with adding them in this repository.

More comments will follow.

DrDaveD:
> cvmfs_rsync does little more than add a few options to rsync, and Jump Trading has been using these tools for a while, I think we can replace it .

But no, there are multiple scripts in place that depend on /usr/bin/cvmfs_rsync to be what it is.  I just had somebody ask me about it within the last week.

DrDaveD:
On the other hand, the existing cvmfs_rsync command is part of the cvmfs-server package.  Are these posix tools only expected to be used on client machines?  I would guess so, because otherwise they probably would have noticed that the name was already used.  If that's the case, perhaps it's OK to have a completely different tool by the same name if they're only installed on different machines.  rpm won't let both packages be installed on the same machine.

andrey-utkin:
Some fixes coming soon (maybe Friday afternoon).

cvmfs/catalog_downloader.cc
```diff
@@ -5,11 +5,11 @@ int kCatalogDownloadMultiplier = 32;
 CatalogItem::CatalogItem(const shash::Any &hash) : hash_(hash) { }
 
 void TaskCatalogDownload::Process(CatalogItem *input) {
-  std::string const catalog_path;
-  shash::Any const catalog_hash;
+  std::string catalog_path;
+  shash::Any catalog_hash;
   // will PANIC if download failed
-  // catalog_mgr_->LoadCatalog(PathString("") /* not used */, *input->GetHash(),
-  // &catalog_path, &catalog_hash, NULL);
```
christge-sft, line 12: I understand that this was commented out because download failure was not behaving properly. Now that the code is back, is it tested against this scenario?

cvmfs/ingestion/pipeline.cc
```diff
@@ -258,3 +258,117 @@ void ScrubbingPipeline::Spawn() {
...
+      PANIC(NULL);
+  }
+}
+
+CompressHashPipeline::CompressHashPipeline()
+  : spawned_(false)
+{
+  // we don't need much parallelism for this pipeline
+  const unsigned int nfork_base = 1;
+
```
christge-sft, line 288: What each of these loops is supposed to be doing? Also what's behind the `nfork_base * <coef>`?

cvmfs/swissknife_ingestsql.cc
```diff
@@ -648,7 +648,11 @@ int swissknife::IngestSQL::Main(const swissknife::ArgumentList &args) {
   if (args.find('@') != args.end()) {
     proxy = *args.find('@')->second;
   } else {
-    proxy = retrieve_config(config_map, "CVMFS_HTTP_PROXY");
+    if (config_map.find("CVMFS_HTTP_PROXY") != config_map.end()) {
+      proxy = retrieve_config(config_map, "CVMFS_HTTP_PROXY");
+    } else {
+      proxy = "DIRECT";
```
christge-sft, line 671: I think that it's a deliberate decision to discourage users working on `DIRECT` mode, and somehow "force" them to set up their proxies in order to relieve Stratum1's from extreme workloads. Do we want to loose on this policy here?

cvmfs/swissknife_ingestsql.cc
```diff
@@ -441,9 +442,12 @@ static XattrList marshal_xattrs(const char *acl_string) {
...
   if (!equiv_mode) {
-    CUSTOM_ASSERT(
-        aclobj.Set("system.posix_acl_access", string(binary_acl, binary_size)),
-        "failed to set system.posix_acl_access (ACL size %ld)", binary_size);
+    if (!aclobj.Set("system.posix_acl_access",
+                    string(binary_acl, binary_size))) {
+      LogCvmfs(kLogCvmfs, kLogStderr,
+               "failed to set system.posix_acl_access (ACL size %ld)",
+               binary_size);
+    }
```
christge-sft, line 451: In the new version, execution won't abort after logging. And in this case we can't change the return value of the function to indicate failure.

cvmfs/swissknife_ingestsql.cc
```diff
@@ -517,6 +521,7 @@ static vector<string> get_file_list(string &path) {
 extern bool g_log_with_time;
 
 int swissknife::IngestSQL::Main(const swissknife::ArgumentList &args) {
+  int ret;
```
christge-sft, line 526: where may we find a return value interpretation?

christge-sft:
I tried to install the tools as described inside the `README.md` and installation failed. (I figured this out since all tests were failing individually). 

I tried both to install them all together with `sudo go install ./...` in the parent repo as well as one by one. Has anyone else faced the same problem? The installation script didn't issue any warning.
```
$ sudo go install ./cmd/cvmfs_chmod ; echo $? ; whereis cvmfs_chmod
0
cvmfs_chmod:
```

cvmfs/catalog_downloader.cc
andrey-utkin: No such test AFAIK.
This comment refers to the fact that `SimpleCatalogManager::LoadCatalog()` using PANIC() for every unsuccessful condition. I am going to add graceful runtime error handling there.

cvmfs/ingestion/pipeline.cc
andrey-utkin: This allocates the workers for a segmented pipeline.
The `nfork_base * <coef>` determines how many workers are created, `coef` represents either an empyrical or hunch-based optimal value. I will consult with the authors for more.
Probably makes sense for very high load or big batch jobs.

cvmfs/swissknife_ingestsql.cc
andrey-utkin: If this is the maintainers' policy, I am ready to follow it and change this. @vvolkl please confirm.

cvmfs/swissknife_ingestsql.cc
andrey-utkin: My fault, I will add error handling, changing the function's signature.

cvmfs/swissknife_ingestsql.cc
andrey-utkin: Here it's a variable to store the return value of called functions.
I can move its declaration closer to its use, this just tends to cause annoyance later sometimes, when similar value storing and using is needed earlier in the function, and ...::Main is a long one.

andrey-utkin:
Huge thanks for the review, @christge-sft !


> I tried to install the tools as described inside the `README.md` and installation failed. (I figured this out since all tests were failing individually).

I will check README.md and correct as needed.
The way to build and install these tools is to add `-DBUILD_POSIX_TOOLS` to cmake options.

DrDaveD:
> perhaps it's OK to have a completely different tool by the same name if they're only installed on different machines.  rpm won't let both packages be installed on the same machine.

Note that stratum 0 machines install both the cvmfs client and cvmfs-server, so these would need to be packaged in a separate package and not be in the same package with the cvmfs client.

cvmfs/catalog_downloader.cc
andrey-utkin: I've replaced `PANIC()`s with returning an error in `SimpleCatalogManager::LoadCatalog()`.
[It's unclear how to raise an error from `TaskCatalogDownload::Process()` as it returns void and it's an inherited method](https://github.com/cvmfs/cvmfs/issues/4048).
Apparently it's not even necessary to check return value of it in the caller, `TaskCatalogDownload::Process()`, because it then calls `WritableCatalogManager::CatalogDownloadCallback()` which does `catalog_download_map_.find(result.hash)` and finds out. The problem is that it just does an assert() on it so basically PANIC again :) I've now replaced it with `return;` as below.
I will work on writing a test for this.

andrey-utkin:
Hi @christge-sft , I think I have addressed the questions and change requests, could you please give another round of review?

cvmfs/ingestion/pipeline.cc
andrey-utkin: > This allocates the workers for a segmented pipeline. The `nfork_base * <coef>` determines how many workers are created, `coef` represents either an empyrical or hunch-based optimal value. I will consult with the authors for more. Probably makes sense for very high load or big batch jobs.

So Matt Harvey confirmed this. All the coefficients are ad-hoc picked, and we are free to change them. The only hard fact is that we in principle can allocate the workers in any quantity we want.

PR #3965, opened 2025-08-21
Replace AUTHORS with CITATION.cff

vvolkl:
No update to the content. The main reason for this change, apart from CITATION.cff being the commonly accepted format on github and Zenodo nowadays, is to aid the creation of new entries in Zenodo. Previously this would just scan the git history for authors, but with a CITATION.cff it honors our author list.

Unfortunately citation-file-format v1.3.0 is not released yet. It would have included more fine-grained contributor roles besides Authors. Until it is available I leave the further acknowledgements as a comment.


DrDaveD (approved):
Looks fine to me.  I didn't double-check the the orcid ids.

That zenodo link sure is out of date.

PR #3972, opened 2025-08-26
[rpm] silence a mkdir warning when /cvmfs already exists

vvolkl:
When cvmfs was previously installed from source, or /cvmfs/ exists due to some other reason, mkdir would print a non-fatal error message that isn't necessary.

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -791,12 +791,12 @@ systemctl daemon-reload
 
 %changelog
 # - When using fuse3, require at least version 3.3.0 (for premounting).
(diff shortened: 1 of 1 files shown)


packaging/rpm/cvmfs-universal.spec
```diff
@@ -793,6 +793,7 @@ systemctl daemon-reload
 # - When using fuse3, require at least version 3.3.0 (for premounting).
 * Thu Jul 31 2025 Valentin Volkl <vavolkl@cern.ch>> - 2.13.3
```
DrDaveD: You might want to update this date.

DrDaveD (approved):

PR #3973, opened 2025-08-28
Add extended info with cvmfs_config status <repo>

### LINKED ISSUE #3856: Command to show how far behind stratum 1s  are
It could be helpful for there to be a cvmfs_config command that shows how far behind at least the current stratum 1 is.  Maybe extend `cvmfs_config status` if someone adds a repository name.  Use `attr -g host /cvmfs/<repo>` as the host name and probe `http://<host>/.cvmfs_status.json`.  As a bonus compare it to `attr -g host_list /cvmfs/<repo>` and show which of that list is current and when they were all updated and their revision number (from `.cvmfspublished`) along with `attr -g revision` for comparison.

Maybe even throw in the logbuffer, or at least the last N lines of it.

DrDaveD:
- Fixes #3856

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -270,7 +270,7 @@ cvmfs_config_usage() {
   echo ""
   echo "  stat [-v | <repository>]"
   echo ""
-  echo "  status"
+  echo "  status [<repository list>]"
   echo ""
   echo "  probe [<repository list>]"
   echo ""
@@ -1150,14 +1150,30 @@ cvmfs_stat() {
 
 cvmfs_status() {
   local list
+  local repos="$@"
   list=""
 
   RETVAL=0
 
   local mounted_repos
+  local mountpoints
   local processed
-  mounted_repos=`list_mounts | awk '{print $3}'`
-  for mountpoint in $mounted_repos
+  mounted_repos="$(echo $(list_mounts | awk '{print $3}'))"
+  if [ -n "$repos" ]; then
+    cvmfs_readconfig || die "Failed to read CernVM-FS configuration"
+    for repo in $repos; do
+      mountpoint="$CVMFS_MOUNT_DIR/$repo"
+      if [[ " $mounted_repos " != *" $mountpoint "* ]]; then
+        echo $mountpoint is not mounted
+        RETVAL=1
+      else
+        mountpoints="$mountpoints $mountpoint"
+      fi
+    done
+  else
+    mountpoints="$mounted_repos"
+  fi
+  for mountpoint in $mountpoints
   do
     local fqrn=
     local pid=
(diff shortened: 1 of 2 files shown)


vvolkl:
I think it's good to have the option to list <repo> on the command line, but should this command also work with CVMFS_REPOSITORIES if that's set? That's the usual mechanism to specify repositories to operate on with cvmfs_config.

Also, this won't work for repositories mounted on non-standard paths (i.e, not in /cvmfs). There isn't really an easy and fast way to do this, as I've noted [elsewhere](https://github.com/cvmfs-contrib/prometheus-cvmfs/pull/5#issue-3269184874)

> Getting the list of mounted repos via cvmfs_config status is a surprisingly slow operation. I bypass this to just parse findmnt and assume all repos are mounted under their name in /cvmfs . There's an option to use the old behavior in case you use non-standard mountpoints. For a future release of cvmfs I think I'll fix this properly by exposing the repository name as a subtype.

I'll open an issue with this idea of adding the repo name as a fs subtype, then the repos can be easily parsed from findmnt.

DrDaveD:
> I think it's good to have the option to list on the command line, but should this command also work with CVMFS_REPOSITORIES if that's set? That's the usual mechanism to specify repositories to operate on with cvmfs_config.

That would be a surprising change to existing behavior.  I think the default behavior should stay unchanged and only show the mounted repositories.

> Also, this won't work for repositories mounted on non-standard paths (i.e, not in /cvmfs). There isn't really an easy and fast way to do this, as I've noted [elsewhere](https://github.com/cvmfs-contrib/prometheus-cvmfs/pull/5#issue-3269184874)

Oh.  There are other places in the script that use $CVMFS_MOUNT_DIR.  Should this call cvmfs_readconfig and use $CVMFS_MOUNT_DIR?

vvolkl:
> That would be a surprising change to existing behavior. I think the default behavior should stay unchanged and only show the mounted repositories.

No, I meant to say to use CVMFS_REPOSITORIES as an alternative to the list given on the command line `local repos="$@"`, if no  argument is given. You'd still check if those are mounted.

> There are other places in the script that use $CVMFS_MOUNT_DIR. Should this call cvmfs_readconfig and use $CVMFS_MOUNT_DIR?

I think that's better, yes. As mentioned, I'll add a better way to check for mounted repos, but that's for a future PR.

vvolkl:
> That would be a surprising change to existing behavior. I think the default behavior should stay unchanged and only show the mounted repositories.

No, I meant to say to use CVMFS_REPOSITORIES as an alternative to the list given on the command line `local repos="$@"`, if no  argument is given. You'd still check if those are mounted.

> There are other places in the script that use $CVMFS_MOUNT_DIR. Should this call cvmfs_readconfig and use $CVMFS_MOUNT_DIR?

DrDaveD:
> No, I meant to say to use CVMFS_REPOSITORIES as an alternative to the list given on the command line local repos="$@", if no argument is given. You'd still check if those are mounted.

Ok but that would still be a surprising change of behavior, since currently it checks all repos.

> I think that's better, yes. 

Ok I added a commit to use $CVMFS_MOUNT_DIR.

vvolkl:
LGTM, apart from the deletion of test 071. Was that intentional? At least a rudimentary test would be good to have.

DrDaveD:
> LGTM, apart from the deletion of test 071. Was that intentional? At least a rudimentary test would be good to have.

No, that was supposed to be a rename.  I must have accidentally left out the new name.  Fixed.

vvolkl (approved):

DrDaveD:
Should this get backported for 2.13.3 or wait for 2.14?

PR #3967, opened 2025-08-21
Change geoip database to openhtc source by default

DrDaveD:
This maintains the maxmind geoip database as an option for the stratum 1 geo api but changes the default to reading via https://geoipdb.openhtc.io which currently redirects to the download area of the [cvmfs-contrib/config-repo geoipdb tag release assets](https://github.com/cvmfs-contrib/config-repo/releases/tag/geoipdb).

PR #3982, opened 2025-09-04
Update geo api for either mmdb name

DrDaveD:
Unfortunately I didn't fully test the new geo db download in #3967, and we apparently didn't notice that the cloudtest 614-geoservice would have failed.  This fixes that situation, including removing the alternate database name when switching between the two different types.

Test 614-geoservice takes about 3 minutes so it isn't exactly "quick" but I wonder if we should include it in the quick tests anyway so it will always run.  Alternatively, maybe the Jenkins cloudtests configuration can be changed to use the new geo database.


vvolkl (approved):
I agree that the geodb test should run regularly, three minutes is acceptable so no problem with adding it to the quick suite.

PR #3988, opened 2025-09-15
add 614-geoservice test to quick suite

DrDaveD:
As discussed in PR #3982.


DrDaveD:
Actually it looks like test 614-geoservice makes some assumptions about where it is run and will need some fixing to be able to succesfully run from github actions.

vvolkl (approved):

vvolkl:
Thanks Dave!

PR #3991, opened 2025-09-18
S3FanoutManager: unhardcode "Cache-Control: max-age" value for PUT .cvmfs*

andrey-utkin:
It had hardcoded value of 61 second, which is remembered by e.g. Minio S3 server, which is inconvenient for testing.

S3FanoutManager can set max-age to any non-negative integer value. If no such override value is given, it looks for environment variables CVMFS_MAX_TTL_SEC or CVMFS_MAX_TTL; if not found, it uses the fallback value (same as was, 61).

To make `cvmfs_config mkfs` conform to desired setting, it can be put into /etc/cvmfs/default.d/99-something.conf .


andrey-utkin:
Updated wrt SEC vs SECS

Issue #3971, opened 2025-08-25
CVMFS_VERSION and CVMFS_VERSION_NUMERIC not available in config


DrDaveD:
The new CVMFS_VERSION and CVMFS_VERSION_NUMERIC variables added in #3934 unfortunately are not available to configuration scripts as intended.

The other variable set in the same function CVMFS_ARCH is also not available in configuration scripts.

vvolkl:
That's unfortunate, I took the code to set CVMFS_ARCH as a template and didn't expect it to be buggy. Will take a look and add an integration test.

PR #3994, opened 2025-09-22
Setup global environment params inside the OptionsManager

christge-sft:
This is an attempt to solve [issue-3971](https://github.com/cvmfs/cvmfs/issues/3971)

What is happening here basically is we're transferring the `SetupGlobalEnvironmentParams()` inside the `OptionsManager`.

The PR is marked as draft since probably some further cleanup is needed (Now the same functionality is provided from the`OptionsManager` and the `Filesystem`).

To test that the environment is setup properly, we provide a temporary `/etc/cvmfs/default.local` that reports if `CVMFS_VERSION` and `CVMFS_ARCH` are properly set.


test/src/108-global-config/main
```diff
@@ -0,0 +1,51 @@
...
+    fi 
+
+    sudo mv "${CVMFS_CONFIG_FILE}.bak" "${CVMFS_CONFIG_FILE}"
+  else
+    echo "config doesn't exist"
+  fi
+
+  sudo rm -rf $logfile
+
+#  cvmfs_mount "atlas.cern.ch,atlas-condb.cern.ch" \
```
vvolkl: Clean up commented code

test/src/108-global-config/main
```diff
@@ -0,0 +1,51 @@
...
+    if ! grep -q -- "CVMFS_ARCH" "$logfile"; then
+      return 2
+    fi 
+
+    sudo mv "${CVMFS_CONFIG_FILE}.bak" "${CVMFS_CONFIG_FILE}"
+  else
+    echo "config doesn't exist"
+  fi
+
+  sudo rm -rf $logfile
```
vvolkl: Move to a cleanup function that can be trapped

test/src/108-global-config/main
```diff
@@ -0,0 +1,51 @@
...
+
+cvmfs_test_name="Global Configuration Variables"
+
+CVMFS_CONFIG_FILE=/etc/cvmfs/default.local
+REPO=/cvmfs/atlas.cern.ch
+
+cvmfs_run_test() {
+  logfile=$1
+
+  local logfile=/tmp/version_env_test.log
```
vvolkl: Should eventually use a workspace directory that can be set from run.sh - but that's to be fixed for a  lot of integration tests,  fine for now.

test/src/108-global-config/main
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+cvmfs_test_name="Global Configuration Variables"
```
vvolkl: Not sure about the naming. I think I prefer "System Configuration Variables"

test/src/108-global-config/default.local
```diff
@@ -0,0 +1,10 @@
+CVMFS_REPOSITORIES=atlas-condb.cern.ch
+CVMFS_HTTP_PROXY="http://ca-proxy.cern.ch:3128"
```
vvolkl: This shouldn't be hardcoded, we need to use CVMFS_TEST_PROXY in order to be able to run the integration tests outside the CERN network. That's done in the common setup, I think it's better to just append to the default.local instead of overwriting it in the test.

cvmfs/options.cc
```diff
@@ -277,13 +277,30 @@ bool OptionsManager::HasConfigRepository(const string &fqrn,
   return false;
 }
 
+void OptionsManager::SetupGlobalEnvironmentParams() {
+  setenv("CVMFS_ARCH", GetArch().c_str(), 1 /* overwrite */);
```
vvolkl: I think it may be better to use the options_mgr SetValue function here, so these variables are also visible when doing `cvmfs_talk parameters`

vvolkl (changes requested):
Looks good, thanks! Some minor comments inline. Also, I'm not sure what happens when CVMFS is upgraded - will the VERSION variable reflect that? May need to add something to the reload code to make sure it does.

test/src/108-global-config/default.local
christge-sft: I adjusted the approach on that: `CVMFS_TEST_PROXY` is used now.
I think that having a very minimal default.local makes the test more readable. Although if you think that appending on the current configuration is better we can do that. Although, there should be a minimal configuration in case default.local is not  there. 

I'm resolving the trhead for now and we're opening it again if necessary.

test/src/108-global-config/main
christge-sft: For that with the new version I'm using the logfile provided through `$1`. So it's the responsibility of the wrapper to setup a proper logging path.

test/src/108-global-config/main
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -e
+cvmfs_test_name="System Configuration Variables"
```
vvolkl, line 3:
```suggestion
cvmfs_test_name="System Configuration Variables"
cvmfs_test_suites="quick"
```

DrDaveD (changes requested):
Please also remove the `FileSystem::SetupGlobalEnvironmentParams` function from cvmfs/mountpoint.cc as extended in #3934.

christge-sft:
> Please also remove the `FileSystem::SetupGlobalEnvironmentParams` function from cvmfs/mountpoint.cc as extended in #3934.

Done! I split the original commit into two. I kept the part where `CVMFS_VERSION` and `CVMFS_VERSION_NUMERIC` are set in `CMakeLists.txt` and I reverted the rest.

DrDaveD (changes requested):
Somehow a lot of irrelevant commits have been pulled in, touching many irrelevant files.  Please clean it up.  Perhaps you'll need to recreate your branch from scratch and cherry-pick just your changes.

I see you are using your own devel branch for this.  That can be OK but pull requests usually work most smoothly when you create a new branch for them and then delete them after the PR is merged and you have updated your devel branch from upstream.

christge-sft:
> Somehow a lot of irrelevant commits have been pulled in, touching many irrelevant files. Please clean it up. Perhaps you'll need to recreate your branch from scratch and cherry-pick just your changes.

The reason for that is that the original commit was doing both (setting `CVMFS_VERSION` and `CVMFS_VERSION_NUMERIC` in CMakeLists.txt and the setenv inside `FileSystem::SetupGlobalEnvironmentParams`.

So this is the commit I split, and this is why you see the history after that splitting changed. But the commits themselves are not affected at all.
> 
> I see you are using your own devel branch for this. That can be OK but pull requests usually work most smoothly when you create a new branch for them and then delete them after the PR is merged and you have updated your devel branch from upstream.

Let's decide on the final version, and I can create a feature branch on my fork to issue the PR from. Although, my fork is mostly (if not only) used by myself, so it's not an issue to do a hard reset after the PR is accepted.

DrDaveD:
That may be the explanation but as this is I can't see what is actually changed so it's really unreviewable.  I also don't recommend doing a revert at all, because in addition to only wanting to undo part of it, there are other parts that are not covered.  Instead, please remove `FileSystem::SetupGlobalEnvironmentParams` and anywhere it is called.

PR #3986, opened 2025-09-10
GSoC'25: Virtme-ng integration for CVMFS tests across kernels

ShivamMadlani:
This is a contribution done through GSoC, for more details on implementation and steps I followed, please refer to my blog post [here](https://shivammadlani.github.io/GSoC_project_blog/2025/10/03/extending-kernel-support.html)

This PR introduces a setup for automated testing on kernel builds for CVMFS test suite using Virtme-ng (a tool to build and boot to a kernel from its source code).

Key additions include:

- Shell scripts for building kernels and running tests (`build_kernel.sh`, `run.sh`, `guest/run_tests.sh`)
- Github workflow integration (`.github/workflows/ci_vng.yaml`)
- Multiple default kernel versions and configurations under `kernel/` and `configs/`
- Modular test cases in tests/ covering kernel checks, probe tests, fuse version checks, and more
- Logging and result tracking in `results/`
- Documentation

Benefits:

- Helps validate kernel functionality and CVMFS compatibility in a safe and isolated virtualized environment
- Early detection of integration issues between kernels and CVMFS ensures more stable releases
- Easy addition of kernel versions
- Centralized logging and result tracking

PR #3996, opened 2025-09-25
Dynamic vendoring of libfuse

ShivamMadlani:
This PR contains a feature addition to dynamically vendor libfuse using cmake.

Key features includes:
- New cmake flag `LIBFUSE_VERSION`
- Fetching, building, installing and linking given libfuse version to cvmfs libraries
- Documentation update for example usage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ RemoteSystemsTempFiles/
 # Externals
 externals/build_*/
 externals/.decompressionDone
+externals/libfuse*
 externals_build*
 externals_install*
 externals/*.tar*
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,6 +479,56 @@ if (BUILD_CVMFS OR BUILD_LIBCVMFS OR BUILD_LIBCVMFS_CACHE OR
   set (INCLUDE_DIRECTORIES ${INCLUDE_DIRECTORIES} ${PROTOBUF_INCLUDE_DIRS})
 endif ()
 
+# Fetch and build libfuse if a specific version is requested
+if (LIBFUSE_VERSION)
+  include(ExternalProject)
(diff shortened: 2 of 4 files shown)


DrDaveD:
What problem is this trying to solve?  I would like to be able to consider whether there is a better solution.

vvolkl:
This is related to a GSoC project - Shivam has been working this summer on allowing us to run tests against specific or custom kernel versions (https://github.com/cvmfs/cvmfs/pull/3986). The different platforms we are testing give us a few different kernels, but being able to choose the kernel version as a parameter can be helpful, in particular for new fuse features.

So when testing on a custom kernel, we might also want a custom libfuse - this is what this PR is trying to do. It should certainly make it clear that it is only for testing, the cmake argument probably shouldn't even be documented.

.gitignore
```diff
@@ -30,6 +30,7 @@ RemoteSystemsTempFiles/
 # Externals
 externals/build_*/
 externals/.decompressionDone
+externals/libfuse*
```
vvolkl, line 33: This change is probably leftover, please remove it

vvolkl:
@ShivamMadlani I don't think you should set FUSE3_INCLUDE_DIR and FUSE3_LIBRARY manually, or have a separate libfuse_external target. Once you've built libfuse, it should be picked up and linked automatically from here:

https://github.com/cvmfs/cvmfs/blob/59dd97a431e5a8f99f32145b955e1f5cd36c7e3e/CMakeLists.txt#L499

You might need to adjust the CMAKE_PREFIX_PATH and possibly the FindFUSE3.cmake .

.gitignore
ShivamMadlani: Ah, my bad!

ShivamMadlani:
@vvolkl Apologies for the misunderstanding, I have drafted #3998, and it address the scope more clearly.

vvolkl:
Superseded by #3998

PR #3999, opened 2025-09-28
Setup global environment params inside the OptionsManager

christge-sft:
This is an attempt to solve https://github.com/cvmfs/cvmfs/issues/3971, also a cleaner version of closed [PR-3994](https://github.com/cvmfs/cvmfs/pull/3994)

What is happening here basically is we're transferring the SetupGlobalEnvironmentParams() inside the OptionsManager.

The PR is marked as draft since probably some further cleanup is needed (Now the same functionality is provided from theOptionsManager and the Filesystem).

To test that the environment is setup properly, we provide a temporary /etc/cvmfs/default.local that reports if CVMFS_VERSION and CVMFS_ARCH are properly set.


DrDaveD (approved):
This is looking better to me.  I believe the cloudtest failures will be addressed in #4000.  I'm not sure about the other failures.

vvolkl (approved):
The test failures are unrelated, looks like a memory issue in the mockfuse test setup.

cvmfs/mountpoint.h
```diff
@@ -689,3 +686,4 @@ class MountPoint : SingleCopy, public BootFactory {
 };  // class MointPoint
 
 #endif  // CVMFS_MOUNTPOINT_H_
+
```
vvolkl:
```suggestion
```

PR #4001, opened 2025-09-30
Add publisher support to add and remove file bundle specs

jblomer:
File bundle specifications are linked to a bundle trigger file in the same directory.


vvolkl (approved):

vvolkl:
Changed base to use the `filebundle` feature branch

PR #4002, opened 2025-10-02
File bundle prefetching

vvolkl:
# File bundle prefetching

Introduce *file bundles*: a way to speedup application startup by declaring
 that opening one file should trigger prefetching of a set of related files.

A bundle is described by a `.cvmfsbundle-<filename>` specification file
placed next to its trigger `<filename>` in the same directory. The spec is a
small versioned JSON document listing the files to prefetch:

```json
{
  "name": "CVMFS_BUNDLE",
  "version": "1.0.0",
  "encoding": "UTF-8",
  "dependencies": [
    "/absolute/path/from/repo/root/libfoo.so",
    "/absolute/path/from/repo/root/data.bin"
  ]
}
```

On the publisher side, the sync mediator picks up these specification files
and links each bundle to its trigger file (exposed via the `bundle_trigger`
magic xattr). When a client then `open()`s the trigger file, a dedicated
fetcher thread pool prefetches and decompresses the listed dependencies in
parallel.

## User-facing changes

- **Client option `CVMFS_PREFETCH_FILEBUNDLES`** (boolean, off by default)
   to enable bundle prefetching on all repositories. 
- **`.cvmfsbundle-<filename>` specification files** are recognized by the
  publisher; `cvmfs_server` add/remove of these specs links them to their
  trigger file.
- New magic xattr `bundle_trigger` (`1`/`0`) indicating whether a file is a
  bundle trigger.

Tested with `test/src/709-cvmfsbundles` and unit tests for the bundle and
bundle-file managers.

For more details, see also [this contribution at CHEP 2026](https://indico.cern.ch/event/1471803/contributions/6967096/).

Assisted-by: Claude (claude-opus-4-8)


cvmfs/cvmfs.cc
```diff
@@ -1179,11 +1179,19 @@ static void cvmfs_open(fuse_req_t req, fuse_ino_t ino,
 
   perf::Inc(file_system_->n_fs_open());  // Count actual open / fetch operations
 
-  if (dirent.IsBundleTrigger()) {
+  std::string is_prefetching_enabled;
+  if (options_mgr_) {
+    options_mgr_->GetValue("CVMFS_FUSE_NOTIFY_INVALIDATION",
```
DrDaveD: I'm confused by this name.  What does that have to do with prefetching?

vvolkl: Ys, that's a copy-paste artefact. I have some updates to this branch that my presentation is based on, will push soon.

PR #4003, opened 2025-10-03
test: add 904-umount-should-stop-cvmfs2

andrey-utkin:
An unexpectedly failing test which unmounts the actual cvmfs mount and expects cvmfs2 processes to be gone, and for success in mounting it again.

Still wondering: is it just me? Or is it "normal" that they linger? Or is there a setting?

### DIFF
--- a/test/src/904-umount-should-stop-cvmfs2/main
+++ b/test/src/904-umount-should-stop-cvmfs2/main
@@ -1,0 +1,19 @@
+#!/bin/bash
+cvmfs_test_name="umount-should-stop-cvmfs2"
+cvmfs_test_autofs_on_startup=false
+cvmfs_test_suites="quick"
+
+cleanup() {
+  true
+}
+
+cvmfs_run_test() {
+  [[ -v TEST_ROOT ]] # expected /home/sftnight/cvmfs/test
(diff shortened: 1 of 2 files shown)


vvolkl:
It think that this isn't really related to cvmfs, but rather is a "feature" of overlayfs ( `/var/spool/cvmfs/<repo>/rdonly` is the lowerdir for the overlay on `/cvmfs/<repo>` ). I've tried this out, and the cvmfs2 processes go away when I do `sudo umount /cvmfs/$CVMFS_TEST_REPO` - it's kind of reasonable to first unmount the overlay before unmounting its lower dir, no?

Also note that these mounts are typically managed via the cvmfs_server command (which itself uses `cvmfs_suid_helper` for that).

andrey-utkin:
Thanks for the clarification!
Indeed when I unmount the overlayfs, it works well (the cvmfs2 briefly stay after umount is completed, but they never hang).
It also reliably works if I unmount cvmfs first, and overlayfs later.

It would be less confusing if unmounting cvmfs from under overlayfs just didn't work, saying "is busy" as it usually does. In this case, cvmfs is successfully unmounted (the line is gone from `/proc/mounts`), but the processes stay.

PR #4006, opened 2025-10-07
Implement FUSE passthrough

andrey-utkin:
Passthrough is a recent FUSE feature (kernel 6.17.0+, libfuse 3.17+)
enabling bypass of userspace on all the read requests. When the FUSE
daemon provides a "backing file descriptor" to FUSE on "open" request,
the kernel reads directly from it without calling into the FUSE daemon.

This is a win for performance. The test supplied demonstrates
measurable speedup:

```
+ dd status=progress bs=512 if=/cvmfs/test.repo.dumbdumb/big.bin of=/dev/null
10485760 bytes (10 MB, 10 MiB) copied, 3.35429 s, 3.1 MB/s
+ dd status=progress bs=512 if=/cvmfs/test.repo.passthru/big.bin of=/dev/null
10485760 bytes (10 MB, 10 MiB) copied, 0.073326 s, 143 MB/s
```

```
+ pushd /cvmfs/test.repo.dumbdumb
+ parallel -N0 'sha1sum --check --status CHECKSUMS' ::: {1..100}
real    0m3.440s
user    0m4.782s
sys     0m3.345s
+ pushd /cvmfs/test.repo.passthru
+ parallel -N0 'sha1sum --check --status CHECKSUMS' ::: {1..100}
real    0m1.221s
user    0m3.904s
sys     0m1.709s
```

Enabling new functionality requires
* building with sufficiently recent libfuse to have FUSE_CAP_PASSTHROUGH
* running with similarly recent enough libfuse
* running on Linux kernel 6.17.0 or newer
* specifying mount option fuse_passthrough (or fuse_passthru)

Related documentation update:
https://github.com/cvmfs/doc-cvmfs/pull/254


andrey-utkin:
`macos_build_and_test` CI failure:
```
src/108-fuse-passthru/main: line 11: conditional binary operator expected
src/108-fuse-passthru/main: line 11: syntax error near `TEST_ROOT'
src/108-fuse-passthru/main: line 11: `  [[ -v TEST_ROOT ]] # expected /home/sftnight/cvmfs/test'
```

Does there happen to be a plan to swithch to running this in `bash`? Surely bash is available and would enable more elegant coding than POSIX sh?

vvolkl:
Very nice, thank you!

I think it may be worthwhile to add a config option CVMFS_FUSE_PASSTHROUGH or similar.  With autofs it can be a bit awkward to add mount options.

>Does there happen to be a plan to swithch to running this in bash? Surely bash is available and would enable more elegant coding than POSIX sh?

This is running in bash in MacOS, just a really ancient version (3.2 or something) which doesn't support `-v`. I think a recent version can be installed with homebrew though.

cvmfs/loader.cc
```diff
@@ -169,8 +173,11 @@ static void Usage(const string &exename) {
                             "before mounting (required for autofs)\n"
     "  -o parse             Parse and print cvmfs parameters\n"
     "  -o cvmfs_suid        Enable suid mode\n"
+    "  -o debug             Enable debug to CVMFS_DEBUGLOG\n"
```
vvolkl, line 177: Does this actually work?

andrey-utkin: Yes, but.

I picked this up from other tests and such. It does work, but only if CVMFS_DEBUGLOG is in client.conf; no effect if put into client.local. I intend to check this again and submit a bug with a test.

andrey-utkin:
> I think it may be worthwhile to add a config option CVMFS_FUSE_PASSTHROUGH or similar.

Do you mean a cmake/compile option, or runtime config option?
I agree to both.

andrey-utkin:
Please review.

cvmfs/loader.cc
andrey-utkin: > but only if CVMFS_DEBUGLOG is in client.conf; no effect if put into client.local

I mistakenly thought that client.local path is /etc/cvmfs/..., but it's /var/spool/cvmfs/... .
So basically `-o debug` + `CVMFS_DEBUGLOG` work.

DrDaveD:
This is odd that that there's only 3 CI checks pending and none of them are running.  I'm guessing that this needs a rebase on the current devel branch.

vvolkl (approved):
LGTM, thanks!

jblomer (commented):
Sorry for commenting only now: I think there needs to be a check for the cache manager type. If it's not the normal POSIX cache manager, the returned file descriptor is artificial and cannot be used for passthrough.

Issue #3993, opened 2025-09-22
hanging clients with 2.13.2


thdesy:
Dear developers,

over the weekend we observed a number of our grid workers to experience again stuck CVMFS clients/mounts.
Installed kernel and packages are as of [1]

We have the Prometheus metric exporter rolled out. Primarily an observation and probably to be expected wrt cvmfs2 processes being unresponsive, but the Prometheus metric exporter is reporting only a very stripped down number of metrics on affected clients [2]. After killall'ing the cvmfs processes, the metrics recover [3]

A bugreport tarball of an affected client is available under
https://syncandshare.desy.de/index.php/s/sBdR4295XjefzxW
password for the download is  `a52697b81d363611`  (to deter bots a bit)
(`cvmfs_config status,stat` and `df` timeouted with the kernel and being unhappy)

At least this time there are some remaining cvmfs2 processes being uninterruptable sleeping [4]. On the nodes here, I had been inspecting, there were remaining mount processes wrt lhcb-condb.cern.ch (causal connection with the repo??). `strace` tends to unresponsive when attaching to the mount processes and does not react itself to SIGTERM but only hard SIGKILL.
I have included perf reports on the mount and cvmfs2 processes from node `batch1019` but the mount process seems to be primarily waiting for its cvmfs2 child, which  looks to deathloop in wait/pselect [5]

I am keeping the two nodes batch1004 & batch1019 disabled and non'killall'ed in case some more debug information needs to be selected.

Cheers,
  Thomas

[1]
```
5.14.0-570.44.1.el9_6.x86_64

cvmfs-2.13.2-1.el9.x86_64
cvmfs-config-egi-2.7-1.8.obs.el9.noarch
cvmfs-fuse3-2.13.2-1.el9.x86_64
cvmfs-libs-2.13.2-1.el9.x86_64

fuse-2.9.9-17.el9.x86_64
fuse3-3.10.2-9.el9.x86_64
fuse3-libs-3.10.2-9.el9.x86_64
fuse-common-3.10.2-9.el9.x86_64
fuse-libs-2.9.9-17.el9.x86_64
```

[2]
```
[root@batch1015 ~]# curl localhost:9868
# HELP cvmfs_sys_maxfd Shows the maximum number of file descriptors available to file system clients.
# TYPE cvmfs_sys_maxfd gauge
cvmfs_sys_maxfd{repo="config-egi.egi.eu"} 
[...]

PR #4009, opened 2025-10-16
use umount handler in loader; should avoid autofs race

vvolkl:
#3993

### DIFF
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -181,14 +181,14 @@ static void Usage(const string &exename) {
 
 void UmountOnCleanup(const string* mountpoint_) {
   if (!mountpoint_) {
-    LogCvmfs(kLogCvmfs, kLogSyslogErr, "crash cleanup handler: no mountpoint");
+    LogCvmfs(kLogCvmfs, kLogSyslogErr, "unmount on cleanup: no mountpoint");
     return;
   }
 
   std::vector<std::string> all_mountpoints = platform_mountlist();
   if (all_mountpoints.empty()) {
     LogCvmfs(kLogCvmfs, kLogSyslogErr,
-             "crash cleanup handler: "
+             "unmount on cleanup: "
              "failed to read mount point list");
     return;
   }
@@ -206,7 +206,7 @@ void UmountOnCleanup(const string* mountpoint_) {
     }
   }
   if (!still_mounted) {
-    LogCvmfs(kLogCvmfs, kLogSyslog, "crash cleanup handler: %s not mounted",
+    LogCvmfs(kLogCvmfs, kLogSyslog, "unmount on cleanup: %s not mounted",
              mountpoint_->c_str());
     return;
   }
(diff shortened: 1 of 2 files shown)


DrDaveD (commented):
I thought that there was going to be a common function used here and by UmountOnCrash in auto_umount.cc.   Did that turn out to be impractical?

cvmfs/loader.cc
```diff
@@ -179,6 +179,77 @@ static void Usage(const string &exename) {
...
+      closedir(dirp);
+    LogCvmfs(kLogCvmfs, kLogSyslog,
+             "unmount un cleanup: "
+             "%s seems not to be stalled (%d)",
+             mountpoint_->c_str(), errno);
+    return;
+  }
+
+  // sudo umount -l *mountpoint_
+  if (!SwitchCredentials(0, getegid(), true)) {
```
DrDaveD, line 216: If you're doing the SwitchCredentials here then it doesn't also need to be done before calling this function.

cvmfs/loader.cc
```diff
@@ -179,6 +179,77 @@ static void Usage(const string &exename) {
      exename.c_str());
 }
 
+void UmountOnCleanup(const string* mountpoint_) {
+  if (!mountpoint_) {
+    LogCvmfs(kLogCvmfs, kLogSyslogErr, "crash cleanup handler: no mountpoint");
```
DrDaveD: Here and later in the function there are more "crash cleanup handler" messages that should be changed to "umount on cleanup".

DrDaveD (changes requested):

vvolkl:
> I thought that there was going to be a common function used here and by UmountOnCrash in auto_umount.cc. Did that turn out to be impractical?

Yes, that's the primary reason this is a draft pull request, I wanted to quickly have a patch to test, will still need to figure out where to put the common function.

cvmfs/loader.cc
DrDaveD: That difference in messages could be best handled by a parameter to a common function.

DrDaveD:
I pushed the current state of my version here.  It doesn't hang, but I'm not yet convinced it's doing what it's supposed to because I'm not seeing the debug messages I expect.

vvolkl:
Closing in favor of #4014

DrDaveD:
I updated this one to just search for mounted repositories before unmounting.  I think that might still be a viable option if #4014 turns out to not work, so I'm reopening for now to see how this does with the tests.

DrDaveD:
There's an unexplained, not locally reproducible problem with test 018, or perhaps it's an issue with the previous test 016.  Here's the log snippet for the record:
```
-- Testing DNS Unreachable (Wed Oct 22 12:12:48 UTC 2025 / test number 016)
restarting autofs...
using systemd
trying to mount atlas.cern.ch first
check if the repo is listable
repo
test1
unmounting and cleaning
trying to mount again with unreachable DNS
timeout was 28723 seconds
timeout was too long: 28723 (expected at most 2000)
check if repo is _not_ listable
Hangup
KILLING:39085
execution took 300.018 seconds
Testcase timed out

-- Testing DNS Timeout (Wed Oct 22 12:17:48 UTC 2025 / test number 017)
test case not part of selected suites

-- Testing HTTP Server Unreachable (Wed Oct 22 12:17:48 UTC 2025 / test number 0
18)
                     USER        PID ACCESS COMMAND
/cvmfs/atlas.cern.ch:
                     root      mount /cvmfs
                     USER        PID ACCESS COMMAND
/cvmfs/cvmfs-config.cern.ch:
                     root      mount /cvmfs
failed to clean environment (retval: 100)
failed to setup environment (retval: 101)
```

Closing since we're going with #4014 instead.

PR #4010, opened 2025-10-17
ingestsql settings convenience

andrey-utkin:


vvolkl:
As a general comment, it's intentional that CVMFS_HTTP_PROXY doesn't have a DIRECT default value  - it's to nudge users in a shared, distributed architecture to configure a proxy, in order to take load of the (shared) stratum-1s. In our high enery physics infrastructure, it still happens from time to time that stratum-1s get overloaded by misconfigured clusters of clients and admins have to intervene.

For the ingestsql tool that's probably not a concern, though maybe @DrDaveD has comments

andrey-utkin:
I agree for merging only the first commit.

andrey-utkin:
And I can drop `default_val` parameter as well then.

DrDaveD:
Server side parameters are different; it's the client where we want to force people to set a proxy.

Having said that, there should be no reason for it to be explicitly set in this code because the client.conf server configuration already includes CVMFS_HTTP_PROXY=DIRECT.   

Under what scenario would this PR be needed?  Likewise, CVMFS_STRATUM0 should also always be in server.conf.

andrey-utkin:
> Under what scenario would this PR be needed?

Config option not being set should not cause an abort and core dump, which it currently does.
This PR makes it print an error and exit gracefully with non-zero exit code.

> Having said that, there should be no reason for it to be explicitly set in this code because the client.conf server configuration already includes CVMFS_HTTP_PROXY=DIRECT.

ingestsql consults `/etc/cvmfs/gateway-client/$REPO_NAME/config` (where `/etc/cvmfs/gateway-client/` part can be replaced through `-C dir` argument). I understand it should do it differently to align with the way of other components of CVMFS. I will look into what other client tools do, but I understand there is a list of files and dirs to source, and then actual env vars override the settings from the files. But I'm not sure what's the correct list of files, given that "Server side parameters are different". `/etc/cvmfs/default.conf` describes an order, is it correct and does it apply to client tools? And does it mean /etc/cvmfs/repositories.d/ should not be sourced?

```
[root@824289a06535 /]# cat /etc/cvmfs/default.conf
# Don't edit here.  Create /etc/cvmfs/default.local.
# As a rule of thumb, overwrite only parameters you find in here.
# If you look for any other parameter, check /etc/cvmfs/domain.d/<your_domain>.(conf|local)
# and /etc/cvmfs/config.d/<your_repository>.(conf|local)
#
# Parameter files are sourced in the following order
#
# /etc/cvmfs/default.conf
# /etc/cvmfs/default.d/*.conf (in alphabetical order)
# $CVMFS_CONFIG_REPOSITORY/etc/cvmfs/default.conf (if config repository is set)
# /etc/cvmfs/default.local
#
# $CVMFS_CONFIG_REPOSITORY/etc/cvmfs/domain.d/<your_domain>.conf (if config repository is set)
# /etc/cvmfs/domain.d/<your_domain>.conf
# /etc/cvmfs/domain.d/<your_domain>.local
#
# $CVMFS_CONFIG_REPOSITORY/etc/cvmfs/config.d/<your_repository>.conf (if config repository is set)
# /etc/cvmfs/config.d/<your_repository>.conf
# /etc/cvmfs/config.d/<your_repository>.local
```

DrDaveD:
`default.conf` shows the default order used when the `cvmfs2` client is used as an ordinary client mounting repositories from stratum 1s, normally via proxy caches.  The `cvmfs_server` command changes that default list when it uses the client for its own purposes.  It uses the `config=` mount option which you can usually see in `etc/fstab` or at least by doing a `ps` and looking at the mount options passed to `cvmfs2` processes.

DrDaveD:
Oh and `cvmfs_server` uses a different list of configuration files to load the environment for `cvmfs_swissknife`.   Basically it is just `/etc/cvmfs/repositores.d/<repo>/server.conf` for a stratum0-style repository for publishing (made by `mkfs`) and `/etc/cvmfs/repositories.d/<repo>/replica.conf` for stratum1-style replicas (made by `add-replica`).  I'm not sure where `/etc/cvmfs/gateway-client` is coming from.  I'm not very familar with the gateway feature and do not see it mentioned at all in `cvmfs_server`.

PR #4014, opened 2025-10-21
loader: do not attempt umount when fuse exits with ret 0

vvolkl:
Another possible fix for #3993 . Here we do not even  attempt  the umount when the fuse loop exits with return value 0, very effectively avoiding the complications of a possible race with another following mount attempt.

However with  this patch,  a cvmfs repository needs to be manually unmounted when the filesystem connection in /sys/fs/fuse is aborted. Since this is rare enough and will only happen on admin intervention, it should be fine.

### DIFF
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -1231,20 +1231,20 @@ int FuseMain(int argc, char *argv[]) {
   config_files_ = NULL;
   socket_path_ = NULL;
 
-  if (retval != 0)
+  // Decide whether to attempt a umount before exiting.
+  // The fuse main loop exists with 0 either when it is already umounted,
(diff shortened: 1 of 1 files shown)


cvmfs/loader.cc
```diff
@@ -1238,7 +1238,11 @@ int FuseMain(int argc, char *argv[]) {
 
 #if CVMFS_USE_LIBFUSE != 2
   if (premount_fd >= 0) {
-    goto cleanup;
+    if (retval == kFailOk) {
+      goto fuse_exit_success;
```
DrDaveD: Why not just close premount_fd and continue here?  I don't see a need for `fuse_exit_success`.

PR #4015, opened 2025-10-22
loader: delay mount until fuse session invocation

vvolkl:
Doing the mount just before where libfuse would do it; needs to elevate/drop permissions an additional time but avoids a potential deadlock in InitSystemFs when trying to mount over an existing cvmfs repository.

### DIFF
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -1024,7 +1024,7 @@ int FuseMain(int argc, char *argv[]) {
 #if CVMFS_USE_LIBFUSE != 2
   if (!premounted_ && !suid_mode_ && getuid() == 0) {
     // If not already premounted or using suid mode, premount the fuse
-    // mountpoint before dropping privileges to avoid the need for fusermount.
+    // mountpoint  to avoid the need for fusermount.
     // Requires libfuse >= 3.3.0.
     //
     if ((uid_ != 0) || (gid_ != 0)) {
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -884,7 +884,9 @@ int FuseMain(int argc, char *argv[]) {
 
 
   int fd_mountinfo = -1;  // needs to be declared before start using goto
+#if CVMFS_USE_LIBFUSE != 2
   int premount_fd = -1;
+#endif
 
   // Drop credentials
   if ((uid_ != 0) || (gid_ != 0)) {
@@ -1025,11 +1027,13 @@ int FuseMain(int argc, char *argv[]) {
     // mountpoint before dropping privileges to avoid the need for fusermount.
     // Requires libfuse >= 3.3.0.
     //
-    if (!SwitchCredentials(0, getgid(), true)) {
-      LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogErr,
-               "failed to re-gain root permissions for mounting");
-      retval = kFailPermission;
-      goto cleanup;
+    if ((uid_ != 0) || (gid_ != 0)) {
+      if (!SwitchCredentials(0, getgid(), true)) {
+        LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogErr,
+                 "failed to re-gain root permissions for mounting");
+        retval = kFailPermission;
+        goto cleanup;
+      }
     }
     platform_stat64 info;
     // Need to know if it is a directory or not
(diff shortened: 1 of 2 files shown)


cvmfs/loader.cc
```diff
@@ -882,52 +882,9 @@ int FuseMain(int argc, char *argv[]) {
...
-      LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogErr,
-               "Failed to mount -t fuse -o %s cvmfs2 %s (%d)", opts,
-               mount_point_->c_str(), errno);
-      return kFailPermission;
-    }
-  }
-#endif
 
   int fd_mountinfo = -1;  // needs to be declared before start using goto
+  int premount_fd = -1;
```
DrDaveD, line 888: Should be surrounded by `#ifdef CVMFS_USE_LIBFUSE != 2`

cvmfs/loader.cc
```diff
@@ -1062,6 +1019,69 @@ int FuseMain(int argc, char *argv[]) {
     }
   }
 
+#if CVMFS_USE_LIBFUSE != 2
+  if (!premounted_ && !suid_mode_ && getuid() == 0) {
+    // If not already premounted or using suid mode, premount the fuse
+    // mountpoint before dropping privileges to avoid the need for fusermount.
```
DrDaveD: This comment isn't quite correct anymore because it's not before dropping privileges

cvmfs/loader.cc
```diff
@@ -1062,6 +1019,69 @@ int FuseMain(int argc, char *argv[]) {
...
+    if (mount("cvmfs2", mount_point_->c_str(), "fuse", flags, opts) == -1) {
+      LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogErr,
+               "Failed to mount -t fuse -o %s cvmfs2 %s (%d)", opts,
+               mount_point_->c_str(), errno);
+      return kFailPermission;
+    }
+  }
+  // Drop credentials
+  if ((uid_ != 0) || (gid_ != 0)) {
+    LogCvmfs(kLogCvmfs, kLogStdout, "CernVM-FS: running with credentials %d:%d",
```
DrDaveD: Remove this duplicate message

cvmfs/loader.cc
```diff
@@ -1062,6 +1019,69 @@ int FuseMain(int argc, char *argv[]) {
...
+             uid_, gid_);
+    const bool retrievable = (suid_mode_ || !disable_watchdog_);
```
DrDaveD: Wait .. if this is only dropped under these conditions, it should only be raised under the same conditions.

DrDaveD (changes requested):

cvmfs/loader.cc
DrDaveD: Oh, sorry, I was in a hurry.  This is not about whether or not to call SwitchCredentials, it's just whether or not it sets the real uid or effective uid.   So in any case, the other comment is that this whole `if` block is already checking for `!suid_mode_`, so that will never be true here.  But it doesn't really hurt.

cvmfs/loader.cc
```diff
@@ -1062,6 +1021,69 @@ int FuseMain(int argc, char *argv[]) {
     }
   }
 
+#if CVMFS_USE_LIBFUSE != 2
+  if (!premounted_ && !suid_mode_ && getuid() == 0) {
+    // If not already premounted or using suid mode, premount the fuse
+    // mountpoint  to avoid the need for fusermount.
```
DrDaveD: Nit double-blank in comment.

DrDaveD (approved):

cvmfs/loader.cc
```diff
@@ -1062,6 +1021,69 @@ int FuseMain(int argc, char *argv[]) {
...
+    if (mount("cvmfs2", mount_point_->c_str(), "fuse", flags, opts) == -1) {
+      LogCvmfs(kLogCvmfs, kLogStderr | kLogSyslogErr,
+               "Failed to mount -t fuse -o %s cvmfs2 %s (%d)", opts,
+               mount_point_->c_str(), errno);
+      return kFailPermission;
+    }
+  }
+  // Drop credentials
+  if ((uid_ != 0) || (gid_ != 0)) {
+    const bool retrievable = (suid_mode_ || !disable_watchdog_);
```
DrDaveD: This could just be set to true, because if it wasn't retrievable going into this block, we wouldn't have been able to retrieve the credentials at the beginning of the block.

PR #4017, opened 2025-10-23
add CVMFS_PREMOUNT_FUSE option to allow fallback to fusermount

vvolkl:


DrDaveD (approved):

vvolkl:
Should have an integration test, but I've tested locally that it invokes fusermount if set to no, and premounts if set to yes or unset.

PR #4018, opened 2025-10-23
Add test of cleanup after soft kill

DrDaveD:
This adds a test for one of the apparently few ways that a premount is now unmounted by loader.cc since #4014.


vvolkl (approved):
Thanks! The code path is yet different when mounting under uid=0, right? I may add another test for that.

DrDaveD:
Yes, with uid=0 it lets libfuse do the mounting and unmounting so it is another code path.

PR #4022, opened 2025-10-27
Cleanly handle when there is no geo db desired

DrDaveD:
While writing up the documenation for the new geoip db source I noticed that the documented `CVMFS_GEO_DB_FILE=NONE` and the allowed `CVMFS_UPDATEGEO_SOURCE=none` both worked but printed disconcerting errors each time snapshot ran.  This removes those error messages.

### DIFF
--- a/cvmfs/server/cvmfs_server_add_replica.sh
+++ b/cvmfs/server/cvmfs_server_add_replica.sh
@@ -149,7 +149,7 @@ cvmfs_server_add_replica() {
         echo "       user ID and run update-geodb monthly as that user"
         echo "    4. Use another update tool such as Maxmind's geoipupdate and"
         echo "       set CVMFS_GEO_DB_FILE to point to the downloaded file"
-        echo "    5. Disable the geo api with CVMFS_GEO_DB_FILE=none"
+        echo "    5. Disable the geo api with CVMFS_GEO_DB_FILE=NONE"
         echo "  See 'Geo API Setup' in the cvmfs documentation for more info."
       fi
     else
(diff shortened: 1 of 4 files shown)


vvolkl:
What was the error message? Would be good to add it to the commit message in case someone searches for it.

Are `none` and `NONE` case sensitive for CVMFS_UPDATE_GEOSOURCE and CVMFS_GEO_DB_FILE respectively? Can we  make them case insensitive?

DrDaveD:
They are case sensitive.  I'll add checks for both NONE and none; I don't think we need to support other case combinations.

The error message for `CVMFS_UPDATEGEO_SOURCE=none` was
```
Installing GeoIP Database... CVMFS_UPDATEGEO_URLSUFFIX () ends in unrecognized suffix
fail
```

Now I'm not sure about an error with `CVMFS_GEO_DB_FILE=NONE` because I wasn't able to reproduce it.  It was making a nonsense symlink though, pointing the db file name in /var/lib/cvmfs-server/geo to `NONE`.

DrDaveD:
@vvolkl This is ready for review.

DrDaveD:
@vvolkl Please review this

vvolkl:
LGTM, except for that I'm not sure I understand what the issue with `snapshot_2_log` is in the integration test and why you remove it

DrDaveD:
I'm sorry I forgot to respond to this earlier.  

I removed the `snapshot_2_log` test because it was testing for incorrect behavior.  When the `CVMFS_GEO_DB_FILE` is `NONE` it is not supposed to link in any database.  The changes to `cvmfs_server_util.sh` made it do nothing as it should instead of the bad behavior.

vvolkl (approved):

PR #4029, opened 2025-11-05
[client] Skip open files on cache cleanup when possible

christge-sft:
On the Algorithmic side open files are handled as if they were pinned.

The tricky part is how the PosixQuotaManager (PQM) will get to know which files are open. In this approach, open hashes are communicated to the PQM using a dedicated magic xattr.

Firstly the PQM needs to know its managed mount points. This is happening early, when the PQM is acquired.
To get the open hashes of a mount point, this approach introduces the magic xattr `user.list_open_hashes`.
To get the mount points managed from the shared PQM,  this approach introduces the magic xattr `user.list_managed_mountopints`.
To get the open hashes of all the mount points associated to the shared PQM, this approach introduces the magic xattr `user.list_groups_open_hashes`. This last magic xattr uses exactly the same mechanism that is employed in `DoCleanup()`.

By default on this stage the feature is turned off. To enable it add `CVMFS_CACHE_CLEANUP_NONOPENLRU=yes` in the configuration.

### DIFF
--- a/cvmfs/cache_posix.cc
+++ b/cvmfs/cache_posix.cc
@@ -269,9 +269,9 @@ PosixCacheManager *PosixCacheManager::Create(
     const string &cache_path,
     const bool alien_cache,
     const RenameWorkarounds rename_workaround,
-    const bool do_refcount) {
+    const bool do_refcount, const bool cleanup_unused_first) {
   UniquePtr<PosixCacheManager> cache_manager(
-      new PosixCacheManager(cache_path, alien_cache, do_refcount));
+      new PosixCacheManager(cache_path, alien_cache, do_refcount,cleanup_unused_first));
   assert(cache_manager.IsValid());
 
   cache_manager->rename_workaround_ = rename_workaround;
--- a/cvmfs/cache_posix.cc
+++ b/cvmfs/cache_posix.cc
@@ -266,12 +266,11 @@ bool PosixCacheManager::InitCacheDirectory(const string &cache_path) {
 }
 
 PosixCacheManager *PosixCacheManager::Create(
-    const string &cache_path,
-    const bool alien_cache,
-    const RenameWorkarounds rename_workaround,
-    const bool do_refcount, const bool cleanup_unused_first) {
-  UniquePtr<PosixCacheManager> cache_manager(
-      new PosixCacheManager(cache_path, alien_cache, do_refcount,cleanup_unused_first));
+    const string &cache_path, const bool alien_cache,
+    const RenameWorkarounds rename_workaround, const bool do_refcount,
+    const bool cleanup_unused_first) {
+  UniquePtr<PosixCacheManager> cache_manager(new PosixCacheManager(
+      cache_path, alien_cache, do_refcount, cleanup_unused_first));
   assert(cache_manager.IsValid());
 
   cache_manager->rename_workaround_ = rename_workaround;
@@ -648,3 +647,4 @@ int64_t PosixCacheManager::Write(const void *buf, uint64_t size, void *txn) {
   transaction->size += written;
   return written;
 }
+
--- a/cvmfs/magic_xattr.cc
+++ b/cvmfs/magic_xattr.cc
@@ -835,7 +835,7 @@ void ListOpenHashesMagicXattr::FinalizeValue() {
         auto empty = hash_map.empty_key();
         auto *keys = hash_map.keys();
         for (size_t i = 0; i < hash_map.capacity(); ++i) {
-          shash::Any &hash = keys[i];
+          const shash::Any &hash = keys[i];
           if (hash != empty) {
             result += hash.ToString() + "\n";
           }
--- a/cvmfs/magic_xattr.cc
+++ b/cvmfs/magic_xattr.cc
@@ -8,7 +8,7 @@
 #include <string>
 #include <vector>
 
-#include "cache_posix.h"  // PosixCacheManager
+#include "cache_posix.h"
 #include "catalog_mgr_client.h"
 #include "crypto/signature.h"
 #include "fetch.h"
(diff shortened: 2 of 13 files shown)


cvmfs/cache_posix.cc
```diff
@@ -269,9 +269,9 @@ PosixCacheManager *PosixCacheManager::Create(
     const string &cache_path,
     const bool alien_cache,
     const RenameWorkarounds rename_workaround,
-    const bool do_refcount) {
+    const bool do_refcount, const bool cleanup_unused_first) {
   UniquePtr<PosixCacheManager> cache_manager(
-      new PosixCacheManager(cache_path, alien_cache, do_refcount));
+      new PosixCacheManager(cache_path, alien_cache, do_refcount,cleanup_unused_first));
```
vvolkl:
```suggestion
      new PosixCacheManager(cache_path, alien_cache, do_refcount, cleanup_unused_first));
```
Whitespace. Can you rerun clang-fmt on changed files?

cvmfs/magic_xattr.cc
```diff
@@ -8,11 +8,13 @@
 #include <string>
 #include <vector>
 
+#include "cache_posix.h"  // PosixCacheManager
```
vvolkl: This type of comment is likely to go stale quickly, when another PR uses something from the header but does not update the comment, let's remove these.

cvmfs/mountpoint.cc
```diff
@@ -306,6 +306,12 @@ FileSystem::PosixCacheSettings FileSystem::DeterminePosixCacheSettings(
     settings.do_refcount = false;
   }
 
+  if (options_mgr_->GetValue(
+          MkCacheParm("CVMFS_CACHE_CLEANUP_NONOPENLRU", instance), &optarg)
+      && options_mgr_->IsOn(optarg)) {
+    settings.cleanup_unused_first = settings.do_refcount;
```
vvolkl, line 314: Add a comment to explain that nonopenlru is not possible without the refcounted cache manager

cvmfs/mountpoint.cc
```diff
@@ -682,7 +688,7 @@ CacheManager *FileSystem::SetupPosixCacheMgr(const string &instance) {
       settings.is_alien,
       settings.avoid_rename ? PosixCacheManager::kRenameLink
                             : PosixCacheManager::kRenameNormal,
-      settings.do_refcount));
+      settings.do_refcount,settings.cleanup_unused_first));
```
vvolkl:
```suggestion
      settings.do_refcount, settings.cleanup_unused_first));
```
whitespace

cvmfs/quota_posix.cc
```diff
@@ -27,6 +27,9 @@
 #include <stdint.h>
 #include <sys/dir.h>
 #include <sys/stat.h>
+#include <sys/xattr.h>
+
+#include <algorithm>  //std::all_of
```
vvolkl: Remove comment (see above)

cvmfs/quota_posix.cc
```diff
@@ -35,6 +38,7 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#include <algorithm>
```
vvolkl, line 40: This include is duplicated, right?

cvmfs/quota_posix.cc
```diff
@@ -2106,3 +2256,91 @@ void PosixQuotaManager::ManagedReadHalfPipe(int fd, void *buf, size_t nbyte) {
...
+      and a_after_b(reference, current)) {
+    // as long as there are still threads that haven't joined yet
+    // and for 10 seconds
+    if (not joined[i]) {
+      const int s = pthread_tryjoin_np(*threads[i], NULL);
+      if (s == 0) {
+        joined[i] = true;
+      }
+    }
+    i = (++i) % handlers.size();
```
vvolkl: There's a compiler warning here:
```
warning: operation on ‘i’ may be undefined [-Wsequence-point]
```
Apparently the order of ++ and % is ambiguous here.

cvmfs/magic_xattr.cc
```diff
@@ -72,6 +74,13 @@ MagicXattrManager::MagicXattrManager(
 
   Register("user.authz", new AuthzMagicXattr());
   Register("user.external_url", new ExternalURLMagicXattr());
+
+  Register("user.cleanup_unused_first", new CleanupUnusedFirstMagicXattr());
+  Register("user.list_open_hashes", new ListOpenHashesMagicXattr());
+  Register("user.list_groups_open_hashes",
```
vvolkl: `"group" `is a slightly confusing name in this context. I would replace it with `"all_repositories"`

vvolkl: This is useful for debugging, but it doesn't need to be an xattr. I would suggest to make it a cvmfs_talk command instead, something like:

```
cvmfs_talk -i <repo> cache openhashes
```

cvmfs/magic_xattr.cc
```diff
@@ -72,6 +74,13 @@ MagicXattrManager::MagicXattrManager(
...
+           new ListGroupsOpenHashesMagicXattr());
+  Register("user.list_managed_mountpoints",
```
vvolkl: Same here, having this as an xattr for a specific repository is a bit unexpected, I would also make this a cvmfs_talk command

vvolkl (changes requested):
Very nice! Some comments inline.

As discussed, a future optimization could be to only exchange partial hashes, since a collision has negligible downsides. Can be done in a future PR together with the std::set optimization.

cvmfs/quota_posix.h
```diff
@@ -372,6 +385,29 @@ class PosixQuotaManager : public QuotaManager {
...
+   */
+  // TODO(gchr): it would be faster if it was a std::set. Needs a comparison
+  // operator for shash::Any
+  pthread_mutex_t *lock_open_files_;
+  std::vector<shash::Any> open_files_;
+
+  bool cleanup_unused_first_;
+  std::vector<std::string> mountpoints_;
+
+  std::vector<shash::Any> CollectGroupsHashes();
```
vvolkl: I still don't like this name. What about 
```suggestion
  std::vector<shash::Any> CollectAllOpenHashes();
```
?

vvolkl (changes requested):
See comment inline

vvolkl (approved):
LGTM!

PR #4033, opened 2025-11-12
[rpm] fix rpm for fedora 43

### LINKED ISSUE #4027: Fedora 43 :: install not working
Hi! I have a few problems trying to install cvmfs on fedora 43:
1. doing `dnf install -y https://cvmrepo.s3.cern.ch/cvmrepo/yum/cvmfs-release-latest.noarch.rpm` will install a 6-3 package 
which at the following update will not work as `$releasever == 43` is not known .. 

2. replacing $releasever with 42 and doing `dnf update` will work BUT will install cvmfs-release 7.1
that have the content based on EL:
`baseurl=http://cvmrepo.s3.cern.ch/cvmrepo/yum/cvmfs/EL/$releasever/$basearch/ http://cvmrepo.web.cern.ch/cvmrepo/yum/cvmfs/EL/$releasever/$basearch/
`

3. ok, puting back the fedora/42 base urls i get 
```
root@hal: yum.repos.d # dnf install cvmfs
Updating and loading repositories:
Repositories loaded.
Package                                            Arch         Version                                            Repository                       Size
Installing:
 cvmfs                                             x86_64       2.13.3-1.fc42                                      cernvm                        5.2 MiB
Installing dependencies:
 cvmfs-config-default                              noarch       2.2-1                                              cernvm                        7.5 KiB
 cvmfs-fuse3                                       x86_64       2.13.3-1.fc42                                      cernvm                       44.6 MiB
 cvmfs-libs                                        x86_64       2.13.3-1.fc42                                      cernvm                       35.6 MiB

Transaction Summary:
 Installing:         4 packages

Total size of inbound packages is 24 MiB. Need to download 24 MiB.
After this operation, 85 MiB extra will be used (install 85 MiB, remove 0 B).
Is this ok [y/N]: y
[1/4] cvmfs-0:2.13.3-1.fc42.x86_64                                                                              100% |   2.2 MiB/s |   1.3 MiB |  00m01s
[2/4] cvmfs-config-default-0:2.2-1.noarch                                                                       100% |  65.5 KiB/s |   9.0 KiB |  00m00s
[3/4] cvmfs-libs-0:2.13.3-1.fc42.x86_64                                                                         100% |   8.9 MiB/s |   9.9 MiB |  00m01s
[4/4] cvmfs-fuse3-0:2.13.3-1.fc42.x86_64                                                                        100% |   9.8 MiB/s |  13.3 MiB |  00m01s
--------------------------------------------------------------------------------------------------------------------------------------------------------
[4/4] Total                                                                                                     100% |  17.9 MiB/s |  24.4 MiB |  00m01s
Running transaction
Transaction failed: Rpm transaction failed.
  - package cvmfs-config-default-2.2-1.noarch does not verify: no digest
```

vvolkl:
Fixes a "Nothing provides group(cvmfs)" error when installing cvmfs packages on Fedora 43. There may be a better way for user creation in rpms (https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/) but this works.

Should go into the cvmfs-2.13 branch as well.

Fixes https://github.com/cvmfs/cvmfs/issues/4027

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -76,7 +76,7 @@ Summary: CernVM File System
 Name: cvmfs
 Version: 2.13.3
 %global base_version %(echo %{version} | cut -d'~' -f1)
-Release: 1%{?dist}
+Release: 2%{?dist}
 URL: https://cernvm.cern.ch/fs/
 Source0: https://ecsft.cern.ch/dist/cvmfs/%{name}-%{version}/%{name}-%{version}.tar.gz
 %if 0%{?selinux_cvmfs}
@@ -176,6 +176,8 @@ Requires: util-linux
 Requires: cvmfs-config
 Requires: cvmfs-libs = %{version}-%{release}
 Requires: cvmfs-fuse3 = %{version}-%{release}
+Provides: group(cvmfs)
+Provides: user(cvmfs)
 
 # SELinux integration
 # These are needed to build the selinux policy module.


DrDaveD (changes requested):
The right way to do this is with a sysusers.d file.  Then on redhat/fedora there are just a few macros needed to manage it.  See for example all the occurrences of `sysusers` in [openbao.spec](https://github.com/openbao/openbao-fedora/blob/master/openbao.spec).  Since cvmfs-universal.spec is also used on SUSE it would need some #if around the use of those macros.  SUSE might support sysusers.d, I'm not sure, but I don't see the macros when I start an opensuse/leap container.

vvolkl:
Yes, the Fedora packaging guide I linked says the same thing. But I've already built 2.13.3-2.fc43, I'll use this workaround for 2.13 and add a sysuser.d file in 2.14.

vvolkl:
@cernvm-bot unittest

PR #4039, opened 2025-11-26
[CLIENT] Prefetch file bundles

christge-sft:
This PR providing the mechanism via which a `Fetch()` will be invoked for the dependencies of a "trigger file".

A BundleMgr is responsible to spawn a pool of threads (`bm_fetcher`s) based on the size of the file bundle, and hand them  over their portion of hashes to invoke a `Fetch()` on.

The trigger file is in json format. `BundleFileMgr` is responsible for parsing the file and feeding the `BundleMgr()` with the file dependencies in the form of `CacheManager::LabeledObject`s. 

At this moment,  BundleFileMgr is mocked only to demonstrate how the BundleMgr is handling the load.

### DIFF
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -106,6 +106,7 @@ void BundleMgr::SpawnFetchers() {
     pthread_t thread;
     const int res = pthread_create(&thread, nullptr, EstablishConnection, this);
     if (res != 0) {
+      LogCvmfs(kLogBungleMgr, kLogDebug, "Thread creation failed!");
       continue;
     }
     int fd;
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -104,7 +104,7 @@ void BundleMgr::SpawnFetchers() {
   const size_t size = 1 + (bfm_->Size() / 30);  // Spawn at least one fetcher
   for (size_t i = 0; i < size; ++i) {
     pthread_t thread;
-    const int res = pthread_create(&thread, nullptr, EstablishConnection, this);
+    const int res = pthread_create(&thread, nullptr, MainBundleMgrFetcher, this);
     if (res != 0) {
       LogCvmfs(kLogBundleMgr, kLogDebug, "Thread creation failed!");
       continue;
@@ -123,7 +123,7 @@ void BundleMgr::SpawnFetchers() {
   }
 }
 
-void *BundleMgr::EstablishConnection(void *data) {
+void *BundleMgr::MainBundleMgrFetcher(void *data) {
   pthread_setname_np(pthread_self(), "bm_fetcher");
   BundleMgr *mgr = static_cast<BundleMgr *>(data);
   const int& wfd = mgr->pipe_bm_[1];
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -1,0 +1,7 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include "bundle_mgr.h"
+
+void BundleMgr::Fetch(){}
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -99,11 +99,13 @@ void BundleMgr::SpawnFetchers() {
 void *BundleMgr::EstablishConnection(void *data) {
   pthread_setname_np(pthread_self(), "bm_fetcher");
   BundleMgr *mgr = static_cast<BundleMgr *>(data);
-  int rfd = mgr->pipe_bm_[0];
+  int wfd = mgr->pipe_bm_[1];
   int back_channel[2];
   MakePipe(back_channel);
 
-  WritePipe(rfd, back_channel + 1, sizeof(int));
+  WritePipe(wfd, &back_channel[1], sizeof(int));
+
+  int& rfd=back_channel[0];
 
   Command cmd;
   while (read(rfd, &cmd, sizeof(Command)) == sizeof(Command)) {
@@ -128,6 +130,7 @@ void *BundleMgr::EstablishConnection(void *data) {
 }
 
 CacheManager::LabeledObject BundleMgr::ReceiveLabeledObject(int fd) const {
+  (void)fd;
   return CacheManager::LabeledObject(shash::Any{}, CacheManager::Label());
 }
 
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -10,8 +10,8 @@
 
 #include "fetch.h"
 #include "util/inode.h"
-#include "util/posix.h"
 #include "util/pointer.h"
+#include "util/posix.h"
 
 BundleMgr::BundleMgr(MountPoint *mp, fuse_ino_t ino) : mount_point_(mp) {
   is_valid_ = cvmfs::GetPathForInode(mp, mp->file_system(), ino, &path_);
@@ -103,7 +103,7 @@ void BundleMgr::JoinFetchers() {
 void BundleMgr::SpawnFetchers() {
   MakePipe(pipe_bm_);
 
-  size_t size = 1+( bfm_->Size() / 30 ); // Spawn at least one fetcher
+  size_t size = 1 + (bfm_->Size() / 30);  // Spawn at least one fetcher
   for (size_t i = 0; i < size; ++i) {
     pthread_t thread;
     int res = pthread_create(&thread, nullptr, EstablishConnection, this);
@@ -116,8 +116,8 @@ void BundleMgr::SpawnFetchers() {
     // Make the write operation to the return pipe non blocking
     // According to the man (7) page of write, when attempting to write
     // n<=PIPE_BUF data on a non blocking pipe, it will either write all of them
-    // or errno will be set to EAGAIN. PIPE_BUF is at least 512bytes on and linux
-    // 4096bytes.
+    // or errno will be set to EAGAIN. PIPE_BUF is at least 512bytes on and
+    // linux 4096bytes.
     int flags = fcntl(fd, F_GETFL);
     fcntl(fd, F_SETFL, flags | O_NONBLOCK);
     fetcher_pool_.push_back({thread, fd});
@@ -162,13 +162,28 @@ CacheManager::LabeledObject BundleMgr::ReceiveLabeledObject(int fd) const {
   return CacheManager::LabeledObject(shash::Any{}, CacheManager::Label());
 }
 
-void BundleMgr::SendLabeledObject(
-    int fd, UniquePtr<CacheManager::LabeledObject>& obj) const {
-  (void)obj;
+bool BundleMgr::SendLabeledObject(
+    int fd, UniquePtr<CacheManager::LabeledObject> &obj) const {
+  return true;
 }
 
-bool BundleMgr::TrySendData(
-    int fd, UniquePtr<CacheManager::LabeledObject>& obj) const {
+bool BundleMgr::TrySendData(int fd,
+                            UniquePtr<CacheManager::LabeledObject> &obj) const {
+  Command cmd = Command::kFetch;
+  size_t n = write(fd, &cmd, sizeof(Command));
+  if (n != sizeof(Command)) {
+    if (not(errno == EAGAIN || errno == EWOULDBLOCK)) {
+      LogCvmfs(kLogBungleMgr,
+               kLogDebug,
+               "write() on back channel failed unexpectedly");
+    }
+    return false;
+  } else {
+    while (SendLabeledObject(fd, obj) != true) {
+      // If a Fetcher receives a kFetch command should receive the Labeled Object also.
+    }
+  }
+
   return true;
 }
 
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -139,8 +139,12 @@ void *BundleMgr::EstablishConnection(void *data) {
     bool terminate = false;
     switch (cmd) {
       case Command::kFetch: {
-        auto object = mgr->ReceiveLabeledObject(rfd);
-        mgr->fetcher_->Fetch(object);
+        auto obj = mgr->ReceiveLabeledObject(rfd);
+        if (not obj.IsValid()) {
+          LogCvmfs(kLogBungleMgr, kLogDebug, "Received a null object");
+          break;
+        }
+        mgr->fetcher_->Fetch(*obj);
       } break;
       case Command::kTerminate:
       default:
@@ -156,13 +160,29 @@ void *BundleMgr::EstablishConnection(void *data) {
   pthread_exit(nullptr);
 }
 
-CacheManager::LabeledObject BundleMgr::ReceiveLabeledObject(int fd) const {
-  (void)fd;
-  return CacheManager::LabeledObject(shash::Any{}, CacheManager::Label());
+UniquePtr<CacheManager::LabeledObject> BundleMgr::ReceiveLabeledObject(
+    int fd) const {
+  shash::Any id = BlockingReceive<shash::Any>(fd);
+  CacheManager::Label label;
+  label.flags=BlockingReceive<int>(fd);
+  label.size=BlockingReceive<uint64_t>(fd);
+  label.zip_algorithm=BlockingReceive<zlib::Algorithms>(fd);
+  label.range_offset=BlockingReceive<off_t>(fd);
+  label.path=BlockingReceive(fd);
+  CacheManager::LabeledObject* obj = new CacheManager::LabeledObject(id,label);
+  assert(obj!=nullptr);
+  return UniquePtr<CacheManager::LabeledObject>(obj);
 }
 
 bool BundleMgr::SendLabeledObject(
-    int fd, UniquePtr<CacheManager::LabeledObject> &obj) const {
+    int fd, const UniquePtr<CacheManager::LabeledObject> &obj) const {
+  BlockingSend(fd, obj->id);
+  BlockingSend(fd, obj->label.flags);
+  BlockingSend(fd, obj->label.size);
+  BlockingSend(fd, obj->label.zip_algorithm);
+  BlockingSend(fd, obj->label.range_offset);
+  const std::string &path=obj->label.path;
+  BlockingSend(fd, path);
   return true;
 }
 
@@ -179,7 +199,8 @@ bool BundleMgr::TrySendData(int fd,
     return false;
   } else {
     while (SendLabeledObject(fd, obj) != true) {
-      // If a Fetcher receives a kFetch command should receive the Labeled Object also.
+      // If a Fetcher receives a kFetch command should receive the Labeled
+      // Object also.
     }
   }
 
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -36,7 +36,7 @@ void BundleMgr::Fetch() {
 
   auto it = fetcher_pool_.begin();
   if (it == fetcher_pool_.end()) {
-    LogCvmfs(kLogBungleMgr,
+    LogCvmfs(kLogBundleMgr,
              kLogDebug,
              "The pool of fetchers is empty. Can't fetch dependencies. "
              "Aborting Op.");
@@ -75,7 +75,7 @@ void BundleMgr::JoinFetchers() {
 
     struct timespec ts;
     if (clock_gettime(CLOCK_REALTIME, &ts) == -1) {
-      LogCvmfs(kLogBungleMgr,
+      LogCvmfs(kLogBundleMgr,
                kLogDebug,
                "Failed to read CLOCK_REALTIME. Detaching fetchers.");
       detach_mode = true;
@@ -88,7 +88,7 @@ void BundleMgr::JoinFetchers() {
       Command cmd = Command::kTerminate;
       WritePipe(fd, &cmd, sizeof(Command));
       if (pthread_timedjoin_np(thread, nullptr, &ts) != 0) {
-        LogCvmfs(kLogBungleMgr,
+        LogCvmfs(kLogBundleMgr,
                  kLogDebug,
                  "Fetcher is busy for too long. Detaching.");
         pthread_detach(thread);
@@ -106,7 +106,7 @@ void BundleMgr::SpawnFetchers() {
     pthread_t thread;
     const int res = pthread_create(&thread, nullptr, EstablishConnection, this);
     if (res != 0) {
-      LogCvmfs(kLogBungleMgr, kLogDebug, "Thread creation failed!");
+      LogCvmfs(kLogBundleMgr, kLogDebug, "Thread creation failed!");
       continue;
     }
     int fd;
@@ -141,7 +141,7 @@ void *BundleMgr::EstablishConnection(void *data) {
       case Command::kFetch: {
         auto obj = mgr->ReceiveLabeledObject(rfd);
         if (not obj.IsValid()) {
-          LogCvmfs(kLogBungleMgr, kLogDebug, "Received a null object");
+          LogCvmfs(kLogBundleMgr, kLogDebug, "Received a null object");
           break;
         }
         mgr->fetcher_->Fetch(*obj);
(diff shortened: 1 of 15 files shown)


cvmfs/file_bundle.h
```diff
@@ -0,0 +1,26 @@
...
+ * This class implements the format for .cvmfsbundle files 
+ */
+
+#ifndef CVMFS_FILE_BUNDLE_H_
+#define CVMFS_FILE_BUNDLE_H_
+
+
+/*
+
+The .cvmfsbundle file servers both as a file list and as a trigger for loading a bundle. The convention is to call it .cvmfsbundle.<filename>, where <filename> should trigger the bundle.
```
jblomer: servers --> serves

cvmfs/bundle_mgr.h
```diff
@@ -0,0 +1,34 @@
...
+#define CVMFS_BUNDLE_MGR_H_
+#include "duplex_fuse.h"  // fuse_ino_t
+#include "mountpoint.h"   //MountPoint*, FileSystem*
+
+namespace cvmfs {
+/*
+ * functions defined inside /cvmfs/cvmfs.cc
+ */
+bool GetPathForInode(const fuse_ino_t ino, PathString *path);
+bool GetDirentForInode(const fuse_ino_t ino, catalog::DirectoryEntry *dirent);
```
jblomer: I think we should avoid exposing these functions (which are quite closely coupled to the fuse logic).

cvmfs/bundle_mgr.h
```diff
@@ -0,0 +1,34 @@
...
+}  // namespace cvmfs
+
+class BundleMgr {
```
jblomer: Perhaps better to make the class long-lived, managed by the mountpoint. Then `Fetch()` would take the inode related params and return a bool (or error code).

cvmfs/cvmfs.cc
```diff
@@ -1176,6 +1177,17 @@ static void cvmfs_open(fuse_req_t req, fuse_ino_t ino,
...
+    BundleMgr bundle_mgr(mount_point_, file_system_, ino);
+    if (bundle_mgr) {
+      bundle_mgr.Fetch();
+    } else {
+      LogCvmfs(kLogCvmfs, kLogDebug,
+               "Couldn't fetch bundle associated to file %s", path.c_str());
+    }
+  }
+
   perf::Inc(file_system_->n_fs_open());  // Count actual open / fetch operations
```
jblomer: I think it's better after the `perf::Inc()` call

cvmfs/bundle_mgr.h
```diff
@@ -20,6 +23,13 @@ class BundleMgr {
   BundleMgr(MountPoint *mp, FileSystem *fs, fuse_ino_t ino) : is_valid_(true) {
     is_valid_ = cvmfs::GetPathForInode(ino, &path_);
     is_valid_ &= cvmfs::GetDirentForInode(ino, &dirent_);
+    fname_ = GetFileName(path_);
+    parent_path_ = GetParentPath(path_);
+
+    // There is a naming convention regarding the name of the file with the
+    // contents of the bundle
+    bundle_file_path_ = PathString(parent_path_.ToString() + "/.cvmfsbundle."
```
jblomer: Make the path assembly a utility function (we may need it elsewhere)

cvmfs/bundle_mgr.cc
```diff
@@ -4,4 +4,36 @@
...
+#include <pthread.h>
+
+#include <vector>
+
+void BundleMgr::Fetch() {
+  std::vector<pthread_t> fetcher_pool;
+
+  for (PathString next_f = bfm_->GetNext(); not next_f.IsEmpty();
+       next_f = bfm_->GetNext()) {
+    pthread_t fetcher_t;
```
jblomer: Looks like `fetcher_t` is a type, perhaps rather call it `fetcher_thread`

cvmfs/bundle_mgr.cc
```diff
@@ -4,4 +4,36 @@
 
 #include "bundle_mgr.h"
 
-void BundleMgr::Fetch(){}
+#include <pthread.h>
+
+#include <vector>
+
+void BundleMgr::Fetch() {
+  std::vector<pthread_t> fetcher_pool;
```
jblomer: I'm surprised that this works. I don't think it's portable, `pthread_t` is an opaque type and not necessarily copyable or movable. Perhaps use `std::vector<pthread_t *>` instead.

cvmfs/bundle_mgr.cc
```diff
@@ -4,4 +4,36 @@
...
+
+  for (PathString next_f = bfm_->GetNext(); not next_f.IsEmpty();
+       next_f = bfm_->GetNext()) {
+    pthread_t fetcher_t;
+
+    // TODO(christge): DoFetch requires real arguments
+    pthread_create(&fetcher_t, nullptr, DoFetch, nullptr);
```
jblomer: Check the return value

cvmfs/bundle_mgr.cc
```diff
@@ -4,4 +4,36 @@
...
+      int is_joinable = pthread_tryjoin_np(thread, &retval);
+      if (is_joinable == 0) {
+        // fetch has finished
+        fetcher_pool.erase(it);
+      } else {
+        // still busy
+        ++it;
+      }
+    }
+  }
```
jblomer: Since we anyway wait for all of them to finish, I think that can be easier: you can just `pthread_join()` them one by one and remove the vector altogether at the end.

cvmfs/bundle_mgr.h
```diff
@@ -40,6 +40,8 @@ class BundleMgr {
   explicit operator bool() const { return is_valid_; }
 
  private:
+  static void *DoFetch(void *data);
```
jblomer: We have a convention of calling a thread's main function `Main...`, e.g. `MainFetch()` in this case.

cvmfs/bundle_mgr.h
```diff
@@ -22,7 +22,7 @@ bool GetDirentForInode(const fuse_ino_t ino, catalog::DirectoryEntry *dirent);
 
 class BundleMgr {
  public:
-  BundleMgr(MountPoint *mp, FileSystem *fs, fuse_ino_t ino) : is_valid_(true) {
+  BundleMgr(MountPoint *mp, fuse_ino_t ino) : is_valid_(true) {
```
jblomer: Consider rebasing and folding this into the original commit.

cvmfs/util/inode.h
```diff
@@ -0,0 +1,16 @@
...
+#ifndef CVMFS_UTIL_INODE_H_
+#define CVMFS_UTIL_INODE_H_
+#include "directory_entry.h"  // catalog::DirectoryEntry
+#include "duplex_fuse.h"
+#include "mountpoint.h"
+namespace cvmfs {
+bool GetDirentForInode(MountPoint *mountpoint, FileSystem *filesystem,
+                       const fuse_ino_t ino, catalog::DirectoryEntry *dirent);
+bool GetPathForInode(MountPoint *mountpoint, FileSystem *filesystem,
+                     const fuse_ino_t ino, PathString *path);
```
jblomer, line 13: As discussed, I believe these two functions should not be exposed and are not actually utility functions (they are not free-standing).

test/unittests/CMakeLists.txt
```diff
@@ -748,6 +748,7 @@ set(CVMFS_UNITTEST_MOCKFUSE_SOURCES
   ${CVMFS_SOURCE_DIR}/wpad.cc
   ${CVMFS_SOURCE_DIR}/xattr.cc
   ${CVMFS_SOURCE_DIR}/bundle_mgr.cc
+  ${CVMFS_SOURCE_DIR}/util/inode.cc
```
jblomer, line 755: Add alphabetically

test/unittests/CMakeLists.txt
```diff
@@ -747,6 +747,7 @@ set(CVMFS_UNITTEST_MOCKFUSE_SOURCES
   ${CVMFS_SOURCE_DIR}/whitelist.cc
   ${CVMFS_SOURCE_DIR}/wpad.cc
   ${CVMFS_SOURCE_DIR}/xattr.cc
+  ${CVMFS_SOURCE_DIR}/bundle_mgr.cc
```
jblomer, line 754: Add alphabetically

cvmfs/bundle_mgr.cc
```diff
@@ -8,6 +8,20 @@
 
 #include <vector>
 
+BundleMgr::BundleMgr(MountPoint *mp, fuse_ino_t ino) : is_valid_(true) {
```
jblomer: Rebase and fold into corresponding previous commits

test/unittests/CMakeLists.txt
```diff
@@ -447,6 +447,7 @@ set(CVMFS_UNITTEST_SOURCES
   t_whitelist.cc
   t_wpad.cc
   t_xattr.cc
+  t_bundle_mgr.cc
```
jblomer: Alphabetical order

test/unittests/CMakeLists.txt
```diff
@@ -503,6 +504,8 @@ set(CVMFS_UNITTEST_SOURCES
   ${CVMFS_SOURCE_DIR}/libcvmfs_legacy.cc
   ${CVMFS_SOURCE_DIR}/libcvmfs_options.cc
   ${CVMFS_SOURCE_DIR}/magic_xattr.cc
+  ${CVMFS_SOURCE_DIR}/bundle_mgr.cc
+  ${CVMFS_SOURCE_DIR}/util/inode.cc
```
jblomer: Alphabetical order

cvmfs/bundle_mgr.cc
```diff
@@ -51,3 +51,4 @@ void BundleMgr::Fetch() {
   }
 }
 
+void* BundleMgr::DoFetch(void* data){}
```
jblomer: Move into corresponding previous commit

test/unittests/t_bundle_mgr.cc
```diff
@@ -0,0 +1,67 @@
+/**
+ * This file is part of the CernVM File System.
+ */
+
+#include <gtest/gtest.h>
+#include "mountpoint.h"
+#include "util/uuid.h"
+#include "testutil.h"
+#include "options.h"
+#include "bundle_mgr.h"
```
jblomer: Include alphabetically

test/unittests/t_bundle_mgr.cc
```diff
@@ -0,0 +1,67 @@
...
+
+using namespace std;  // NOLINT
+
+class T_BundleMgr: public ::testing::Test {
+ protected:
+  virtual void SetUp() {
+    repo_path_ = "repo";
+    uuid_dummy_ = cvmfs::Uuid::Create("");
+    used_fds_ = GetNoUsedFds();
+    fd_cwd_ = open(".", O_RDONLY);
```
jblomer: Not quite sure what this is used for

cvmfs/file_bundle.h
```diff
@@ -31,7 +32,21 @@ class BundleFileMgr {
   // TODO(christge): this is to be reverted. It's the basic interface needed to
   // interact with the file bundle. Now there are some mocks for prototyping
   BundleFileMgr(const PathString &bf) { }
-  PathString GetNext() { return PathString{}; };
+  CacheManager::LabeledObject *GetNext() {
```
jblomer: This should not compile anymore as since `GetNext()` is already used. Ideally, each commit will compile and successfully run the tests.

cvmfs/util/logging_internal.h
```diff
@@ -108,7 +108,8 @@ enum LogSource {
   kLogReflog,
   kLogKvStore,
   kLogTelemetry,
-  kLogCurl
+  kLogCurl,
+  kLogBungleMgr
```
jblomer: This must come with a corresponding change to the module name, as indicated in the commit.

cvmfs/bundle_mgr.cc
```diff
@@ -8,47 +8,131 @@
...
+  bool detach_mode = false;
+  // Join fetcher threads.
+  // Give a fetcher thread 10 seconds to join.
+  for (auto it = fetcher_pool_.begin(); it != fetcher_pool_.end(); ++it) {
+    auto &tuple = *it;
+    auto thread = std::get<0>(tuple);
+    auto fd = std::get<1>(tuple);
+
+    struct timespec ts;
+    if (clock_gettime(CLOCK_REALTIME, &ts) == -1) {
```
jblomer, line 77: Use our `platform_monotonic_time()` helpers

cvmfs/bundle_mgr.cc
```diff
@@ -8,47 +8,131 @@
...
+      LogCvmfs(kLogBungleMgr,
+               kLogDebug,
+               "Failed to read CLOCK_REALTIME. Detaching fetchers.");
+      detach_mode = true;
```
jblomer, line 81: This is hard to test. I think we should just expect to get a timestamp and otherwise crash.

cvmfs/bundle_mgr.cc
```diff
@@ -8,47 +8,131 @@
...
+    } else {
+      ts.tv_sec += 10;
+      Command cmd = Command::kTerminate;
+      WritePipe(fd, &cmd, sizeof(Command));
+      int res = pthread_timedjoin_np(thread, nullptr, &ts);
+      if (res != 0) {
+        LogCvmfs(kLogBungleMgr,
+                 kLogDebug,
+                 "Fetcher is busy for too long. Detaching.");
+        pthread_detach(thread);
```
jblomer, line 94: I'm not sure that we should go this way. We have no way of controlling detached threads piling up. Let's add a TODO, I think we need to discuss the thread management.

cvmfs/bundle_mgr.cc
```diff
@@ -8,47 +8,131 @@
...
       }
     }
   }
+  ClosePipe(pipe_bm_);
+}
+
+void BundleMgr::SpawnFetchers() {
+  MakePipe(pipe_bm_);
+
+  size_t size = bfm_->Size() / 30;
```
jblomer: Why divided by 30? Better to make it a named constant and to comment on the choice of the value.

cvmfs/bundle_mgr.cc
```diff
@@ -8,47 +8,131 @@
...
+  for (size_t i = 0; i < size; ++i) {
+    pthread_t thread;
+    int res = pthread_create(&thread, nullptr, EstablishConnection, this);
+    if (res != 0) {
+      continue;
```
jblomer, line 110: Rather: crash

cvmfs/bundle_mgr.cc
```diff
@@ -8,47 +8,131 @@
...
+    }
+    int fd;
+    ReadPipe(pipe_bm_[0], &fd, sizeof(int));
+    fetcher_pool_.push_back({thread, fd});
+  }
+}
+
+void *BundleMgr::EstablishConnection(void *data) {
+  pthread_setname_np(pthread_self(), "bm_fetcher");
```
jblomer, line 127: Should not be necessary

cvmfs/bundle_mgr.cc
```diff
@@ -8,47 +8,131 @@
...
+      break;
+    }
+  }
+
+  ClosePipe(back_channel);
+  pthread_exit(nullptr);
+}
+
+CacheManager::LabeledObject BundleMgr::ReceiveLabeledObject(int fd) const {
+  return CacheManager::LabeledObject(shash::Any{}, CacheManager::Label());
```
jblomer: Add a TODO to make clear that this method is still a stub

cvmfs/bundle_mgr.h
```diff
@@ -4,17 +4,24 @@
...
 
 #include "file_bundle.h"
 #include "mountpoint.h"
 #include "shortstring.h"
 #include "util/pointer.h"
+#include "util/single_copy.h"
+
+class BundleMgr : SingleCopy {
+  friend class T_BundleMgr;
+  FRIEND_TEST(T_BundleMgr, Simple);
```
jblomer: Do we need this if the fixture already is a friend?

test/unittests/t_bundle_mgr.cc
```diff
@@ -2,16 +2,75 @@
  * This file is part of the CernVM File System.
  */
 
+#include <gmock/gmock.h>
```
jblomer, line 5: Thinking about it, I'm not sure that generous mocking is the best way to go here. In principle it's a viable approach but in the end there is quite a bit of mechanics to mock (it's not exactly lean) and it also results in virtual methods being sprinkled around. I would prefer setting up the actual manager objects on a small, on the fly created test repository. This has been done in other unit tests, too.

cvmfs/bundle_mgr.h
```diff
@@ -17,7 +17,6 @@
 
 class BundleMgr : SingleCopy {
   friend class T_BundleMgr;
-  FRIEND_TEST(T_BundleMgr, Simple);
```
jblomer: Fold into previous commit

cvmfs/bundle_mgr.cc
```diff
@@ -99,11 +99,13 @@ void BundleMgr::SpawnFetchers() {
...
   BundleMgr *mgr = static_cast<BundleMgr *>(data);
-  int rfd = mgr->pipe_bm_[1];
+  int wfd = mgr->pipe_bm_[1];
   int back_channel[2];
   MakePipe(back_channel);
 
-  WritePipe(rfd, &back_channel[1], sizeof(int));
+  WritePipe(wfd, &back_channel[1], sizeof(int));
+
+  int& rfd=back_channel[0];
```
jblomer: Why an integer reference?

cvmfs/bundle_mgr.cc
```diff
@@ -128,6 +130,7 @@ void *BundleMgr::EstablishConnection(void *data) {
 }
 
 CacheManager::LabeledObject BundleMgr::ReceiveLabeledObject(int fd) const {
+  (void)fd;
```
jblomer: Fold into previous commit

test/unittests/t_bundle_mgr.cc
```diff
@@ -216,5 +216,6 @@ class T_BundleMgr : public ::testing::Test {
 TEST_F(T_BundleMgr, Fetch) {
   bfm_->Reset(entries_);
   bundle_mgr_->Fetch();
+  EXPECT_EQ(bfm_->it,entries_.end());
```
jblomer: Rebase & fold into previous relevant commit

cvmfs/bundle_mgr.cc
```diff
@@ -35,11 +35,28 @@ void BundleMgr::Fetch() {
   SpawnFetchers();
 
   CacheManager::LabeledObject *obj;
+  auto it = fetcher_pool_.begin();
+  if (it == fetcher_pool_.end()) {
```
jblomer, line 38: Can this ever happen? If not, just assert

cvmfs/bundle_mgr.cc
```diff
@@ -35,11 +35,28 @@ void BundleMgr::Fetch() {
...
+    auto wfd = std::get<1>(pair);
+    // Find the first available Fetcher to send the data
+    while (not TrySendData(wfd, *obj)) {
+      if (( ++it ) == fetcher_pool_.end()) {
+        it = fetcher_pool_.begin();
+      }
+      pair = *it;
+      wfd = std::get<1>(pair);
+    }
+    if (( ++it ) == fetcher_pool_.end()) {
```
jblomer: Isn't this a busy loop? Maybe better to block if there are no more fetchers

cvmfs/bundle_mgr.cc
```diff
@@ -92,6 +109,14 @@ void BundleMgr::SpawnFetchers() {
...
     int fd;
     ReadPipe(pipe_bm_[0], &fd, sizeof(int));
+
+    // Make the write operation to the return pipe non blocking
+    // According to the man (7) page of write, when attempting to write
+    // n<=PIPE_BUF data on a non blocking pipe, it will either write all of them
+    // or errno will set to EAGAIN. PIPE_BUF is at least 512bytes on and linux
+    // 4096bytes.
+    int flags = fcntl(fd, F_GETFL);
+    fcntl(fd, F_SETFL, flags | O_NONBLOCK);
```
jblomer, line 121: We have utility functions for this in posix.cc

cvmfs/bundle_mgr.cc
```diff
@@ -100,7 +103,7 @@ void BundleMgr::JoinFetchers() {
 void BundleMgr::SpawnFetchers() {
   MakePipe(pipe_bm_);
 
-  size_t size = bfm_->Size() / 30;
+  size_t size = 1+( bfm_->Size() / 30 ); // Spawn at least one fetcher
```
jblomer: Fold into previous relevant commit

cvmfs/bundle_mgr.cc
```diff
@@ -162,13 +162,28 @@ CacheManager::LabeledObject BundleMgr::ReceiveLabeledObject(int fd) const {
...
+    int fd, UniquePtr<CacheManager::LabeledObject> &obj) const {
+  return true;
 }
 
-bool BundleMgr::TrySendData(
-    int fd, UniquePtr<CacheManager::LabeledObject>& obj) const {
+bool BundleMgr::TrySendData(int fd,
+                            UniquePtr<CacheManager::LabeledObject> &obj) const {
+  Command cmd = Command::kFetch;
+  size_t n = write(fd, &cmd, sizeof(Command));
```
jblomer: We have utility functions for this (SafeWrite etc.)

cvmfs/bundle_mgr.h
```diff
@@ -35,6 +38,53 @@ class BundleMgr : SingleCopy {
                          UniquePtr<CacheManager::LabeledObject> &obj) const;
   bool TrySendData(int fd, UniquePtr<CacheManager::LabeledObject> &obj) const;
 
+  // CT stands for contiguous type
+  template<typename CT>
+  void BlockingSend(int fd, const CT &obj, size_t size = sizeof(CT)) const {
```
jblomer, line 44: Nice! Could go into our utility functions.

cvmfs/bundle_mgr.cc
```diff
@@ -139,8 +139,12 @@ void *BundleMgr::EstablishConnection(void *data) {
     bool terminate = false;
     switch (cmd) {
       case Command::kFetch: {
-        auto object = mgr->ReceiveLabeledObject(rfd);
-        mgr->fetcher_->Fetch(object);
+        auto obj = mgr->ReceiveLabeledObject(rfd);
+        if (not obj.IsValid()) {
```
jblomer: Can this happen? If not, assert

cvmfs/bundle_mgr.h
```diff
@@ -4,12 +4,12 @@
 
 #ifndef CVMFS_BUNDLE_MGR_H_
 #define CVMFS_BUNDLE_MGR_H_
-#include <gtest/gtest_prod.h>
 
 #include <tuple>
 #include <type_traits>
 #include <vector>
 
+#include "duplex_testing.h"
```
jblomer, line 12: fold into previous relevant commit

jblomer (commented):
Nice work! I like that the commits are all small and well reviewable. I'd suggest to rebase the history, fold in fixups etc. so that the result can be nicely reviewed commit by commit and every commit compiles and runs the unit tests.

cvmfs/bundle_mgr.cc
christge-sft: I think that the conclusion from the in person conversation is:
_not sure we need a pool of threads here but if we do, they better detach so that the main thread can return and open the trigger file._

test/unittests/t_bundle_mgr.cc
christge-sft: TBH I thought it would be useful when I initialized the unit tests mimicking similar ones, but it's not useful indeed.

cvmfs/util/logging_internal.h
christge-sft: @jblomer  If you are referring to the typo, it's fixed. If not, I'm not sure I'm getting the point.

cvmfs/bundle_mgr.cc
christge-sft: I think what we concluded eventually is that it's better to detach those threads anyway.

cvmfs/bundle_mgr.cc
christge-sft: Is crashing really desired here? Just one thread is enough to do the work (sub-optimally ofc)

cvmfs/bundle_mgr.cc
christge-sft: Not necessary indeed, but I think it simplifies debugging.

christge-sft:
History is now in much better shape

christge-sft:
After https://github.com/cvmfs/cvmfs/pull/4149 this is now deprecated. Most of the feedback is incorporated there.

PR #4049, opened 2025-12-05
[CLIENT] Parse Bundle File

christge-sft:
This is a complement of https://github.com/cvmfs/cvmfs/pull/4039.

It is the mechanism that parses the trigger file and returns `CacheManager::LabeledObject - s` to the `BundleMgr()` in order to trigger fetching.

The trigger file is assumed to have the following format:
```json
{                                                                                      
"labeled_objects":[{"id":{"algorithm":%d,"suffix":"%c","digest":"%s"},"label":{"flags":%d,"size":%PRIu64,"zip_algorithm":%d,"range_offset":%d,"path":"%s"}}]                  
}                                                                                      
```

At the moment I have provided multiple different constructors for the `BundleFileMgr` because we haven't yet converged on how will the trigger file be propagated to the `BundleMgr` and the `BundleFileMgr`.

### DIFF
--- a/cvmfs/CMakeLists.txt
+++ b/cvmfs/CMakeLists.txt
@@ -244,6 +244,7 @@ if (BUILD_CVMFS)
        notify/subscriber_supervisor.cc
        notify/subscriber_sse.cc
        quota_listener.cc
+       util/inode.cc
        shortstring.cc
        supervisor.cc
        talk.cc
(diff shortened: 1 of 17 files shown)


christge-sft:
After https://github.com/cvmfs/cvmfs/pull/4149 this is now deprecated. Most of the feedback is incorporated there.

PR #4053, opened 2025-12-12
Add upstream cache performance metrics

andrey-utkin:
If CVMFS_PROXY_CACHE_STATUS_HEADER is set to a non-empty string, CVMFS
client, when it requests data from upstream server or proxy, will look
for an HTTP response header with such name, parsing out whether it as a
cache hit or a cache miss for that upstream server. These results are
aggregated into the following metrics to guide performance tuning:

download.n_proxy_cache_hit
download.n_proxy_cache_miss
download.n_proxy_cache_unc
download.sz_proxy_cache_hit
download.sz_proxy_cache_miss
download.sz_proxy_cache_unc
download-external.n_proxy_cache_hit
download-external.n_proxy_cache_miss
download-external.n_proxy_cache_unc
download-external.sz_proxy_cache_hit
download-external.sz_proxy_cache_miss
download-external.sz_proxy_cache_unc

n_proxy_cache_* metrics show the number of fetch requests fulfilled.
sz_proxy_cache_* show the total amount of bytes delivered.

*_proxy_cache_hit counters are for responses where upstream server indicates
that it was a cache hit.
*_proxy_cache_miss counters are for responses where upstream server indicates
that it was a cache miss.
*_proxy_cache_miss counters are for responses where upstream server has not
provided the expected indication - either because the header was missing, or
its value did not parse.

To enable this feature, define CVMFS_PROXY_CACHE_METRICS, e.g.:

    cmake . -DCFLAGS="-DCVMFS_PROXY_CACHE_METRICS=1" [...]

### DIFF
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -1390,6 +1390,11 @@ bool MountPoint::CreateCatalogManager() {
 
 bool MountPoint::CreateDownloadManagers() {
   string optarg;
+
+  if (options_mgr_->GetValue("CVMFS_PROXY_CACHE_STATUS_HEADER", &optarg)) {
+    download::ProxyCacheCounters::status_hdr = std::string(optarg);
+  }
+
   download_mgr_ = new download::DownloadManager(
       kDefaultNumConnections,
       perf::StatisticsTemplate("download", statistics_));
@@ -1567,6 +1572,8 @@ void MountPoint::CreateFetchers() {
       external_download_mgr_,
       backoff_throttle_,
       perf::StatisticsTemplate("fetch-external", statistics_));
+  proxy_cache_total_counters = new download::ProxyCacheCounters(
+      perf::StatisticsTemplate("download-total", statistics_));
 }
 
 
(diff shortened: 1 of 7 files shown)


andrey-utkin:
I am going to change the test code to avoid use of [cvmfs-posix-tools (cvmfs_touch) which are not merged yet](https://github.com/cvmfs/cvmfs/pull/3964) - I'll use ingestsql I think.

I also intend to add ifdef so that this feature doesn't get compiled in if not desired.

UPDATE 2025-12-17 all above done. Please review.

andrey-utkin:
Please review.

vvolkl:
Apologies it took me a while to come around to this. For my understanding, are these metrics specific to varnish or any proxy (squid?) will have them?

It's a nice addition, but I'd like to see if there's an alternative way to structure the code  to avoid all the #ifdefs , they make the code somewhat hard to read I find.

andrey-utkin:
> are these metrics specific to varnish or any proxy (squid?) will have them?

They are not builtin, but using such metrics is an established technique and there are tutorials on setting that up - official and not.

> It's a nice addition, but I'd like to see if there's an alternative way to structure the code to avoid all the #ifdefs , they make the code somewhat hard to read I find.

ifdefs have been dropped now, I just thought this feature won't be accepted without.

PR #4061, opened 2026-01-03
XattrList::CreateFromFile(): add compile-time option to use fewer syscalls

andrey-utkin:
If CVMFS_GET_XATTR_FAST is defined, the method uses automatic (on-stack) 64 KiB buffer with platform_lgetxattr(), using no more than one syscall to retrieve each key, and no dynamic memory allocation.

By default, it uses up to 3 syscalls per key retrieval and up to one dynamic memory allocation per key: it starts with 255 bytes buffer, and if it's not sufficient, it makes a syscall with buffer size argument 0, which returns the size of necessary buffer, which is then dynamically allocated, and used for the third platform_lgetxattr() call.

Link: https://github.com/cvmfs/cvmfs/pull/3622#discussion_r2294968712


andrey-utkin:
Not tested yet. CI run welcome.

vvolkl (commented):
Looks good. Do you want to add a cmake option to define CVMFS_GET_XATTR_FAST?

PR #4062, opened 2026-01-07
Bump quota manager protocol revision for recently added commands

DrDaveD:
After #4029, upgrading a running mount crashes.  I tracked it down to a Panic() on the running quota manager because it did not recognize the new `kSetCleanupPolicy` command.  This PR bumps the protocol revision number so a new cvmfs module won't send the new commands to an old quota manager, avoiding the crash.

### DIFF
--- a/cvmfs/quota.cc
+++ b/cvmfs/quota.cc
@@ -16,7 +16,7 @@
 
 using namespace std;  // NOLINT
 
-const uint32_t QuotaManager::kProtocolRevision = 2;
+const uint32_t QuotaManager::kProtocolRevision = 3;
 
 void QuotaManager::BroadcastBackchannels(const string &message) {
   assert(message.length() > 0);
--- a/cvmfs/quota.h
+++ b/cvmfs/quota.h
@@ -41,6 +41,9 @@ class QuotaManager : SingleCopy {
    *  - backchannel command 'R': release pinned files if possible
    * Revision 2:
    *  - add kCleanupRate command
+   * Revision 3:
+   *  - add kRegisterMountpoint, kGetMountpoints, kGetGroupHashes, and
+   *    kSetCleanupPolicy commands
    */
   static const uint32_t kProtocolRevision;
 
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -426,7 +426,9 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
     LogCvmfs(kLogQuota, kLogDebug, "failed to start cache manager");
     return NULL;
   }
-  LogCvmfs(kLogQuota, kLogDebug, "new cache manager pid: %d", new_cachemgr_pid);
+  LogCvmfs(kLogQuota, kLogDebug,
+           "new cache manager pid: %d protocol revision %d",
+           new_cachemgr_pid, QuotaManager::kProtocolRevision);
   quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
   const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(),
                                   O_RDWR | O_TRUNC, 0600);
@@ -792,6 +794,8 @@ uint32_t PosixQuotaManager::GetProtocolRevision() {
 }
 
 void PosixQuotaManager::SetCleanupPolicy(bool cleanup_unused_first) {
+  if (protocol_revision_ < 3) return;
+
   LruCommand cmd;
   cmd.command_type = kSetCleanupPolicy;
   WritePipe(pipe_lru_[1], &cmd, sizeof(cmd));
@@ -800,6 +804,8 @@ void PosixQuotaManager::SetCleanupPolicy(bool cleanup_unused_first) {
 }
 
 void PosixQuotaManager::RegisterMountpoint(const std::string &mountpoint) {
+  if (protocol_revision_ < 3) return;
+
   LruCommand cmd;
   cmd.command_type = kRegisterMountpoint;
   WritePipe(pipe_lru_[1], &cmd, sizeof(cmd));
@@ -810,6 +816,8 @@ void PosixQuotaManager::RegisterMountpoint(const std::string &mountpoint) {
 }
 
 std::string PosixQuotaManager::GetMountpoints() {
+  if (protocol_revision_ < 3) return "";
+
   int pipe_mp[2];
   MakeReturnPipe(pipe_mp);
 
@@ -825,6 +833,8 @@ std::string PosixQuotaManager::GetMountpoints() {
 }
 
 std::string PosixQuotaManager::GetGroupHashes() {
+  if (protocol_revision_ < 3) return "";
+
   int pipe_gh[2];
   MakeReturnPipe(pipe_gh);
 
--- a/cvmfs/quota_posix.h
+++ b/cvmfs/quota_posix.h
@@ -129,7 +129,7 @@ class PosixQuotaManager : public QuotaManager {
     kListVolatile,
     kCleanupRate,
     kSetLimit,
-    // After non open aware LRU cleanup
+    // as of protocol revision 3
     kRegisterMountpoint,
     kGetMountpoints,
     kGetGroupHashes,


vvolkl:
That's a good catch, thanks!

DrDaveD:
One thing I'm not clear on is how a new command could be received by an old quota manager, because I thought the quota manager got shut down during a reload before new code is loaded.

christge-sft:
For documentation purposes: trying to reproduce the aforementioned behavior was not straight forward. After @DrDaveD 's consulting, we concluded that the behavior is mostly visible when upgrading from v2.13.3 to https://github.com/cvmfs/cvmfs/pull/3730. I didn't see any crashing processes, but I had an incident where previously mounted repos where not there after the upgrade.

DrDaveD:
It appears the problem is intermittent, probably caused by some kind of race condition.  I found that the problem happens more reliably with multiple repositories mounted, and it fails when upgrading to the current devel branch and does not require #3730.

Here's my reproducer:
1. Install cvmfs-2.13.3.
2. Run the following:
```
for repo in /cvmfs/*; do sudo umount $repo; done; ls -d /cvmfs/cms.cern.ch/. /cvmfs/atlas.cern.ch/. /cvmfs/sft.cern.ch/.
```
3. Install cvmfs from the current devel branch.

If you run gdb on the lower-number cachemgr process before step 3, it gets an abort with this stack:
```
#0  0x00007f69b551b52f in raise () from /lib64/libc.so.6
#1  0x00007f69b54eee65 in abort () from /lib64/libc.so.6
#2  0x00007f69b5ece09b in __gnu_cxx::__verbose_terminate_handler() [clone .cold.1] () from /lib64/libstdc++.so.6
#3  0x00007f69b5ed454c in __cxxabiv1::__terminate(void (*)()) ()
   from /lib64/libstdc++.so.6
#4  0x00007f69b5ed45a7 in std::terminate() () from /lib64/libstdc++.so.6
#5  0x00007f69b5ed4808 in __cxa_throw () from /lib64/libstdc++.so.6
#6  0x00007f69b459f028 in Panic (coordinates=<optimized out>,
    source=<optimized out>, mask=<optimized out>, format=<optimized out>)
    at /usr/include/c++/8/ext/new_allocator.h:86
#7  0x00007f69b459f0a5 in Panic (coordinates=<optimized out>,
    _source=3034198993, nul=0x4 <error: Cannot access memory at address 0x4>)
    at /builddir/build/BUILD/cvmfs-2.13.3/cvmfs/util/exception.cc:56
#8  0x00007f69b4bf0d88 in PosixQuotaManager::ProcessCommandBunch (
    this=0x7ffed5beb510, this@entry=0x7ffed5bef8e0, num=num@entry=1,
    commands=commands@entry=0x7ffed5beb790,
    descriptions=descriptions@entry=0x7ffed5bebd90 "")
    at /builddir/build/BUILD/cvmfs-2.13.3/cvmfs/quota_posix.cc:1836
#9  0x00007f69b4bf24b5 in PosixQuotaManager::MainCommandServer (
    data=data@entry=0x7ffed5bef8e0)
    at /builddir/build/BUILD/cvmfs-2.13.3/cvmfs/quota_posix.cc:1550
#10 0x00007f69b4bf2ea3 in PosixQuotaManager::MainCacheManager (
    argc=argc@entry=11, argv=argv@entry=0x7ffed5beff78)
    at /builddir/build/BUILD/cvmfs-2.13.3/cvmfs/quota_posix.cc:1208
#11 0x00007f69b4c04583 in AltProcessFlavor (argc=11, argv=0x7ffed5beff78)
    at /builddir/build/BUILD/cvmfs-2.13.3/cvmfs/cvmfs.cc:2561
#12 0x00007f69b52ad481 in FuseMain (argc=11, argv=0x7ffed5beff78)
    at /builddir/build/BUILD/cvmfs-2.13.3/cvmfs/loader.cc:704
#13 0x00000000004042b1 in main (argc=11, argv=0x7ffed5beff78)
    at /builddir/build/BUILD/cvmfs-2.13.3/cvmfs/fuse_main.cc:145
```
If you then go up to stack level 8 and do `print commands[0].command_type` it shows 24, which is the number for the new `kSetCleanupPolicy` command.

christge-sft:
> One thing I'm not clear on is how a new command could be received by an old quota manager, because I thought the quota manager got shut down during a reload before new code is loaded.

Does bumping the protocol revision force trigger an update of the QuotaManager? Because if not, the protocol version checks on the QuotaManager ~~probably won't  be effective~~ will hide the fact that t he QuotaManger remains "old"

DrDaveD:
> Does bumping the protocol revision force trigger an update of the QuotaManager? Because if not, the protocol version checks on the QuotaManager probably will hide the fact that the QuotaManager remains "old"

Changing the revision does not trigger an update.  The revision field here and in loader.cc and cvmfs.cc are just used for keeping things compatible when revisions on the other side get out of sync.   As far as I understand it, there is always an update of the QuotaManager when there's a reload.  At least, the quota manager process always has a new start time.  So I don't know how the new code in the loaded cvmfs module gets out of sync from the quota manager, since I believe the quota manager code is in the same loadable module.  On the other hand the loader and cvmfs modules are separate so they can get out of sync by design.

DrDaveD:
I expanded the debug message that prints [`new cache manager pid`](https://github.com/cvmfs/cvmfs/blob/cvmfs-2.13.3/cvmfs/quota_posix.cc#L425) to also print out the protocol revision, and it is clear when debugging is enabled that as long as the old loader is running, even after doing more `cvmfs_config reload` commands, it is still running old code.  I don't know why that's happening, and it doesn't explain why the crash isn't reliably repeatable.

```
diff --git a/cvmfs/quota_posix.cc b/cvmfs/quota_posix.cc
index f0f3d9894..8417c656d 100644
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -425,7 +425,9 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
     LogCvmfs(kLogQuota, kLogDebug, "failed to start cache manager");
     return NULL;
   }
-  LogCvmfs(kLogQuota, kLogDebug, "new cache manager pid: %d", new_cachemgr_pid);
+  LogCvmfs(kLogQuota, kLogDebug,
+           "new cache manager pid: %d protocol revision %d",
+           new_cachemgr_pid, QuotaManager::kProtocolRevision);
   quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
   const int fd_lockfile_rw = open((workspace_dir + "/lock_cachemgr").c_str(),
                                   O_RDWR | O_TRUNC, 0600);
```

It looks to me like quota_posix.cc is only linked with the loadable cvmfs module, not in the stub which includes the loader.  Also, if I look at `/proc/<pid>/exe` for each of the cvmfs2 processes after an upgrade, only the primary process that includes the loader shows '/usr/bin/cvmfs2 (deleted)' indicating that it is still running code from a deleted binary.  The `__cachemgr__` processes show just `/usr/bin/cvmfs2` so I would think that should mean it is running new code.

DrDaveD:
Closing & reopening to re-run unittest.

DrDaveD:
And again.  The ubuntu2204 x86_64 test keeps failing even though no tests are marked as failed. Very frustrating.

DrDaveD:
And again, same problem

vvolkl:
@cernvm-bot unittest

DrDaveD:
Ubuntu 22.04 has been removed from the unittest so this time it should work.  I just rebased on the current devel and force-pushed.

DrDaveD:
> I expanded the debug message that prints [new cache manager pid](https://github.com/cvmfs/cvmfs/blob/cvmfs-2.13.3/cvmfs/quota_posix.cc#L425) to also print out the protocol revision, and it is clear when debugging is enabled that as long as the old loader is running, even after doing more cvmfs_config reload commands, it is still running old code.

I have an explanation for this at least now.  I was misinterpreting the debug log.  In fact the problem there was that the debug log stopped at reload time (because of a bug that only recognizes the debug flag at reload time if it is set for all repositories).  I was searching back from the end for the most recent message, not noticing it was an old message.   So I was incorrect in concluding it was still running old code.

In fact the debug mode shows it is running new code.  I also have not been able to reproduce the problem this week; the reproducer isn't causing the failure anymore.

vvolkl (changes requested):
Missing a comment about Revision 3 in `quota.h`: https://github.com/cvmfs/cvmfs/blob/d5ff84e272fb1d5d03bec16de6308abcfecdddff/cvmfs/quota.h#L38-L44

vvolkl (approved):

PR #4064, opened 2026-01-19
Migrate from bootstrap.sh to cmake: libarchive

ShivamMadlani:
This PR is the beginning of the migration from bootstrap.sh to cmake for building vendored dependencies.

Library involved: libarchive
Current version: 3.3.2

- added a new cmake module to handle dependencies (this is the replacement for `bootstrap.sh`)
- removed `libarchive` from the build process of `bootstrap.sh`
- removed the `find_package(libarchive)` function (no longer required as we can directly link target from `FetchContent`)
- handled `libarchive` config and build in `cmake/Modules/ExternalDependencies.cmake`
- linked `archive_static` target with `cvmfs_swissknife`, `cvmfs_server` and `cvmfs_receiver`
- removed redundant environment variable `LibArchive_LIBRARY`


cmake/Modules/ExternalDependencies.cmake
```diff
@@ -0,0 +1,44 @@
...
+
+# Required to maintain backwords compatibility with vendored dependencies
+set(CMAKE_POLICY_VERSION_MINIMUM 3.5)
+
+set(LIBARCHIVE_VERSION "3.3.2")
+set(LIBARCHIVE_LOCAL_PATH "${CMAKE_SOURCE_DIR}/externals/libarchive/libarchive-${LIBARCHIVE_VERSION}.tar.gz")
+set(LIBARCHIVE_URL "")
+
+# Resolve local path/URL of archive
+function(get_source_url NAME LOCAL_PATH REMOTE_URL RESULT_VAR)
```
vvolkl: This function isn't needed. You can give two arguments to `URL` in `FetchContent_Declare and it will behave just this way, falling back to the second if the first is not available

cmake/Modules/ExternalDependencies.cmake
```diff
@@ -0,0 +1,44 @@
+# CMake module for vendored dependencies
+include(FetchContent)
+
+# Required to maintain backwords compatibility with vendored dependencies
```
vvolkl:
```suggestion
# Required to maintain backwards compatibility with vendored dependencies
```

cmake/Modules/ExternalDependencies.cmake
```diff
@@ -0,0 +1,44 @@
...
+    message(STATUS "${NAME}: Using remote URL: ${REMOTE_URL}")
+    set(${RESULT_VAR} "${REMOTE_URL}" PARENT_SCOPE)
+  else()
+    message(FATAL_ERROR "${NAME}: Local archive not found and no remote URL provided!")
+  endif()
+endfunction()
+
+# -> libarchive <-
+# NOTE: This specific version of libarchive exports `archive` and `archive_static` as build targets.
+# NOTE: If upgrading to newer verion verify target names and make sure they match with usage in cvmfs/CMakeLists.txt
```
vvolkl:
```suggestion
# NOTE: If upgrading to newer version verify target names and make sure they match with usage in cvmfs/CMakeLists.txt
```

cmake/Modules/ExternalDependencies.cmake
```diff
@@ -0,0 +1,44 @@
...
+                patch -p0 < "${CMAKE_SOURCE_DIR}/externals/libarchive/src/libarchive_cmake.patch"
+  DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+)
+
+# Configuration
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
+set(ENABLE_TEST OFF CACHE BOOL "" FORCE)
+set(ENABLE_CPIO OFF CACHE BOOL "" FORCE)
+set(ENABLE_TAR OFF  CACHE BOOL "" FORCE)
+set(ENABLE_CAT OFF  CACHE BOOL "" FORCE)
```
vvolkl: Better to move this before `FetchContent_Declare`, see CMP0077

Also, there are more options set in bootstrap.sh, no? Better to add the full list.

CMakeLists.txt
```diff
@@ -300,6 +300,14 @@ if (BUILTIN_EXTERNALS)
   message("System prefix path: ${CMAKE_SYSTEM_PREFIX_PATH}")
 endif(BUILTIN_EXTERNALS)
 
+#
+# Migrate bootstrap.sh to cmake
+#
+if(BUILTIN_EXTERNALS)
+  message(STATUS "Enterning external dependencies setup")
```
vvolkl:
```suggestion
  message(STATUS "Entering external dependencies setup")
```

cmake/Modules/ExternalDependencies.cmake
```diff
@@ -0,0 +1,44 @@
...
+# NOTE: This specific version of libarchive exports `archive` and `archive_static` as build targets.
+# NOTE: If upgrading to newer verion verify target names and make sure they match with usage in cvmfs/CMakeLists.txt
+
+get_source_url("libarchive" "${LIBARCHIVE_LOCAL_PATH}" "${LIBARCHIVE_URL}" LIBARCHIVE_SRC)
+FetchContent_Declare(
+  LibArchive
+  URL "${LIBARCHIVE_SRC}"
+  PATCH_COMMAND patch -p0 < "${CMAKE_SOURCE_DIR}/externals/libarchive/src/fix-new-glibc.patch" &&
+                patch -p0 < "${CMAKE_SOURCE_DIR}/externals/libarchive/src/libarchive_cmake.patch"
+  DOWNLOAD_EXTRACT_TIMESTAMP TRUE
```
vvolkl: Add `URL_HASH`

cmake/Modules/ExternalDependencies.cmake
```diff
@@ -0,0 +1,44 @@
+# CMake module for vendored dependencies
```
vvolkl: Maybe better to rename to ExternalLibarchive.cmake or similar, and create one file per Dependency

vvolkl (changes requested):
Thanks Shivam for this contribution, that's great! Some comments inline. You can also remove makeHook.sh and configureHook.sh.

I think the build failure is because some compiler flags are leaking through to the main build, this needs to be fixed.

It may be a good idea to try and update libarchive to the latest version directly instead of patching the build system of a really old version.

cmake/Modules/ExternalDependencies.cmake
ShivamMadlani: Ah okay. Do we intend to keep the tarballs in the `externals/`. Or we want to fetch directly from a URL?

ShivamMadlani: I checked by adding 2 args in URL. 
Turns out we cannot mix local path (file://) and remote (http://). A helper function would be required in this case.

ShivamMadlani:
Thank you for the comments! I have addressed them in new commits. 

I have fixed the compiler flags leak as well as you suggested through CMP0077

I bumped up version to 3.8.0 for now. If everything goes well, ill remove the configurehook, makehook, patches and findlibarchive.cmake in my following commits

cmake/Modules/ExternalDependencies.cmake
vvolkl: Right, thanks for checking. Then indeed `get_source_url` is the right solution.

vvolkl:
Currently this is mixing the libarchive versions, 3.3.2 from local, but 3.8.0 from remote. Just update the local one too by adding the source tarball from 3.8.0  to the repository for now. I'm thinking if we can remove the checked-in copies of the tarballs - that should be ok if we keep a backup copy on one of our servers. We'll need to add a step to download them in the script that creates the CVMFS source tarball, as the build often needs to work without network when building packages like rpms.

ShivamMadlani:
@vvolkl the CI/cloudtests is failing even after adding tarball locally. I tried finding the issue by looking at the logs and it seems like an rpm packaging issue. I am not familiar with rpm packaging, can you point me to the direction for a fix?

Also the workflows *_ubuntu seems to be stuck.

vvolkl:
Right, that's because libarchive is doing an install step and installing files next to the cvmfs ones. rpmbuild checks that only the files we declare in the package are installed, which is good because we don't want to install libarchive, it's statically linked. Let's try adding `ENABLE_INSTALL=OFF` to the libarchive cmake options, I think the target should still work.

ShivamMadlani:
> Right, that's because libarchive is doing an install step and installing files next to the cvmfs ones. rpmbuild checks that only the files we declare in the package are installed, which is good because we don't want to install libarchive, it's statically linked. Let's try adding `ENABLE_INSTALL=OFF` to the libarchive cmake options, I think the target should still work.

Yes, this fits the picture now! Added the `ENABLE_INSTALL=OFF` and handled the variable scope by unsetting at the end of module.

vvolkl:
As a note to myself, the vendored libarchive tarball was manually minified (see https://github.com/cvmfs/cvmfs/tree/cvmfs-2.13.3/externals/libarchive/src#readme), I should check if the static library increases drastically in size.

vvolkl:
I'm looking into removing the local tarball alltogether. I think that will work if we:

* add a copy to a self-hosted server. Done already for libarchive: https://ecsft.cern.ch/dist/cvmfs/build_externals/libarchive-3.8.5.tar.gz ( I went for 3.8.5 instead of 3.8.0, I don't see a reason not to use the latest one if it works )
* add a script to `ci/` that downloads it (and the other externals that will be rewritten like this. For this it would be good to move the url string to a file where it can be easily sourced both from bash and cmake)
* run this script when preparing the source tarball for package builds

ShivamMadlani:
> I'm looking into removing the local tarball alltogether. I think that will work if we:
> 
>     * add a copy to a self-hosted server. Done already for libarchive: https://ecsft.cern.ch/dist/cvmfs/build_externals/libarchive-3.8.5.tar.gz ( I went for 3.8.5 instead of 3.8.0, I don't see a reason not to use the latest one if it works )
> 
>     * add a script to `ci/` that downloads it (and the other externals that will be rewritten like this. For this it would be good to move the url string to a file where it can be easily sourced both from bash and cmake)
> 
>     * run this script when preparing the source tarball for package builds

Yeah, makes sense. Before that should we address the test timeout? I looked at the logs, test 79-dnsroaming is the issue. the timeout is happening right after cvmfs_mount is being called in the test. not sure if this is because of the cmake changes as other tests like 2-probe are passing.

079-dnsroaming/main
```
  echo "*** mounting sft.cern.ch"
  cvmfs_mount sft.cern.ch "CVMFS_DNS_ROAMING=ON" || return 1

  echo "*** getting default value of DNS server (should be unset)"
```
logs
```
-- Testing DNS roaming (Wed Jan 21 10:13:35 UTC 2026 / test number 079)
restarting autofs...
using systemd
*** mounting sft.cern.ch
Hangup
KILLING:99145
execution took 300.016 seconds
Testcase timed out
```

vvolkl:
Yes, I think the test failure is unrelated. Would be good to investigate, but I think that may have been a temporary network problem.

I've thought a bit more about what we need  for the externals and how to get CMake to do that, and  while a few things aren't super elegant I don't think there is a blocking issue:

- [x] Remove externals tarball from the git repository and move them to a server, but keeping the possibility to download them separately for caching or preparing complete source tarballs.
  - For this we can just set 
  ```
  DOWNLOAD_DIR ${EXTERNALS_LIB_LOCATION}/download
  ```
    in FetchContent, and then if `externals/download` is populated with a tarball with the right name and hash, CMake will use that one and skip the download. To download the external dependency source tarballs when preparing the cvmfs source tarball we can either invoke the configure step of cmake (takes a bit longer but is simplest) with the right arguments or write another script, that should probably grep for the tarball url in the CMakeLists.
- [x] Allow some way of caching the binaries of the vendored externals
  - A nice feature of `bootstrap.sh` was that we could cache the `externals_install` folder, and if it was restored in the right location the build system would just find the libraries and skip the build. This can be emulated with CMake by just setting (here for libarchive):
      ```
      set(ENABLE_INSTALL  ON CACHE BOOL "" FORCE)
     set(CMAKE_INSTALL_PREFIX ${EXTERNALS_INSTALL_LOCATION})
     ```
     That way, the internals would only install at the same time with cvmfs when running `make install`, but there's even a way to add a custom target to install only the deps, which is pretty nice. For this we can set
     ```
     set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME external_deps)
     ```
     before FetchContent, and if the external package hasn't messed with components, we can install only it with:
     ```
     add_custom_target(install_external
     COMMAND ${CMAKE_COMMAND} --install ${CMAKE_BINARY_DIR} 
          --component external_deps 
          --prefix ${EXTERNALS_INSTALL_LOCATION}
    COMMENT "Installing external dependencies to     ${EXTERNAL_INSTALL_LOCATION}"
     )
     ``` 
    This makes preparing a cache pretty easy, but to pick up the cached dependencies on the next build we still need the last point:
  - [ ] Flexibly choose between building the package on the spot and using find_package
    - This can be done with the [`FIND_PACKAGE_ARGS` option to FetchContent_Declare](https://cmake.org/cmake/help/latest/module/FetchContent.html#integrating-with-find-package), and the `FETCHCONTENT_TRY_FIND_PACKAGE_MODE` variable. That one is global though (applying to all FetchContent calls) whereas we'd want to be flexible in choosing for any external if it should be picked up from the system or built from source. So here there's a little bit more code to write that parses BUILTIN_EXTERNALS_LIST and BUILTIN_EXTERNALS_EXCLUDE and sets FETCHCONTENT_TRY_FIND_PACKAGE_MODE for each package accordingly.

cmake/Modules/FetchLibArchive.cmake
```diff
@@ -0,0 +1,74 @@
...
+  LibArchive
+  URL "${LIBARCHIVE_SRC}"
+  URL_HASH "${LIBARCHIVE_HASH}"
+  DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+)
+
+# Make libarchive available
+FetchContent_MakeAvailable(LibArchive)
+
+# Unset variables from cache to avoid bleeding into other packages
```
vvolkl: Instead of unsetting, how about moving the content of this file to `externals/libarchive/CMakeLists.txt`
and calling `add_subdirectory`in the main CMakeLists.txt? Then this runs in its own scope automatically.

vvolkl (changes requested):

ShivamMadlani:
I am investigating the unit test timeout
```
[----------] 4 tests from T_FileWatcher
[ RUN      ] T_FileWatcher.NoEventStop
```

I first verified if this is because of my changes by:
checking out to devel branch -> building cvmfs with unit tests -> running `./cvmfs_unittests` on my local machine.

--> the filewatcher test passes almost instantly

but on repeating the test on the current branch, the test enters a constant wait (hangs). below is the backtrace for gdb
```
(gdb) bt
#0  0x00007fc8afca432a in ?? () from /usr/lib/libc.so.6
#1  0x00007fc8afd2cbaa in read () from /usr/lib/libc.so.6
#2  0x00007fc8b0c58eee in ReadHalfPipe (fd=5, buf=buf@entry=0x7fff01e2a087, nbyte=nbyte@entry=1, timeout_ms=timeout_ms@entry=0) at /home/white/projects/cvmfs/cvmfs/util/posix.cc:547
#3  0x0000558c7dacf7a9 in file_watcher::FileWatcher::Spawn (this=this@entry=0x558ca1e39220) at /home/white/projects/cvmfs/cvmfs/file_watcher.cc:73
#4  0x0000558c7dc6baa7 in T_FileWatcher_NoEventStop_Test::TestBody (this=<optimized out>) at /home/white/projects/cvmfs/test/unittests/t_file_watcher.cc:70
#5  0x0000558c7dcad085 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (object=0x558ca1e38890, method=<optimized out>,
    location=0x558c7de1a651 "the test body") at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:2653
#6  testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=object@entry=0x558ca1e38890, method=<optimized out>,
    location=location@entry=0x558c7de1a651 "the test body") at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:2700
#7  0x0000558c7dc9bfe3 in testing::Test::Run (this=this@entry=0x558ca1e38890) at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:2739
#8  0x0000558c7dc9c41c in testing::Test::Run (this=0x558ca1e38890) at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:2730
#9  testing::TestInfo::Run (this=0x558ca1e37be0) at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:2885
#10 0x0000558c7dc9dc10 in testing::TestSuite::Run (this=0x558ca1e37d00) at /usr/include/c++/15.2.1/bits/stl_vector.h:1282
#11 0x0000558c7dca3e06 in testing::TestSuite::Run (this=<optimized out>) at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:3018
#12 testing::internal::UnitTestImpl::RunAllTests (this=this@entry=0x558ca1da05b0) at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:6054
#13 0x0000558c7dca46a3 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x558ca1da05b0,
    method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x558c7dca2f10 <testing::internal::UnitTestImpl::RunAllTests()>,
    location=0x558c7de6c7d0 "auxiliary test code (environments or event listeners)") at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:2653
#14 testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x558ca1da05b0, method=<optimized out>,
    location=0x558c7de6c7d0 "auxiliary test code (environments or event listeners)") at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:2700
#15 testing::UnitTest::Run (this=<optimized out>) at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/src/gtest.cc:5594
#16 0x0000558c7d32d028 in RUN_ALL_TESTS () at /home/white/projects/cvmfs/build/_deps/googletest-src/googletest/include/gtest/gtest.h:2334
#17 main (argc=<optimized out>, argv=0x7fff01e2a6b8) at /home/white/projects/cvmfs/test/unittests/main.cc:24
```

This hang does not occur in devel branch, and on looking at above backtrace,
```
/home/white/projects/cvmfs/cvmfs/util/posix.cc:547
    num_bytes = read(fd, buf, nbyte);
```

getting blocked at read syscall, can happen if the background thread created never writes to the buffer, causing to wait indefinitely at read()
```
cvmfs/file_watcher.cc::67
  assert(pthread_create(&thread_, NULL, &FileWatcher::BackgroundThread, this)
         == 0);
```

I am still trying to figure out how these changes affect the threads. Please let me know your thoughts on this.

vvolkl:
I think that may be because FetchContent for  libarchive is still adding an -O3 flag to the main build. It's a bit strange that this causes the unittests to lock up, but let's check first what happens if we get rid of this and build with -Os as before

PR #4065, opened 2026-01-21
gw: annotate cvmfs_receiver process with lease path

### LINKED ISSUE #3399: RFE: Gateway quality-of-life improvement -- annotate cvmfs_receiver process
Could the lease path also be passed as an (unused) argument to `cvmfs_reciver` - that makes it unambiguous what a given receiver process is working on, very useful if one gets stuck and we need to investigate.

orbaps:
This PR addresses issue #3399 by adding an optional `-l` argument to the `cvmfs_receiver` process.

### Description
The `cvmfs_receiver` process is spawned by the gateway. This change passes the lease path as an unused `-l` command-line argument to the receiver process. This allows administrators to easily identify which receiver process corresponds to which lease path using standard process listing tools, aiding in debugging and monitoring.

### Changes
*   **`cvmfs/receiver/receiver.cc`**: Updated argument parsing to accept an optional `-l` flag.
*   **`gateway/internal/gateway/receiver/pool.go`**: Updated the worker pool to extract the lease path from the task and pass it to the receiver process.

### Testing
*   Verified that the gateway compiles and tests pass.
*   The extra argument is ignored by the receiver logic, ensuring no side effects on functionality.

Fixes #3399

### DIFF
--- a/cvmfs/receiver/receiver.cc
+++ b/cvmfs/receiver/receiver.cc
@@ -30,6 +30,8 @@ swissknife::ParameterList MakeParameterList() {
       'd', "Path to debug log. Ignored if the non-debug version runs, "
            "Default: "
                + std::string(kDefaultDebugLog)));
+  params.push_back(swissknife::Parameter::Optional(
+      'l', "Lease path (unused, for process identification)"));
   return params;
 }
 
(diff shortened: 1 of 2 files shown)


vvolkl (commented):
Right, this is a very straightforward solution, and is a nice improvement. I'm wondering though if we can use one of the methods discussed [here](https://stackoverflow.com/questions/6082189/change-process-name-in-linux/55584492#55584492) to annotate the process name without introducing spurious arguments.

Issue #3340, opened 2023-07-25
Improve test setup for gateway


jblomer:
The gateway test setup is currently somewhat clunky, e.g. it works with a hardcoded `test.repo.org` repository.  It would be nicer to have the gateway setup orthogonal to the other server operations. I.e., it should be possible to create repositories as usual and then to add a gateway.

See also https://github.com/cvmfs/cvmfs/pull/3339

Agamya-Samuel:
Hi @jblomer, I am an Open Source contributor, and I plan to contribute to this project.
I have read the docs, let me know if this PR #4067 needs any improvemeent.

PR #4067, opened 2026-01-24
Add timeout in test setup for gateway

Agamya-Samuel:
Replace unreliable sleep and multiple status checks in set_up_repository_gateway with a proper polling loop that waits for the gateway to be ready with a timeout.

This change:
- Replaces fixed sleep 1 and three consecutive status calls with a polling loop
- Adds a 30-second timeout to prevent infinite loops
- Returns an error if the gateway doesn't start in time
- Suppresses status output during polling for cleaner logs
- Matches the pattern used in restart_repository_gateway for consistency

Part of #3340


vvolkl (changes requested):
Thanks for taking a look at this!

How about using `until` instead of while, and the `timeout` command directly:

```bash
if ! timeout 30 bash -c "until $gateway_script status &>/dev/null; do sleep 1; done"; then
    echo "Gateway failed to start within 30 seconds"
    return 1
fi
```
Seems clearer to me.

Also, the message should be echo'd to stderr, not stdout.

Agamya-Samuel:
> Thanks for taking a look at this!
> 
> How about using `until` instead of while, and the `timeout` command directly:
> 
> ```shell
> if ! timeout 30 bash -c "until $gateway_script status &>/dev/null; do sleep 1; done"; then
>     echo "Gateway failed to start within 30 seconds"
>     return 1
> fi
> ```
> 
> Seems clearer to me.
> 
> Also, the message should be echo'd to stderr, not stdout.

sure @vvolkl I'll update this and let you know.

Agamya-Samuel:
I have read the overview docs at - https://cvmfs.readthedocs.io/en/stable/
and - https://chrisdown.name/2024/02/05/reliably-creating-d-state-processes-on-demand.html

what would be the best I can contribute and understand this project better ?
should I setup the entire enviroment locally - https://github.com/cvmfs/cvmfs/blob/devel/doc/developer/10-overview.md

or there is a slightly better approach ?

test/test_functions
```diff
@@ -2691,11 +2691,12 @@ EOF'
 
     echo "  Starting repository gateway"
     sudo $gateway_script start
-    # Let the service boot up
-    sleep 1
-    $gateway_script status
-    $gateway_script status
-    $gateway_script status
+
+    # Wait for gateway to be ready with timeout (use timeout + until as suggested)
```
vvolkl:
```suggestion
    # Wait for gateway to be ready with timeout
```

PR #4076, opened 2026-01-29
feat(ducc): support publishing to CVMFS repository subdirectories

### LINKED ISSUE #3780: cvmfs_ducc: add option for putting images into a subdirectory

clelange:
feat(ducc): support publishing to CVMFS repository subdirectories

## Description
This PR implements support for specifying a target subdirectory within a CVMFS repository for `ducc` unpacking (e.g., `unpacked.cern.ch/compat`). This addresses the feedback from PR #3806 and ensures robust handling of repository paths.

## Changes
- **Utilities**: Added `GetRepoAndSubdir` in [ducc/cvmfs/utils.go](https://github.com/clelange/cvmfs/blob/3f5dc356924f589477a404a22c879392a2011a0e/ducc/cvmfs/utils.go) to correctly split repository names and subdirectories, supporting both production strings and test harness paths.
- **Core Logic**: Updated `CreateSneakyChain` in [ducc/cvmfs/cvmfs.go](https://github.com/clelange/cvmfs/blob/3f5dc356924f589477a404a22c879392a2011a0e/ducc/cvmfs/cvmfs.go) to use the base repository name for scratch directory resolution (`/var/spool/cvmfs/REPO/scratch/current`) while preserving the subdirectory structure for the internal layout.
- **Command Handling**: Updated [ducc/cvmfs/raw.go](https://github.com/clelange/cvmfs/blob/3f5dc356924f589477a404a22c879392a2011a0e/ducc/cvmfs/raw.go) to pass the full path to `cvmfs_server transaction` (supporting subdir locking) while stripping the subdirectory for `publish` and `abort` commands.
- **Tests**:
  - Added unit tests for path splitting in [ducc/cvmfs/utils_test.go](https://github.com/clelange/cvmfs/blob/3f5dc356924f589477a404a22c879392a2011a0e/ducc/cvmfs/utils_test.go).
  - Extended the mock `cvmfs_server` to verify subdirectory transaction arguments.
  - Added a new integration test `TestPublishToSubdir` to verify the end-to-end flow.

## Verification
- **Unit/Integration Tests**: Verified using the local Docker-based test harness (`golang:1.24.0-bookworm`) with `TEST_DUCC_ONLINE=yes`. All tests passed, including the new integration test.
- **Build**: Verified standard CMake build succeeds.

Fixes #3780

### DIFF
--- a/ducc/cvmfs/utils.go
+++ b/ducc/cvmfs/utils.go
@@ -11,17 +11,19 @@ import (
 // component is the repository name and the rest is the subdirectory path.
 //
 // For testing with mock repositories (relative paths like "../../tmp/mockrepo"),
-// use a colon separator: "../../tmp/mockrepo:subdir" to avoid ambiguity with
-// the slashes in the relative path itself.
+// use a triple-slash separator: "../../tmp/mockrepo///subdir" to avoid ambiguity
+// with the slashes in the relative path itself. The string stays a valid path
+// (multiple / are treated as one by the OS) and avoids headaches with filepaths
+// that could contain literal colons.
 //
 // Examples:
-//   - "unpacked.cern.ch"           → ("unpacked.cern.ch", "")
-//   - "unpacked.cern.ch/images"    → ("unpacked.cern.ch", "images")
-//   - "../../tmp/mock:subdir"      → ("../../tmp/mock", "subdir")
+//   - "unpacked.cern.ch"              → ("unpacked.cern.ch", "")
+//   - "unpacked.cern.ch/images"       → ("unpacked.cern.ch", "images")
+//   - "../../tmp/mock///subdir"       → ("../../tmp/mock", "subdir")
 func GetRepoAndSubdir(cvmfsRepo string) (repoName, subDir string) {
-	if strings.Contains(cvmfsRepo, ":") {
-		repoName, subDir, _ = strings.Cut(cvmfsRepo, ":")
-		subDir = strings.TrimSuffix(subDir, "/")
+	if idx := strings.Index(cvmfsRepo, "///"); idx >= 0 {
+		repoName = cvmfsRepo[:idx]
+		subDir = strings.TrimSuffix(cvmfsRepo[idx+3:], "/")
 		return
 	}
 	if strings.HasPrefix(cvmfsRepo, "..") || strings.HasPrefix(cvmfsRepo, "/..") {
--- a/ducc/cvmfs/utils.go
+++ b/ducc/cvmfs/utils.go
@@ -27,6 +27,13 @@ func GetRepoAndSubdir(cvmfsRepo string) (repoName, subDir string) {
 	if strings.HasPrefix(cvmfsRepo, "..") || strings.HasPrefix(cvmfsRepo, "/..") {
 		return cvmfsRepo, ""
 	}
+	if strings.HasPrefix(cvmfsRepo, "/") {
+		if strings.HasPrefix(cvmfsRepo, "/cvmfs/") {
+			cvmfsRepo = strings.TrimPrefix(cvmfsRepo, "/cvmfs/")
+		} else {
+			return strings.TrimSuffix(cvmfsRepo, "/"), ""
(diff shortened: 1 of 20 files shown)


clelange:
The tests were incorrectly parsing relative paths such as `../../repo/subdir` by treating the entire string as the repository name. I now added support for a `:` separator to unambiguously separate the repository path and the subdirectory. Let's see if this works now.

clelange:
OK, I hope it's better now. Also cleaned up my commits.

ducc/cvmfs/utils.go
```diff
@@ -0,0 +1,30 @@
...
+
+import "strings"
+
+// GetRepoAndSubdir splits a CVMFSRepo specification into repository name and subdirectory.
+//
+// In production, repository paths look like "repo.cern.ch/subdir" where the first
+// component is the repository name and the rest is the subdirectory path.
+//
+// For testing with mock repositories (relative paths like "../../tmp/mockrepo"),
+// use a colon separator: "../../tmp/mockrepo:subdir" to avoid ambiguity with
```
vvolkl: If pathstrings like these need to be separated, I find it best to use `///` as a separator, so `../../tmp/mockrepo///subdir`. Like this the string stays a valid path (multiple `/` are treated as one by the os) and there are no headaches with filepaths containing actual`:`s (which is unlikely to be the case in our tests but is in principle allowed in file paths)

vvolkl (changes requested):
Thanks for the initiative! This is not yet quite finished. When testing this with a live repo `/cvmfs/test.cern.ch/subdir` I get multiple duplicated subdirs. It's for sure not passing the repo and path strings correctly to `cvmfs_server ingest` in `IngestintoCVMFS` and there may be more places to patch still. Unfortunately the mock testing won't catch this.

vvolkl:
PS: you might want to hold off with updating this PR until https://github.com/cvmfs/cvmfs/pull/4092 is in production. This is what I waited for in the first place, it'll let us eliminate the template transactions which also makes this feature easier to implement.

clelange:
OK, no problem, I'll wait for the other PR.
In the meantime, I'll give it a try again in a local installation. So far I had relied on CI tests.

clelange:
OK, my tests on a light-weight VM didn't work. Need to switch to a proper one.

clelange:
I *think* that things should be OK now. I've also fixed some missing requirements and made the tests more robust. Happy to squash commits if it's useful.

clelange:
I managed to make all the tests work locally on a ubuntu machine (with the exception of the graphdriver one). I had to make some more fixes for this, e.g. provide a valid `storage.conf` for podman. Also happy to first fix the tests separately to keep this pull request focussed and since it's becoming difficult to review it.

vvolkl:
Yes, best to fix the tests in a separate PR. For this PR I'd prefer to first merge #4092, which just needs the test failure adressed, but I can take  care of rebasing this PR afterwards.

clelange:
OK, let me extract the test fixes and move them to a separate PR. Might take me a bit though due to other commitments.

clelange:
@vvolkl could you please approve the workflows just to make that things are OK?

vvolkl:
Went ahead with resolving the conflicts and adressing my comment about the separator (original branch still here: https://github.com/cvmfs/cvmfs/tree/ducc_subdir). For me this is OK now, I'll go ahead and merge this PR,  will do a bigger validation once  the multiarch support is merged as well.

vvolkl (approved):

clelange:
Great, thanks, that makes things easier for me :-)

Issue #4069, opened 2026-01-27
Overhaul of build system for vendored dependencies


vvolkl:
The current build system invokes bootstrap.sh, which builds some vendored dependencies from source tarballs checked into the repository and installs them to `externals/externals_install.${platform}-${arch}` so CMake can then find and (statically) link them. This has worked very well, but has a few downsides that are more obvious the more often you need to build:

- The compilation of the externals happens during the configure step, making this step long and cpu-intensive. It's nicer to keep the configure step short and do compilations during the build step
- It's harder to propagate settings like `-jX` or cross-compilation settings ( for example, this makes it harder to build universal binaries on macos )
- Updating the version of vendored dependencies means checking in more binary blobs into the git repo, slowing down future git clone operations
- In order to upstream CVMFS to package repositories, it's good to reduce the number of vendored dependencies

As a solution we came up with #4064 that uses FetchContent to adress all these things on the example of libarchive. It also fullfills the additional requirements that a source tarball that bundles all vendored dependencies can be easily created (by downloading the vendored dependencies to externals/download prior to zipping), and that the vendored dependencies can be easily cached (either by using ccache or installing with `make install_externals`, and caching the `externals/external_install.*` dir. It's also flexible in choosing between taking the dependency from the system vs building it from source, by using EXTERNALS_INSTALL_LIST and EXTERNALS_INSTALL_EXCLUDE.

There are a few small improvements that are  yet to be done:

- [ ] add some instructions in doc/developer
- [ ] when using FETCHCONTENT_TRY_FIND_PACKAGE_MODE=ALWAYS, you could in theory pick up an installation from externals/externals_install when you were intending to link to a system package. That shouldn't break things but may be surprising, so we could manipulate the CMAKE_PREFIX_PATH in that case
- [ ]  modify the script creating the source tarball to populate externals/download

The same recipe can be applied to most other vendored dependencies as well. I'll comment below for the individual cases.

Thanks to @ShivamMadlani for the nice initiative that got this started!

vvolkl:
- `libarchive`: can be taken from system. #4064

vvolkl:
- [x] `vjson`: should be replaced. I think a good candidate would be simdjson which can be taken from the system
    - [x] Done by @ShivamMadlani  in #4083 , with nlohmann-json as replacement as the API is more favorable. 
    - nlohmann-json can be taken from the system, but there's not too much benefit as it's header-only

vvolkl:
[...]

PR #4083, opened 2026-01-31
Migrate:  vjson->nlohmann json and bootstrap->cmake

ShivamMadlani:
## Migration phase 2: Successor to #4064 

View issue #4069

### Migrations:
- vjson -> nlohmann json
- bootstrap.sh -> cmake build system

### Progress:
- [x] add vendoring and building logic for nlohmann json
- [x] replace older `json_parse()` with nlohmann json equivalent logic in `cvmfs/authz/helper_util.cc`, `cvmfs/json_document.cc`
- [x] link to target nlohmann_json::nlohmann_json instead of ${VJSON_LIBRARIES} in `cvmfs/`, `test/unittests/` and `test/micro-benchmarks`
- [x] remove find_package call to vjson and its FindVJSON.cmake module
- [x] remove `externals/vjson/`
- [x] fix json key ordering, use nlohmann::ordered_json instead
- [x] fix floating point formatting. update test/force precision
- [x] strict parsing: update unit tests to stop using {} without key value pair

Please feel free to add/remove points or files I missed


vvolkl:
We've been discussing using nlomann-json instead since the migration seems to be simpler.

ShivamMadlani:
There are total of 72 callers of JsonDocument::SearchInObject() in cvmfs. nlohmann provides APIs to do the same. I can think of 2 ways to proceed:

1. Replace all 72 instances of SearchInObject with nlohmann json api. This will clean up a lot of manual memory management and offload logic to the json library itself
2. Keep the 72 callers untouched, but modify the function and the signature itself in json_document.cc to internally use nlohmann json. This will be less invasive than first approach but still use the legacy logic

I think the first approach is better on the long run at the expense of effort

externals/json/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# CMake module for vendoring json
+include(FetchContent)
+
+set(JSON_VERSION "3.12.0")
+set(JSON_URL "https://github.com/nlohmann/json/archive/refs/tags/v${JSON_VERSION}.tar.gz")
+# TODO: add json tarball to cern servers
```
vvolkl:
```suggestion
```
done:  https://ecsft.cern.ch/dist/cvmfs/build_externals/json-v3.12.0.tar.gz

vvolkl:
I think we want to take nlohmann-json from the system by default. Are you up for adding it to rpm/deb packaging? For that I think you need to add a `Requires: json` in https://github.com/cvmfs/cvmfs/blob/devel/packaging/rpm/cvmfs-universal.spec both for the cvmfs and server package, as well as adding `nlohmann-json3` to the Depends list for both cvmfs and cvmfs-server in https://github.com/cvmfs/cvmfs/blob/devel/packaging/debian/cvmfs/control

Maybe we can also set the default find strategy to always go for the system package, and error when it is not available.

ShivamMadlani:
> I think we want to take nlohmann-json from the system by default. Are you up for adding it to rpm/deb packaging? For that I think you need to add a `Requires: json` in https://github.com/cvmfs/cvmfs/blob/devel/packaging/rpm/cvmfs-universal.spec both for the cvmfs and server package, as well as adding `nlohmann-json3` to the Depends list for both cvmfs and cvmfs-server in https://github.com/cvmfs/cvmfs/blob/devel/packaging/debian/cvmfs/control

Yeah sure, I'll try and fix the packaging.

> Maybe we can also set the default find strategy to always go for the system package, and error when it is not available.

Does this mean we wont be vendoring it?

vvolkl:
> Does this mean we wont be vendoring it?

Not when building the regular rpm/deb packages, no. But it's important to have the option to do so for other builds - we may try to do a completely statically linked build in the future.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -174,6 +174,7 @@ Requires: util-linux
   %endif
 %endif
 Requires: cvmfs-config
+Requires: nlohmann-json-devel
```
vvolkl: I think it's just called `json-devel` on rpm-based patforms:

```
yum -y install nlohman-json-devel
Last metadata expiration check: 1:24:17 ago on Tue 17 Feb 2026 02:05:22 PM CET.
No match for argument: nlohman-json-devel
Error: Unable to find a match: nlohman-json-devel
[root@cvm-test-18 ~]# yum -y install json-devel
Last metadata expiration check: 1:24:30 ago on Tue 17 Feb 2026 02:05:22 PM CET.
Dependencies resolved.
======================================================================================
 Package               Arch        Version                       Repository      Size
======================================================================================
Installing:
 json-devel            x86_64      3.6.1-2.el8                   epel           142 k
```

ShivamMadlani: you are right! sorry about that. Updated.

packaging/rpm/cvmfs-universal.spec
```diff
@@ -174,6 +174,7 @@ Requires: util-linux
   %endif
 %endif
 Requires: cvmfs-config
+Requires: json-devel
```
vvolkl: Forgot another thing: I think we should split this up in 
```
BuildRequires: json-devel
```
and
```
Requires: json
``` 
Like this the json-devel package (which includes header files) is only needed when building the cvmfs package, but when a user does 
`yum install cvmfs`
only the smaller runtime package is needed. Similarly for debian

ShivamMadlani: Right, makes sense.

but only `json-devel` package exists in fedora (rpm based distros)
```
bash-5.3# yum install json
Updating and loading repositories:
Repositories loaded.
Failed to resolve the transaction:
No match for argument: json
You can try to add to command line:
  --skip-unavailable to skip unavailable packages
bash-5.3# yum install json-devel
Updating and loading repositories:
Repositories loaded.
Package                       Arch     Version                       Repository            Size
Installing:
 json-devel                   x86_64   3.11.3-4.fc43                 fedora             1.0 MiB
Installing dependencies:
 cmake-filesystem             x86_64   3.31.10-3.fc43                updates            0.0   B
 libpkgconf                   x86_64   2.3.0-3.fc43                  fedora            78.1 KiB
 libstdc++-devel              x86_64   15.2.1-7.fc43                 updates           37.3 MiB
 pkgconf                      x86_64   2.3.0-3.fc43                  fedora            88.5 KiB
 pkgconf-m4                   noarch   2.3.0-3.fc43                  fedora            14.4 KiB
 pkgconf-pkg-config           x86_64   2.3.0-3.fc43                  fedora           989.0   B

Transaction Summary:
 Installing:         7 packages
 ```
 
 this is because this library is a header only package. same case with debian package, only `-dev` is available

Comment (vvolkl) [reply packaging/rpm/cvmfs-universal.spec]: Ah right - then we only need the `BuildRequires:`

Comment (ShivamMadlani) [reply packaging/rpm/cvmfs-universal.spec]: maybe we also need to add json-devel here in dnf install -y?
https://github.com/cvmfs/cvmfs/blob/93c1033fd2b891d39461480bf9d77a1e4c8c9a4d/test/common/container/Dockerfile-dev#L10-L22

Comment (vvolkl) [reply packaging/rpm/cvmfs-universal.spec]: Yes, I'll just update and  rebuild the gitlab-registry.cern.ch/cernvm/build-images/centos_x86_64:9 image. Maybe we should actually just use a base image here - we have the `ci/build_install_builddeps.sh` script now that makes installing build dependencies pretty easy. But it'd be good to add some caching to the docker build

**Comment by vvolkl** [comment]
On MacOS the build system still tries to find vjson:
```
 --> Copying vjson
cp: /Users/runner/work/cvmfs/cvmfs/externals/vjson/src/*: No such file or directory
CMake Error at CMakeLists.txt:275 (message):
  bootstrapping failed
```

File: externals/bootstrap.sh [hunk 4]
```diff
@@ -285,7 +285,7 @@ build_lib() {
 
 # Build a list of libs that need to be built
 # Check if BUILTIN_EXTERNALS_LIST is set and override missing_libs
-if [ x"$BUILTIN_EXTERNALS_LIST" != x"" ] && ! echo ";${BUILTIN_EXTERNALS_LIST};" | grep -qE ";(libarchive);"; then
+if [ x"$BUILTIN_EXTERNALS_LIST" != x"" ] && ! echo ";${BUILTIN_EXTERNALS_LIST};" | grep -qE ";(libarchive|json);"; then
```
vvolkl: Shouldn't this be:
```suggestion
if [ x"$BUILTIN_EXTERNALS_LIST" != x"" ] && ! echo ";${BUILTIN_EXTERNALS_LIST};" | grep -qE ";(libarchive|vjson);"; then
```

externals/bootstrap.sh
vvolkl:
```suggestion
    missing_libs="libcurl libcrypto pacparser zlib sparsehash leveldb maxminddb protobuf sqlite3 sha3"
```

vvolkl (approved):

PR #4085, opened 2026-02-03
Make recognition of CVMFS_DEBUGLOG at reload be per-repo

DrDaveD:
I noticed in https://github.com/cvmfs/cvmfs/pull/4062#issuecomment-3844380032 that debug mode gets lost at reload time if it isn't set for all repositories.  This has been a problem since #3327 where debug mode was implemented to be enabled or disabled at reload time.  This PR makes it separately enable or disable debug mode per repo.


vvolkl (approved):
That's very useful - thanks Dave!

PR #4092, opened 2026-02-08
New tools for overlays on catalog level

vvolkl:
This should let us make the unpacker simpler and more efficient.  

First draft as of now, but basic functionality is working. Still to do:

- [x] complete test with some docker images, checking that we get the same overlay (can be done with integration tests when adapting ducc to use these tools.
-  nested subcatalogs in the layers are not handled as of yet (not needed for the immediate future).

### DIFF
--- a/.github/workflows/ci_container_unpacker.yaml
+++ b/.github/workflows/ci_container_unpacker.yaml
@@ -63,9 +63,9 @@ jobs:
           ./ci/build_package.sh $PWD /tmp/build-ubuntu cvmfs
       - name: Install packages
         run: |
-          sudo apt-get -y remove "cvmfs*"
           sudo apt-get -y install /tmp/build-ubuntu*/*.deb
           sudo cvmfs_config setup
+          sudo systemctl stop autofs
           sudo systemctl start apache2
           sudo systemctl status apache2
       - name: Setup Go
--- a/.github/workflows/ci_container_unpacker.yaml
+++ b/.github/workflows/ci_container_unpacker.yaml
@@ -1,0 +1,98 @@
+name: container_unpacker
+on:
+  pull_request:
+    branches: [devel, cvmfs*]
+  push:
+  workflow_dispatch:
+
+jobs:
+  validate_unpacker:
+    runs-on: ubuntu-latest
+    steps:
+
+      - id: lsb-release
+        run: |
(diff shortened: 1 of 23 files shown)


DrDaveD:
I'm not really understanding how this works or how it helps make things simpler & more efficient.  Could you please give an example?

vvolkl:
> I'm not really understanding how this works or how it helps make things simpler & more efficient. Could you please give an example?

Sure. The unpacker works by first unpacking all the tarballs corresponding to the layers of the image to the `.layers` subdirectory,  and then creates the overlayed "flat image" in `.flat` and symlinks it to a human readable location, which is then typically given to apptainer to run. Until now the unpacker did the overlay in the following way: It would do a template transaction from the first layer to an intermediate directory. While this transaction was open, it would copy the contents of the second layer to the upperdir, so the result of the transaction would be intermediate directory 1 that is the  overlay of layers 1 and 2. It would then repeat this procedure with intermediate directory 1 and layer 3 to produce intermediate directory 2 until all layers. are overlaid. The intermediate directories are called "chains" and serve also as a cache for the n-th overlay. The procedure to produce the chains was called a "sneaky transaction".

In this PR I replace this operation with a tool that does the same thing by overlaying the catalogs of the layers directly, following the same logic as logic as overlayfs regarding whiteout files etc. The usage is something like 
```
cvmfs_server overlay -l /.layers/<path with hash of first layer>,/.layers/<path with hash of second/layer,... -d /.flat/<path to resulting flat image>
```
 It's already a pretty fast operation, but can also cache intermediate overlays on local disc, so crucially it no longer needs to save "chains" in the repository and basically allows us to delete the `.chains` subdir in unpacked.cern.ch. It also drastically cuts down on the number of transactions needed to add an image.
 
 To be seen how much this will reduce the repository size. We'll certainly need some usage quotas as well. For this part I'm planning to  to integrate the unpacker with harbor, which already supports everything necessary for proper operations.

vvolkl:
On second thought, I think I'll remove the caching from the first draft of this tool. It's a bit tricky to get right, and would actually really need an efficient way to attach catalogs at new locations. The new overlay is also so much faster that it's hardly needed. For https://registry.hub.docker.com/library/python:3.9 which has 8 layers the new swissknife_overlay takes 4s for creating the flat image, the old unpacker using template transaction takes 45s.

PR #4099, opened 2026-02-10
Add cvmfs_server ingestsql utility

vvolkl:
@andrey-utkin

### DIFF
--- a/cvmfs/server/cvmfs_server_util.sh
+++ b/cvmfs/server/cvmfs_server_util.sh
@@ -1241,7 +1241,7 @@ is_subcommand() {
     resign list info tag list-tags lstags check transaction enter abort snapshot    \
     skeleton migrate list-catalogs diff checkout update-geodb gc catalog-chown      \
     eliminate-hardlinks eliminate-bulk-hashes fix-stats update-info update-repoinfo \
-    mount fix-permissions masterkeycard ingest overlay merge-stats print-stats"
+    mount fix-permissions masterkeycard ingest ingestsql overlay merge-stats print-stats"
 
   for possible_command in $supported_commands; do
     if [ x"$possible_command" = x"$subcommand" ]; then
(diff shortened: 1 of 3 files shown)


cvmfs/server/cvmfs_server_util.sh
```diff
@@ -1415,6 +1415,28 @@ Supported Commands:
                   Extract the content of the tarfile inside the base directory,
                   in the same transaction it also delete the required folders.
                   Use '-' as -t argument to read the tarball from STDIN.
+<<<<<<< HEAD
```
andrey-utkin: conflict markers got committed into the code.

andrey-utkin:
Sorry for not responding for so long, even though I have looked at this earlier.

I am not sure why would such relatively simple wrapper is needed. Is it a convention that user is not supposed to run swissknife commands directly, and needs a cvmfs_server wrapper? I am a bit concerned about possible confusion from having two ways to run the same functionality, with slightly different behaviour. Also two places to update documentation about flags and behaviour.

I can look into integrating some specific conveniences into the original tool code.

andrey-utkin (commented):
see my other comment. regarding code, just this small issue.

vvolkl:
> Sorry for not responding for so long, even though I have looked at this earlier.
> 
> I am not sure why would such relatively simple wrapper is needed. Is it a convention that user is not supposed to run swissknife commands directly, and needs a cvmfs_server wrapper? I am a bit concerned about possible confusion from having two ways to run the same functionality, with slightly different behaviour. Also two places to update documentation about flags and behaviour.
> 
> I can look into integrating some specific conveniences into the original tool code.

Thanks for taking a look. Indeed, the convention is that cvmfs_swissknife is not user facing, but there is in most cases a corresponding cvmfs_server command that makes cli usage easier.  See `cvmfs_server ingest` vs `cvmfs_swissknife ingest`, `cvmfs_server snapshot` vs. `cvmfs_swissknife sync` etc. The server script typically reads the configuration from /etc/cvmfs and uses that to fill in some arguments to swissknife, and does some additional checks and sanitations. We've discussed a couple of times to rewrite these scripts and migrate more of them to C++, but that's something for the medium term future, so I think it's good to have this for ingestsql as well, I'm certainly happy if I have to type less when invoking it. 

Ingestsql of course has its own wrapper with the posix tools, but I think the server command can be helpful too.

PR #4108, opened 2026-02-13
Make busy snapshot during -a not be an ERROR

DrDaveD:
Stratum 1s that use snapshot -a end up getting many occurrences in the snapshot log of messages like this even though it is a very normal situation:
```
another update is in progress... aborting
ERROR from cvmfs_server snapshot!
```
Especially that ERROR message makes it more difficult to search for real errors in the log.  This PR turns "aborting" into "skipping" and removes the ERROR message.  It also removes the extraneous message in /var/log/messages.

(I thought there was probably an open issue about fixing this but I wasn't able to locate one).


vvolkl (approved):
Looks good to me, thanks Dave.

Issue #4118, opened 2026-02-25
Optimising garbage collection


chrisburr:
Opening this to group together my thoughts based on the chat I just had with @vvolkl.

I see a path towards optimising the GC in three steps:

- [x] Bulk S3 deletes with 1000 objects in a single POST instead of per-object DELETE requests (https://github.com/cvmfs/cvmfs/pull/4122)
- [ ]  Make the sweep phase concurrent so multiple catalog trees are swept in parallel using the existing thread pool (https://github.com/cvmfs/cvmfs/pull/4196)
- [ ]  Async/lock-free garbage collection where the expensive mark phase (catalog traversal) runs outside the lock, then the lock is briefly acquired only for a delta-mark and sweep

As mentioned I'll start opening PRs for this series of improvements.

Longer term if the above isn't enough I could imagine going further with:

- Fully asynchronous garbage collection where objects are only removed in the windows between transactions (e.g. the third step above, except able to catch up multiple times)
- Checkpointing or caching to allow garbage collection to be stopped and then continue from where it's left off

I'm not convinced these would be needed and it's not immediately obvious to me how to implement these in the current code.

DrDaveD:
@bbockelm Is also interested in optimizing CVMFS garbage collection.  In the last couple of weeks he wrote his own from scratch in go to try out some concepts and it was very much faster.  His code is [here](https://github.com/bbockelm/go-cvmfs-stratum1/tree/main/gc).

vvolkl:
> [@bbockelm](https://github.com/bbockelm) Is also interested in optimizing CVMFS garbage collection. In the last couple of weeks he wrote his own from scratch in go to try out some concepts and it was very much faster. His code is [here](https://github.com/bbockelm/go-cvmfs-stratum1/tree/main/gc).

Very interesting! It's certainly true that for big repository it's an issue to keep all hashes in memory. Even if it's faster, I think it's still worthwhile to look into approaches that make gc less blocking.

bbockelm:
It was a fun bit of work:
- lexicographically sort the entries in the catalogs, using a few tricks around external merge sort to ensure they're never fully in memory.
- lexicographically walk the filesystem tree, also cheap because of the directory structure.
- take advantage of the fact that comparing two lexicographically sorted streams is O(n).

For disk-backed origins, I was able to get the locked part to a minute or so (from what Dave mentioned were a few hours).

[...]

PR #4122, opened 2026-02-25
Batched S3 deletes

chrisburr:
This is the first step of https://github.com/cvmfs/cvmfs/issues/4118.

The basic idea is that it's more efficient to do a bulk delete operation via POST with up to 1000 objects instead of doing one DELETE request per object. See [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) for more details.

### DIFF
--- a/cvmfs/network/s3fanout.cc
+++ b/cvmfs/network/s3fanout.cc
@@ -24,6 +24,101 @@ using namespace std;  // NOLINT
 
 namespace s3fanout {
 
+/**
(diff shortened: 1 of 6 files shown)


vvolkl:
Looks good to me, thanks Chris! There's some updates to the test infrastructure for s3 I've been meaning to do, I'll see if that's possible to do before merging this. In any case it's good for 2.14.

vvolkl (approved):

PR #4128, opened 2026-02-27
Allow cvmfs_server resign to be done without .crt if .cvmfswhitelist exists

DrDaveD:
It was reported on Mattermost that a `cvmfs_server resign` requires a repository's `.crt` file to be present, which was surprising.  This surprise can be avoided by actually resigning an existing `.cvmfswhitelist` file rather than recreating it.

However, this does not avoid the need for the configuration of the repo to be present; only the `-w` option does that.  There was no hint about that in the usage info though, so this PR also adds that.

### DIFF
--- a/cvmfs/server/cvmfs_server_resign.sh
+++ b/cvmfs/server/cvmfs_server_resign.sh
@@ -100,7 +100,7 @@ cvmfs_server_resign() {
 
         local wl="$(get_upstream_config $CVMFS_UPSTREAM_STORAGE)/.cvmfswhitelist"
         local crt=/etc/cvmfs/keys/$name.crt
-        if [ ! -f "$wl" ] || [ -f $crt ]; then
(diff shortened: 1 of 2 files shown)


cvmfs/server/cvmfs_server_resign.sh
```diff
@@ -98,8 +98,14 @@ cvmfs_server_resign() {
 
       else
 
+        local wl="$(get_upstream_config $CVMFS_UPSTREAM_STORAGE)/.cvmfswhitelist"
+        local crt=/etc/cvmfs/keys/$name.crt
+        if [ ! -f "$wl" ] || [ -f $crt ]; then
```
vvolkl:
```suggestion
        if [ ! -f "$wl" ] || [ -f "$crt" ]; then
```
Quote (mainly for consistency)

vvolkl (approved):

PR #2545, opened 2020-04-23
Add a hidden `cvmfs_server eliminate-bulk-hashes` command

jblomer:
The command removes bulk hashes from files that also exist in chunked form.

PR #4129, opened 2026-02-27
Ignore bulk hashes

DrDaveD:
This makes catalog access for snapshot (pull), gc, and check ignore any bulk hashes when there are chunks available.  This cleans out all the old legacy bulk hashes from the early days of CVMFS without needing to run `cvmfs_server eliminate-bulk-hashes`.

The repository most affected is cms.cern.ch which has about 2T of these old bulk hashes.  I first tried using `cvmfs_server eliminate-bulk-hashes` on a copy of the repository and for some reason it was not successful in removing the old hashes.

The caveat with the approach in this PR is that if someone tries to replicate from a repository that has been cleaned out with this while using an older version of cvmfs-server, they will run into missing files.  There are no longer any existing clients that still use the old bulk hashes.  That was already the case when the `eliminate-bulk-hashes` command was added almost 6 years ago in #2545.

### DIFF
--- a/cvmfs/catalog_sql.cc
+++ b/cvmfs/catalog_sql.cc
@@ -1369,8 +1369,9 @@ SqlAllChunks::SqlAllChunks(const CatalogDatabase &database) {
                + StringifyInt(shash::kSuffixMicroCatalog) + " END "
                + "AS chunk_type, " + flags2hash + "," + flags2compression
                + "FROM catalog WHERE (hash IS NOT NULL) AND "
-                 "(flags & "
-               + StringifyInt(SqlDirent::kFlagFileExternal) + " = 0)";
+                 " (flags & "
+               + StringifyInt(SqlDirent::kFlagFileExternal |
+                              SqlDirent::kFlagFileChunk) + " = 0)";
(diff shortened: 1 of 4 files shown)


vvolkl:
Looks good to me. I wonder if it is worthwhile to add a config option that will revert to the old behavior. In case some problem pops up it saves people from downgrading.

DrDaveD:
How about just a `cvmfs_swissknife snapshot` option?  Then we could tell people to edit their /usr/bin/cvmfs_server if needed.

DrDaveD:
Or were you thinking about an environment variable?  Maybe that would be easy.

vvolkl:
Just getting a variable from environ is easiest, but means we cannot use tools like showconfig to easily debug a system, so for that reason I would prefer a proper `CVMFS_` config variable. It does mean adding an argument to swissknife sync / check as well unfortunately.

DrDaveD:
I don't expect to be able to get back to this for a few weeks so maybe you could add the option.

vvolkl:
> I don't expect to be able to get back to this for a few weeks so maybe you could add the option.

Done

DrDaveD (commented):
Looks good

PR #4133, opened 2026-03-04
golang: Remove vendor dirs from git

### LINKED ISSUE #4132: No longer commit vendored golang dependencies to repo
There are a few pending updates to do to the dependencies of the golang components, and I would prefer not checking those into the repository to decrease git noise.  There are [several arguments](https://victoriametrics.com/blog/vendoring-go-mod-vendor/#should-i-commit-the-vendor-directory) for and against commiting the vendor directory. Crucially, the CernVM-FS source tarball created for releases should still contain a copy of the vendored sources, so builds can still be done offline. With regards to compilation speeds, it should be fine to rely on the go module cache.

vvolkl:
Fixes #4132


vvolkl:
Split off the build system changes into #4137

PR #4134, opened 2026-03-04
Fix linkage of s3benchmark.cc

### LINKED ISSUE #4125: Main build system doesn't see header of vendored libarchive
Related but different observations to #4072 .
Apparently it's the fallout of the recent libarchive vendoring change #4064 .
I build current master (6168e392fabbc890c8ccdbf1e290f43827969366) in Fedora 43. It happens to have libarchive package installed but not libarchive-devel, so no header file.
My previously working build command now failed with `cvmfs/duplex_libarchive.h:12:10: fatal error: archive.h: No such file or directory`. I checked the externals_install dir and eventually found out about the recent transition.

From logs I see it built the library under _deps/, but apparently it is not making sure the main cvmfs cmake script finds archive.h header file.

I tried adding `-DBUILTIN_EXTERNALS=yes` but I suspect it doesn't matter.
```
...
-- External [LibArchive]: Mode OPTIONAL (Try System First)
-- Could NOT find LIBB2 (missing: LIBB2_LIBRARY LIBB2_INCLUDE_DIR)
-- Could NOT find LZ4 (missing: LZ4_LIBRARY LZ4_INCLUDE_DIR)
-- Checking for one of the modules 'libzstd'
-- Could NOT find ZSTD (missing: ZSTD_LIBRARY ZSTD_INCLUDE_DIR)
-- Building LibArchive from vendored sources.
...
[ 23%] Building CXX object cvmfs/CMakeFiles/cvmfs_server.dir/ingestion/item_mem.cc.o
In file included from /usr/local/src/cvmfs/cvmfs/ingestion/ingestion_source.h:18,
                 from /usr/local/src/cvmfs/cvmfs/upload.h:118,
                 from /usr/local/src/cvmfs/test/stress/s3benchmark.cc:10:
/usr/local/src/cvmfs/cvmfs/duplex_libarchive.h:12:10: fatal error: archive.h: No such file or directory
   12 | #include <archive.h>
      |          ^~~~~~~~~~~
compilation terminated.
```

Changing the environment to have libarchive-devel package, and thus the header, makes the compilation successful.

Good luck with overhauling the dependency vendoring code. But speaking as a former distro packager (Gentoo, also maintained inhouse deb packages), cmake scripts for dependency vendoring always look awful, each project writes their own but it's never state of the art, super verbose, wrong assumptions, things named wrongly, parameters not general enough, and often cmake itself pushes you the wrong way, by providing wrong examples, naming things confusingly, and not providing the right information or commands. It's just sad view almost always, it usually takes at least a hundred of lines of code to get what you can have almost for nothing. The constructive part of my rant would be to lean on pkgconfig data of the dependencies, and pkgconfig tooling available in cmake. It doesn't seem like the reworked libarchive vendoring code uses that at the moment.

vvolkl:
Fixes #4125

### DIFF
--- a/test/stress/CMakeLists.txt
+++ b/test/stress/CMakeLists.txt
@@ -29,6 +29,7 @@ target_link_libraries (s3benchmark
                        ${ZLIB_LIBRARIES}
                        ${OPENSSL_LIBRARIES}
                        ${VJSON_LIBRARIES}
+                       LibArchive::LibArchive
                        pthread
                        dl
 )


andrey-utkin:
The fix works, thanks

PR #4147, opened 2026-03-08
Reapply the fix of linkage of s3benchmark.cc

andrey-utkin:
Commit 9b0f21ab4f88 reverted the fix.

(cherry picked from commit 79e375ae4e3e8e400d49d307693a01de4b8787b9)

Fixes: 9b0f21ab4f88 ("golang: Prepare for removal of vendor dir, remove --mod=vendor (#4137)")


vvolkl (approved):
Good catch, my bad!

Issue #3888, opened 2025-06-24
chksetup fails to find executables on Fedora 42


vvolkl:
Fedora 42 symlinks `/usr/sbin -> /usr/bin`, so `sudo which cvmfs_config` yields `/usr/sbin/cvmfs_config` instead of `/usr/bin/cvmfs_config`, which confuses the logic that looks for other installed cvmfs files.

It's questionable if this check is really needed - and package managers for other platforms have already complained about it.  Probably it's enough to check that the executables are in the PATH and the same version.

vvolkl:
I'll add a workaround for 2.13.1 and review the checks in 2.14

PR #4148, opened 2026-03-08
cvmfs_config: get its own location correctly

andrey-utkin:
"Try to cd into the path I'm invoked with (if any), and regardless of success, print current directory" is not an adequate way to deduce script location. It is guaranteed to work only if invoked with absolute path, or if CWD in an actual script location.

On Fedora 43, execution goes as follows, when invoked as
- `sudo cvmfs_config chksetup` (used in test/src/001-chksetup)
- `bash cvmfs_config chksetup`
- `sudo bash cvmfs_config chksetup`
- `sudo bash -x cvmfs_config chksetup`

But not if invoked as
- `cvmfs_config chksetup`
- `/usr/bin/cvmfs_config chksetup`
- `bash -c 'cvmfs_config chksetup'`

```
+++ dirname cvmfs_config
++ cd .
++ pwd
+ SCRIPT_LOCATION=/usr/local/src/cvmfs/test ++ echo /usr/local/src/cvmfs/test
++ sed -e 's/^\(.*\)\/bin$/\1/'
+ INSTALL_BASE=/usr/local/src/cvmfs/test ++ '[' x/usr/local/src/cvmfs/test = x/usr ']'
++ echo /usr/local/src/cvmfs/test/sbin
+ SBIN_BASE=/usr/local/src/cvmfs/test/sbin
```

---8<---

I tried to change `test/src/001-chksetup` in a way to provoke the same error in the actions runner, for demonstration and solution testing purposes, but so far I failed at provoking it. I can confirm that this patch solves the issue in my Fedora 43 container where I run tests.

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Common configuration tasks for CernVM-FS
 
-SCRIPT_LOCATION=$(cd "$(dirname "$0")"; pwd)
+SCRIPT_LOCATION=$(dirname "$(command -v "$0")")
 INSTALL_BASE=$(echo "$SCRIPT_LOCATION" | sed -e 's/^\(.*\)\/bin$/\1/')
 # workaround for fedora 42
 INSTALL_BASE=$(echo "$INSTALL_BASE" | sed -e 's/^\(.*\)\/sbin$/\1/')


vvolkl:
Happy to merge, but  the better fix is to remove this alltogether. As far as I can recall, the location is only used to check that other cvmfs are located in certain paths relative to  cvmfs_config. That is not strictly needed and makes problems on some platforms that do unexpected things with bin and sbin, but otherwise work fine

vvolkl:
See also https://github.com/cvmfs/cvmfs/issues/3888

PR #4149, opened 2026-03-09
Prefetch files from bundle when trigger is opened.

christge-sft:
The approach here is to have an individual thread spawned on the spot that will do the prefetching.

There are two important entities here, the `BundleFileMgr` (that is iterating through the bundle file) and the `BundleMgr` (that is responsible to manage the fetcher thread).

With this PR https://github.com/cvmfs/cvmfs/pull/4049 and https://github.com/cvmfs/cvmfs/pull/4039 are now deprecated. I have tried to incorporate most of the remarks of @jblomer.

In case that performance is not enough, the `BundleMgr` is designed to easily go from a single thread to a pool of threads.


christge-sft:
@vvolkl as discussed, I'm merging this to the feature branch.

PR #4181, opened 2026-03-29
client: add CVMFS_REPOSITORIES_NOMOUNT param to help with syslog noise

vvolkl:
This PR implements a longstanding request to help with log noise from mount attempts on projects.cern.ch. This is a repository that can only be mounted on an internal network, but since it's in a search path for experiment software, there are frequent mount attempts when using autofs in an external network, which cannot easily be avoided by fixing the software. This causes lots of noise in the syslogs since each failed mount attempt is logged like:

```
cvmfs2[2714102]: (no.such.repo) failed to download repository manifest (2 - malformed URL)
cvmfs2[2714102]: (no.such.repo) Failed to initialize root file catalog (16 - file catalog failure)
 ```
 
 We've thought about several other solutions and workarounds, including configuring the squids to redirect external requests to another server url that serves an existing but empty repo, but ultimately I think it's best to do it explicitly and fail early in the loader.
 
 I'm still moderately worried about getting tickets where this feature is mistaken for a cvmfs issue, but I've tried to add clear debug log messages for the problem. Care should be taken by the admins to only deploy this config site-locally in the right environment ( in order to not ignore projects.cern.ch in the cern network for example). Please also add appropriate comments so it is not copied blindly.

### DIFF
--- a/cvmfs/loader.cc
+++ b/cvmfs/loader.cc
@@ -768,14 +768,14 @@ int FuseMain(int argc, char *argv[]) {
     premount_fuse_ = false;
   }
 
-  if (options_manager->GetValue("CVMFS_IGNORED_REPOS", &parameter)
+  if (options_manager->GetValue("CVMFS_REPOSITORIES_NOMOUNT", &parameter)
       && !parameter.empty()) {
     const vector<string> ignored_repos = SplitString(parameter, ',');
     for (unsigned i = 0, s = ignored_repos.size(); i < s; ++i) {
       if (Trim(ignored_repos[i]) == *repository_name_) {
         LogCvmfs(kLogCvmfs, kLogStdout,
                  "CernVM-FS: mount attempt for %s ignored "
-                 "(listed in CVMFS_IGNORED_REPOS)",
+                 "(listed in CVMFS_REPOSITORIES_NOMOUNT)",
                  repository_name_->c_str());
         delete options_manager;
         fuse_opt_free_args(mount_options);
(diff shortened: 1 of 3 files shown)


vvolkl:
So with this current patch we actually don't even attempt to mount any repos in the ignored list. Maybe it's worthwhile to have a mode where the mount is still attempted, and only the log message is suppressed on error? That way this config could be set, but would still work at CERN (in this example). There's a trade-off here between efficiency (failing earlier is better) and having a config that works in a wider scope.

deesto:
Just saw your MM post @vvolkl ... if you can accurately detect the domain on which the client requests to mount the projects repo, my suggestion would be not to attempt the mount, and return your current 
[mount attempt ignored message](https://github.com/cvmfs/cvmfs/pull/4181/changes#diff-fb0d29a2d9228dc98918da423936eb6c0407171da8d68194c6ca00477d89404dR87) or otherwise inform the client that this is an invalid request from this client/domain.

vvolkl:
> Just saw your MM post @vvolkl ... if you can accurately detect the domain on which the client requests to mount the projects repo, my suggestion would be not to attempt the mount, and return your current [mount attempt ignored message](https://github.com/cvmfs/cvmfs/pull/4181/changes#diff-fb0d29a2d9228dc98918da423936eb6c0407171da8d68194c6ca00477d89404dR87) or otherwise inform the client that this is an invalid request from this client/domain.

Thanks for taking a look. There is no detection of the domain at the moment, this feature relies on the configuration parameter being set in the right scope ( a local config file deployed only at the right site ). Hence my comment about making sure it is not being copied. I think any attempt to detect the domain is in the end basically equivalent to actually attempting the mount, at least I cannot think of a good solution.

DrDaveD (approved):

PR #4192, opened 2026-04-02
selinux: provide a default context if unset

### LINKED ISSUE #4190: service container provided mounts have `?` selinux type
```shell
podman run --rm -it -e CVMFS_HTTP_PROXY='http://squid.fnal.gov:3128;DIRECT'  -e CVMFS_REPOSITORIES='fermilab.opensciencegrid.org,' --cap-add SYS_ADMIN   --device /dev/fuse   --volume /cvmfs:/cvmfs:shared,z   cvmfs/service:latest
```

Then:
```shell
$ ls -alZ /cvmfs/fermilab.opensciencegrid.org/
total 33
drwxr-xr-x 9 49375 9125 ? 4096 Aug 12  2014 .
drwxr-xr-x 6 49375 9125 ? 4096 Sep  2  2025 cosmiq
-rw-r--r-- 1 49375 9125 ?  374 Nov 11  2024 .cvmfsdirtab
drwxr-xr-x 3 49375 9125 ? 4096 Mar  3 15:34 ebd
drwxrwsr-x 8 49375 9125 ? 4096 Apr 29  2025 genie
drwxr-xr-x 5 49375 9125 ? 4096 May 21  2025 mars
drwxr-xr-x 7 49375 9125 ?   31 Mar 18  2020 noble
drwxr-xr-x 3 49375 9125 ? 4096 Dec  4  2023 packages
drwxr-xr-x 5 49375 9125 ? 4096 Aug 10  2021 products
```

I'd expect the default selinux context to be one of `system_u:object_r:fusefs_t:s0`, `system_u:object_r:container_file_t:s0`, `system_u:object_r:container_ro_file_t:s0`, or `system_u:object_r:public_file_t:s0`.

Personally, I like `system_u:object_r:fusefs_t:s0`.

The host doesn't have `cvmfs-selinux` installed because I'm trying for a pure container workflow.

This persists regardless of use with a disk volume mounted under `/var/lib/cvmfs` or not.

jcpunk:
The `default_t` context is a suitable non-value.  It should map into any selinux policy with limited consequences and no realy change in current behavior.

The major limitation is `default_t` doesn't actually permit execution of confined processes by default.

If we don't need RHEL7 compat, the context should default to `fusefs_t` to match the policy in modern systems.

I'm not clear there are any confined selinux users of this container on RHEL7, since it won't allow access.  If this is desired, `usr_t` might be more appropriate as that is the default context for `/opt` and its content.

The PR currently implements "just enough" to make the undefined context errors go away, but should be modified to select an actually useful default context.

Fixes: https://github.com/cvmfs/cvmfs/issues/4190

### DIFF
--- a/cvmfs/cvmfs.cc
+++ b/cvmfs/cvmfs.cc
@@ -1928,7 +1928,12 @@ static void cvmfs_getxattr(fuse_req_t req, fuse_ino_t ino, const char *name,
   if (!magic_xattr.IsNull()) {
     attribute_result = magic_xattr->GetValue(attr_req_page, xattr_mode);
   } else {
-    if (!xattrs.Get(attr, &attribute_result.second)) {
+    /* security.selinux is excluded from catalog hashing and storage (see
+     * HashMeta). Synthesize a host-local default at query time rather than
+     * storing policy-dependent labels in the content-addressed catalog. */
+    if (strcmp(attr.c_str(), "security.selinux") == 0) {
+      attribute_result.second = "system_u:object_r:fusefs_t:s0";
+    } else if (!xattrs.Get(attr, &attribute_result.second)) {
       fuse_reply_err(req, ENOATTR);
       return;
     }


vvolkl:
Thanks for tackling this issue! I would assume that we don't need RHEL7 compat from now on. We can designate 2.13.3 as a stable version for RHEL7 should it be needed.

jcpunk:
> We can designate 2.13.3 as a stable version for RHEL7 should it be needed.

I'll update the patch to set to `fusefs_t`. It doesn't break anything on RHEL7, but rather doesn't fix anything on RHEL7 either.

jcpunk:
I do not understand the test failures...

vvolkl:
> I do not understand the test failures...

Right, the integration tests compare the ouput of `ls -l` of content published to a CVMFS repository to a reference directory, so the reference would need to be updated. Maybe this patch is a bit too general in setting the selinux xattrs though, it might be better to add a flag based on mount or config option to this patch - Let me think about it a bit more.

jcpunk:
The missing `.` at the end puzzles me as that is the selinux metadata signifier and that is explicitly what I'm trying to add.

From test.log:
```
-drwxr-xr-x. runner runner foo/
+drwxr-xr-x runner runner foo/
```

My goal with this patch set is to add `fusefs_t` when no other attribute is present so there is always a context...

vvolkl:
> The missing `.` at the end puzzles me as that is the selinux metadata signifier and that is explicitly what I'm trying to add.
> 
> From test.log:
> 
> ```
> -drwxr-xr-x. runner runner foo/
> +drwxr-xr-x runner runner foo/
> ```
> 
> My goal with this patch set is to add `fusefs_t` when no other attribute is present so there is always a context...

It's the other way around, no? You are adding the `.`, the reference doesn't have it, so that's somewhat expected. As I said, I think it makes sense to put this behind a mount option, so it can be enabled specifically for the service container, or when selinux is actually enabled on the host.

jcpunk:
My understanding of the behavior is that, on non-selinux systems, the selinux labels aren't queried so it shouldn't cause any issues to have them present.

PR #4196, opened 2026-04-08
Add concurrency support to sweep phase of GC

chrisburr:
Thread-safe primitives and concurrent sweep for garbage collection (step 2 of #4118).

Two commits, intended to be reviewed in order:
1. Adds thread-safety primitives (`ContainsOrInsert`, `ShardedHashFilter`, atomic counters, `log_mutex_`).
2. Flips callback serialization off during sweep, with three new parallel-specific tests.

### DIFF
--- a/cvmfs/garbage_collection/hash_filter.h
+++ b/cvmfs/garbage_collection/hash_filter.h
@@ -125,4 +125,60 @@ class SmallhashFilter : public AbstractHashFilter {
   bool frozen_;
 };
 
+
+//------------------------------------------------------------------------------
+
+
+/**
+ * Thread-safe AbstractHashFilter backed by MultiHash with per-shard locks.
+ */
+class ShardedHashFilter : public AbstractHashFilter {
+ public:
+  ShardedHashFilter() {
+    // zero_element is MD5("unobtanium")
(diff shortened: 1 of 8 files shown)


cvmfs/garbage_collection/hash_filter.h
```diff
@@ -125,4 +125,60 @@ class SmallhashFilter : public AbstractHashFilter {
...
+/**
+ * Thread-safe AbstractHashFilter backed by MultiHash with per-shard locks.
+ */
+class ShardedHashFilter : public AbstractHashFilter {
+ public:
+  ShardedHashFilter() {
+    // zero_element is MD5("unobtanium")
+    const shash::Any zero_element(
+        shash::kMd5, shash::HexPtr("d61f853acc5a39e01f3906f73e31d256"));
+    // 255 is MultiHash's uint8_t shard-index limit; more shards reduce
```
chrisburr, line 141: For completeness, I did some microbenchmarking of this and found that it was always better to use more shards so long as the number of threads is >=2 and even for the single threaded case the impact was minimal.

There is probably room for further performance improvements by modifying `shash` but I think this is good enough as-is.

PR #4199, opened 2026-04-10
Fix deadlock with batched deletes on s3

vvolkl:
@chrisburr The batched deletes introduced in #4122 clash with the `jobs_in_flight_` logic and deadlocked gc - fix is to count per batch instead of per key.  

Also shows that we need to run integration tests against s3 - I'm about to add a new containerized setup using garage for that.


chrisburr:
Thanks for debugging this and sorry for the bug! Would you like any help with the integration test?

chrisburr:
I've made a PR targeting this branch which adds a regression test and fixes what I think is a bug: https://github.com/vvolkl/cvmfs/pull/6

PR #4200, opened 2026-04-10
fix: handle unbounded RLIMIT_NOFILE in CloseAllFildes

chrisburr:
On macOS `sysconf(_SC_OPEN_MAX)` returns `LONG_MAX` when `RLIMIT_NOFILE` is `unlimited`. Truncating that to `int` yielded a negative value, causing `CloseAllFildes` to fail.

This PR widens `max_fd` to `long` and macOS now has the same fallback rather than always iterating.


vvolkl (approved):
Good catch, thanks!

PR #4209, opened 2026-04-16
fix json -> nlohmann_json in cmake, so find_package works with system

vvolkl:


DrDaveD (approved):

DrDaveD:
I confirm that with this patch 2.14.0-pre3 can be built on osg koji (where network access is cut off).

PR #4227, opened 2026-05-01
Reduce output for non-interactive status, and use proxy when contacting stratum 1s

DrDaveD:
We had a site cause hits direct on stratum 1s from worker nodes because they had a script running `cvmfs_config status repo` on all the mounted repos and the status command did not use the local proxy.  Change the status command to not print the extra details when run non-interactively and also use the proxy for its queries to stratum 1s.

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -1214,8 +1214,12 @@ cvmfs_status() {
       RETVAL=1
       continue
     fi
-    export http_proxy="$(attr -qg proxy $mountpoint)"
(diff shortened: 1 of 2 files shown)


vvolkl (changes requested):
Thanks. I think test 071 needs to be adapted.

DrDaveD:
> Thanks. I think test 071 needs to be adapted.

Done.

vvolkl:
@DrDaveD I think there was the danger of DIRECT being used as a hostname. Added a commit with a possible fix, please check and adapt if necessary.

DrDaveD:
You're right, good find & fix.

vvolkl:
I've added a -v flag, so this command can be tested from a script without emulating an interactive session.

Issue #4226, opened 2026-04-29
Segfault on cvmfs client when upgrading OSG RPM to 2.14.0~pre3-2.1.osg24up.el9


jthiltges:
We're doing initial testing of cvmfs 2.14 at Nebraska for an upcoming OSG release ([SOFTWARE-6332](https://opensciencegrid.atlassian.net/browse/SOFTWARE-6332)) with the assistance of @DrDaveD . We encountered a segfault in cvmfs2 when upgrading an active client:
- from `2.13.3-1.osg24.el9`
- to `2.14.0~pre3-2.1.osg24up.el9`

The RPM upgrade caused 7x /usr/bin/cvmfs2 processes to segfault and the associated /cvmfs FUSE mounts to fail with “Transport endpoint is not connected”. After manually unmounting the /cvmfs filesystems, autofs remounted them, and new jobs appear to be running OK.

Debug info:
- cvmfs_config bugreport: [cvmfs-bugreport_2026-04-29_red-c0821.unl.edu.tar.gz](https://github.com/user-attachments/files/27214394/cvmfs-bugreport_2026-04-29_red-c0821.unl.edu.tar.gz)
- `coredumpctl info` of one of the processes: [coredumpctl-info-3952922.txt](https://github.com/user-attachments/files/27214395/coredumpctl-info-3952922.txt)

I'm not sure if it's relevant, but `info proc mappings` of one of the corefiles showed the process had both `libcvmfs_fuse3_stub.so.2.13` and `libcvmfs_fuse3.so.2.14.0` loaded. [maps-3952922.txt](https://github.com/user-attachments/files/27214402/maps-3952922.txt)

vvolkl:
Thanks @jthiltges for the report. To me it looks exactly like the problem https://github.com/cvmfs/cvmfs/pull/4062 was trying to prevent, but apparently doesn't. Investigating.

vvolkl:
Right, #4062 won't help because unlike for the fuse processes, the quota manager doesn't even get hotpatched during a reload, it just stays on the old version. Instead, this looks like a concurrency bug due to non-atomic writes over the pipe.

DrDaveD:
No, the quota manager does get restarted during reload.  I verified it by installing 2.13.3, freshly mounting one repo which also mounts the config repo, waiting a minute, and doing `cvmfs_config reload`.  The `__cachemgr__` processes then show that they started the later minute.

[...]

PR #4234, opened 2026-05-06
Make LRU commands atomic to avoid race conditions

christge-sft:
With the addition of the open file aware LRU commands, we've seen race conditions happening cause the cvmfs process to crash https://github.com/cvmfs/cvmfs/issues/4226.

This is probably due to the fact that in the current implementation a single command would be completed with multiple pipe writes. With this commit we pass the extra information through the command description following the example of the `DoInsert()`.

### DIFF
--- a/cvmfs/quota_posix.cc
+++ b/cvmfs/quota_posix.cc
@@ -425,8 +425,8 @@ PosixQuotaManager *PosixQuotaManager::CreateShared(
     return NULL;
   }
   LogCvmfs(kLogQuota, kLogDebug,
-           "new cache manager pid: %d protocol revision %d",
-           new_cachemgr_pid, QuotaManager::kProtocolRevision);
+           "new cache manager pid: %d protocol revision %d", new_cachemgr_pid,
+           QuotaManager::kProtocolRevision);
   quota_mgr->SetCacheMgrPid(new_cachemgr_pid);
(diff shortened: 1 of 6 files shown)


cvmfs/quota_posix.cc
```diff
@@ -1377,9 +1400,7 @@ void *PosixQuotaManager::MainCommandServer(void *data) {
 
     // Set Cleanup Policy
     if (command_type == kSetCleanupPolicy) {
-      bool policy = false;
-      ReadPipe(quota_mgr->pipe_lru_[0], &policy, sizeof(bool));
-      quota_mgr->cleanup_unused_first_ = policy;
+      quota_mgr->cleanup_unused_first_ = (description_buffer[0] == 'S') ? true : false;
```
vvolkl: Is `description_buffer[0]` correct here? Shouldn't it be `description_buffer[kMaxDescription * num_commands]`?

vvolkl (changes requested):

cvmfs/quota_posix.cc
christge-sft: Hmm what you've spotted is right @vvolkl ! Same for the mountpoint registration.

cvmfs/quota_posix.cc
vvolkl: `buf` needs to be NULL terminated, right?

PR #4235, opened 2026-05-06
Zstd compression

andrey-utkin:
This PR picks up where @HereThereBeDragons "blueprint" branch has left off. https://github.com/cvmfs/cvmfs/pull/3667

From now on, CVMFS only gets told which compression algorithm to use to actually compress and store data. It doesn't need to get told how to *decompress*: for directory entries it always uses catalog metadata, and for metadata outside of catalogs it can reliably guess by the first byte of the downloaded object. **The goal is for there to be no such thing as failure due to misconfiguration**, on either publisher or client side. (Except for the case of client being not built with support of the compression schemes used.) See new test 823 for demonstration of this new flexibility: https://github.com/andrey-utkin/cvmfs/blob/12b21edd75e21252685d91a821ceda561d12155b/test/src/823-compress-alg-migration/main

ci_ubuntu workflow run results (green):
https://github.com/andrey-utkin/cvmfs/actions/runs/25461404663/job/74704242998

The object-oriented interface for Compressor and Decompressor classes is mostly as it was in the "blueprint" branch. I see an upside in its integration with core Sink class, but I must admit the I/O with these classes is was not easy to learn and to debug. I am open to any suggestions regarding changing these interfaces, including recombination with current compression.h stuff.


andrey-utkin:
These are ci_ubuntu CI results for the same changes but with changed compile-time setting for default compression:

https://github.com/andrey-utkin/cvmfs/actions/runs/25465156204/job/74716557708 (zstd)

https://github.com/andrey-utkin/cvmfs/actions/runs/25465169002/job/74716597849 (none)

cvmfs/compression/compressor_echo.cc
```diff
@@ -0,0 +1,182 @@
...
+  }
+  input->SetIdxInsideChunk(input->GetIdxInsideChunk() + written);
+  if (!input->HasInputLeftInChunk() && !input->has_chunk_left() && flush) {
+    return kStreamEnd;
+  } else {
+    return kStreamContinue;
+  }
+}
+
+#if 0
```
christge-sft: Hi @andrey-utkin This is an example of a dead code section.

PR #4239, opened 2026-05-08
test/run.sh: shuffle the test cases order

andrey-utkin:
This is to stress the assumptions that the tests do not have dependency on each other and that the cleanup after each test is sufficient.

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -207,7 +207,7 @@ if [ $test_exclusions -ne 0 ]; then
   shift # get rid of '--'
 fi
 
-testsuite="$@"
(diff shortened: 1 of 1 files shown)


andrey-utkin:
This results in test `108-global-config` failing:

https://github.com/andrey-utkin/cvmfs/actions/runs/25526703994

vvolkl:
I think this is a nice option to have, but am not convinced it should be the default behavior. The test suite has grown a bit organically and the order isn't perfect, but It still makes sense to run 000-dummy and 001-chksetup before some of the more specific and involved tests.

Issue #4236, opened 2026-05-07
High RSS with large number of open directory fds.


mharvey-jt:
When  a large number (1000s) of directory handles are held open on directories with moderately large entry counts (1000s+) the client RSS can become very high - GBs - apparently due to inefficiency in the sparsehash used to cache the listings.

vvolkl:
Yes, I'm measuring 68kb per open handle. We've also recently had a report of 50GB+ used by cvmfs on a very big node that did several recursive directory listings, so that can add up, but will also be released on close. From a first brief investigation there are several possible optimizations, but nothing that is completely straightforward. Except that we could shrink the buffer when opendir completes, since

https://github.com/cvmfs/cvmfs/blob/29ba451e42b7b6d9fae998a7fba887011cdf3fa7/cvmfs/bigvector.h#L83

likely overshoots. I'll investigate the other options further and propose something.

PR #4243, opened 2026-05-11
Shrink dir listing buffer

andrey-utkin:
This is an in-situ fix for the issue of high memory usage with open dir fds: https://github.com/cvmfs/cvmfs/issues/4236

At the cost of one more reallocation and copying, we'll avoid holding onto unneeded headroom which we tend to have due to DoubleCapacity() being used to grow BigVector.

Passes the existing tests, but the memory footprint impact is yet to be measured, and the scenario is yet to be covered with a test. Going to append these when ready.


andrey-utkin:
Hm, with the alleged "fix" the test still passes and shows that RSS growth is "lean" - exactly by BYTES_PER_DIRENT per dirent.

PR #4246, opened 2026-05-19
Suggestions on the pool of fetchers version

christge-sft:
1. I change the pool of fetchers from a `std::vector<pthread_t>*` to a `std::vector<std::unique_ptr<pthread_t>>`
The motivation here is that there is no need to keep the pool vector via a pointer (minor) 

2. I changed the pool creation policy.
In the old version first we'ld default construct threads for the entire pool, and in case of an initialization failure we'ld default destruct what was not initialized properly. 
Now I just increment the pool with the properly initialized fetcher threads.

+ Extra:
Determines if feature is used during `Init()`. This should make the decision on whether we should look for a trigger file in the first place faster (important when feature is disabled).

### DIFF
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -134,7 +134,7 @@ BundleMgr::BundleMgr(MountPoint *mp, const PathString &path)
     }
   }
 
-  SpawnFetcher();
+  SpawnFetcherPool();
 }
 
 void BundleMgr::Fetch() {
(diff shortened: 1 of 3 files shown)


christge-sft:
In C++20 there is this nice [likely/unlikely](https://en.cppreference.com/cpp/language/attributes/likely) compiler attribute introduced. We could make to a zero overhead for the default `cvmfs_open()`  just by declaring the prefetching decision as `[[ unlikely ]]`

vvolkl (approved):

Issue #4249, opened 2026-05-24
Latest curl (8.20) breaking CVMFS


ktht:
Initially reported [in the comments](https://aur.archlinux.org/packages/cvmfs) of the CVMFS package on AUR, it looks like the latest curl build ([8.20-7](https://gitlab.archlinux.org/archlinux/packaging/packages/curl/-/tree/8.20.0-7?ref_type=tags) to be exact) causes CVMFS to hang indefinitely when trying to list a directory. The exact commit in curl that that broke CVMFS is [`39036c9`](https://github.com/curl/curl/commit/39036c90216e059bbee64b86b198eae3135e0cda) in the curl repo. Interestingly, CVMFS hangs only when running in daemon mode (i.e., when it's mounted with `mount -t cvmfs ...`), but there are no issues when running it in foreground (i.e., `cvmfs2 -f ...`). Problems also go away when rebuilding curl with `--disable-threaded-resolver` flag and linking it to CVMFS libs.

Stacktrace of the process that gets stuck (when running, e.g., `ls /cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase`):

```gdb
Thread 9 (Thread 0x7f5e6ffff6c0 (LWP 842279) "cvmfs2"):
#0  0x00007f5e822a0a52 in ?? () from /usr/lib/libc.so.6
#1  0x00007f5e82294abc in ?? () from /usr/lib/libc.so.6
#2  0x00007f5e82294e07 in ?? () from /usr/lib/libc.so.6
#3  0x00007f5e8229a00d in ?? () from /usr/lib/libc.so.6
#4  0x00007f5e81278a73 in Curl_thread_join (hnd=0x56400a77f9f8) at curl_threads.c:90
#5  0x00007f5e8130a646 in thrdpool_join_zombies (tpool=0x56400a76cdf8) at thrdpool.c:207
#6  0x00007f5e8130abcd in Curl_thrdpool_destroy (tpool=0x56400a76cdf8, join=true) at thrdpool.c:347
#7  0x00007f5e8130b7d3 in thrdq_unlink (tqueue=0x56400a770bb8, locked=true, join=true) at thrdqueue.c:183
#8  0x00007f5e8130bad4 in Curl_thrdq_destroy (tqueue=0x56400a770bb8, join=true) at thrdqueue.c:253
[...]

PR #4252, opened 2026-05-25
Avoid reusing curl easy handle after fork

ktht:
Addresses https://github.com/cvmfs/cvmfs/issues/4249

Tested with curl 8.20 (which is the version where we'd run into issues) as well as with curl 8.19 on the latest CVMFS release (version 2.13.3).

### DIFF
--- a/cvmfs/network/download.cc
+++ b/cvmfs/network/download.cc
@@ -645,7 +645,7 @@ void *DownloadManager::MainDownload(void *data) {
                                    &still_running);
         } else {
           // Return easy handle into pool and write result back
-          download_mgr->ReleaseCurlHandle(easy_handle, true);
+          download_mgr->ReleaseCurlHandle(easy_handle, true /* allow_reuse */);
 
           DataTubeElement *ele = new DataTubeElement(kActionStop);
           info->GetDataTubePtr()->EnqueueBack(ele);
--- a/cvmfs/network/download.cc
+++ b/cvmfs/network/download.cc
@@ -2020,7 +2020,12 @@ Failures DownloadManager::Fetch(JobInfo *info) {
       }
     } while (VerifyAndFinalize(retval, info));
     result = info->error_code();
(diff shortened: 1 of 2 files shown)


cvmfs/network/download.cc
```diff
@@ -645,7 +645,7 @@ void *DownloadManager::MainDownload(void *data) {
                                    &still_running);
         } else {
           // Return easy handle into pool and write result back
-          download_mgr->ReleaseCurlHandle(easy_handle);
+          download_mgr->ReleaseCurlHandle(easy_handle, true);
```
jblomer:
```suggestion
          download_mgr->ReleaseCurlHandle(easy_handle, true /* allow_reuse */);
```

cvmfs/network/download.cc
```diff
@@ -2020,7 +2020,7 @@ Failures DownloadManager::Fetch(JobInfo *info) {
       }
     } while (VerifyAndFinalize(retval, info));
     result = info->error_code();
-    ReleaseCurlHandle(info->curl_handle());
+    ReleaseCurlHandle(info->curl_handle(), false);
```
jblomer:
```suggestion
    // Prevent the handle from being added to the idle list, which
    // avoids that it is mistakenly picked up by the multi handle later
    // in multi-threaded context. This, in turn, would fail to wait for
    // for resolver threads that meanwhile vanished due to a fork()
    // in Daemonize()
    ReleaseCurlHandle(info->curl_handle(), false /* allow_reuse */);
```

jblomer (approved):
Looks good to me. I guess we are good with respect to the vanishing resolver threads because they should be sleeping at the time of fork().

I'll leave merging to @vvolkl

vvolkl (approved):

vvolkl:
Thank you for this nice contribution! We're planning to upgrade curl also in our provided packages, so that's very helpful.

Issue #4253, opened 2026-06-01
Tests using inflate_file() are inefficient and may time out


andrey-utkin:
Is there any reason to not replace the logic of "use this little poetry text to fill up this megabytes-big data vessel, with control loop in bash" with `dd if=/dev/urandom` or something?

I can see the odd resulting size as a potential feature. Occasional ascii-ness shouldn't be of importance. Anything else?

This happens repeatedly.
```
++ stat -c %s big_file/1megabyte
+ '[' 449820 -lt 1000000 ']'
+ cat /bin/ls
++ stat -c %s big_file/1megabyte
+ '[' 449869 -lt 1000000 ']'
+ cat /bin/ls
++ stat -c %s big_file/1megabyte
++ stop_test_tree HUP
++ local signal=HUP
++ '[' -z 240 ']'
++ kill -s HUP -- -240
++ sleep 1
[1]+  Hangup                  bash -c '
	set -e &&
	set -o pipefail &&
	set -x &&
	set -u &&
	cd "$4" &&
	. ./test_functions &&
	. "$1/main" &&
	cd "$2" &&
	cvmfs_run_test "$3" "$4/$1"
	retval=$?
	mangle_test_retval "$retval"
	exit $?
      ' run_test_child "$1" "$2" "$3" "$4"
++ kill -KILL -- -240
++ true
++ wait 240
++ true
++ exit 124
execution took 301.020 seconds
Testcase timed out
```

```
 # grep -RnIl 'inflate_file[(][)]' . | sort
./510-filechunks/main
./511-removefilechunks/main
./512-movechunkstonested/main
./513-mergechunksfromnested/main
./514-changechunkedfile/main
./516-createsnapshotadvanced/main
./521-createmultiplesnapshots/main
./528-recreatespoolarea/main
./530-recreatespoolarea_defaultkey/main
./542-storagescrubbing/main
./543-storagescrubbing_scriptable/main
./557-basicgarbagecollect/main
./558-garbagecollectnestedcatalogs/main
./559-garbagecollectnamedtags/main
./561-garbagecollectmultihash/main
./573-garbagecollecttimestamp/main
./574-garbagecollecttimestampwithtags/main
./576-garbagecollectstratum1/main
./581-snapshotgarbagecollectedrepo/main
./660-store_publish_statistics/main
./670-listreflog/main
```

christge-sft:
Hello Andrey,

[...]

PR #4255, opened 2026-06-02
test: inflate_file() performance improvement

andrey-utkin:
See https://github.com/cvmfs/cvmfs/issues/4253

So I replaced "append small files into bigger file" loop with something faster, specifically `dd if=/dev/urandom bs=10000`. I noticed that `dd` with fixed `count` and `bs` may not always produce exactly the expected size in bytes, so I added assertions for the final file size. But avoidance of a loop with `stat(1)` on every lap is a big part of performance optimization.

Test 660 was a pickle and I restored it to produce exactly the same contents, but faster. I don't see how to tweak the reference values, because the resulting value actually changes every run - it's 14, 15 or 16. I am surprised to see content-dependent chunking - that's something new for me.

```
+ expected_value='expected_values_5[7]'
+ '[' x8 '!=' x15 ']'
+ echo '*** Test 5.7 - n_chunks_added -               (15 vs ref:8) FAILED!'
*** Test 5.7 - n_chunks_added -               (15 vs ref:8) FAILED!
+ return 100
```

I believe it's better for quality assurance to exercise the software on random data, so I prefer to not do the treatment I did to test 660 to all the other ones.

### DIFF
--- a/test/src/510-filechunks/main
+++ b/test/src/510-filechunks/main
@@ -91,7 +91,7 @@ cvmfs_run_test() {
   produce_files_in $repo_dir || return 3
 
   echo "putting exactly the same stuff in the scratch space for comparison"
-  produce_files_in $reference_dir || return 4
+  cp -a "$repo_dir"/* "$reference_dir" || return 4
 
   echo "creating CVMFS snapshot"
   publish_repo $CVMFS_TEST_REPO || return $?
--- a/test/src/510-filechunks/main
+++ b/test/src/510-filechunks/main
@@ -3,17 +3,6 @@ cvmfs_test_name="Creating File Chunks"
 cvmfs_test_autofs_on_startup=false
 cvmfs_test_suites="quick"
 
-inflate_file() {
-  local destination_file=$1
-  local source_file=$2
-  local desired_file_size=$3
-
-  touch $destination_file
-  while [ $(stat -c %s $destination_file) -lt $desired_file_size ]; do
-    cat $source_file >> $destination_file
-  done
-}
-
 produce_files_in() {
 	local working_dir=$1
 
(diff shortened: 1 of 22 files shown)


andrey-utkin:
This breaks test 514. I don't understand what the failing assertion relies on, and what's the mental model we could use to update the assertion.

```
+ local big_file=50megabyte
++ echo '   ATTACH '\''0e1751ebc3640209fe4b4289cc6ce3612338f330C.uncompressed'\'' AS old;   ATTACH '\''dad902048df178a20fb5528720e3b44a61c2ac77C.uncompressed'\'' AS new;   CREATE TEMPORARY TABLE old_offsets (rownum INTEGER PRIMARY KEY AUTOINCREMENT, offset INTEGER, size INTEGER);   CREATE TEMPORARY TABLE new_offsets (rownum INTEGER PRIMARY KEY AUTOINCREMENT, offset INTEGER, size INTEGER);   INSERT INTO old_offsets (offset, size) SELECT offset, size FROM old.chunks WHERE md5path_1 = (SELECT md5path_1 FROM old.catalog WHERE name = '\''50megabyte'\'') ORDER BY offset;   INSERT INTO new_offsets (offset, size) SELECT offset, size FROM new.chunks WHERE md5path_1 = (SELECT md5path_1 FROM new.catalog WHERE name = '\''50megabyte'\'') ORDER BY offset;   SELECT new.size - old.size FROM old_offsets AS old, new_offsets AS new WHERE old.rownum = new.rownum GROUP BY new.size - old.size ORDER BY count(new.size - old.size) DESC LIMIT 1;'
++ sqlite3
+ most_common_size_difference=1485201
+ '[' 1485201 -eq 0 ']'
+ return 7
[1]+  Exit 7                  bash -c '
```


https://github.com/cvmfs/cvmfs/blob/devel/test/src/514-changechunkedfile/main#L115-L128

PR #4261, opened 2026-06-06
test: do not use $USER in tests

andrey-utkin:
Most tests use CVMFS_TEST_USER which the test harness itself sets. USER is a standard system variable but it may be unset, particularly in container environments, and the test harness doesn't set it.

This change brings the few divergences from the standard practice into uniformity.

The entire change was generated by the following command:

 grep -l '[$]USER\>' test/src -RnI | xargs sed -i -e 's:[$]USER\>:$CVMFS_TEST_USER:g'

Co-authored-by: sed
Co-authored-by: grep
Co-authored-by: xargs
Co-authored-by: bash

This will be more important if/when the project adopts `set -u` (abort on referencing undefined variable).

### DIFF
--- a/test/src/400-ducc-ingest_images/main
+++ b/test/src/400-ducc-ingest_images/main
@@ -132,7 +132,7 @@ EOL
 
     # crete the repository where to store the content
     echo -n "*** Creating CVMFS repo..."
-    create_empty_repo $CVMFS_TEST400_REPOSITORY $USER || return $?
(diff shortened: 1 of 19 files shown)


vvolkl (approved):
LGTM. As a follow-up, we could make the user argument to `create_empty_repo` optional and use `$CVMFS_TEST_USER` as default.

Issue #4044, opened 2025-12-02
Improve error message when mounting fails due to unreachable manifest


vvolkl:
As suggested by LHCb: When trying to do a manual mount with a typo in the repository name, the error message one would get is:
```
~# mount -t cvmfs lhcb-conddb.cern.ch /cvmfs/lhcb-conddb.cern.ch
CernVM-FS: running with credentials 133:147
CernVM-FS: loading Fuse module... Failed to initialize root file catalog (16 - file catalog failure)
```
This is technically correct, but could be more helpful, in this case saying that the manifest could not be downloaded and printing the url that was tried.

We could also think about an api on the stratum-1 that returns the list of available repos and fuzzy match against it for a "Did you mean ... ?"  message.

vvolkl:
Duplicate of https://its.cern.ch/jira/projects/CVM/issues/CVM-663

PR #4271, opened 2026-06-11
Append the chain of Stratum1 URLs that where tried to fetch the manif…

christge-sft:
Related to https://github.com/cvmfs/cvmfs/issues/4044

The error msg in case of failure now becomes:
```sh
$ sudo mount -t cvmfs lhcb-conddb.cern.ch /cvmfs/lhcb-condb.cern.ch
CernVM-FS: running with credentials 964:964
CernVM-FS: loading Fuse module... Failed to initialize root file catalog. Server chain: http://cvmfs-stratum-one.cern.ch/cvmfs/lhcb-conddb.cern.ch|http://cernvmfs.gridpp.rl.ac.uk/cvmfs/lhcb-conddb.cern.ch|http://cvmfs-s1bnl.opensciencegrid.org/cvmfs/lhcb-conddb.cern.ch|http://cvmfs-s1fnal.opensciencegrid.org/cvmfs/lhcb-conddb.cern.ch| (16 - file catalog failure)
```

### DIFF
--- a/cvmfs/mountpoint.cc
+++ b/cvmfs/mountpoint.cc
@@ -15,6 +15,7 @@
 #include <climits>
 #include <cstring>
 #include <vector>
+#include <string>
 
 #include "duplex_fuse.h"  // IWYU pragma: keep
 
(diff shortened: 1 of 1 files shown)


cvmfs/mountpoint.cc
```diff
@@ -1326,6 +1327,13 @@ bool MountPoint::CreateCatalogManager() {
   if (!retval) {
     boot_error_ = "Failed to initialize root file catalog";
     boot_status_ = loader::kFailCatalog;
+    std::vector<std::string> host_chain;
+    download_mgr_->GetHostInfo(&host_chain,nullptr,nullptr);
+    boot_error_ += ". Server chain: ";
+
+    for(auto url :host_chain){
+      boot_error_+=url+"|";
```
DrDaveD, line 1335: I suggest using `;` instead of `|` because at least for CVMFS_HTTP_PROXY a vertical bar means random order and a semicolon means in that order.

vvolkl:
I think we can further improve error logging by propagating error codes here, but for 2.14 maybe let's just log a hint to run chksetup without logging any urls at all

PR #4274, opened 2026-06-16
fix(gw): don't hold leaseMutex during commit

### LINKED ISSUE #4103: /api/v1/leases often extremely slow
On lhcbdev.cern.ch we often see that `/api/v1/leases` takes 20+ seconds to respond. I don't think this used to be the case and this causes various issues for us not being able to view the status of the system.

I suspect it's related to when the gateway is busy but I don't have any concrete evidence of that.

chrisburr:
`CommitLease` held the global `leaseMutex` for its whole duration, including the slow external receiver call. Because `GetLeases` acquires the same mutex, listing leases blocked behind any in-flight commit, causing `/api/v1/leases` to be extremely slow when the gateway is busy.

Split `CommitLease` into three phases so `leaseMutex` only guards the short SQLite reads/writes: look up + validate the lease, run the receiver commit without `leaseMutex` (still serialised per-repository by the existing per-repo commit lock), then delete the committed lease. This also removes the nested leaseMutex -> repo-lock acquisition.

Fixes #4103

### DIFF
--- a/gateway/internal/gateway/backend/lease_service_test.go
+++ b/gateway/internal/gateway/backend/lease_service_test.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	gw "github.com/cvmfs/gateway/internal/gateway"
+	"github.com/cvmfs/gateway/internal/gateway/receiver"
 )
 
 func TestLeaseServiceNewLease(t *testing.T) {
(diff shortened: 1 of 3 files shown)


gateway/internal/gateway/backend/lease_service_test.go
```diff
@@ -281,3 +282,61 @@ func TestLeaseServiceCommitLease(t *testing.T) {
 		}
 	})
 }
+
+// TestCommitLeaseDoesNotBlockGetLeases is a regression test for issue #4103: a
```
chrisburr, line 286: Arguably this test is overkill but it was useful for verifying the fix behaved the way I expected. I'm happy to remove it if you prefer?

vvolkl:
Thanks Chris for the patch. The test is fine by me, but I'll have to take a closer look if this could potentially re-introduce errors like https://github.com/cvmfs/cvmfs/issues/3259 - which this leaseMutex was intended to fix. Now we know that in 3259 we locked more than we needed too, but I do want to double-check for conditions like this.

chrisburr:
I'm pretty confident this PR won't reintroduce that issue.

I was mostly paranoid if there was any implicit serialisation caused by this mutex but I didn't find anything.

vvolkl:
> I'm pretty confident this PR won't reintroduce that issue.
> 
> I was mostly paranoid if there was any implicit serialisation caused by this mutex but I didn't find anything.

No, I don't think there's any danger of that issue being reintroduced, but I'm not (yet) 100% certain there are no races when a lease expires.

vvolkl:
Yes, this PR allows the following unlikely but bad sequence of events:

- Publisher A holds a lease on `/subdirectory`, modifies `/subdirectory/file1`. The lease expires while it's doing a slow commit.
- Publisher B takes a lease on `/subdirectory`, is granted because A's lease expired, modifies a different part of `/subdirectory/file1`. Previously the leaseMutex made this is impossibly, but this PR allows it.
- Publisher B uses the wrong base for its commit, A's changes to `file1` are lost.

The result is a fairly confusing situation where parts of the publication of A are lost. I think it'd be fine to drop the publication of A entirely, since the lease expired. I think I'll add a small check to the receiver to doublecheck the expiration timestamp when it's commiting, to make this a bit  more unlikely, then I think we can include this in 2.14.

For the next release I think it maybe good to split `Commit` into two parts, one that does the merge and prepares and uploads all the catalog`, and one that does the update to the repository manifest in one short atomic operation. That way a check of lease validity before this step should be much more robust.

PR #4276, opened 2026-06-16
tidy: Make template impl headers self-contained

vvolkl:
The *_impl.h template implementation headers are included at the bottom of their primary header, after the corresponding class is declared. They did not include that primary header themselves, so they could not be parsed as a standalone translation unit. This made clang-tidy emit spurious parse errors (undeclared identifiers, protected-member access, out-of-line definition without definition) whenever clang-tidy-diff fed one of these headers to the compiler directly.

Add the self-include guarded by the existing include guard, matching the convention already used by catalog_mgr_impl.h and catalog_diff_tool_impl.h.

### DIFF
--- a/cvmfs/catalog_counters_impl.h
+++ b/cvmfs/catalog_counters_impl.h
@@ -8,6 +8,7 @@
 #include <map>
 #include <string>
 
+#include "catalog_counters.h"
 #include "catalog_sql.h"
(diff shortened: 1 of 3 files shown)


vvolkl:
Folded into https://github.com/cvmfs/cvmfs/pull/4223

PR #4277, opened 2026-06-17
Summarize cvmfs_server check errors

### LINKED ISSUE #3389: `cvmfs_server check -r` should summarize errors at end
As requested in the coordination meeting by @deesto : Running the checks produces a lot of output, and usually needs to be redirected to a file and grepped to find errors. We should print a summary at the end.

vvolkl:
Fixes #3389

### DIFF
--- a/cvmfs/server/cvmfs_server_check.sh
+++ b/cvmfs/server/cvmfs_server_check.sh
@@ -110,7 +110,62 @@ __do_check() {
                      $(get_swissknife_proxy)           \
                      $(get_follow_http_redirects_flag) \
                      $with_reflog"
-  $user_shell "$check_cmd"
+
+  # Set CVMFS_SERVER_CHECK_SUMMARY=0 to restore the previous behavior.
+  if [ "${CVMFS_SERVER_CHECK_SUMMARY:-1}" = "0" ]; then
+    $user_shell "$check_cmd"
+    return $?
+  fi
+
+  # The check produces a lot of progress output on stdout while problems are
+  # reported on stderr.  Mirror stderr into a temporary file so that a concise
+  # summary can be printed at the end without changing the live output.
+  local check_error_log
+  check_error_log="$(mktemp)" || die "Failed to create temporary file"
+  local check_error_pipe
+  check_error_pipe="$(mktemp)" || die "Failed to create temporary file"
+  rm -f "$check_error_pipe"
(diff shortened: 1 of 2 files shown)


cvmfs/server/cvmfs_server_check.sh
```diff
@@ -110,7 +110,62 @@ __do_check() {
...
+  # reported on stderr.  Mirror stderr into a temporary file so that a concise
+  # summary can be printed at the end without changing the live output.
+  local check_error_log
+  check_error_log="$(mktemp)" || die "Failed to create temporary file"
+  local check_error_pipe
+  check_error_pipe="$(mktemp)" || die "Failed to create temporary file"
+  rm -f "$check_error_pipe"
+  mkfifo "$check_error_pipe" || die "Failed to create temporary pipe"
+
+  tee -a "$check_error_log" < "$check_error_pipe" >&2 &
```
DrDaveD, line 147: This trickery with running tee in the background is basically so you can check the first command in the pipeline for an error exit code, is that correct?  If so that might deserve a comment.  If you can depend on bash, it has a couple of simpler options for this; the `PIPESTATUS` array contains exit codes for all the pipe commands, and the `pipefail` option will return an exit code if any command in a pipeline fails.

cvmfs/server/cvmfs_server_check.sh
```diff
@@ -110,7 +110,62 @@ __do_check() {
...
+  $user_shell "$check_cmd" 2>"$check_error_pipe"
+  local check_retval=$?
+  wait "$tee_pid"
+  set -e
+
+  rm -f "$check_error_pipe"
+  if [ $check_retval -ne 0 ]; then
+    __print_check_summary "$name" "$check_error_log"
+  fi
+  rm -f "$check_error_log"
```
DrDaveD, line 160: Is there any chance that the function could exit with an error before this point?  If so this temp file and possibly the other one might not get deleted.  I would probably use a trap to make sure that didn't happen.

DrDaveD (approved):
Approving since my suggestions are optional improvements.

cvmfs/server/cvmfs_server_check.sh
vvolkl: Implemented, but this function already had a trap for the gc lock, so that needs to be handled here too. It's a bit more complex than I'd like, but should work

Issue #3465, opened 2023-12-05
Give CVMFS_AUTO_TAG_TIMESPAN a default on newly created repositories


DrDaveD:
I think it's a problem that by default generic tags stay around indefinitely.  It's especially a problem when garbage collection is enabled because it interferes with garbage collection.  I suggest setting the default to "4 weeks ago".

PR #4278, opened 2026-06-17
Set 2 week default for autotag timespan

### LINKED ISSUE #4116: CVMFS_AUTO_TAG_TIMESPAN should have a default
Currently the default when creating repositories to publish is to tag every publish and never delete them, which means nothing ever gets deleted even on fresh stratum 1 snapshots (or garbage collection, although that warns if the auto tag timespan is not set).  The CVMFS documentation recommends setting `CVMFS_AUTO_TAG_TIMESPAN="2 weeks ago"`.  This should be the default.

vvolkl:
Fixes #4116 and #3465

### DIFF
--- a/cvmfs/server/cvmfs_server_mkfs.sh
+++ b/cvmfs/server/cvmfs_server_mkfs.sh
@@ -213,7 +213,6 @@ cvmfs_server_mkfs() {
   local volatile_content=0
   local autotagging=true
   local auto_tag_timespan=
-  local auto_tag_timespan_set=0
   local unionfs
   local hash_algo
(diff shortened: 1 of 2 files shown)


DrDaveD (approved):
I think it would be good enough to just see if `auto_tag_timespan` is not empty instead of adding another variable when it is set, but this way works too.

vvolkl:
Right, I was trying to keep it possible to enable autotagging, but disable the autotag cleanup. However the user can simply delete the config line that mkfs sets from server.conf, if that's really needed, and I doubt it will be.

PR #4313, opened 2026-07-08
Add cvmfs_config docker-run

vvolkl:
This is  convenience for the docker run invocation needed for the service container, which can be a bit complicated: 

```
docker run -d --rm   -e CVMFS_CLIENT_PROFILE=single   -e CVMFS_REPOSITORIES=sft.cern.ch,...   --cap-add SYS_ADMIN   --device /dev/fuse   --volume /cvmfs:/cvmfs:shared
```
`cvmfs_config docker-run` prints this to stdout so users don't need to look it up in the docs - It may need even more flags for SELinux, which are included here as well.

### DIFF
--- a/packaging/container/mount_cvmfs.sh
+++ b/packaging/container/mount_cvmfs.sh
@@ -25,8 +25,10 @@ usage() {
   cat <<EOF
 CernVM-FS service container ${VERSION:+version $VERSION}
 
-Mounts CernVM-FS repositories under /cvmfs. The mode is selected by the
-CVMFS_REPOSITORIES environment variable:
+Mounts CernVM-FS repositories under /cvmfs for export to the host.
+
+Will either mount a given list of repositories, or mount repositiories
+on demand, depending on the CVMFS_REPOSITORIES environment variable:
 
   unset  automount mode  repositories are mounted on first access
   set    explicit mode   the listed repositories are pre-mounted at startup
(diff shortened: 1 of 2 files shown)


packaging/container/mount_cvmfs.sh
```diff
@@ -21,6 +21,49 @@ cleanup() {
   exit 0
 }
 
+usage() {
+  cat <<EOF
+CernVM-FS service container ${VERSION:+version $VERSION}
+
+Mounts CernVM-FS repositories under /cvmfs. The mode is selected by the
```
DrDaveD: Does this really mount cvmfs repositories, or does it just bind mount cvmfs into a docker container?  I am confused by the terminology.

vvolkl: Indeed there are a few different scenarios with cvmfs + containers, but this is for our "service container" (https://cvmfs.readthedocs.io/en/latest/cpt-quickstart/#other-platforms) which does run cvmfs and mount the repositories inside the container. This is intended to be bind-mounted to the host (and possibly into other containers), so replaces the cvmfs installation on the host. I believe NET2 is using cvmfs this way.

DrDaveD: Ah.  Then maybe it could say it is for export to the host.

vvolkl: Done

PR #4314, opened 2026-07-08
Feature: Add D-State Debugger

Physic69:
Added `dcoredumper` utility to extract  core dumps directly from `/proc` interfaces, bypassing `ptrace()` so it doesn't hang on `D-state processes`. Also includes the cvmfs_config dcoredump wrapper and CMake integration.

` (GSoC 2k26 PR)`

TODO:

- [x] Add Dump levels
- [x] refactor memory management
- [x] User manual, explaning code & usage


cvmfs/Dcoredumper.cc
```diff
@@ -0,0 +1,943 @@
...
+#include <fcntl.h>
+#include <unistd.h>
+#include <dirent.h>
+#include <elf.h>
+#include <errno.h>
+#include <stdint.h>
+#include <sys/types.h>
+#include <sys/utsname.h>
+
+#define NOTE_ALIGN(sz) (((sz) + 3) & ~3)
```
christge-sft: I'ld go with a type safe alternative

cvmfs/Dcoredumper.cc
```diff
@@ -0,0 +1,943 @@
...
+#define PAGE_SIZE 4096UL
```
christge-sft: Maybe a type safe constexpr?

cvmfs/Dcoredumper.cc
```diff
@@ -0,0 +1,943 @@
...
+  size_t      reg_sp;         /* index of stack pointer in regs       */
+  size_t      reg_fp;         /* index of frame pointer in regs       */
+  size_t      prstatus_size;  /* total size of NT_PRSTATUS descriptor */
+  size_t      prstatus_pid;   /* byte offset of pr_pid in prstatus    */
+  size_t      prstatus_reg;   /* byte offset of pr_reg in prstatus    */
+  size_t      prpsinfo_size;  /* total size of NT_PRPSINFO descriptor */
+  size_t      prpsinfo_pid;   /* byte offset of pr_pid in prpsinfo    */
+  size_t      prpsinfo_fname; /* byte offset of pr_fname in prpsinfo  */
+  size_t      prpsinfo_args;  /* byte offset of pr_psargs in prpsinfo */
+} arch_desc_t;
```
christge-sft: What about this being a templated class and arch_x86_64 and arch_aarch64 being specifications, with the static field being compile time constants?

cvmfs/Dcoredumper.cc
```diff
@@ -0,0 +1,943 @@
...
+  136,  /* prpsinfo_size                     */
+  24,   /* prpsinfo_pid                      */
+  40,   /* prpsinfo_fname                    */
+  56    /* prpsinfo_args                     */
+};
+
+
+
+/* g_arch is set at startup by detect_arch() via uname(). */
+static const arch_desc_t *g_arch = &arch_x86_64;
```
christge-sft: Does this really need to be a global? What about a ctx structure initialized in main and passed around wherever necessary?

cvmfs/Dcoredumper.cc
```diff
@@ -0,0 +1,943 @@
...
+
+/**
+ * Dump level controls which memory regions are written to the core.
+ * NT_FILE always covers all file-backed regions regardless of level
+ * so GDB can reload .text from disk.
+ */
+typedef enum {
+  DUMP_STANDARD,  /**< skip read-only file-backed regions (default) */
+  DUMP_FULL       /**< dump everything readable */
+} dump_level_t;
```
christge-sft: why not enum classes? type safety.

cvmfs/Dcoredumper.cc
```diff
@@ -0,0 +1,943 @@
...
+
+static dump_level_t g_level = DUMP_STANDARD;
```
christge-sft: same for globals.

christge-sft:
As a general comment, this is clearly  C-code. @vvolkl what do you think about it? I am always in favor of using all nice things that come along with C++. Although, if you think that this as a standalone tool is good enough, we can leave it as is.

christge-sft:
Also, as of @vvolkl 's recommendation in the dedicated mattermost channel, I'ld also suggest folding this executable inside the `cvmfs2` process.

Issue #4177, opened 2026-03-25
CI builds with koji/mock


vvolkl:
I just checked manually for 2.14.0~pre1 that it builds in chroot/without network access from the source tarball, but for future releases we should add automatic testing.

PR #4320, opened 2026-07-10
externals: Move maxminddb to fetchcontent 

vvolkl:


DrDaveD:
@vvolkl Will this download from the internet while building an rpm?  The OSG koji build server requires everything to be pre-downloaded, it does not allow network access during the build.

vvolkl:
> @vvolkl Will this download from the internet while building an rpm? The OSG koji build server requires everything to be pre-downloaded, it does not allow network access during the build.

We use the same mechanism for all vendored dependencies now, so it's the same as for libarchive that was already migrated in 2.14. Since that worked, I don't expect any problems. Yes, to fetch the sources to vendor the dependencies network access is needed, but that's only for building the source tarball/srpm. That will then include the source tarballs for all the vendored dependencies in EXTERNALS_DOWNLOAD_DIR, which the build systems will automatically reuse if the hash matches, so no network is needed to build the rpms. We should add a ci workflow to test this though, as I've already mentioned in  https://github.com/cvmfs/cvmfs/issues/4177.

vvolkl:
Turns out we were still missing the script that downloads the necessary source tarballs, but I added it along with the CI workflow that now succeeds for an offline build: https://github.com/cvmfs/cvmfs/pull/4326

That script may still download more dependency tarballs as needed, as some dependencies may be taken from the system, but I think that's OK.

PR #4324, opened 2026-07-12
Fix issues in cvmfs_config showconfig -j

SotirisKav:
This PR fixes issues regarding the JSON output:

- **Shell environment leak:** `showconfig -j` outputs what is actually set in the config files, ignoring variables exported in the local shell.
- **EscapeShell quotes:** Fixed the bug where semicolon-separated URLs (like in `CVMFS_SERVER_URL`) kept their shell-protection single quotes in the JSON output.
- **Control characters:** Added mappings for `\n`, `\r`, `\t`, etc., in the `json_string()` function so config files with weird line endings (that Windows adds) don't result in invalid JSON syntax.
- **Tests:** Renewed the test no. 046 (defaultd) so it tests these edge cases (by leaking a variable in shell, and by testing the EscapeShell bug against grid.cern.ch which is configured to have 4 server urls, separated by ';').
- **Documentation** Fixed a small inaccuracy in the `showconfig -s` usage description (ignores empty-valued parameters -> ignores not set parameters).

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -266,7 +266,7 @@ cvmfs_config_usage() {
   fi
   echo "  chksetup"
   echo ""
-  echo "  showconfig [-s(hort output, only non-empty parameters)] [-j|--json] [<repository>]"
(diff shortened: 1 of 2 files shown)


vvolkl (approved):
Thanks for the fix!

PR #4333, opened 2026-07-16
Fix compiling s3mockserver

andrey-utkin:
Without this, it complains for absense of curl/curl.h

Fixes: 1e7bacc448cf ("externals: Move curl and c-ares to FetchContent (#4322)")

### DIFF
--- a/test/stress/CMakeLists.txt
+++ b/test/stress/CMakeLists.txt
@@ -48,6 +48,7 @@ add_executable(s3mockserver
 
 target_link_libraries (s3mockserver
                        cvmfs_util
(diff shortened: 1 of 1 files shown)


vvolkl (approved):
Thank you. I thought I added a build with -DBUILD_ALL=ON last time around, but this directory still has no CI coverage apparently.

vvolkl:
Ah, I see, all the github actions workflow cache the externals, which masked this. I added one workflow without caching to catch this: https://github.com/cvmfs/cvmfs/pull/4337

PR #4339, opened 2026-07-18
feat: support building libcvmfs-crypto against the system OpenSSL

chrisburr:
This PR adds a cleaner way (i.e. `-DBUILTIN_EXTERNALS_EXCLUDE=libcrypto`) of supporting OpenSSL for packaging in ecosystems where statically linking LibreSSL is discouraged/forbidden. For example Gentoo has shipped an equivalent patch since 2.10.


vvolkl:
With https://github.com/cvmfs/cvmfs/pull/3576 we'll be able to remove libressl alltogether, once I get around to update the SHAKE-128 implementation. Let's wait and see if this is still needed.

chrisburr:
In that case I'll close this and keep it downstream until the libnettle migration is done. 😄

PR #4340, opened 2026-07-18
Support non-default toolchains

chrisburr:
I found this was needed to build libcvmfs for conda-forge and I think it's generally applicable to all builds.


vvolkl (changes requested):
The current crypto externals are fast becoming obsolete, see https://github.com/cvmfs/cvmfs/pull/3576. Can you drop those changes and only keep the patch for cvmfs_compiler.cmake?

chrisburr:
Thanks for the prompt reviews, done 👍

vvolkl (approved):

PR #4351, opened 2026-07-20
client: Fix file bundle trigger inconsistency

chrisburr:
As discussed on mattermost with @vvolkl the current handling was inconsistent between `.cvmfsbundle.` and `.cvmfsbundle-`.

Also fixes the test and a log message typo.

### DIFF
--- a/cvmfs/bundle_mgr.cc
+++ b/cvmfs/bundle_mgr.cc
@@ -27,7 +27,7 @@
 namespace {
 constexpr size_t kDefaultBundlePoolSize = 8;
 
-// Read the .cvmfsbundle.<basename> file via the cvmfs cache
+// Read the .cvmfsbundle-<basename> file via the cvmfs cache
 BundleFileMgr *LoadBundleFromCvmfs(MountPoint *mp,
                                    const PathString &bundle_file_path) {
   catalog::DirectoryEntry dirent;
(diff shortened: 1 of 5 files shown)


cvmfs/bundle_mgr.cc
```diff
@@ -142,13 +142,28 @@ void BundleMgr::Fetch() {
...
+ * resolved relative to the directory holding the bundle file.
+ */
+PathString BundleMgr::NormalizeDependencyPath(const PathString &path) const {
+  std::string p = path.ToString();
+  if (!p.empty() && p[0] == '/')
+    return path;
+  if (p.compare(0, 2, "./") == 0)
+    p = p.substr(2);
+  return PathString(parent_path_.ToString() + "/" + p);
+}
```
vvolkl: For consistency I'd prefer to implement this with `PathString`:
```suggestion
PathString BundleMgr::NormalizeDependencyPath(const PathString &path) const {
  if (path.StartsWith(PathString("/", 1)))
    return path;
  const PathString relative = path.StartsWith(PathString("./", 2))
                                  ? path.Suffix(2)
                                  : path;
  PathString normalized(parent_path_);
  normalized.Append("/", 1);
  normalized.Append(relative.GetChars(), relative.GetLength());
  return normalized;
}
}
```

test/src/709-cvmfsbundles/main
```diff
@@ -12,18 +12,29 @@ produce_dependency_files() {
...
 
   popd >/dev/null || return 1
 }
 
+# Path of a file's content object in the local cache, derived from its
+# hash extended attribute (a metadata lookup that does not fetch content)
+cache_object_path() {
+  local cache=$1
+  local file=$2
+  local hash=$(attr -g hash "$file" | cut -d: -f2 | xargs)
```
vvolkl:
```suggestion
  local hash=$(get_xattr hash "$file")
```

vvolkl (changes requested):
Good catch. There is a workaround for this issue by publishing both file formats, but great to have this fixed in time for 2.14.1. Looks correct to me, two small code comments inline.

vvolkl:
@chrisburr I think we should look at the last few commits separately. Can you create new PRs?

chrisburr:
Sorry that was supposed to be a separate branch but pushed accidentally to the wrong name. Will do.

vvolkl (approved):

PR #4352, opened 2026-07-20
server: do not abort on bundle dependency paths reaching PIPE_BUF

chrisburr:
This PR adds a failing test and fixes it by removing the assert. There is still a limit on the sender side (64KiB on Linux, less macOS) but I think that's unlikely to be a problem in practice.


cvmfs/bundle_mgr.h
vvolkl:
```suggestion
```

Header no longer needed with this patch.

vvolkl (approved):

PR #4358, opened 2026-07-22
fix: keep libarchive available for unit test builds

chrisburr:
In #4338 I missed that `libarchive` is also needed for the unit tests.


chrisburr:
The failure is from the unreliable nettle source URL and is fixed in https://github.com/cvmfs/cvmfs/pull/4360

PR #4362, opened 2026-07-23
fix: merge nested catalogs by column name instead of position

chrisburr:
🤖 Below is written by an LLM but I think summarises what I found well enough:

Observed on a gateway setup with CernVM-FS 2.13.3 while removing a nested catalog from `lhcbdev.cern.ch` (a repository old enough that its catalog databases predate schema revision 7; the nested catalog had been created recently, i.e. with the fresh layout):

```
$ cvmfs_server transaction lhcbdev.cern.ch/testfiledb-mirror
$ rm /cvmfs/lhcbdev.cern.ch/testfiledb-mirror/.cvmfscatalog
$ cvmfs_server publish lhcbdev.cern.ch
...
Lease end request - error reply: {"reason":"worker 'commit' call failed: possible that the receiver crashed: EOF","status":"error"}
SessionContext: could not commit session. Aborting.
```

The publisher-side `cvmfs_swissknife sync` runs `MergeIntoParent()` to dissolve the nested catalog, producing a corrupted root catalog (767 shifted rows) that it uploads to the backend. During commit, the receiver diffs old vs. new revision: the old listing of the directory has the real entry names, the corrupted new listing has eight directories all named `"1"`. The diff reports them as additions; the second `AddDirectory("/testfiledb-mirror/1")` inserts a duplicate md5 path hash, the SQLite UNIQUE constraint fails, and the `assert(retval)` in `WritableCatalog::AddEntry()` aborts the receiver. The crash is 100% reproducible on every retry.

Note that the receiver crash is what *prevented* the corruption from being published: on a non-gateway repository the same merge runs in `cvmfs_swissknife sync` and the corrupted catalog would be published silently.

## Receiver stack trace

From `/tmp/cvmfs_receiver/stacktrace.*` (2.13.3, paths shortened):

```
Signal: 6 (SIGABRT)
#4  __pthread_kill_implementation
#5  raise
#6  abort
#7  __assert_fail_base.cold
#8  __assert_fail
#9  catalog::WritableCatalog::AddEntry (entry=..., xattrs=...,
        entry_path="/testfiledb-mirror/1", parent_path="/testfiledb-mirror")
        at cvmfs/catalog_rw.cc:166
#10 catalog::WritableCatalogManager::AddDirectory (parent_directory="testfiledb-mirror")
        at cvmfs/catalog_mgr_rw.cc:479
#11 receiver::CatalogMergeTool<WritableCatalogManager, SimpleCatalogManager>::ReportAddition
#12 CatalogDiffTool<catalog::SimpleCatalogManager>::DiffRec at cvmfs/catalog_diff_tool_impl.h:182
#13 CatalogDiffTool<catalog::SimpleCatalogManager>::DiffRec at cvmfs/catalog_diff_tool_impl.h:232
#14 CatalogDiffTool<catalog::SimpleCatalogManager>::Run at cvmfs/catalog_diff_tool_impl.h:88
#15 receiver::CatalogMergeTool<...>::Run at cvmfs/receiver/catalog_merge_tool_impl.h:81
#16 receiver::CommitProcessor::Process (lease_path="lhcbdev.cern.ch/testfiledb-mirror", ...)
        at cvmfs/receiver/commit_processor.cc:201
#17 receiver::Reactor::HandleCommit at cvmfs/receiver/reactor.cc:438
#18 receiver::Reactor::HandleRequest at cvmfs/receiver/reactor.cc:508
#19 receiver::Reactor::Run at cvmfs/receiver/reactor.cc:186
#20 main
```

The corruption is directly visible in the uploaded root catalog — the children of the merged directory, where `name` holds the former `flags` value and `flags` received the (empty) `mtimens`:

```
sqlite> SELECT name, flags, size, mode FROM catalog WHERE parent_1 = ... ;
1|<NULL>|20|16877          -- a directory (kFlagDir = 1)
1|<NULL>|53|16877
...
4|<NULL>|122736|33188      -- a regular file (kFlagFile = 4)
68|<NULL>|634618908|33188  -- a chunked file (kFlagFile | kFlagFileChunk = 68)
```

### DIFF
--- a/cvmfs/catalog_rw.cc
+++ b/cvmfs/catalog_rw.cc
@@ -700,12 +700,25 @@ void WritableCatalog::CopyToParent() {
                                           "AS other;");
   retval = sql_attach.Execute();
   assert(retval);
-  retval = SqlCatalog(database(), "INSERT INTO other.catalog "
-                                  "SELECT * FROM main.catalog;")
+  // The two databases can have a different physical column order: fresh
+  // databases are created with mtimens between mtime and flags, while
+  // databases migrated from schema revision < 7 have mtimens appended at the
+  // end (ALTER TABLE can only append).  A positional "SELECT *" copy would
+  // silently shuffle the values into the wrong columns, so the columns have
+  // to be named explicitly.
+  const string catalog_columns =
+      "md5path_1, md5path_2, parent_1, parent_2, hardlinks, hash, size, "
+      "mode, mtime, mtimens, flags, name, symlink, uid, gid, xattr";
+  retval = SqlCatalog(database(),
+                      "INSERT INTO other.catalog (" + catalog_columns
+                          + ") SELECT " + catalog_columns
+                          + " FROM main.catalog;")
                .Execute();
   assert(retval);
-  retval = SqlCatalog(database(), "INSERT INTO other.chunks "
-                                  "SELECT * FROM main.chunks;")
+  const string chunks_columns = "md5path_1, md5path_2, offset, size, hash";
+  retval = SqlCatalog(database(),
+                      "INSERT INTO other.chunks (" + chunks_columns
+                          + ") SELECT " + chunks_columns + " FROM main.chunks;")
                .Execute();
   assert(retval);
   retval = SqlCatalog(database(), "DETACH other;").Execute();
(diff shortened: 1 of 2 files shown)


vvolkl:
Good catch! That must have been there since https://github.com/cvmfs/cvmfs/pull/3513 . I'm a bit surprised nobody has been hit by this before, we must have quite a few repositories that old that would either hit the gateway error or show some very visible corruption.

chrisburr:
I guess it's rare to remove a catalog but not the entire tree (even in my case, I didn't intend to)

PR #4370, opened 2026-07-28
swissknife_lease: add "refresh" action

andrey-utkin:


vvolkl:
How do you intend this flag to be used? swissknife commands are not intended to be user facing, so it should be wrapped in a `cvmfs_server` command somewhere.

andrey-utkin:
Here's a snippet I currently rely on in my testing:

https://github.com/andrey-utkin/cvmfs/blob/ce0d93592e3d3e3b9da512c8f18d3405fadf9cc9/test/src/901-ingestsql/run_test#L61-L73

Sometimes, under load (most probably with I/O being the bottleneck), with the gateway lease intentionally shortened from 2hr to 1min to speed up the test's progress, initial `publish` would have the lease timeout, breaking the repeated testing loop. I don't intend to upstream [this patch to the test](https://github.com/andrey-utkin/cvmfs/commit/ce0d93592e3d3e3b9da512c8f18d3405fadf9cc9), but the lease command may be useful to others in a similar scenario.

vvolkl:
Wait a second, I already added something to cvmfs_server myself:  https://github.com/cvmfs/cvmfs/pull/4294 . So you could use `cvmfs_server refresh-lease` in your script already. It does makes sense to have this functionality in swissknife_lease, but  then we should invoke it from cvmfs_server, otherwise there will be two disconnected code paths doing the same thing. Could you try updating `cvmfs_server refresh-lease`?

andrey-utkin:
> So you could use `cvmfs_server refresh-lease` in your script already

oh cool, I didn't notice it's there :)

PR #4371, opened 2026-07-28
Reduce package size: make gdb optional, strip debug symbols into separate package.

### LINKED ISSUE #4186: Lightweight CVMFS client for diskless compute nodes
We are running CVMFS on diskless compute nodes (RHEL 9, aarch64) and would like to keep the installation as small as possible.

When installing the `cvmfs` RPM from the official repository, we noticed that it pulls in debug libraries and debugging packages:

```shell
======================================================================
 Package                   Architecture Version                  Size
======================================================================
Installing:
 cvmfs                     aarch64      2.13.3-1.el9             15 M
Installing dependencies:
 autofs                    aarch64      1:5.1.7-65.el9          380 k
 boost-regex               aarch64      1.75.0-13.el9_7         257 k
 cvmfs-config-default      noarch       2.2-1                   9.0 k
 cvmfs-fuse3               aarch64      2.13.3-1.el9             14 M
 cvmfs-libs                aarch64      2.13.3-1.el9             11 M
 fuse                      aarch64      2.9.9-17.el9             83 k
 fuse-common               aarch64      3.10.2-9.el9            8.5 k
 gdb                       aarch64      16.3-2.el9              153 k
 gdb-headless              aarch64      16.3-2.el9              4.8 M
 initscripts-service       noarch       10.11.8-4.el9            13 k
 libsss_autofs             aarch64      2.9.7-4.el9_7.1          35 k
 source-highlight          aarch64      3.1.9-12.el9            666 k

Transaction Summary
======================================================================
Install  13 Packages

Total download size: 46 M
Installed size: 165 M
```

The debug libraries consume 50.9 MB and gdb/gdb-headless packages add another 16.6 MB:

```shell
$ ls -la /usr/lib64/libcvmfs*_debug.so.*
-rwxr-xr-x. 1 root root  6707280 23. Okt 20:22 /usr/lib64/libcvmfs_crypto_debug.so.2.13.3
-rwxr-xr-x. 1 root root 22616224 23. Okt 20:31 /usr/lib64/libcvmfs_fuse3_debug.so.2.13.3
-rwxr-xr-x. 1 root root 22615752 23. Okt 20:27 /usr/lib64/libcvmfs_fuse_debug.so.2.13.3
-rwxr-xr-x. 1 root root  1508680 23. Okt 20:22 /usr/lib64/libcvmfs_util_debug.so.2.13.3

$ ls -la /usr/lib64/libcvmfs*_debug.so* | awk '{sum+=$5} END {print sum/1024/1024 " MB"}'
50.972 MB
$ rpm -qa --queryformat '%{NAME} %{SIZE}\n' | grep -E '^gdb |^gdb-headless' | awk '{sum+=$2} END {print sum/1024/1024 " MB"}'
16.5694 MB
```

Environment:
- RHEL 9
- aarch64
- CVMFS version: cvmfs-2.13.3-1.el9 (installed from the official CVMFS RPM repository)
- Total installed size: 165 MB

It would be very helpful in our environment to separate the debug and production libraries as it is done in many other packages. Is there any particular reason preventing that separation? If it's a matter of tweaking the spec files, we could explore contributing a fix.

Thank you for your help and for maintaining CVMFS!

vvolkl:
For regular users this hopefully leaves everything unchanged, but with
```
yum install cvmfs --setopt=install_weak_deps=False
```
users who care about package size can save some 80 MB.

Fixes https://github.com/cvmfs/cvmfs/issues/4186

### DIFF
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -66,8 +66,10 @@
 %define hardlink /usr/bin/hardlink
 %endif
 
-%define __strip /bin/true
-%define debug_package %{nil}
+# Move DWARF to -debuginfo while keeping .symtab for watchdog backtraces.
+%global _find_debuginfo_opts -g
+# Prevent the later brp pass from stripping .symtab.
+%global __brp_strip %{nil}
 %if 0%{?el6} || 0%{?el5} || 0%{?el4}
 %define __os_install_post %{nil}
 %endif
@@ -160,7 +162,8 @@ Requires: zlib
 %if !0%{?suse_version}
 Requires: protobuf
 %endif
-Requires: gdb
+# gdb adds details to watchdog crash reports.
+Recommends: gdb
 %if 0%{?suse_version}
 Requires: libarchive13
 %else
--- a/packaging/rpm/cvmfs-universal.spec
+++ b/packaging/rpm/cvmfs-universal.spec
@@ -162,8 +162,9 @@ Requires: zlib
 %if !0%{?suse_version}
 Requires: protobuf
 %endif
-# gdb adds details to watchdog crash reports.
+# gdb and matching debug symbols add details to watchdog crash reports.
 Recommends: gdb
+Recommends: %{name}-debuginfo%{?_isa} = %{version}-%{release}
 %if 0%{?suse_version}
 Requires: libarchive13
 %else
--- a/packaging/debian/cvmfs/control
+++ b/packaging/debian/cvmfs/control
@@ -48,8 +48,9 @@ Depends:
  ${shlibs:Depends}
 Recommends:
  autofs (>= 5.1.2),
-# gdb adds details to watchdog crash reports.
- gdb
+# gdb and matching debug symbols add details to watchdog crash reports.
+ gdb,
+ cvmfs-dbg (= ${binary:Version})
 Homepage: http://cernvm.cern.ch
 Description: CernVM File System
  HTTP File System for Distributing Software to CernVM.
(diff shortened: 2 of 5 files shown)


packaging/rpm/cvmfs-universal.spec
```diff
@@ -160,7 +162,9 @@ Requires: zlib
 %if !0%{?suse_version}
 Requires: protobuf
 %endif
-Requires: gdb
+# gdb and matching debug symbols add details to watchdog crash reports.
+Recommends: gdb
+Recommends: %{name}-debuginfo%{?_isa} = %{version}-%{release}
```
DrDaveD, line 167: What about `%{name}-libs-debuginfo` and `%{name}-fuse3-debuginfo`?

DrDaveD (changes requested):

packaging/rpm/cvmfs-universal.spec
DrDaveD: And for each of the other subpackages, `-ducc`, `-server`, `-gateway`, `-shrinkwrap`, `-snapshotter`, each Recommend their own `-debuginfo`?   It's probably best to treat each `%package` the same, these and `-libs` and `-fuse3`.  There are additional subpackages `-devel` and `-unittests` but I don't see any corresponding `-debuginfo` for those.  In fact the `-unittests` package is not built when I tested it.

vvolkl:
Thanks Dave for taking a look. Maybe having the debuginfo packages as a week dependency is actually not needed. With gdb but no debuginfo we still get a stacktrace with function names, just no line numbers. I think that'd be acceptable.

DrDaveD:
I don't know.  Those line numbers make stack traces much more useful.

PR #4373, opened 2026-07-28
Fix deploying golang

andrey-utkin:
Clean slate build fails on Ubuntu 24.04 (seemed to not happen earlier). The first commit bash enables strict mode and tracing (`set -x`) in `externals/bootstrap.sh` and `externals/*/src/*.sh` and adjusts the code. The second one addresses the initial failure.


kadircanyildirm-crypto:
Independent reproduction on a clean `ubuntu:24.04` container — no WSL, no
pre-existing toolchain.

Steps:

```bash
docker run --rm -it ubuntu:24.04 bash
apt-get update && apt-get install -y sudo git ca-certificates
git clone --depth 1 https://github.com/cvmfs/cvmfs.git && cd cvmfs   # 746d9e4
sudo ci/build_install_builddeps.sh
apt-get install -y ninja-build      # not covered by the builddeps script
mkdir build && cd build
cmake -G Ninja -D BUILD_UNITTESTS=ON -D BUILD_GATEWAY=ON ../
```

Output:

```
--> Building golang_rev2
rm: cannot remove '/src/cvmfs/externals_install.x86_64.ubuntu24.04/bin/go': No such file or directory
rm: cannot remove '/src/cvmfs/externals_install.x86_64.ubuntu24.04/bin/gofmt': No such file or directory
ln: failed to create symbolic link '/src/cvmfs/externals_install.x86_64.ubuntu24.04/bin/go': No such file or directory
ln: failed to create symbolic link '/src/cvmfs/externals_install.x86_64.ubuntu24.04/bin/gofmt': No such file or directory
CMake Error at CMakeLists.txt:291 (message):
  bootstrapping failed
```

I first hit this on WSL2 / Ubuntu 24.04 and assumed it might be specific to
that environment. The container run above shows it is not.

As far as I can tell the cause is that `externals/bootstrap.sh` prepares the
install directory with

```sh
mkdir -p $externals_build_dir
mkdir -p $externals_install_dir/include
mkdir -p $externals_install_dir/lib
```

but not `bin`, while `externals/golang_rev2/src/makeHook.sh` symlinks into
`bin/`. Creating the directory there also resolves it:

```diff
 mkdir -p $externals_install_dir/lib
+mkdir -p $externals_install_dir/bin
```

Happy to test any variant of the fix if that is useful.

vvolkl:
@andrey-utkin I think you'll need something like:

```diff
diff --git a/externals/bootstrap.sh b/externals/bootstrap.sh
index cfbd99f77..2189c84d1 100755
--- a/externals/bootstrap.sh
+++ b/externals/bootstrap.sh
@@ -214,7 +214,7 @@ if [[ -v BUILTIN_EXTERNALS_LIST ]] && ! echo ";${BUILTIN_EXTERNALS_LIST};" | gre
 else
     missing_libs="libcrypto sha3"

-    if [ x"$BUILD_GATEWAY" != x ] || [ x"$BUILD_DUCC" != x ] || [ x"$BUILD_SNAPSHOTTER" != x ]; then
+    if [ x"${BUILD_GATEWAY:-}" != x ] || [ x"${BUILD_DUCC:-}" != x ] || [ x"${BUILD_SNAPSHOTTER:-}" != x ]; then
         required_go_minor_version="23"
         if [ -n "$(command -v go)" ]; then
           go_minor_version=`go version | { read _ _ v _; echo ${v#go}; } | cut -d '.' -f2`
```

Just to note, all of the changed files are fast becoming obsolete. `boostrap.sh` we can remove after the update to the crypto externals I have in the pipeline. For golang I think I haven't noticed this failure  because I'm always using `ci/build_install_buildeps.sh`, which will use `ci/bootstrap_golang.sh` if needed.

kadircanyildirm-crypto:
> For golang I think I haven't noticed this failure because I'm always using `ci/build_install_buildeps.sh`, which will use `ci/bootstrap_golang.sh` if needed.

I tried that path to double-check, and it still ends in the same failure: the Go
that builddeps bootstraps is not visible to a subsequent plain `cmake` run.
Clean `ubuntu:24.04` container, `devel` @ 746d9e4, with `ninja-build` added on
top — the builddeps script does not install it, and without it cmake stops
before it ever reaches bootstrapping.

**1.** `sudo ci/build_install_builddeps.sh` does the right thing and exits 0
(output abridged):

```
[builddeps] go not found — bootstrapping go
[bootstrap-golang] Installing go 1.25.8 into /src/cvmfs/externals_install.x86_64.ubuntu24.04
[…]
[bootstrap-golang] go 1.25.8 installed successfully
[builddeps] go bootstrapped at: /src/cvmfs/externals_install.x86_64.ubuntu24.04/go/bin/go
```

**2.** Immediately afterwards that Go is unreachable for whatever runs next —
`command -v go` finds nothing, while the binary itself is fine:

```console
$ /src/cvmfs/externals_install.x86_64.ubuntu24.04/go/bin/go version
go version go1.25.8 linux/amd64
```

The `export CVMFS_GO_DIR` in
[`build_install_builddeps.sh#L80-L83`](https://github.com/cvmfs/cvmfs/blob/746d9e41c181ced0ca6bb348ace7c28891b0d0af/ci/build_install_builddeps.sh#L80-L83)
is scoped to that script's own process — which the developer docs additionally
run under `sudo` — so it cannot reach the caller's shell, and `PATH` is left
untouched. `CVMFS_GO_DIR` is read by the packaging scripts, but nothing on the
cmake/externals side reads it.

**3.** `cmake -G Ninja -D BUILD_UNITTESTS=ON -D BUILD_GATEWAY=ON ../` then
decides whether to vendor Go purely from `command -v go`
([`externals/bootstrap.sh#L216-L226`](https://github.com/cvmfs/cvmfs/blob/746d9e41c181ced0ca6bb348ace7c28891b0d0af/externals/bootstrap.sh#L216-L226)).
It never checks `$externals_install_dir/go/bin/go` — its own install target,
where the builddeps Go is already sitting — so `golang_rev2` is selected anyway
(output filtered to the relevant lines):

```
Bootstrap - Using default externals list: libcrypto sha3 golang_rev2
[…]
Downloading Go Binaries ...
Downloading https://go.dev/dl/go1.24.2.linux-amd64.tar.gz ...
[…]
--> Building golang_rev2
rm: cannot remove '/src/cvmfs/externals_install.x86_64.ubuntu24.04/bin/go': No such file or directory
ln: failed to create symbolic link '/src/cvmfs/externals_install.x86_64.ubuntu24.04/bin/go': No such file or directory
[…]
CMake Error at CMakeLists.txt:291 (message):
  bootstrapping failed
```

Besides the missing-`bin/` failure this PR fixes, that re-download is 74.8 MB
and it is a *different* Go:
[`externals/bootstrap.sh#L8`](https://github.com/cvmfs/cvmfs/blob/746d9e41c181ced0ca6bb348ace7c28891b0d0af/externals/bootstrap.sh#L8)
pins 1.24.2,
[`ci/bootstrap_golang.sh#L27`](https://github.com/cvmfs/cvmfs/blob/746d9e41c181ced0ca6bb348ace7c28891b0d0af/ci/bootstrap_golang.sh#L27)
pins 1.25.8, and the tarball is
[extracted into the same prefix](https://github.com/cvmfs/cvmfs/blob/746d9e41c181ced0ca6bb348ace7c28891b0d0af/externals/bootstrap.sh#L104-L110).
Reading `…/externals_install.x86_64.ubuntu24.04/go/bin/go version` in the same
container, before and after the cmake run:

| stage | reported version |
|---|---|
| after `build_install_builddeps.sh` | `go1.25.8 linux/amd64` |
| after `cmake …` (which fails)      | `go1.24.2 linux/amd64` |

So the bootstrapped toolchain is silently downgraded on the way to the error.

This may also be why the builddeps route looks clean elsewhere: with a system Go
≥ 1.23 on `PATH`, `golang_rev2` is never selected at all; and the packaging flow
has explicit wiring for the bootstrapped one —
[`ci/cvmfs/deb.sh#L120-L136`](https://github.com/cvmfs/cvmfs/blob/746d9e41c181ced0ca6bb348ace7c28891b0d0af/ci/cvmfs/deb.sh#L120-L136)
auto-detects `${_install_dir}/go/bin` and passes it to `debuild` via
`--prepend-path`, and
[`ci/cvmfs/rpm.sh#L69-L77`](https://github.com/cvmfs/cvmfs/blob/746d9e41c181ced0ca6bb348ace7c28891b0d0af/ci/cvmfs/rpm.sh#L69-L77)
prepends it to `PATH` directly. A plain `cmake` after builddeps — what
`doc/developer/30-building-from-source.md` describes — has no equivalent, so it
breaks precisely when there is no usable system Go, i.e. the case the bootstrap
exists for.

Since `bootstrap.sh` is on its way out this may well be moot shortly; flagging it
mainly because the documented from-source path currently fails even when the Go
bootstrap itself succeeds, in case that changes the priority. If an interim fix
is useful I would be glad to open a small PR teaching `externals/bootstrap.sh`
to accept `$externals_install_dir/go/bin/go`, mirroring the detection already in
`deb.sh`/`rpm.sh` — and either way, happy to re-test once the crypto-externals
rework lands.

vvolkl:
`ci/build_install_builddeps.sh` sets CVMFS_GO_DIR. The ci scripts that build packages respect that already. For a `cmake; make install` workflow we could do the following:

```diff
diff --git i/CMakeLists.txt w/CMakeLists.txt
index 9383533e2..f141b8b63 100644
--- i/CMakeLists.txt
+++ w/CMakeLists.txt
@@ -601,7 +601,19 @@ endif ()
 # First check if we have golang available
 #
 if (BUILD_GATEWAY OR BUILD_DUCC OR BUILD_SNAPSHOTTER)
-  find_program(GO_COMPILER go)
+  if (NOT CVMFS_GO_DIR AND NOT "$ENV{CVMFS_GO_DIR}" STREQUAL "")
+    set (CVMFS_GO_DIR "$ENV{CVMFS_GO_DIR}")
+  endif ()
+  find_program(GO_COMPILER go
+               HINTS ${CVMFS_GO_DIR} "${EXTERNALS_INSTALL_LOCATION}/go/bin")
   if(NOT GO_COMPILER)
     message(FATAL_ERROR "go compiler not found")
   else()
```

kadircanyildirm-crypto:
Thanks, that looks much cleaner than my bootstrap.sh idea. If it's OK with you,
I'll test it with a clean build and open a PR with this change.

andrey-utkin:
https://github.com/cvmfs/cvmfs/actions/runs/30695107606/job/91359485631?pr=4373

> /Users/runner/work/cvmfs/cvmfs/externals/bootstrap.sh: line 210: conditional binary operator expected

```
if [[ -v BUILTIN_EXTERNALS_LIST ]] && ! echo ";${BUILTIN_EXTERNALS_LIST};" | grep -qE ";(libarchive|vjson);"; then
```

I will work around this, but can we please have modern Bash on macos?

PR #4378, opened 2026-07-29
Fix metalink bugs

DrDaveD:
This fixes a few bugs I found with testing metalink now that I have actual servers to test against.
- Failures that happened on the first redirect were always considering the metalink lookup as the problem even if the problem was actually on the redirected server.
- A retry on a metalink server added to the existing host lists rather than resetting it from scratch.
- The index into the list of hosts was left at -1 instead of setting it to 0 after the redirect.

In addition, the "switching host"/"switching metalink" log message was missing the intended manager prefix.  I also added a new log message after metalink hosts are received, and made the "Trying again" debug message more intuitive by leaving out options that had false booleans.

### DIFF
--- a/cvmfs/network/download.cc
+++ b/cvmfs/network/download.cc
@@ -1424,6 +1424,17 @@ void DownloadManager::ProcessLink(JobInfo *info) {
   if (host_list.size() > 0) {
     SetHostChain(host_list);
     opt_metalink_timestamp_link_ = time(NULL);
+    // Set the index to the first linked host because the protocol includes
+    // redirecting to the first host
+    info->SetCurrentHostChainIndex(0);
+    // Don't process the same links again if there are retries
+    info->SetLink("");
+    LogCvmfs(kLogDownload, kLogDebug | kLogSyslog,
+                 "(manager '%s' - id %" PRId64 ") "
+                 "received %d hosts from metalink server, starting with %s",
+                 name_.c_str(), info->id(),
+                 host_list.size(),
+                 host_list[0].c_str());
   }
 }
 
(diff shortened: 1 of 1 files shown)


cvmfs/network/download.cc
```diff
@@ -1985,9 +2003,6 @@ Failures DownloadManager::Fetch(JobInfo *info) {
     info->GetHashContextPtr()->buffer = alloca(info->hash_context().size);
   }
 
-  // In case JobInfo object is being reused
-  info->SetLink("");
```
vvolkl, line 1989: @DrDaveD  I may have been a bit hasty in merging this. Are you sure that this line should be removed? Indeed JobInfo objects are long-lived and may be re-used, so I think it's correct to reset the links.

DrDaveD: It's not needed anymore because that is instead done in the `ProcessLink()` function immediately after they are processed.

vvolkl: Hmmm, even if that is the case, why remove it? That puts all the responsibility on ProcessLink to clear this correctly in all possible circumstances. And I don't think it does. It seems like it's correct for the expected usage, but I believe it'd be theoretically possible for any non-metalink server to set a Link header, which won't be cleared by ProcessLink. That will then lead to aggregations that seem very hard to debug.

DrDaveD: It's true that non-metalink servers could theoretically send `Link` headers, but ProcessLink is the only consumer of that information so it won't have any effect until a metalink request is done.  I guess you're worried about a case of a non-metalink request sending a `Link` header followed by a metalink request.

It seems to me that a better solution might be to ignore `Link` headers if it was not a metalink request.

vvolkl: Yes, and if the value won't be used it could grow, theoretically without bounds. For the patch release I'd be happy if we kept the explicit clear.

DrDaveD: Restored in #4382

PR #3062, opened 2022-09-14
adding test for S3-azureblob over https

hmeiland:
adding test to verify S3 azureblob usage over https; updating the main test script now first; not yet activated in the different test environments....

PR #4379, opened 2026-07-29
S3/Azure: add Managed Identity (MSI) bearer-token authentication for Azure Blob

hmeiland:
## Summary

Adds support for authenticating the S3/Azure Blob uploader with an Azure
**Managed Identity (MSI)** bearer token instead of a shared storage account key.

Today an Azure-Blob-backed Stratum 1 must store a long-lived
`CVMFS_S3_SECRET_KEY` (the storage account key) in plaintext on disk. On any
Azure VM with an assigned managed identity, this change lets the publisher
obtain a short-lived OAuth token from the Instance Metadata Service (IMDS) and
authenticate blob writes with `Authorization: Bearer …`, so **no storage key is
required**.

This complements the existing Azure Blob SharedKey support (#2590, #3062) — the
SharedKey path is unchanged and remains the default.

## What changed

New config keys (both optional):

| Key | Meaning |
|-----|---------|
| `CVMFS_S3_AZURE_MSI` | `true` enables MSI bearer auth; implies the Azure authz method and makes `CVMFS_S3_SECRET_KEY` optional |
| `CVMFS_S3_AZURE_MSI_CLIENT_ID` | Optional client id to select a specific user-assigned identity (omit for the system-assigned identity) |

Example Stratum 1 S3 config (no secret key):

    CVMFS_S3_HOST=<account>.blob.core.windows.net
    CVMFS_S3_ACCESS_KEY=<account>
    CVMFS_S3_BUCKET=<container>
    CVMFS_S3_DNS_BUCKETS=false
    CVMFS_S3_FLAVOR=azure
    CVMFS_S3_USE_HTTPS=true
    CVMFS_S3_AZURE_MSI=true

Implementation:

- **`cvmfs/network/s3fanout.{cc,h}`**
  - `S3Config` gains `azure_msi` and `azure_mi_client_id`.
  - New `RefreshAzureToken()` fetches an OAuth token for
    `resource=https://storage.azure.com/` from IMDS
    (`169.254.169.254`, header `Metadata: true`, `NOPROXY` for the link-local
    address). The token and its `expires_on` are cached and re-minted 300 s
    before expiry; access is mutex-guarded for the multi-threaded fanout.
    Response JSON is parsed with the existing `JsonDocument`.
  - `MkAzureAuthz()` gains an MSI branch emitting `x-ms-date`,
    `x-ms-version: 2017-11-09`, `Authorization: Bearer <token>` and
    `x-ms-blob-type: BlockBlob`. The SharedKey branch is untouched.

- **`cvmfs/upload_s3.{cc,h}`**
  - Parses the two new keys; `CVMFS_S3_AZURE_MSI=true` selects the Azure authz
    method and relaxes the secret-key requirement.

No new dependencies (uses the existing libcurl and `JsonDocument`).

## Testing

Validated end-to-end on an Azure VM with a system-assigned managed identity
granted **Storage Blob Data Contributor** on the target storage account:

- Built `cvmfs` (server) from this branch; warning-clean.
- `cvmfs_swissknife upload` with an MSI-only config (no secret key) uploaded a
  blob to a real Azure Blob container; independent read-back returned the exact
  payload (HTTP 200).
- Confirmed the emitted request carries a valid identity JWT
  (`aud=https://storage.azure.com/`), `x-ms-version: 2017-11-09` and
  `x-ms-blob-type: BlockBlob`.
- Token refresh/expiry path exercised; SharedKey path regression-checked
  (unchanged behavior).

## Notes / open questions

- Token resource + `x-ms-version` are currently fixed constants; happy to make
  them configurable if preferred.
- IMDS uses libcurl directly (consistent with the existing fanout); can wrap in
  the download manager instead if that's the house style.

### DIFF
--- a/cvmfs/network/s3fanout.cc
+++ b/cvmfs/network/s3fanout.cc
@@ -12,9 +12,11 @@
 #include <cassert>
 #include <cerrno>
 #include <cstring>
+#include <ctime>
 #include <utility>
 
 #include "crypto/hash.h"
+#include "json_document.h"
 #include "util/exception.h"
 #include "util/platform.h"
 #include "util/posix.h"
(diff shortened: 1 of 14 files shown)


vvolkl:
Hugo, thanks for the patch. That seems quite nice actually. I've looked into token-based authentication for publishing, and while AWS seems to have a `GetAccessToken`, our CEPH deployment doesn't support it, or at least it's never been tried out. So I'm currently looking into adding some kind of custom token provider to the gateway, but this is really because we are missing this functionality.

I guess without an azure VM this is not easily testable on our side, right? I'm happy to merge, but I'll be trusting you to test it.

hmeiland:
Pushed an additional commit to this branch: **`a4c356b` — s3: stage large Azure blobs via Put Block / Put Block List**.

### Why
A single-shot `Put Blob` is capped at **64 MiB** under SharedKey (`x-ms-version 2011-08-18`) and **256 MiB** under MSI (`2017-11-09`). CVMFS objects larger than that limit were rejected by Azure with **HTTP 413**, which blocked replication of larger catalogs/data objects to an Azure-blob Stratum-1.

### What
Oversized blobs are now uploaded as a sequence of `Put Block` requests (4 MiB SharedKey / 64 MiB MSI) followed by a `Put Block List` commit that re-applies the original blob content-type:

- New request types `kReqPutAzureBlock` / `kReqPutAzureBlockList`, driven on the existing easy handle via `VerifyAndFinalize` (each block re-initializes URL, signature and `INFILESIZE`).
- Fixed-width base64 block ids for unambiguous commit ordering; the signer uses the **unescaped** id, the request URL uses a **URL-escaped** id.
- `Put Block List` signs `x-ms-blob-content-type` (sorts before `x-ms-date` in CanonicalizedHeaders).
- Retries re-stream the same block window from offset 0, replacing the uncommitted block.
- Staging engages only when `authz_method == kAuthzAzure` and the object exceeds the single-PUT limit; small objects and the S3 path are unchanged.

### Validation
Ran on the eessi.io Azure-blob-backed Stratum-1 (SharedKey backend):

- Full seed of **14,495,247 processed chunks (1,449,878 newly uploaded)** across all `2023.06` / `2025.06` / `2026.06` arch + GPU-accel trees.
- **Zero HTTP 413s** across the entire seed log.
- Manifest published and served: `HTTP 200`, **revision 21625**.

Since this commit adds large-object staging on top of MSI auth, the PR now covers both Azure features — happy to update the PR title/description to reflect that if you'd prefer, or split it into a separate PR.

hmeiland:
Per reviewer preference, I have **split the large-blob staging work into its own PR: #4395**.

This PR (#4379) is now scoped to **Azure MSI bearer-token auth only** (commits `9ecd72c`, `618135e`); the previously-pushed staging commit `a4c356b` has been moved out and force-pushed off this branch.

#4395 is stacked on top of this one (it reuses `MkAzureAuthz` / `RefreshAzureToken` / `config_.azure_msi`), so it is opened as a **draft** and depends on #4379 merging first.

hmeiland:
Thanks @vvolkl — really appreciate the willingness to merge on trust. On your "not easily testable without an Azure VM" point: good news — **it can be tested in CI without any Azure VM**, and I've built a runnable harness to prove it, so you don't have to take the Azure path on faith.

**How:** the write side is emulated by **Azurite** (Microsoft's official storage emulator) run with `--oauth basic`, which validates the bearer token's `aud`/`iss`/`exp` but not its signature — exactly what a test double needs. It speaks the full Blob REST dialect, including the `Put Block` / `Put Block List` staging. The token side is a ~100-line stdlib **mock IMDS** that mints a JWT with the right audience (`https://storage.azure.com/`) and an `sts.windows.net` issuer.

**One small enabler:** `RefreshAzureToken()` currently hard-codes the IMDS endpoint at the link-local `169.254.169.254`, which is what makes it un-testable off-Azure. A tiny opt-in override does the trick — unset ⇒ byte-for-byte identical behaviour on a real VM:

```cpp
const char *imds_env = getenv("CVMFS_AZURE_IMDS_ENDPOINT");
const string imds_base = (imds_env && imds_env[0])
                         ? string(imds_env) : string("http://169.254.169.254");
string url = imds_base + "/metadata/identity/oauth2/token?api-version=2018-02-01"
             "&resource=https%3A%2F%2Fstorage.azure.com%2F";
```

With that in place, a CI job is just: start Azurite `--oauth basic` + the mock IMDS, point `CVMFS_AZURE_IMDS_ENDPOINT` at the mock, and publish. My harness replays the exact requests this PR emits and passes end-to-end:

- IMDS token fetch → `Bearer` container create → single-PUT `BlockBlob` → `Put Block` × N → `Put Block List` commit → byte-exact read-backs, all green.

That also exercises the large-object staging in #4395. (This is on top of the real-Azure production validation already noted above — a live `software.eessi.io` Stratum 1 on Azure Blob under MSI.)

I'm happy to **contribute this as a CI job** (an emulator smoke test that always runs, plus a full `cvmfs_server` publish wired into the existing S3/garage integration harness), and to stay the tester-of-record for the Azure path since you have no Azure CI. Would you like me to add it to this PR, or as a follow-up?

hmeiland:
Follow-up: I've pushed the harness itself onto this branch (`cb84fb0`) under [`test/azure-msi/`](https://github.com/cvmfs/cvmfs/tree/feature-azblob_msi/test/azure-msi) so it's runnable straight from the PR — no Azure VM needed:

```bash
cd test/azure-msi && ./run-local.sh   # Azurite --oauth basic + mock IMDS + verifier, self-cleaning
```

`verify_msi_path.py` replays the exact requests this PR emits — IMDS token fetch → `Bearer` `BlockBlob` PUT → `Put Block` / `Put Block List` → byte-exact read-back — and passes end-to-end. `README.md` there covers the details, plus a `publisher-e2e` job that does a real `cvmfs_server` publish against the emulator (it applies the small `CVMFS_AZURE_IMDS_ENDPOINT` override in `s3fanout-imds-endpoint.patch`; happy to inline that into `s3fanout.cc` instead if you prefer).

It's a plain directory for now — not wired into `.github/workflows/` — so it doesn't trigger anything on CI until you want it. Glad to move it into the S3 integration harness whenever suits.

PR #4384, opened 2026-07-31
rpm: remove json-devel dependency

vvolkl:
On el10 and newer, json-devel moved to epel-release. Since nlohmann-json is header-only, rather than
adding epel to the build dependencies, I think it's better to just use the vendored library.


vvolkl:
I left it as a system dependency for Debian since there it seems to be more generally available. Maybe it's better to be consistent, but I think it's fine either way.

PR #4390, opened 2026-08-01
Derive development and package versions from Git

vvolkl:
Development versions now use 
`dev{order}.{sequence_number}.{timestamp}.git{commit_hash[.dirty]}` 
e.g. 
`2.15.0~dev1.0.20260731145317.gitf8d9cd7e63`.

For proper releases everything should stay the same, and it's possible to override the generated version.

More details in `doc/developer/32-versioning.md`.


.github/workflows/ci_koji_mock.yaml
```diff
@@ -45,8 +45,10 @@ jobs:
 
               cp "$tarball" rpmbuild/SOURCES/
               cp packaging/rpm/cvmfs.te packaging/rpm/cvmfs.fc rpmbuild/SOURCES/
+              cp packaging/rpm/cvmfs-universal.spec rpmbuild/cvmfs-universal.spec
+              sed -i "s/^Version: .*/Version: ${version}/" rpmbuild/cvmfs-universal.spec
```
DrDaveD: I think this approach is problematic, requiring every process that builds a cvmfs rpm to repeat the process.  I would rather see packaging/rpm/cvmfs-universal.spec be renamed to packaging/rpm/cvmfs-universal.spec.in and make it a requirement to run a script to generate cvmfs-universal.spec and check it into cvmfs.   A CI check can verify that re-running the script creates the identical output.  I use that technique in several other packages.

Having said that, I can see why you would be hesitant to do that for just this one parameter.  I have my own rpm build script for development and it was easy enough for me to add an override of the Version since I was already overriding the Release.  In the other packages where I use a `.in` file there are quite a few more parameters that are generated.  In at least one of the cases the OpenSUSE Build System reads the .spec directly and there's no option there to apply a script (although there is an option there for defining rpm macros).  My build script overrides the Release number using a very similar algorithm as you're using for the Version, using a release of `0.0.$(date +%Y%m%d%H%M%S).$LOGNAME.$(git rev-parse HEAD)git`.

vvolkl: That's a good point, but I also think there's some disadvantages to doing a .in generation, often it's just convenient to be able to work with a spec file directly without a pre-processing step. I've now replaced the sed with a cvmfs_version rpm macro, and use a VERSION file for the default, let me know what you think about that.

DrDaveD: I guess that's a good compromise.  I wasn't aware that you could have an rpm macro that could execute a shell command.

vvolkl:
Thanks for the comments. Will merge now, we can iterate if needed (the next pre-release should show if that works out).

PR #4394, opened 2026-08-03
Drop support for libfuse2

### LINKED ISSUE #3886: Remove support for libfuse2
#3879 removed libfuse2 support for new platforms, but we should proceed in a minor release to remove it for all platforms.  That then removes the need for the separate cvmfs-fuse3 subpackage, moving the libcvmfs_fuse3 library to the main package.

vvolkl:
Includes a check so we don't reload any existing fuse2 mounts on upgrade. Maybe overkill.

Fixes #3886


vvolkl:
Still to do:

- [ ] Remove fuse3 package (after checking if that makes complications with puppet etc)

cvmfs/loader.h
```diff
@@ -198,7 +198,8 @@ struct LoaderExports {
   // As of CernVM-FS 2.7, this has been rebranded from
   // struct fuse_chan **fuse_channel  to
   // void **fuse_channel_or_session
-  // in order to work with both libfuse2 and libfuse3
+  // in order to work with both libfuse2 and libfuse3.  Since the removal of
+  // libfuse2 support it always points to a struct fuse_session *.
   void **fuse_channel_or_session;
```
DrDaveD: Why not rename this to `fuse_session`?

DrDaveD (approved):

cvmfs/loader.h
vvolkl: Good catch, done.

Issue #4327, opened 2026-07-15
`deprecated-enum-enum-conversion` Compiler warnings in logging system


vvolkl:
Gcc 16.2 complains about:

```
cvmfs/util/logging_internal.h:116:36: warning: bitwise operation between different enumeration types ‘LogFacilities’ and ‘LogFlags’ is deprecated [-Wdeprecated-enum-enum-conversion]
  116 | const int kLogInfoMsg = kLogStdout | kLogShowSource | kLogInform;
```

To be silenced, or adressed in as larger refactoring of the logging system.

PR #4396, opened 2026-08-03
logging: silence deprecated enum-enum conversion warnings

kadircanyildirm-crypto:
With your permission, I would like to try this.

gcc 16 (defaulting to a newer C++ standard) warns with
-Wdeprecated-enum-enum-conversion for every log mask combining
LogFacilities, LogFlags and LogLevels, as reported in #4327. These
enums intentionally occupy disjoint bit ranges of one integer mask,
so this adds the six mixed-type operator| overloads centrally in
logging_internal.h: all ~50 call sites compile warning-free in both
C++17 and C++20 mode, with no behavior change.

Verified on Ubuntu 24.04 / gcc 13.3: a probe including the header
goes from 5 warnings to 0 with -std=gnu++20 and stays at 0 with
gnu++17; clang-format reports no changes and the T_Logging unit
tests pass.


manishpaulish:
I had a fix for #4327 ready when I found this PR, and the approach turned out to be identical, six `operator|` overloads returning `int`. Before dropping mine I tested this branch, and there is one thing I think is worth catching before it merges.

**The overloads themselves look right.** All six orderings, `int` result matching the mask parameter, and it keeps every call site unchanged, which is much better than casting at each of them. On this branch the `-Wdeprecated-enum-enum-conversion` count across the tree goes to zero.

**But `kLogWarning`, `kLogInfoMsg` and `kLogVerboseMsg` stop being compile-time constants.** They are declared just below the operators:

```cpp
const int kLogWarning = kLogStdout | kLogShowSource | kLogNormal;
```

Once `operator|` is an ordinary function rather than a built-in, that initialiser is no longer an integral constant expression in C++03, which is the standard CONTRIBUTING.md pins. Since this header is included across essentially the whole code base, every translation unit picks up a static initialiser.

Compiling `int GetWarning() { return kLogWarning; }` at `-O2`:

| | `GetWarning()` | static initialisers | still a constant expression |
|---|---|---|---|
| `devel` | `mov w0, 9218` | 0 | yes |
| this branch | `adrp` + `ldr` | 4 | no |

The constant expression part is checkable directly:

```cpp
char probe[kLogWarning == 9218 ? 1 : -1];
```

compiles on `devel` and fails on this branch with `size of array 'probe' is not an integral constant-expression`.

To be clear about what this is and is not: the mask values are unchanged, so nothing logs differently, and nothing in the tree currently uses these three in a constant expression, so it is not a build break. It is a silent change to three exported constants that no warning reports, which is why I thought it worth raising rather than leaving to be found later.

**The fix is small.** Keeping an explicit cast on the first operand selects the built-in operator, so those three stay foldable:

```cpp
const int kLogWarning =
    static_cast<int>(kLogStdout) | kLogShowSource | kLogNormal;  // NOLINT
```

That restores `mov w0, 9218` and zero static initialisers, with the warning count still at zero. The `// NOLINT` is for cpplint's indent-within-namespace rule, which fires on the wrapped continuation line; CONTRIBUTING.md already sanctions that idiom for `using namespace std;`.

Happy to push that as a commit to this branch if that is easiest, or you can fold it in directly, whichever you prefer.

One other thing I checked while I was in there, in case it is useful: `cvmfs/authz/helper_log.h` has its own `kLogAuthz*` mask, but it uses `const unsigned` rather than an enum, so it is unaffected. That accounts for the rest of the logging surface.

kadircanyildirm-crypto:
Good catch, you are right. Folded it in as 5b07e09, with you credited as co-author on the commit.

Before pushing I checked the rest of the surface: those three are the only `const int` in the tree initialised from an enum-enum `|`, and none of them is used anywhere a constant expression is required, so it lines up with what you found. I also confirmed the `// NOLINT` is needed — without it cpplint reports `whitespace/indent_namespace` on all three continuation lines. I added a short comment above the block so the cast does not get tidied away later by someone who reads it as redundant.

Thanks for building the branch and measuring it instead of just opening a competing PR.

vvolkl (approved):
This is a very nice solution, thank you!

PR #4403, opened 2026-08-18
Fix podman additional-store CI validation

vvolkl:
Use explicit storage config; root podman ignored the HOME override.


vvolkl:
@christge-sft That should fix the corresponding CI failures on the open PRs, please rebase the UniquePtr PR.

PR #4405, opened 2026-08-19
Add CUE Configuration Schema

sotiriskavadakis:
Adds a CUE schema for the client and server configuration, plus three
subcommands built on it:

- `cvmfs_config validate [repository]` checks a merged client configuration
- `cvmfs_config explain <parameter>` prints its type, default and description
- `cvmfs_server validate <repository>` does the same for a repository's
  server.conf
- `cvmfs_config generate` writes a `/etc/cvmfs/default.local` tuned to the node's specs
- 
## Examples

An alien cache configured alongside a shared cache and a managed quota:

```
$ cvmfs_config validate atlas.cern.ch
#ClientConfig._rule_alien_cache_conflicts_with_shared_cache: conflicting values false and true
```

```
$ cvmfs_config explain CVMFS_QUOTA_LIMIT
NAME
       CVMFS_QUOTA_LIMIT - client configuration parameter, optional, default 1024

DESCRIPTION
       Soft-limit of the cache in Megabyte.
```

## Layout

- `cvmfs/config/cvmfs_client.cue` declares 152 client parameters
- `cvmfs/config/cvmfs_server.cue` declares 123 server parameters
- `cvmfs/config/gen_client.cue` derives client parameters from node facts
- `cvmfs/config/main.go` embeds all three with `go:embed` and validates the JSON
  from `cvmfs_config showconfig -j`

## Notes

- Values arrive from `showconfig -j` as strings, so the validator preprocesses
  them into the expected types before unifying.
- Built with `-DBUILD_CONFIG_VALIDATOR=ON`, which is off by default and
  requires Go.

## Tests

- `test/src/115-clientconfigvalidate`, `test/src/116-serverconfigvalidate` and
  `test/src/117-clientconfiggenerate` cover the three subcommands


cvmfs/config/cvmfs_client.cue
```diff
@@ -0,0 +1,321 @@
...
+#URL: =~"(?i)^https?://[^ \\t;|,]+$"
+
+// Chain of HTTP(S) URLs: ';' separates failover groups,
+// '|' separates load-balanced members within a group.
+#URLChain: =~"(?i)^https?://[^ \\t;|]+([;|]https?://[^ \\t;|]+)*$"
+
+// Proxy chain: ';' separates failover groups, '|' load-balanced members.
+// An entry is DIRECT (no proxy), auto (WPAD/PAC discovery) or a proxy
+// address with optional scheme and port.
+_#proxyEntry: "(DIRECT|auto|(https?://)?[A-Za-z0-9._-]+(:[0-9]{1,5})?)"
```
vvolkl: Why not use `#Port`?

cvmfs/config/gen_client.cue
```diff
@@ -0,0 +1,47 @@
...
+	// zero when the network could not be measured
+	rtt_ms: #UInt | *0
+}
+
+facts: #Facts
+
+// The parameters to write, in this order.
+tuned: {
+	// Take a quarter of the free disk but no more than 20000MB
+	CVMFS_QUOTA_LIMIT: list.Min([20000, quo(facts.cache_avail_mb, 4)])
```
vvolkl: These heuristics are all a bit problematic I think. A max of 20 GB is way too small, the bigger nodes are routinely using 40-50. If we do need a max, let's set it to 128 GB for now.

cvmfs/config/gen_client.cue
```diff
@@ -0,0 +1,47 @@
...
+	// Take the file-descriptor limit the cvmfs service runs under, leave 1024 spare, but no more than 131072
+	CVMFS_NFILES: list.Min([facts.nfiles_max-1024, 131072])
```
vvolkl: Max is also way too small, and I'm not sure about the logic of `facts.nfiles_max-1024`. `nfiles_max` is taken from  autofs? This one will need some dedicated benchmarking to get right. We'll probably want to use the number of cores as a proxy for the number of repos and files from cvmfs that will be opened. I suggest to comment it and leave it as a TODO.

cvmfs/cvmfs_config
```diff
@@ -1117,6 +1126,153 @@ cvmfs_showconfig() {
...
+  [ -x "$validator_bin" ] || validator_bin="/usr/lib/cvmfs/cvmfs_validator"
+  [ -x "$validator_bin" ] || die "no validation engine available; build \
+CernVM-FS with -DBUILD_CONFIG_VALIDATOR=ON"
+
+  cvmfs_readconfig || die "Failed to read CernVM-FS configuration"
+
+  # server URLs and proxies are per repository, so choose one to test them out first
+  local fqrn=${CVMFS_REPOSITORIES%%,*}
+  [ -n "$fqrn" ] || fqrn=$CVMFS_CONFIG_REPOSITORY
+  [ -z "$fqrn" ] || cvmfs_readconfig "$fqrn" || true
```
vvolkl: Not sure about the `|| true`. CVMFS_REPOSITORIES is supplied by the user, so can contain errors and typos. What happens in that case?

cvmfs/server/cvmfs_server_validate.sh
```diff
@@ -0,0 +1,104 @@
...
+#
+# Implementation of the "cvmfs_server validate" command
+
+# This file depends on functions implemented in the following files:
+# - cvmfs_server_util.sh
+# - cvmfs_server_common.sh
+
+
+# Ensure a string is properly escaped for JSON output.
+_validate_json_string() {
```
vvolkl, line 14: This seems to escape less chars than json_string() from cvmfs_config. Not sure if there's an easy way to refactor, but should at least do the same thing.

cvmfs/server/cvmfs_server_validate.sh
```diff
@@ -0,0 +1,104 @@
...
+  local server_conf="/etc/cvmfs/repositories.d/${name}/server.conf"
+  [ -r "$server_conf" ] || die "Cannot read $server_conf"
+
+  # Build a JSON object from server.conf, read strictly from the file (not
+  # the ambient shell) so a stray CVMFS_*/X509_* variable in the caller's
+  # environment can't leak into the config being validated. Empty values
+  # count as unset and are dropped, matching the config's own convention.
+  local varname value sep=""
+  local json="{"
+  for varname in $(grep -o '^[A-Za-z0-9_]*=' "$server_conf" | tr -d '=' \
```
vvolkl, line 60: What happens if there are duplicated values in server.conf? Will the validator handle it correctly?

vvolkl (changes requested):
Very nice work overall! Some comments inline. The heuristics for the config generation still need some work, but it's already great to have the implementation here. We'll also need to think of a way to keep the config parameters in sync when adding future ones (I think I've already added a new one on devel).

cvmfs/config/cvmfs_client.cue
sotiriskavadakis: Using `#Port` directly wasn't possible as it's an int constraint. Introduced a helper port regex identifier and integrated it into the `_#proxyEntry` regex.

cvmfs/config/gen_client.cue
sotiriskavadakis: set max to 128GB + increased to take half of the available disk.

cvmfs/config/gen_client.cue
sotiriskavadakis: commented it out for now.

cvmfs/cvmfs_config
sotiriskavadakis: First of all now i first use the helper `mkfqrn` function to expand names (`atlas` -> `atlas.cern.ch`), and then if it can't find `CVMFS_SERVER_URL` (in one of the per-domain config files), it warns the user and doesn't write parameters that need access to a valid repo.

cvmfs/server/cvmfs_server_validate.sh
sotiriskavadakis: Now it's doing the same conversions, but because it runs with /bin/sh and not bash, had to use different syntax

cvmfs/server/cvmfs_server_validate.sh
sotiriskavadakis: added `sort -u` so the names are deduplicated

Issue #4402, opened 2026-08-17
cvmfs build missing for sle16


flaruina:
Hello, 
following the instructions to install cvmfs [here](https://cvmfs.readthedocs.io/en/stable/cpt-quickstart.html#linux-officially-supported) on a opensuse 16 machine I noticed the repo for sle16 is not available.

There is no reference to sle16 in [cvmfs-universal.spec ](https://github.com/cvmfs/cvmfs/blob/devel/packaging/rpm/cvmfs-universal.spec), would it be possible to add support for sle16/opensuse16?

vvolkl:
Thanks for the patch! cvmfs-2.14.1-1.sle16 is now available on the package repo ( can be set up with the usual quickstart instructions at https://cvmfs.readthedocs.io/en/stable/cpt-quickstart/ ) - tested in my opensuse/leap:16 test container, let me know if you have any troubles with it.

flaruina:
@vvolkl  tested on a opensuse 16 laptop with success (selinux labels seem ok too). 

If any issue arise, I'll let you know. Thanks!

PR #4406, opened 2026-08-20
Spec to build for sle16 with selinux

flaruina:
Some edits to the spec file to add support for suse linux 16 and selinux (enforcing by default in sle16).
Hope it helps with #4402


vvolkl (approved):
Thanks for the patch, I'll try to build and publish 2.14.1-2 packages. I think in general we can just use suse_version directly instead of the sleXX macros, but I'll refactor that in another PR.

vvolkl (approved):

PR #4408, opened 2026-08-21
fix: handle mountpoint paths in showconfig

### LINKED ISSUE #4080: weird `showconfig` output when passing full mount point
```
$ cvmfs_config showconfig /cvmfs/software.eessi.io
CVMFS_REPOSITORY_NAME=/cvmfs/software.eessi.io
CERNVM_GRID_UI_VERSION=
CVMFS_ALIEN_CACHE=
CVMFS_ALT_ROOT_PATH=
sed: -e expression #1, char 10: unknown option to `s'
sed: -e expression #1, char 11: unknown option to `s'
...
```

Correct use would be `cvmfs_config showconfig software.eessi.io`, but this is an easy mistake to make...

jeetrex17:
`cvmfs_config showconfig /cvmfs/software.eessi.io` currently treats the full mountpoint as the repository name. This also breaks the plain output because the path is passed to `sed`.

Strip `CVMFS_MOUNT_DIR` from mountpoint arguments before loading the repository configuration. Trailing slashes are tolerated and invalid paths abort with an error naming the argument. Plain output now uses shell parameter expansion, as the JSON output already does.

The new test exposed a pre-existing bug: in short mode the last command of the output loop can be a failed `[ ... ]` condition, so `showconfig -s` printed the correct output but exited 1. The plain output branch now returns 0 explicitly, as the JSON branch already does.

Test 046 covers repository names and mountpoints in regular, JSON, and short output, including trailing slashes, invalid paths, and exit codes.

Fixes #4080

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -266,7 +266,8 @@ cvmfs_config_usage() {
   fi
   echo "  chksetup"
   echo ""
-  echo "  showconfig [-s(hort output, only parameters set in the configuration)] [-j|--json] [<repository>]"
+  echo "  showconfig [-s(hort output, only parameters set in the" \
+       "configuration)] [-j|--json] [<repository>|<mountpoint>]"
   echo ""
   echo "  stat [-v | <repository>]"
   echo ""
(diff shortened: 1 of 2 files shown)


vvolkl (approved):
Thank you for this nice contribution! LGTM. I only added some quoting to the logging on fatal error, so the user-supplied strings don't get expanded in the error message.

PR #4410, opened 2026-08-22
Direct-to-S3 ingest for a mountless publisher, with optional support for  Stratum-1 prewarming

pbuncic:
This adds an **ingest path that uploads data objects straight to S3** while the gateway handles only catalogs, so a **mountless publisher** (`cvmfs_server connect-gw -P`) can publish and re-publish a repository without moving object payloads through the gateway. It also lets the publisher learn exactly which objects reached S3, so Stratum-1 replicas can be pre-warmed before the catalog flips.

Everything is opt-in. A stock gateway and the existing `cvmfs_server ingest` behaviour are unchanged; the new behaviour is reached only through the added flags (`-3` direct-to-S3, `--object-list`, catalog fast-delete).

## Motivation

The classic ingest path funnels every object's chunk/compress/upload **and** the catalog commit through the single serial gateway. Offloading object upload to S3 leaves the gateway only the catalog work, which is what lets object throughput scale independently of the gateway. Doing that on a **mountless** publisher surfaced three gaps that this branch closes: the receiver could not create the parent directories of a lease path it had never mounted; a re-publish into an occupied path died on the catalog `UNIQUE` constraint because a mountless delete had no filesystem to classify entries; and the publisher had no way to enumerate the objects it had handed to S3 for Stratum-1 prewarming.

## What this adds

**Direct-to-S3 object upload (`-3`).** `cvmfs_swissknife ingest -3` sends data objects to S3 via a `GatewayS3Uploader` while catalogs still go through the gateway. `-3` **fails closed** if the spooler is not a gateway spooler rather than silently falling back. `SessionContext` gains `DiscardBucket()` so a bucket whose content was uploaded out-of-band is released instead of being carried (and kept allocated) in every subsequent `ObjectPack`. 

**Stratum-1 prewarming (`--object-list <fd>`).** Ingest can report every data object it confirmed into S3 over an inherited pipe, so the publisher can announce those objects to the Stratum-1 replicas and have them pre-fetch before the new catalog is served. Requires `-3` (the objects must be in S3 to enumerate).

**Mountless publisher support.** The receiver can optionally create the missing parent directories of a gateway lease path (walking the ancestor chain once per publish, only when a parent is actually missing, and without stamping synthesised directories with a 1970 mtime). `WritableCatalogManager::HasDirectory` resolves the parent rather than the child, and the sync mediator skips an unmaterialised directory the catalog already has.

**Republish on a mountless publisher (catalog fast-delete).** For a fast delete of a nested-catalog mountpoint the catalog is the authority — a transition point is by definition a directory — so `SyncMediator::Remove` dispatches the delete on the catalog when the filesystem cannot type the entry. Without this a mountless `ingest -f <path>` "succeeds" but removes nothing, and the re-ingest then fails on `catalog.md5path UNIQUE`. The fast delete is restricted to a **leaf** nested catalog — one that mounts no child catalogs (`self.nested_catalogs == 0`, read fail-closed via `LookupCounters`). A structural node (the repository root, or a directory aggregating many published catalogs) is never a leaf, so a request naming a top-level path cannot wipe everything beneath it; a published entry is a
single dedicated leaf catalog.

**Supporting fixes.** CMake refuses to configure a build that defines `NDEBUG`, which would strip `assert()`s that carry side effects the code relies on — among them `cvmfs_server ingest`'s libarchive format registration (without it ingest fails with "No formats registered") and the client's privilege drop. The ingest sources use `std::unique_ptr`. Docs for the ingest workflow, build-from-source, and mountless ingest are added.

## Backward compatibility

Every entry point is opt-in — a new flag or an added receiver capability — and no existing endpoint, request body, or default flag behaviour changes. A stock gateway simply lacks the new frontend behaviour; a stock `cvmfs_server ingest` without `-3`/`--object-list` publishes exactly as before. There is no change to the gateway's commit serialisation, on-disk formats, or the client read path.

## Testing

Exercised end to end on a gateway/publisher testbed against an S3 (minio) backend: a full 170-package software stack (O2 + dependencies) published via direct-to-S3 ingest, and re-published into the now-occupied paths — the catalog fast-delete removes the old subtree and the re-ingest commits. The fast-delete commit's negative control is the "removes nothing" refusal previously observed on the same mountless container; the positive control is a clean republish after the fix. Stratum-1 prewarming was verified by the publisher announcing the `--object-list` objects and the receivers fetching them ahead of the catalog flip.


vvolkl:
I think the object list is a nice addition. The "supporting fixes" should go into separate PRs though, I already started to split them off. The assert handling is a good catch, that should be properly adressed.

vvolkl:
A proper fix for the issue raised in commit [d2c9f61](https://github.com/cvmfs/cvmfs/pull/4410/commits/d2c9f61f8702e3c17b7ab44eae7098297c617b00) is now in https://github.com/cvmfs/cvmfs/pull/4413

vvolkl:
I've now rebased this branch on devel (copy of previous state still here: https://github.com/cvmfs/cvmfs/tree/archive/ingest-direct-s3-objectlist),  removing the unrelated fixes, and harmonized it with #4316 in particular.

I think this is ok to merge, but I want to clearly declare this as an experimental feature. Ultimately I think it'll be better not to have the s3 credentials outside of the gateway. We still need to do some benchmarking of other possible solutions like handing out pre-signed urls from the gateway to the publisher (if it's done in bulk maybe there is actually negligible overhead), that have less potential for operational headaches. We're also still trying to find a good general solutions for authorizing publications when there are no dedicated per-repository publishers that will possibly make this direct-graft idea easier too.

Some more cleanup and detailed review is necessary too. The overly verbose and not very descriptive comments don't really fit into the codebase.

PR #4411, opened 2026-08-24
catalog: name the row when an entry cannot be added

vvolkl:
Include the entry path, parent, catalog, and SQLite error in AddEntry failures, instead of just having a bare assert.


vvolkl:
From #4410

PR #4413, opened 2026-08-24
Replace all asserts that have side effects

vvolkl:
This PR removes all asserts with production logic and replaces them with exceptions or other error handling. This prevents issues with builds where asserts are compiled out.

### DIFF
--- a/cvmfs/util/capabilities.cc
+++ b/cvmfs/util/capabilities.cc
@@ -82,15 +82,33 @@ bool ClearPermittedCapabilities(const std::vector<cap_value_t> &reservecaps,
   const int nreservecaps = (int) reservecaps.size();
   const int ninheritcaps = (int) inheritcaps.size();
 
-  if (!SetpcapCapabilityPermitted()) {
-    if (nreservecaps > 0) {
-      LogCvmfs(kLogCvmfs, kLogDebug,
-        "Capabilities cannot be reserved because setpcap is not permitted.");
+  {
+    cap_t caps_proc = cap_get_proc();
+    if (caps_proc == NULL) {
+      LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
+               "Failed to get process capabilities (errno: %d)", errno);
       return false;
     }
-    LogCvmfs(kLogCvmfs, kLogDebug,
-      "Capabilities are already cleared because setpcap is not permitted.");
-    return true;
+    cap_flag_value_t cap_state;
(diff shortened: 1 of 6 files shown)


vvolkl:
FYI @andrey-utkin, since this touches ingestsql

cvmfs/util/capabilities.cc
```diff
@@ -82,15 +82,33 @@ bool ClearPermittedCapabilities(const std::vector<cap_value_t> &reservecaps,
   const int nreservecaps = (int) reservecaps.size();
   const int ninheritcaps = (int) inheritcaps.size();
 
-  if (!SetpcapCapabilityPermitted()) {
-    if (nreservecaps > 0) {
-      LogCvmfs(kLogCvmfs, kLogDebug,
-        "Capabilities cannot be reserved because setpcap is not permitted.");
+  {
```
DrDaveD: This really makes the code blow up in size, I don't like it very much.  And it doesn't handle __APPLE__.  

Why can't this continue to call SetpcapCapabilityPermitted?  The asserts have been removed from it.

If you really are going to remove the use of that function, you should also remove its definition in this file and the header file. 

There's no reason to still include `<cassert>` in this file.

cvmfs/util/capabilities.cc
```diff
@@ -82,15 +82,33 @@ bool ClearPermittedCapabilities(const std::vector<cap_value_t> &reservecaps,
...
+    cap_t caps_proc = cap_get_proc();
+    if (caps_proc == NULL) {
+      LogCvmfs(kLogCvmfs, kLogSyslogErr | kLogDebug,
+               "Failed to get process capabilities (errno: %d)", errno);
```
DrDaveD: If the problem that this PR is trying to solve is that asserts are compiled out, can it instead invoke a function to die like `PANIC()` or `abort()` instead of returning false?  I'm worried it will cause other bad side effects if allowed to continue.

DrDaveD (changes requested):

vvolkl:
Thanks for the comments, I agree with all of those. Added a commit to adress them.

vvolkl:
@DrDaveD I believe now all comments are adressed

DrDaveD:
@vvolkl There are still 9 cases in ClearPermittedCapabilities where an assert was turned into a log message and return false.  I think it would be better if those were direct PANICs.  Currently they'll all turn into a PANIC because everywhere that ClearPermittedCapabilities returns false PANIC is called, but there's no guarantee that will be the case in the future.

PR #4416, opened 2026-08-24
fix: make cvmfs_server check -r reconstruct reflogs and exit

### LINKED ISSUE #3206: Add UI to reconstruct reflog
Reconstructing the reflog from scratch can be a necessary operation in some cases (e.g., adding a gateway to long-lived repository, see https://cernvm-forum.cern.ch/t/error-on-publisher/283/). The code is there (`cvmfs_swissknife reconstruct_reflog`) but it is not accessible from the UI.

### LINKED ISSUE #3758: cvmfs_server check -r runs too long
Many times reflog problems are immediately fixed with `cvmfs_server check -r` but then it goes on to run a full check when that's not what the admin is interested in.  It should exit after it fixes the reflog.

jeetrex17:
`cvmfs_server check -r` currently repairs reflog problems and then continues with a full repository check. It also reconstructs a missing `.cvmfsreflog` only when the manifest already contains a reflog hash, which excludes legacy repositories.

Make `check -r` stop after reflog handling and reconstruct missing reflogs even without a manifest `Y` entry. Reconstruction failures now return a non-zero exit code. Repair-only runs do not update full-check status, and `check -a -r` is not limited by the normal check interval.

Options that apply only to a full repository check (`-i`, `-t`, and `-s`) are rejected when used with `-r`.

Tests 519, 634, and 688 cover legacy reconstruction, early exit, invalid option combinations, and `check -a -r`.

Fixes #3206
Fixes #3758


cvmfs/server/cvmfs_server_check.sh
```diff
@@ -64,6 +64,12 @@ __do_check() {
   fi
 
   # do it!
+  if [ $repair_reflog -eq 1 ]; then
+    echo "Checking reference log of ${name}..."
+    __check_repair_reflog $name
```
vvolkl, line 69: `check -r` now returns directly from `__check_repair_reflog`. That function hashes
   `curl ... | cvmfs_publish hash`; without pipefail, a failed curl is masked because
   `cvmfs_publish hash` successfully hashes empty input. It then overwrites
   reflog.chksum with the empty-content hash and returns success. Previously, the
   subsequent integrity check would fail. Check the download/pipeline status before
   updating the checksum, or validate the repaired reflog before returning

jeetrex17: Thanks, you're right with the early exit, the subsequent integrity check no longer caught this. I split the download and hash into separate checked steps, so either failure returns non-zero before reflog.chksum is touched. I also added a regression test in 634 that makes the backend reflog unreadable to Apache (while peek still succeeds) and verifies that check -r fails with the checksum unchanged.

vvolkl:
One more minor comment: We have a -x flag to set the temp dir to locations other than ${CVMFS_SPOOL_DIR}/tmp . In some disaster (disk full) scenarios that can be very useful, and it'd be great if that could be used with check -r too. That's not a blocker for this PR though, we can add that at some point.

jeetrex17:
Cool , I've updated the PR so the temporary reflog download uses the `-x` scratch directory too. Test 634 now covers both a missing scratch directory (repair fails, checksum untouched) and a successful repair with a valid one.

vvolkl (approved):
Very nice, when having to restore the reflog it was somewhat inconvenient to either abort the check -r after a few seconds or wait for the full check to complete. Thanks again for the patch!

PR #4419, opened 2026-08-27
fix: add showconfig --all

### LINKED ISSUE #4098: cvmfs_config showconfig currently doesn't show variant symlink configuration
In the `software.eessi.io` repository, there is a variant symlink `EESSI_HOST_INJECTIONS`. I'd like to query `cvmfs_config showconfig` to figure out where it points to (e.g. to figure out if the client overrides the default). However, variant symlinks are currently not shown by `cvmfs_config showconfig`. E.g.:

```
$ cat /etc/cvmfs/default.local | grep EESSI_HOST_INJECTIONS
EESSI_HOST_INJECTIONS=/gpfs/admin/hpc/sw/EESSI/
$ ls -al /cvmfs/software.eessi.io/host_injections
lrwxrwxrwx 1 cvmfs cvmfs 25 Oct  3  2023 /cvmfs/software.eessi.io/host_injections -> /gpfs/admin/hpc/sw/EESSI/
$ cvmfs_config showconfig software.eessi.io | grep -i EESSI_HOST_INJECTIONS
$
```

Talking with @vvolkl on Slack, it turns out that anything not starting with `CVMFS_` is filtered out by `showconfig`. It'd be really helpful for me to be able to check variant symlink configuration. @vvolkl commented that changing the current behavior might break things for people that rely on existing behavior, but having it behind an extra flag should probably be feasible. That seems totally reasonable to me, and would work without any issue for my use case.

jeetrex17:
`cvmfs_config showconfig` currently hides configured variables that do not start with `CVMFS_`, including variables used by variant symlinks such as `EESSI_HOST_INJECTIONS`.

Add `-a` / `--all` to include all parameters returned by the configuration parser. This is opt-in, so the existing output remains unchanged. The flag also works with `-s`, `-j`, mountpoint paths, and the no-repository form.

Extended `046-defaultd` to cover the new flag and unchanged default behavior.

Fixes #4098

### DIFF
--- a/cvmfs/cvmfs_config
+++ b/cvmfs/cvmfs_config
@@ -994,35 +994,39 @@ json_string() {
   printf '"%s"' "$s"
 }
 
+cvmfs_showconfig_all_requested() {
+  local option
+  for option in "$@"; do
+    case "$option" in
+      -a|--all)
+        return 0
+      ;;
+    esac
+  done
+  return 1
+}
+
 cvmfs_showconfig() {
   local fqrn
   local org
   local short_output=0
   local json_output=0
-  local all_output=0
-  local short_opt=
-  local all_opt=
-  # Loop over the output flags and the repository name. Flag order does not
-  # matter.
-  while [ $# -gt 0 ]; do
-    case "$1" in
+  local option
+  # Keep the arguments intact so -a/--all can be checked after configuration
+  # variables have been evaluated in this dynamic scope.
+  for option in "$@"; do
+    case "$option" in
       -s)
         short_output=1
-        short_opt="-s "
-        shift 1
       ;;
       -a|--all)
-        all_output=1
-        all_opt="-a "
-        shift 1
+        continue
       ;;
       -j|--json)
         json_output=1
-        shift 1
       ;;
       *)
-        org=$1
-        shift 1
+        org=$option
       ;;
     esac
   done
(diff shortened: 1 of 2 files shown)


vvolkl (changes requested):
From an AI review:

> The new `-a/--all` flag introduces locals `all_output`/`all_opt` in `cvmfs_showconfig` (cvmfs/cvmfs_config:999-1002). Because bash `local` is dynamically scoped, the second call to `cvmfs_readconfig "$fqrn"` (line 1066) runs `eval "$CVMFS_PARMS"` inside `cvmfs_showconfig`'s own stack frame — and config files parsed via `BashOptionsManager::ParsePath` (cvmfs/options.cc) can legitimately contain arbitrary non-`CVMFS_`-prefixed assignments. If any sourced config file (default.conf, domain.d/*, config.d/*, or .local overrides) happens to define a variable literally named `all_output`, it silently overwrites the flag right before the `if [ $all_output -eq 0 ]` check at line 1075 — so `-a` can silently stop working (or get silently forced on), with no error.

jeetrex17:
Yep, that review is correct, I reproduced the collision with both `all_output=0` and `all_opt=`. Fixed in d3197abb4 by keeping the original arguments intact and checking `-a` / `--all` directly. I also added regression coverage for the direct and recursive forms.

vvolkl (approved):
Thank you very much, looks good to me now!

PR #4424, opened 2026-08-31
Add missing cap library to stubs

barskern:
Symbols from libcap are used in capabilities.{h,cc}, but the library is not explicitly linked. This was giving me undefined symbols errors for e.g. `cap_get_proc` when trying to load the stubs.

```
Aug 31 12:13:21 host1 cvmfs2[44625]: Error: failed to load cvmfs library, tried: './libcvmfs_fuse3_stub.so' '/usr/lib/libcvmfs_fuse3_stub.so' '/usr/lib64/libcvmfs_fuse3_stub.so' './libcvmfs_fuse_stub.so' '/usr/lib/libcvmfs_fuse_stub.so' '/usr/lib64/libcvmfs_fuse_stub.so'
                                     ./libcvmfs_fuse3_stub.so: cannot open shared object file: No such file or directory
                                     /usr/lib/libcvmfs_fuse3_stub.so: undefined symbol: cap_get_proc
                                     /usr/lib64/libcvmfs_fuse3_stub.so: undefined symbol: cap_get_proc
                                     ./libcvmfs_fuse_stub.so: cannot open shared object file: No such file or directory
                                     /usr/lib/libcvmfs_fuse_stub.so: cannot open shared object file: No such file or directory
                                     /usr/lib64/libcvmfs_fuse_stub.so: cannot open shared object file: No such file or directory
```

For more context, this problem arised when I was adapting the provided AUR PKGBUILD to 2.14.1 and my system, trying to build CVMFS following the ArchLinux package conventions:

<details>

<summary>PKGBUILD adapted from https://aur.archlinux.org/packages/cvmfs</summary>

```
# Maintainer: Frank Siegert < frank dot siegert at googlemail dot com >
# Contributor: Sebastien Binet binet-at-cern-ch
# Maintainer: Wainer Vandelli <wainer dot vandelli at gmail dot com>
# Contributor: Konstantin Gizdov < arch at kge dot pw >
# Contributor: Lin Further
# Contributor: vavolkl
# Contributor: barskern
pkgname=cvmfs
pkgver=2.14.1
pkgrel=1
pkgdesc="A client-server file system implemented in FUSE and developed to deliver software distributions onto virtual machines in a fast, scalable, and reliable way."
arch=('x86_64')
url="http://cernvm.cern.ch/portal/filesystem"
license=('BSD-3-Clause')
depends=('bash' 'fuse3' 'curl' 'c-ares' 'pacparser' 'sqlite' 'leveldb' 'libressl' 'libcap' 'protobuf')
makedepends=('cmake' 'make' 'gtest' 'sparsehash' 'help2man')
backup=('etc/cvmfs/default.local')
install=cvmfs.install
options=('!emptydirs')
source=(
    "https://github.com/cvmfs/cvmfs/archive/refs/tags/$pkgname-$pkgver.tar.gz"
    "fix-missing-cap-libraries.patch"
)
md5sums=('f2fe8673a4b05e2e931a780fa9254525'
         '5463c4779c744a1f322e3f0e2fdf8fa4')

# cvmfs tags are called cvmfs-Z.XX.Y
_srcdir="$pkgname-$pkgname-$pkgver"

prepare() {
    patch -d $_srcdir -Np1 -i ../fix-missing-cap-libraries.patch
}

build() {
    # Full relo recommended by namcap and set rpath for libressl, as it is not
    # used on arch elsewhere, but is required by cvmfs. Cleaner to use the
    # libressl package from the arch repos than building libressl as a part of
    # cvmfs, especially if there are other packages that rely on libressl.
    export LDFLAGS="$LDFLAGS -Wl,-z,relro,-z,now -Wl,-rpath,/usr/lib/libressl"

    # Hide numerous warnings from the cvmfs logging headers
    export CXXFLAGS="$CXXFLAGS -Wno-deprecated-enum-enum-conversion"

    local cmake_options=(
        -B build
        -S ${_srcdir}
        -W no-author
        -D CMAKE_BUILD_TYPE=None
        -D CMAKE_INSTALL_PREFIX=/usr
        -D BUILD_CVMFS=ON
        -D BUILD_SERVER=OFF
        -D BUILD_RECEIVER=OFF
        -D BUILD_LIBCVMFS_CACHE=OFF
        -D BUILD_LIBCVMFS=OFF
        # If we need fetch content, add urls to source above
        -D FETCHCONTENT_FULLY_DISCONNECTED=ON
        -D BUILTIN_EXTERNALS_LIST="sha3"
        -D Libcrypto_LIBRARIES=/usr/lib/libressl/libcrypto.so
        -D Libcrypto_INCLUDE_DIRS=/usr/include/libressl
    )
    cmake "${cmake_options[@]}"
    cmake --build build
}

package() {
    DESTDIR="$pkgdir" cmake --install build

    install -Dm644 "${_srcdir}/COPYING" "${pkgdir}/usr/share/licenses/cvmfs/COPYING"

    echo "CVMFS_REPOSITORIES=atlas.cern.ch,atlas-online-nightlies.cern.ch,sft.cern.ch" > $pkgdir/etc/cvmfs/default.local
    echo "CVMFS_HTTP_PROXY=DIRECT" >> $pkgdir/etc/cvmfs/default.local
}
```
</details>


vvolkl:
Hi Ole, thanks for reporting this and contributing to cvmfs!

I'm happy to merge this patch, but strictly speaking I don't think it should be necessary. My reasoning is that the stub is `dlopen`'ed by cvmfs2, which means that when resolving symbols in the stub it should inherit the scope of cvmfs2, which does link libcap. Could it be that your installation accidentally mixed versions, for example had a cvmfs2 from 2.13.3 and a stub library from 2.14.1? That's the only explanation I can think of, because I also couldn't reproduce this issue with the provided PKGBUILD in a fresh arch container.

In any case, I don't see any downside to explicitly linking libcap in the stub, that's certainly correct, so will merge regardless.

barskern:
> Hi Ole, thanks for reporting this and contributing to cvmfs!
> 
> I'm happy to merge this patch, but strictly speaking I don't think it should be necessary. My reasoning is that the stub is `dlopen`'ed by cvmfs2, which means that when resolving symbols in the stub it should inherit the scope of cvmfs2, which does link libcap. Could it be that your installation accidentally mixed versions, for example had a cvmfs2 from 2.13.3 and a stub library from 2.14.1? That's the only explanation I can think of, because I also couldn't reproduce this issue with the provided PKGBUILD in a fresh arch container.
> 
> In any case, I don't see any downside to explicitly linking libcap in the stub, that's certainly correct, so will merge regardless.

Hi @vvolkl , thank you for the quick reply and handling!

Ah, very interesting. This then might be an issue with my system setup. It would be very curious if it got mixed up as I am compiling and installing them together, I should not have the stubs available anywhere else. Anyhow, I will try to reproduce it in a container aswell to isolate away my specific system!

vvolkl (approved):

barskern:
I think I found the cause, I've been using `-fuse-ld=mold` (aka. the [mold](https://github.com/rui314/mold) linker) lately, and when I swap away from that, it starts working again. Hence it seems to be that it does something differently in this case. I don't really have the excess time to delve much deeper into this at the moment, but I suppose it's worth noting and mentioning here anyhow. My current variant of the PKGBUILD is [here](https://gitlab.com/barskern/pkgs/-/blob/master/cvmfs/PKGBUILD) for posterity and reference.

vvolkl:
> I think I found the cause, I've been using `-fuse-ld=mold` (aka. the [mold](https://github.com/rui314/mold) linker) lately, and when I swap away from that, it starts working again. Hence it seems to be that it does something differently in this case. I don't really have the excess time to delve much deeper into this at the moment, but I suppose it's worth noting and mentioning here anyhow. My current variant of the PKGBUILD is [here](https://gitlab.com/barskern/pkgs/-/blob/master/cvmfs/PKGBUILD) for posterity and reference.

Interesting, thanks for sharing!

barskern:
I was too curious so did a low-effort LLM search and it seems to have hit gold. I verified its claims by doing readelf on `cvmfs2` after a build with gnu ld and mold, for the latter the `NEEDED` dependency for `libcap` is gone.

The TLDR is; This is not a `dlopen()` bug in mold. The loaded CVMFS stub uses `cap_get_proc` from `libcap` but did not link against `libcap` directly. With LTO and `--as-needed`, mold correctly removed libcap from `cvmfs2` because the executable had no live references to it, leaving no provider when the stub was loaded. GNU `ld.bfd` happened to retain the accidental dependency. Linking the stub directly with `${CAP_LIBRARIES}`, as PR 4424 does, is the correct fix.

[Full LLM explanation](https://www.perplexity.ai/search/f4a386b5-0b09-48be-8cf4-c2aee55474b2)

PR #4431, opened 2026-09-08
Fix blank configuration settings

### LINKED ISSUE #4430: Metalink feature currently unusable due to cvmfs variable-setting quirky behavior
I want to use CVMFS_EXTERNAL_METALINK now that OSDF redirectors are available and that bugs have been fixed in 2.14.1.  Unfortunately, one of the bugs fixed was that prior to 2.14.1, when CVMFS_EXTERNAL_METALINK is set to a blank string, cvmfs ignores CVMFS_EXTERNAL_URL.  Combined with the cvmfs quirky behavior that if a config variable is set within a conditional statement that is not executed then the variable is set to a blank string, that makes it impossible to define CVMFS_EXTERNAL_METALINK on new enough versions of cvmfs without breaking 2.13.x and 2.14.0.

DrDaveD:
This makes two changes related to configuration variables getting set to blank:
1. Adds `declare -g` to the prefixes recognized as setting variables.  This enables adding a backwards-compatible config variable setting that does not set a value to blank when the setting is only in a false conditional block, because previous versions will ignore the line.  EDIT: Since `declare -g` fails on Debian, instead use `export X VAR=VAL`.
2. If a variable is only set within a false conditional block, it will no longer get set to blank, it will just not be set.

- Fixes #4430

### DIFF
--- a/cvmfs/options.cc
+++ b/cvmfs/options.cc
@@ -64,6 +64,15 @@ string OptionsManager::TrimParameter(const string &parameter) {
   } else if (result.find("export ") == 0) {
     result = result.substr(7);
     result = Trim(result);
+    // This enables a hack to ignore a variable setting in cvmfs versions
+    // prior to 2.14.2 which also stopped setting variables to blank when
+    // they are defined only in false conditionals.  The hack is to use
+    // 'export X VAR=VAL', where VAR will be recognized in 2.14.2 and later
+    // and ignored prior to 2.14.2.
+    const size_t last_space = result.find_last_of(" ");
+    if (last_space != string::npos) {
+      result = result.substr(last_space + 1);
+    }
   } else if (result.find("eval ") == 0) {
     result = result.substr(5);
     result = Trim(result);
--- a/cvmfs/options.cc
+++ b/cvmfs/options.cc
@@ -248,10 +248,12 @@ void BashOptionsManager::ParsePath(const string &config_file,
 
     ConfigValue value;
     value.source = config_file;
-    const string sh_echo = "echo $" + parameter + "\n";
+    const string sh_echo = "echo ${" + parameter + "-xNOTxSETx}\n";
     WritePipe(fd_stdin, sh_echo.data(), sh_echo.length());
     GetLineFd(fd_stdout, &value.value);
-    PopulateParameter(parameter, value);
+    if (value.value != "xNOTxSETx") {
+      PopulateParameter(parameter, value);
+    }
   }
 
   close(fd_stderr);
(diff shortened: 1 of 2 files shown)


test/unittests/t_options.cc
```diff
@@ -94,7 +101,7 @@ TYPED_TEST(T_Options, ParsePath) {
   options_manager.ParsePath(config_file, false);
   options_manager.SwitchTemplateManager(opt_temp_mgr);
 
-  // printf("DUMP: ***\n%s\n***\n", options_manager.Dump().c_str());
+  printf("DUMP: ***\n%s\n***\n", options_manager.Dump().c_str());
```
vvolkl: Either re-comment or  use this facility:

```suggestion
SCOPED_TRACE(options_manager.Dump());
```
That should give output  only on error.

cvmfs/options.cc
```diff
@@ -239,10 +242,13 @@ void BashOptionsManager::ParsePath(const string &config_file,
 
     ConfigValue value;
     value.source = config_file;
-    const string sh_echo = "echo $" + parameter + "\n";
+    const string sh_echo = "if [ -v " + parameter + " ]; then echo $" + \
+                                parameter + "; else echo xNOTxSETx; fi\n";
```
vvolkl: There's a bashism here and we still need to support dash.
```suggestion
const string sh_echo = "if [ -n \"${" + parameter + "+x}\" ]; then echo $"
                       + parameter + "; else echo xNOTxSETx; fi\n";
```

vvolkl (changes requested):

vvolkl:
Actually, `declare -g` will also fail under dash. Will this be an issue for debian?

cvmfs/options.cc
DrDaveD: I assumed that since the class is called `BashOptionsManager` that we could rely on it being bash.  Is that not so?  ... oh, I see it calls `Shell()` which calls `/bin/sh` :-( 

I thought I had tried the `${parameter+x}` style first but my mistake was to read the man page and there it talks about `:+` expansion which is not what I want because it applies to both empty and unset the same.  Ah, the text before the list of expansions says "Omitting the colon results in a test  only  for  a  parameter  that  is unset".  That means a `${parameter-xNOTxSETx}` is the easiest thing to use.

DrDaveD:
> Actually, `declare -g` will also fail under dash. Will this be an issue for debian?

Yes, that's a problem.  Ok, I have an alternative that works for dash and bash: `export X VAR=VAL` will ignore VAR in older versions, but I'll make that be recognized instead of `declare -g`.

DrDaveD:
@vvolkl Please review

